PR #27, opened 2017-06-13
Replacing einsum with iterated `np.tensordot` for speedup

### LINKED ISSUE #24: Checking if a germ set is "amplificationally complete" (AC) takes a long time
In the example germ generation notebook, checking if a 1Q germ set is AC seems to be fine, but for 2Q it takes a while. (I can add more quantitative statements in a bit.)

ampolloreno:
Fixes #24. Hopefully that's because `einsum` is slow, and not because I handled my indices incorrectly.


enielse:
This functionality has been incorporated as a part of a larger update.

PR #29, opened 2017-07-18
Suggested Speedups for Germ Selection

bbloo:
When running germ selection on complicated gatesets, performance of some of these functions starts to matter a lot.


bbloo:
Similar to #27 but tested its use on the beta branch

enielse:
This functionality has been incorporated as a part of a larger update.

PR #42, opened 2018-10-20
adding a space in a docstring

ampolloreno:


enielse:
Thanks for the request - this was fixed in a more recent version of pyGSTi.

PR #71, opened 2019-06-13
Feature timedependence

enielse:
Adds (initial) support for time-dependent circuits and GST, and for continuously-parameterized gates (e.g. an arbitrary-angle X rotation gate).  This adds the basic framework, but routines are slow and need debugging.


enielse:
Fuller support for continuously parameterized gates will require updates to `ProcessorSpec` and the creation of a built-in factory that wraps around function that takes in label-arguments and outputs a unitary matrix.

PR #98, opened 2019-11-22
Feature report jinja templates

enielse:
Upgrades pyGSTi html reports to use jinja templates and produce a single html file without Ajax calls to facilitate better browser compatibility.


robpkelly:
@enielse I pushed a few changes, mostly offloading the new HTML generation to the templates. I would say this branch is good to merge, but I'd like you to check that generated reports still load correctly in Safari with the changes I made. If it works, go ahead and merge.

One note: twice while testing your changes on Chromium, switching tabs at the wrong time seemed to prevent a table from loading. I've tried and I just can't figure out how to reproduce this, but there's most likely a (rare) race condition in the figure loading pipeline. Something to be aware of, I guess.

enielse:
I like the additional HTML templating, but the reports weren't loading without an internet connection for me.  I tracked this down to the onerror handler of <script> tags being fired asynchronously, and this led to the above commits, which keep most of the new templating but tweak the CDN-fallback machinery to something closer to what is was earlier (checking for a library's presence and using document.write to add fallback script tags).  All the tests I've done work now.

Issue #90, opened 2019-09-13
Dropping Python 2 support


robpkelly:
Python 2.7 will reach end-of-life on January 1st, 2020 (see [this melodramatic countdown timer](https://pythonclock.org/)). Additionally, [many major Python projects](https://python3statement.org/) have pledged to drop support for Python 2 on or before that date.

The next major release of pyGSTi (v0.9.9) will drop support for Python 2. In other words, we'll be limiting our future support to Python 3.5 and 3.7. Users running pyGSTi on Python 2 should consult [the official guide](https://docs.python.org/3/howto/pyporting.html) for more information on porting their environment to Python 3.

---

As of ace6d65 our CI no longer builds for Python 2.7. Developers are no longer required to ensure Python 2.7 compatibility in new contributions. Here's what should be done before v0.9.9:

- [X] Drop Python 2.7 builds from TravisCI
- [x] Specify `python_requires` in `setup.py`
- [x] Remove Python 2 backports from dependencies in `setup.py`
  - _Not sure we actually have any strictly for Python 2..._
- [x] Remove Python 2 compatibility duct-tape throughout the source tree, including:
  - [x] `sys.version_info` checks for python 2
  - [x] most if not all `__future__` imports
  - [x] Python 2 compatibility patches in `pygsti.tools.compattools`
- [x] Remove Python 2 static test fixture resources

robpkelly:
#102 closes this, but I'm keeping this issue open until v0.9.9 just to keep users/developers aware

PR #105, opened 2019-12-14
Refactor `pygsti.baseobjs`

robpkelly:
Since we're dropping Python 2.7 support with #90, it's easier to avoid circular imports, meaning we can refactor the names under the `baseobjs` package to other existing packages that make more sense.

- Refactored under `objects`:
  - `baseobjs.smartcache` 
  - `baseobjs.verbosityprinter`
  - `baseobjs.profiler`
  - `baseobjs.basis`
  - `baseobjs.protectedarray`
  - `baseobjs.label`
  - `baseobjs.objectivefns`
- Refactored under `tools`:
  - `baseobjs.opttools`
  - `baseobjs.basisconstructors`
- Refactored under `io`:
  - `baseobjs.circuitparser`
  - `baseobjs.fastcircuitparser`
- Removed:
  - `baseobjs.parameterized`
  - `baseobjs.deprecated_dim`
  - `baseobjs.exceptions`

Since names under `baseobjs` were deliberately exposed in other modules, this should _not_ constitute a major API change. One possible exception is that `basistools` no longer exposes the `basisconstructors` namespace via wildcard import. If this actually is a non-trivial API change, I'd want to either re-expose those names with a deprecation warning, or simply merge `basisconstructors` into `basistools` entirely (I'm a little confused on the semantics of separating them in the first place)


enielse:
Let's restore the `exceptions` module into the `tools` or `objects` sub-package.  The exceptions *are* objects, but they're not primary ingredients in pyGSTi and are used more like utility functions, so I think I lean toward putting them in `tools`.  Unless Rob strongly disagrees, let's do that.

PR #107, opened 2019-12-18
Move source tree to repo root

robpkelly:
This patch moves the pyGSTi source tree from `packages/pygsti` to `pygsti`, and updates relevant paths elsewhere in the repo. Looks big but it really isn't.

The original intention of storing the source tree under `packages` was to allow free sharing of resources between pyGSTi and potential related future packages; I argue that this is harmful to API design. Related future packages should share resources through a well-defined API, and should be controlled in a separate repository under the pyGSTio organization. In the mean time, let's organize our repo in a more usual fashion, with the source tree immediately under the repo root.

**Developers should take care to update their IDE project setup when pulling this patch!**


robpkelly:
@enielse @kmrudin @tjproct @aerusso if you're tagged here, it's because I know you may or may not be actively working on features. Give this comment a :+1:  so I know you read this.

Don't email me being like "hey where'd the source tree go" :stuck_out_tongue:

enielse:
Rob, this looks good - can the unit tests be updated to pass before we merge this in?

PR #110, opened 2019-12-20
Refactor cython extension modules for consistency

robpkelly:
A few pyGSTi units can optionally use optimized Cython implementations, falling back on slower native python implementations if the Cython extensions are not built. This patch refactors these extension/fallback units into subpackages with consistent structure and behavior: the Cython and native modules expose the same names, which may be imported directly if required, but may also be imported more generally from the subpackage. In the latter case, importing a name will import from the optimized Cython extension if it is available, and fall back to the native python implementation if not. When falling back to the native python implementation, a warning is shown on import, but may be suppressed by setting the environment variable `PYGSTI_NO_CYTHON_WARNING`.

This patch:
- refactors `replib`, `fastreplib` and `slowreplib` into `objects.replib`
- refactors `fastopcalc` and the corresponding native python implementations from `polynomial` into `objects.opcalc`
- refactors `fastcircuitparser` and the corresponding native python implementations from `circuitparser` into `io.circuitparser`

---

`tools.fastcalc` has been left untouched because it doesn't have a complete native python implementation. I'm just leaving that behavior as it is, unless someone suggests otherwise.

Previously, `objects.replib` used a pattern somewhat similar to this, but would warn when exposed methods were called rather than warning on import. I argue warning on import is more useful, because a developer would most likely want to know what module they're importing when they import it.


enielse:
I need to look at this more closely, but I think this PR may cause the "cython warning" to get displayed *whenever* a user doesn't have the cython modules built, even if the would-be cython routines are never used.  The desired behavior is that the warning message is only displayed when a would-be cython function is *invoked*, not when its module is loaded (because the modules will always be loaded in the course of importing pygsti, but there's not need to warn the user about slow performance unless the user is doing something that utilizes those routines).

Overall I like the refactoring here and think this is much better organization.  But I'd like to preserved the behavior described above and think this should be possible if it isn't the behavior already.

enielse:
It turns out that this PR *will* change the warning behavior (when cython extensions aren't built) from the previously used warn-on-call to a warn-on-import behavior.  Rob has convinced me this is a better approach that will annoy less people, so despite my initial concerns I'm good with this now and will merge it.

PR #112, opened 2020-01-20
WIP: New modelpack API

robpkelly:
Note that this patch depends on `pygsti.protocols.gst.StandardGSTDesign`, so I'm requesting to merge into `feature-protocols` instead of `develop`. This feels very "agile" which I hear is a good thing!

This patch introduces a new object-oriented API for accessing standard modelpacks, which were previously modules prefixed with `std1Q` and `std2Q` under `pygsti.construction`.

- Legacy modelpacks have been refactored into `pygsti.modelpacks.legacy`
- New modelpacks use multi-qubit labelling conventions
  - Modelpack members can optionally be parameterized with a label (index) factory
- New modelpacks are objects which inherit from a base class (`Modelpack`)
  - Can still be accessed via import mechanism, like modules
- Added stub for `RPEModelpack` type
- Added `scripts/upgrade_stdmodule.py`, which generates python source for SMQ modelpacks from legacy modelpack definitions


robpkelly:
I wasn't sure how to add qubit-label parameterization to `target_model` without breaking the existing caching structure. Work that in if you can.

PR #136, opened 2020-05-21
paths: fix open() syntax for python 3.5

drewrisinger:
The following syntax doesn't work in python 3.5 (tried looking for a doc for a few minutes, but couldn't find the link I found a few days ago). [python3.8 pathlib docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open)
```python
with open(pathlib.Path()) as f:
    ...
``` 
Instead, the following syntax should be used
```python
with pathlib.Path().open() as f:
    ...
```
This "bug" was remedied in python 3.6+, but not in 3.5. Since pygsti nominally supports python3.5, this change should be made.
(I haven't changed this everywhere, just where I ran into it).

Also just an FYI, didn't convert to it here, but pathlib has nice ``[read,write]_text()`` and ``[read,write]_bytes()`` methods.

### DIFF
--- a/install_locally.py
+++ b/install_locally.py
@@ -17,7 +17,7 @@ print("Adding path:", myPackagePath)
 usp = Path(site.getusersitepackages())
 usp.mkdir(parents=True, exist_ok=True)
 uspfile = usp / 'GST.pth'
-with open(uspfile, 'w') as f:
+with uspfile.open('w') as f:
     f.write(pathspec)
     print('Wrote to {}'.format(uspfile))
 
--- a/pygsti/io/metadir.py
+++ b/pygsti/io/metadir.py
@@ -117,7 +117,7 @@ def load_meta_based_dir(root_dir, auxfile_types_member='auxfile_types',
     """
     root_dir = _pathlib.Path(root_dir)
     ret = {}
-    with open(root_dir / 'meta.json', 'r') as f:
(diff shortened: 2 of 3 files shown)


enielse:
These are good changes. Thanks for highlighting this lack of compatibility.  There are multiple ways to fix these bugs; I'd prefer to follow the pattern we've used elsewhere in pyGSTi, where we change `open(fn)` to `open(str(fn))`.  Because this pattern is slightly different than what this PR contains, and since we'd need @drewrisinger to sign a CLA before being able to accept the PR, it seems preferable to just fix this issue internally and not accept this PR.  Let's leave it open, however, until the fixes are made and python 3.5 compatibility is restored.

drewrisinger:
Fine by me. Sorry about going off style, but feel free to send me a CLA. I
might be sending more fixes in the future.

drewrisinger:
Any update? Ran into this issue again today while trying to run more circuits on a different machine with the ``beta`` branch.

enielse:
Sorry for the delay.  Recent commit d4f4639fd804a5981f3621f0e86f81191cd8c4e3 should fix this issue by replacing `open(.)` with `open(str(.))` as this is the pattern we've been using.  This update will be on the **beta** branch until the next release; is this acceptable? (I'll assume so unless I hear otherwise.)

drewrisinger:
Yup, great! Thank you!

PR #148, opened 2020-07-22
Documentation generation cleanup & bugfixes

### LINKED ISSUE #124: ReadTheDocs Manual out of date
**Describe the bug**
Looks like the ReadTheDocs manual is out of date, dated to ~Jan 2019 (commit 9b197b5988499ff1be317db21c1905b8c63c6e49).

**To Reproduce**
1. Go to https://pygsti.readthedocs.io/en/latest/
2. Look at ``Revision ...`` in footer

**Additional context**
I was looking for a few more-recent functions that aren't documented in the manual, such as ``pygsti.io.write_empty_protocol_data``. Found in source, just didn't expect manual to be out of date.

robpkelly:
Closes #124 

This patch cleans up & fixes errors relating to automated project documentation generation:

- Fixes `Basis` being exposed by (and imported from) `tools` instead of `objects`.
- Controls names exposed by `protocols.rpe`.
- Tweaks & cleanup for `sphinx` templates.
  - Adds autodocs for `protocols`, `modelpacks`, `extras.rpe` and `extras.drift` subpackages.
  - Links pyGSTi subpackages directly from documentation root.
  - Includes inherited members in autodocs & shows inheritance.
  - Uses standard `autoclass` for readability.
  - Removes unused/unneeded templates.
  - Uses `objname` instead of `fullname` for headers.
  - Organizes autodoc stubs under `_auto` directory instead of documentation root.
- Removes monkey-patching of `sphinx` autosummary generation (equivalent functionality is built into `sphinx>=2.1`) which was causing documentation builds to fail under certain circumstances.
  - Pins `sphinx>=2.1` in `rtd-requirements.txt`
- Moves documentation to `/doc`
  - Moves developer notes to `/doc/notes`
- Uses package metadata to populate release/version info in docs (ultimately from `setuptools_scm`)
- Removes suppression of most `sphinx` warnings, which developers should probably be aware of.
- Adds links to Apache license & SNL copyright disclaimer to documentation root.

The docs produced by this patch can currently be reviewed [on this _Read the Docs_ version](https://pygsti.readthedocs.io/en/124-bugfix-docgen/index.html), which will be removed after this patch is merged.


pygsti/protocols/rpe.py
```diff
@@ -323,3 +323,9 @@ def raw_angles(self):
 RPEDesign = RobustPhaseEstimationDesign
 RPE = RobustPhaseEstimation
 RPEResults = RobustPhaseEstimationResults
+
+__all__ = [
```
enielse, line 327: Why is this needed?

robpkelly: It's included in this patch just so that autodoc doesn't try to document external names imported in `protocols` from `protocols.rpe`, namely `OrderedDict`, `namedtuple`, `Namespace` and `numpy`.

There are less-invasive ways of doing this but I figure we don't want to expose those anyway so we might as well.

We could also import those names as protected in `protocols.rpe` to match the pattern used elsewhere but I felt that was more invasive.

enielse: Gotcha - If you don't mind, let's switch over to using protected names like we do elsewhere and drop the explicit __all__ assignment.  It's a good kind of invasive :)

robpkelly: Sounds good to me. Resolved with 1c078e8

PR #150, opened 2020-08-08
Make `pygsti.DataSet.add_count_dict` convert all keys in `count_dict`…

lnmaurer:
… to strings. This allows us to use the `collections.Counter` returned by `cirq.TrialResult.histogram` for `count_dict`. This does not affect existing behaviour because if the keys are already strings, then they won't be altered.

### DIFF
--- a/pygsti/objects/dataset.py
+++ b/pygsti/objects/dataset.py
@@ -1400,9 +1400,10 @@ class DataSet(object):
         elif isinstance(count_dict, _OrderedDict):  # then don't sort keys
             outcomeCounts = _ld.OutcomeLabelDict(list(count_dict.items()))
         else:
-            # sort key for deterministic ordering of *new* outcome labels)
+            # sort key for deterministic ordering of *new* outcome labels). If keys are
+            # not strings, they will be converted to strings.
             outcomeCounts = _ld.OutcomeLabelDict([
-                (lbl, count_dict[lbl]) for lbl in sorted(list(count_dict.keys()))])
+                (str(lbl), count_dict[lbl]) for lbl in sorted(list(count_dict.keys()), key=str)])
 
         outcomeLabelList = list(outcomeCounts.keys())
         countList = list(outcomeCounts.values())


lnmaurer:
I think that you can simplify `sorted(list(count_dict.keys()), key=str)` to `sorted(count_dict, key=str)` altho the former is probably easier to interpret. I might at least get rid of the `list(...)`.

    test = {'a': 0, 'b': 1, 'c': 3}
    sorted(test.keys()) -> ['a', 'b', 'c']
    sorted(test) -> ['a', 'b', 'c']

enielse:
The suggested update, which changes `DataSet.add_count_dict` so it converts the keys of `count_dict` into strings, would be problematic for other portions of pyGSTi.  

PyGSTi uses *tuples of strings*, in general, to designate circuit outcomes.  For example, a circuit with an intermediate measurement with outcomes `"A"` and `"B"` and a final measurement with outcomes `"0"` and `"1"` has the outcomes `("A","0")`, `("A","1")`, `("B","0")`, and `("B","1")`.  In the special case when there is only a terminating measurement the outcome is a 1-tuple.  PyGSTi contains logic that seamlessly treats 1-tuple outcomes as strings and vice versa, giving the impression that outcome labels are just strings so long as you never deal with intermediate measurements.  That said, the suggested update would break this framework for the general case.

I'm not entirely clear on the motivation for this change - perhaps just casting or changing the keys `Counter` returned from `cirq.TrialResult.histogram` to strings would offer an alternative solution?

enielse:
Regarding the suggested simplification of `sorted(list(count_dict.keys()))` to `sorted(count_dict)`, I expect you're right - though I'd want to check this works for versions of Python 3.5 and above.  It may be a leftover from some Python 2 version also.  I don't think the creation of the list is a big performance hit.  I'll note this as a low priority update.

enielse:
Closing this PR for the reasons mentioned above.  Feel free to reopen if continued discussion is desired.

PR #155, opened 2020-10-16
Remove mpi4py dependency

aerusso:
This patch adds a code path that does not use MPI.  It needs a code review and testing.


enielse (approved):
Looks good.

PR #156, opened 2020-10-20
Added DataSet.add_cirq_trial_result method to import Cirq.TrialResult objects

lnmaurer:
I ended up adding the import method to DataSet because that seemed to fit well. I can move it elsewhere if you'd like. I also added a jupyter notebook showing how to run everything. Feel free to include it or discard it.


lnmaurer:
(Looks like I have some typos in the notebook, but can iron that out later if needed.)

enielse (approved):
Looks good.  We may want to add a unit test at some point to make sure this doesn't break.

PR #158, opened 2020-12-03
Feature tutorial fix

sserita:
WIP: First pass at getting all tutorials running on the develop/beta branches. All syntax and some minor class refactoring is working.

There are a few outstanding problems remaining that I'd like feedback on:

- [x] `objects/ExplicitModel`: There is no more `product` function in the `Model` class, didn't see an obvious replacement.
- [x] `objects/advanced/ProcessorSpec`: Type mismatch in underlying representation code when doing forward simulation of Clifford circuit
- [x] `algorithms/DriftCharacterization`: There appears to have been a change to `CircuitListDesign` that causes issue with some analysis/report tools, specifically how fid/germ circuits are stored/indexed?
- [x] `algorithms/RB-CliffordRB`: There is a KeyError related to outcomes in the `compute_summary_statistics`. Worth mentioning that DirectRB and MirrorRB do NOT have this problem
- [x] `algorithms/RB-Overview`: Same issue as `algorithms/RB-CliffordRB`
- [x] `algorithms/VolumetricBenchmarks`: The class is commented out in `protocols/vb.py` so this whole tutorial is moot?
- [x] `algorithms/advanced/CliffordRB-Simulation-ExplicitModel`: Similar error to `algorithms/RB-CliffordRB` but now with Circuit labels instead of outcomes. Worth mentioning that ImplicitModel does NOT have this problem
- [x] `algorithms/advanced/ForwardSimulationTypes`: Tone of issues with `statevec` evotype. Matrix simtype raises NotImplemented, map simtype shows similar rep type issues as `objects/advanced/ProcessorSpec`
- [x] `report/ReportGeneration`: LaTeX PDF generation failed, maybe plotly is returning RGB tuple instead of hex code?
- [x] `report/advanced/WorkspaceSwitchboards`: No more `BulkCircuitList`, this is probably just `CircuitList` now. Real issue is I'm not sure what the matching `truncate` call is - functions like `truncate_from_circuits` only take list/set, not sure what the matching logic from `germs` kwarg was

Also there are a few comments (but not errors):
- `objects/advanced/Results`: I needed to construct an Estimate class, which is not exposed in `protocols/__init__.py`. Should it be?
- `objects/advanced/GST-FiducialAndGermSelection`: Just a comment, the `find_germs` and GRASP part is what took a long time (several minutes for me). Hopefully useful data point for profiling any FPR perf issues


sserita:
Removing the `ProcessorSpec` and `ForwardSimulationTypes` issues as blocking for this PR. This issue will be tackled in the future once MPI enhancements of the forward simulators is completed. See Issue #161.

sserita:
Latest commits have fixes for all checked issues except for replib evotype (Issue #161) and the Volumetric Benchmarks. I think it's ready for the handoff to you, @enielse.

enielse (approved):
Everything looks good.  The remaining work involves getting the various forward simulators and evolution-types working again, which will be the topic of a larger effort so we'll just note that here as an item for future work.

PR #160, opened 2020-12-10
Bugfix marginalized povm

sserita:
This includes Erik's suggested fixes + a modification to _is_primitive_povm_circuit_lbl to allow marginalized primitive POVM labels to be detected as primitive as well.

The new test (`test_marginalized_povm`) fails if custom `qubit_labels` are used, as in other `LocalNoiseModel` tests. I think this is a related but separate marginalization bug to what I'm trying to fix in this PR. The relevant portion of error message:
```
  File "/Users/sserita/Documents/repos/pyGSTi/pygsti/objects/povm.py", line 1468, in <listcomp>
    indices_to_keep = set([list(all_sslbls).index(l) for l in sslbls_after_marginalizing])
ValueError: 0 is not in list
```
This can be reproduced by adding `qubit_labels=['qb{}'.format(i) for i in range(nQubits)],` into the constructor call of `mdl_local` in the new test.

### DIFF
--- a/pygsti/objects/implicitmodel.py
+++ b/pygsti/objects/implicitmodel.py
@@ -281,7 +281,7 @@ class ImplicitOpModel(_mdl.OpModel):
 
         return s
 
-    def _default_privitive_povm_layer_lbl(self, sslbls):
+    def _default_primitive_povm_layer_lbl(self, sslbls):
         """
         Gets the default POVM label.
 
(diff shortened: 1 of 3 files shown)


sserita:
I fixed the test for custom labels. It was mostly errors in translating my test case from a notebook into the test. I think this is ready for review (pending tests coming back clean).

enielse (approved):
We're set to go :)

PR #166, opened 2021-01-13
Bugfix action merge

sserita:
The (unnamed workflow) was probably a result of the merge using a pull request under the hood. Apparently the PR uses temporary SHAs that don't work as smoothly with GitHub Actions, or something like that. Related issue can be found [here](https://github.com/IgnusG/jest-report-action/issues/33#issuecomment-639889546).

Anyway, my solution is to do a merge + push instead of a pull. Also, I simplified this logic down a lot and got rid of the external push action. Hopefully that gives less chances for errors, but we will still need to monitor this in production.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,41 +34,41 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    #- name: Set up installation environment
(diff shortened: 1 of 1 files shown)


enielse (approved):
Looks great - I hope it works.

PR #174, opened 2021-04-16
Feature std static op

sserita:
Added the StaticStandardOp class and reworked the crosstalk-free model construction for more explicit error specification.

Probably more work could be done on extending new error specification API to the cloud noise model construction, but that seems less clear since that construction relies heavily on LindbladOps. In the meantime, this PR could already be useful in the feature-chp-revive branch.


enielse (approved):
Everything looks good!

PR #178, opened 2021-05-04
Feature chp revive

sserita:
Integrates Scott Aaronson's CHP program into pyGSTi for forward simulation.

Major infrastructure changes:

- A new WeakForwardSimulator base class designed to run "shots" of circuits to obtain outcome probabilities
- A CHPForwardSimulator using files and subprocess to interface with the CHP executable
- A new `chp` evotype which is integrated with the recently developed StandardStaticOp
- A new "sparse" outcome probabilities interface in ForwardSimulator when no outcomes are provided, potentially allowing one to avoid construction of dense outcome probability arrays (e.g. COPALayout, simplified POVM effects)


pygsti/construction/modelconstruction.py
```diff
@@ -1243,14 +1245,22 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
...
-        idle_op = None
+        global_idle_op = None
 
     prep_layers = {}
-    if 'prep' in all_keys:
+    if evotype == 'chp':
+        assert ('prep' not in all_keys), "Cannot specify 'prep' error specification with CHP evotype"
+
+        # For CHP, prep can only be idle gate
+        rho0 = _op.ComposedOp([_op.EmbeddedOp(range(num_qubits), [i], _op.StaticStandardOp('Gi', 'chp'))
```
enielse, line 1260: It still seems incorrect to me to set 'rho0' to be an operation intstead of a spam vector.

pygsti/extras/rb/simulate.py
```diff
@@ -349,13 +349,14 @@ def depolarizing_errors_circuit_simulator(circuitlist, shots, errormodel, gate_t
...
+        # SS 2021-02-02: Checking whether this is still needed
         # Todo : this is a temp hack to get around a bug in Circuit.
-        if circuit[-1].name[0] == '#':
-            circuit = circuit.copy(editable=True)
-            circuit.delete_layers(-1)
-            circuit.delete_lines('*')
+        #if circuit[-1].name[0] == '#':
+        #    circuit = circuit.copy(editable=True)
+        #    circuit.delete_layers(-1)
+        #    circuit.delete_lines('*')
```
enielse, line 357: This seems good - the hash tag stuff is leftover from an earlier time, and this shouldn't be needed anymore.

pygsti/objects/forwardsim.py
```diff
@@ -187,8 +191,16 @@ def probs(self, circuit, outcomes=None, time=None):
         -------
         probs : OutcomeLabelDict
             A dictionary with keys equal to outcome labels and
-            values equal to probabilities.
+            values equal to probabilities. If no target outcomes provided,
+            only non-zero probabilities will be reported.
         """
+        if outcomes is None:
+            try:
+                # TODO: Patch ralloc up into probs so it can be passed here (and in create_layout)
```
enielse, line 199: I'd like to discuss what you have in mind here re: "patch ralloc up..."

pygsti/objects/operation.py
```diff
@@ -1434,13 +1471,36 @@ def __init__(self, name, evotype):
...
+            nqubits = 2 if any(['c' in n for n in native_ops]) else 1
+
+            rep = replib.CHPOpRep(native_ops, nqubits)
         else:
             raise ValueError("Invalid evotype for a StaticStandardOp: %s" % evotype)
 
         LinearOperator.__init__(self, rep, evotype)
 
+    @property
+    def base(self):
```
enielse, line 1489: I don't think StaticStandardOp should have a .base attribute.  This should be in a to_dense method I think?

pygsti/objects/operation.py
```diff
@@ -2656,11 +2720,15 @@ class StochasticNoiseOp(LinearOperator):
         if not None, a list of `dim-1` initial error rates along each of
         the directions corresponding to each basis element.  If None,
         then all initial rates are zero.
+    
+    seed_or_state : float or RandomState, optional
+        Random seed for RandomState (or directly provided RandomState)
+        for sampling stochastic superoperators with the 'chp' evotype.
```
enielse, line 2726: Maybe say "when using a WeakForwardSimulator" rather than "with the 'chp' evotype"?

pygsti/objects/operation.py
```diff
@@ -2742,7 +2847,7 @@ def copy(self, parent=None, memo=None):
             A copy of this object.
         """
         if memo is not None and id(self) in memo: return memo[id(self)]
-        copyOfMe = StochasticNoiseOp(self.dim, self.basis, self._evotype, self._params_to_rates(self.to_vector()))
+        copyOfMe = StochasticNoiseOp(self.basis.size, self.basis, self._evotype, self._params_to_rates(self.to_vector()))
```
enielse, line 2850: Does self.dim not work?

pygsti/objects/operation.py
```diff
@@ -5859,6 +6044,30 @@ def __init__(self, state_space_labels, target_labels, operation_to_embed, dense_
             assert(nQubits is not None), "State space does not contain a definite number of qubits!"
             rep = replib.SBOpRepEmbedded(self.embedded_op._rep,
                                          nQubits, qubit_indices)
+        elif evotype == "chp":
+            # assert that all state space labels == qubits, since we only know
+            # how to embed cliffords on qubits...
+            assert(len(self.state_space_labels.labels) == 1
+                   and all([ld == 2 for ld in self.state_space_labels.labeldims.values()])), \
+                "All state space labels must correspond to *qubits*"
+            if self.embedded_op._evotype == 'chp':
```
enielse, line 6053: no "else" block corresponding to this?

pygsti/objects/povm.py
```diff
@@ -1075,6 +1080,63 @@ def __str__(self):
...
+        prefix : str
+            A string, usually identitying this POVM, which may be used
+            to prefix the simplified gate keys.
+
+        Returns
+        -------
+        OrderedDict of SPAMVecs
+        """
+        # TODO: CHP short circuit
+        if self._evotype == 'chp':
```
enielse, line 1125: Shouldn't this always be True given __init__?

pygsti/objects/operation.py
sserita: Changed to an assert

pygsti/objects/povm.py
sserita: Removed unused CHPPOVM

pygsti/objects/operation.py
sserita: dim was being used as the number of superoperators in the expansion. This change was reverted and I made sure to use basis.size more careful in StochasticNoiseOp and DepolarizeOp.

pygsti/objects/operation.py
sserita: It definitely shouldn't, this was just a temporary measure to get StaticStandardOp to work everywhere. Inserted a comment noting this, and hope that this will be fixed post evotype refactor.

pygsti/objects/operation.py
sserita: I originally agreed with this, but taking a closer look, this restriction is actually about the representation, not the type of forward simulation. The problem is that CHP only has C, H, and P as native gates, so we have to sample the stochastic map. For example, WeakForwardSimulators with a densitymx evotype could use the usual StochasticNoiseOp representation and therefore not use the seed_or_state kwarg.

Let me know if you still think I should change it.

pygsti/extras/rb/simulate.py
sserita: Removed.

pygsti/construction/modelconstruction.py
sserita: Agreed, but we decided to defer this issue for a more general SPAMVec + noisy op implementation in the near future. Comment added here to indicate that.

pygsti/objects/forwardsim.py
sserita: Finished piping up resource_alloc to probs (which is just passed through to the COPALayout).

pygsti/objects/operation.py
enielse: All these changes/comments look good.

enielse (approved):

Issue #175, opened 2021-04-20
Multiprocessing for RB circuit generation (with POC code)


lriesebos:
**Is your feature request related to a problem? Please describe.**

Circuit generation for Clifford RB can be unpractically slow for a large number of circuits, depths, and citerations. Probably this also holds for Direct RB. Large depths (8k and beyond) is especially common for ion-trap machines.

**Describe the solution you'd like**

We can get some easy speedups using Python [multiprocessing](https://docs.python.org/3/library/multiprocessing.html).
I can not make a pull request according to your contribution guidelines, but I completely sign-off on this untested proof-of-concept (POC) code for the Clifford RB design/protocol that can probably also be applied to direct RB and other RB variants too:

```python
from multiprocessing import Pool

...

class CliffordRBDesign(BenchmarkingDesign):
    """ Experiment design for Clifford randomized benchmarking """

    def __init__(self, pspec, depths, circuits_per_depth, qubit_labels=None, randomizeout=False,
                 citerations=20, compilerargs=[], descriptor='A Clifford RB experiment',
                 add_default_protocol=False, seed=None, verbosity=1, processes=1):
        
        ...

        for lnum, l in enumerate(depths):
            if verbosity > 0:
                print('- Sampling {} circuits at CRB length {} ({} of {} depths)'.format(circuits_per_depth, l,
                                                                                         lnum + 1, len(depths)))
                print('  - Number of circuits sampled = ', end='')
            circuits_at_depth = []
            idealouts_at_depth = []
            
            def mp_fn(n)
                r = _rc.clifford_rb_circuit(pspec, l, qubit_labels=qubit_labels, randomizeout=randomizeout,
                                            citerations=citerations, compilerargs=compilerargs)
                if verbosity > 0: print(n + 1, end=',')
                return r
            
            with Pool(processes) as p:
                output = p.map(mp_fn, list(range(circuits_per_depth)))
            
            for c, iout in output:
                circuits_at_depth.append(c)
                idealouts_at_depth.append((''.join(map(str, iout)),))
                
            circuit_lists.append(circuits_at_depth)
            ideal_outs.append(idealouts_at_depth)
            if verbosity > 0: print('')

            ...
```

**Describe alternatives you've considered**

Multiprocessing can be applied even more aggressively and/or fine-grained over the various dimensions (circuits, depths, citerations). Feel free to modify the POC implementation as desired.

lriesebos:
[...]

PR #179, opened 2021-05-06
175 feature rb design enhancement

sserita:
Several enhancements to speedup RBDesign construction:

- parallelization of circuit creation (Issue #175) 
- more efficient symplectic reps for circuits (fewer checks, internal gate construction)
- a Cython routine for fast Clifford composition

Below are listed some example speedups and other notes from profiling.
Timings were taken on a 2018 MacBook Pro with a 2.2 GHz 6-core Intel i7.

| RB Design | Width | Depth | Circuits | Develop | 1 Proc | 4 Proc | 8 Proc |
| ---              | ---      | ---      | ---                 | ---         | ---      | ---        | --- |
| CliffordRBDesign | 2 | 128 | 10 | 40 s | 33 s | 11.1 s | 9.9 s |
| DirectRBDesign | 4 | 512, 1024, 2048 | 20 | 88 s | 34 s | 13.5 s | 11.3 s |
| MirrorRBDesign | 25 | 512 | 20 | 53 s | 23 s | 8.0 s | 6.5 s |

Before improvements, symplectic_rep_of_clifford_circuits was the bottleneck for DirectRBDesign and MirrorRBDesign.
This is no longer the case, resulting in the first set of speedups from Develop to the current PR on 1 processor.
DirectRBDesign is now bottlenecked by compute_stabilizer_state and compute_stabilizer_measurement, and MirrorRBDesign is now bottlenecked by sample_pauli_layer_as_compiled_circuit.
CliffordRBDesign has been (and is still) bottlenecked by the Clifford compilation.

The multiprocessing proof-of-concept code given in #175 no longer works for Python 3.8 on Mac due to the change in multiprocessing start method (from "fork" to "spawn"), which would require pickling of the RBDesign objects.
Instead a `tools.mptools` with a general mapping function (with kwargs, for convenience) was added.
Parallel speedup is not perfect but certainly helpful.

A list of the next targets to gain further speedups:

- compile_conditional_symplectic for DirectRBDesign
- change_gate_library for MirrorRBDesign and CliffordRBDesign
    - compress_depth_inplace for CliffordRBDesign
    - insert_labels_into_layers_inplace for MirrorRBDesign


pygsti/tools/symplectic.py
```diff
@@ -1026,7 +1038,7 @@ def compute_internal_gate_symplectic_representations(gllist=None):
     return srep_dict
 
 
-def symplectic_rep_of_clifford_circuit(circuit, srep_dict=None, pspec=None):
+def symplectic_rep_of_clifford_circuit(circuit, srep_dict={}, pspec=None):
```
enielse, line 1041: Let's switch this back to None, unless the overhead is really a problem.  Using an empty dictionary as a default argument is bad practice since a dict is a mutable type -- weird stuff can happen if things go awry.  This may mean the docstring gets reverted too?

pygsti/tools/symplectic.py
```diff
@@ -451,8 +460,10 @@ def compose_cliffords(s1, p1, s2, p2):
         The 'phase vector' over the integers mod 4 representing the compsite Clifford
     """
     assert(_np.shape(s1) == _np.shape(s2)), "Input must be Cliffords acting on the same number of qubits!"
-    assert(check_valid_clifford(s1, p1)), "The first matrix-vector pair is not a valid Clifford!"
-    assert(check_valid_clifford(s2, p2)), "The second matrix-vector pair is not a valid Clifford!"
+    if do_checks:
+        assert(check_valid_clifford(s1, p1)), "The first matrix-vector pair is not a valid Clifford!"
+    if do_checks:
```
enielse, line 465: Could this just be included in the "if" block above?

enielse (changes requested):
Minor change requested, otherwise it looks great!

pygsti/tools/symplectic.py
sserita: Excellent point... done!

pygsti/tools/symplectic.py
sserita: Not a problem at all, that's a good point. Changed it back.

enielse (approved):

PR #181, opened 2021-05-11
Feature qasm conversion updates

tjproct:
Very small, local changes. Adds "Gczr" as a standard gate (a controlled Z(theta) rotation gate), adds the ability to automatically convert circuits containing this gate into QASM, and adds some details to QASM-conversion-related functions' docstrings.


enielse (approved):
Looks good!

PR #182, opened 2021-05-12
Feature noiseop spamvec

sserita:
Several changes designed to allow general noisy operations after state prep/before measurement.

- New ComposedSPAMVec/POVM that takes a general LinearOperator and a ComputationalSPAMVec/ComputationalBasisPOVM
- CHPForwardSimulator updated to handle ComposedSPAMVec/POVM (rather than LinearOperator/ComputationalBasisPOVM)
- create_crosstalk_free_model updated to allow Composed instead of Lindblad for SPAMVec/POVM

Currently, model construction avoids the use of TensorProdSPAMVec/POVM as it does not set parameters properly when used with a ComposedSPAMVec of ComposedOps.

### DIFF
--- a/pygsti/construction/modelconstruction.py
+++ b/pygsti/construction/modelconstruction.py
@@ -1253,11 +1253,7 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
     else:
         global_idle_op = None
 
-    # TODO: While TensorProdSPAMVec does not work with ComposedSPAMVec, build up prep/povm as n-qubit ops first
-    # Need qubit labels to make the right EmbeddedOps
-    if qubit_labels is None:
-        qubit_labels = range(num_qubits)
-    
+    # TODO: While TensorProdSPAMVec does not work with ComposedSPAMVec, build up prep/povm as n-qubit ops first    
     prep_layers = {}
     if 'prep' in all_keys:
         # rho_base1Q = _spamvec.ComputationalSPAMVec([0], evotype, 'prep')
@@ -1279,7 +1275,7 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
         # TODO: Could do qubit-specific by allowing different err_gate1Q
         err_gates = [err_gate1Q.copy() for i in range(num_qubits)] if independent_gates else [err_gate1Q,] * num_qubits
         err_gateNQ = _op.ComposedOp([
-            _op.EmbeddedOp(qubit_labels, [qubit_labels[i]], err_gates[i]) for i in range(num_qubits)])
+            _op.EmbeddedOp(range(num_qubits), [i], err_gates[i]) for i in range(num_qubits)])
         # TODO: Could specialize for LindbladOps, although this should also handle that case
         prep_layers['rho0'] = ComposedSPAMVec(rho_baseNQ, err_gateNQ, 'prep')
     else:
@@ -1306,9 +1302,9 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
         # TODO: Could do qubit-specific by allowing different err_gate1Q
         err_gates = [err_gate1Q.copy() for i in range(num_qubits)] if independent_gates else [err_gate1Q,] * num_qubits
         err_gateNQ = _op.ComposedOp([
-            _op.EmbeddedOp(qubit_labels, [qubit_labels[i]], err_gates[i]) for i in range(num_qubits)])
+            _op.EmbeddedOp(range(num_qubits), [i], err_gates[i]) for i in range(num_qubits)])
         # TODO: Could specialize for LindbladOps, although this should also handle that case
-        prep_layers['Mdefault'] = ComposedPOVM(err_gateNQ, Mdefault_baseNQ)
+        povm_layers['Mdefault'] = ComposedPOVM(err_gateNQ, Mdefault_baseNQ)
     else:
         povm_layers['Mdefault'] = _povm.ComputationalBasisPOVM(num_qubits, evotype)
 
--- a/pygsti/construction/modelconstruction.py
+++ b/pygsti/construction/modelconstruction.py
@@ -1010,7 +1010,8 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
                                 depolarization_parameterization='depolarize', stochastic_parameterization='stochastic',
                                 lindblad_parameterization='auto', availability=None, qubit_labels=None, geometry="line",
                                 evotype="auto", simulator="auto", on_construction_error='raise',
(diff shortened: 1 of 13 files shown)


pygsti/construction/modelconstruction.py
```diff
@@ -1254,52 +1253,58 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
...
+        #                            stochastic_error_probs, lindblad_error_coeffs,
+        #                            depolarization_parameterization, stochastic_parameterization,
+        #                            lindblad_parameterization)
+        # if isinstance(err_gate, _op.LindbladOp):
+        #     prep1Q = _spamvec.LindbladSPAMVec(rho_base1Q, err_gate, 'prep')
+        # else:
+        #     prep1Q = ComposedSPAMVec(rho_base1Q, err_gate, 'prep')
+        # prep_factors = [prep1Q.copy() for i in range(num_qubits)] if independent_gates else [prep1Q] * num_qubits
+        # prep_layers['rho0'] = _spamvec.TensorProdSPAMVec('prep', prep_factors)
+        rho_baseNQ = _spamvec.ComputationalSPAMVec([0,]*num_qubits, evotype, 'prep')
```
enielse, line 1270: Do we want to still have the option of having a TensorProdSPAMVec here?  It's probably more performant than a composition of embedded ops, and maybe should be used in cases when it works and when the user desires.  I think it should work whenever the factor spam vectors have a definite num_params, and we can probably fix to work in general.

pygsti/construction/modelconstruction.py
```diff
@@ -1254,52 +1253,58 @@ def create_crosstalk_free_model(num_qubits, gate_names, nonstd_gate_unitaries={}
...
+        #                            stochastic_error_probs, lindblad_error_coeffs,
+        #                            depolarization_parameterization, stochastic_parameterization,
+        #                            lindblad_parameterization)
+        # if isinstance(err_gate, _op.LindbladOp):
+        #     povm1Q = _povm.LindbladPOVM(err_gate, Mdefault_base1Q, "pp")
+        # else:
+        #     povm1Q = ComposedPOVM(err_gate, Mdefault_base1Q)
+        # povm_factors = [povm1Q.copy() for i in range(num_qubits)] if independent_gates else [povm1Q] * num_qubits
+        # povm_layers['Mdefault'] = _povm.TensorProdPOVM(povm_factors)
+        Mdefault_baseNQ = _povm.ComputationalBasisPOVM(num_qubits, evotype)
```
enielse, line 1297: Same comment as above - should there be an option to create a TensorProdPOVM?

pygsti/objects/localnoisemodel.py
```diff
@@ -717,6 +717,8 @@ def __init__(self, num_qubits, gatedict, prep_layers=None, povm_layers=None, ava
         #SPAM (same as for cloud noise model)
         if prep_layers is None:
             pass  # no prep layers
+        elif isinstance(prep_layers, _sv.SPAMVec): # just a single SPAMVec
```
enielse, line 720: Does the docstring state that giving a single spamvec is allowed behavior?  If not, we should add it.

pygsti/objects/operation.py
```diff
@@ -430,8 +430,10 @@ class LinearOperator(_modelmember.ModelMember):
 
     def __init__(self, rep, evotype):
         """ Initialize a new LinearOperator """
-        if isinstance(rep, int):  # For operators that have no representation themselves (term ops)
-            dim = rep             # allow passing an integer as `rep`.
+        # For operators that have no representation themselves (term ops)
+        # allow passing an integer as `rep`.
+        if isinstance(rep, int) or isinstance(rep, _np.int64) or rep == _np.inf:
```
enielse, line 435: Why do we need _np.inf here?

pygsti/objects/spamvec.py
```diff
@@ -2452,6 +2452,7 @@ def from_vector(self, v, close=False, dirty_value=True):
         """
         if self._prep_or_effect == "prep":
             for sv in self.factors:
+                print("Have gpindices: ", sv.gpindices, " v: ", v)
```
enielse, line 2455: Is this just a debug statement?  Maybe remove it?

pygsti/objects/operation.py
```diff
@@ -986,6 +988,45 @@ def transform_inplace(self, s):
         Smx = s.transform_matrix
         Si = s.transform_matrix_inverse
         self.set_dense(_np.dot(Si, _np.dot(self.to_dense(), Smx)))
+    
+    def spam_transform_inplace(self, s, typ):
```
enielse, line 992: What are these spam_transform_inplace updates needed for?  Is it possible to save them until later to ease merging with other planned updates?

enielse (commented):
Overall this looks good.  I'd like to understand the changes to operation.py, however, and wonder if we can postpone some of them until after the upcoming large merge?

pygsti/construction/modelconstruction.py
sserita: I originally just intended this to be a temporary measure, but in hindsight, it's probably a good call for everything coming into develop to have good default behavior.

I've now added a `tensorprod_spamvec_povm` kwarg, controlling which construction is used. The default is the old TensorProd construction, and I can just use False in my runs until we get a proper fix for the TensorProd variants.

pygsti/construction/modelconstruction.py
sserita: Same solution as above.

pygsti/objects/spamvec.py
sserita: Nice catch, removed.

pygsti/objects/operation.py
sserita: I believe these are to enable gauge transformations. These are available in LindbladOp for the LindbladSPAMVec/POVM, so I piped it up to the base class.

However, these can definitely be saved for later. I've removed them for now.

pygsti/objects/operation.py
sserita: This is a workaround for a bug when creating models with 100s of qubits. I think somewhere in StateSpaceLabels, int overflows are handled by setting the dim to np.inf.

I've removed the change here to make merge with operation.py easy, and will probably start a CHP-specific branch where I start to work around these sorts of issues in less hacky way.

pygsti/objects/localnoisemodel.py
sserita: It actually already was but was vague as stated. I modified the docstring to include the allowed types (SPAMVec or POVM).

Note that technically, LocalNoiseModel can still take a single LinearOperator as a prep. I didn't change this behavior, although CHP no longer uses this code and now takes SPAMVecs.

enielse:
I've taken the content from this PR and manually merged it into a new branch & PR that is based on the post-massive-refactor that occurred.  Closing this PR in favor of #192

PR #185, opened 2021-05-25
Guard against empty circuit lists in dataset load.

sserita:
Fixes a bug Tim encountered in the DirectRB tutorial,
where circuits with no outcome:count pairs could not be reshaped
during dataset loading.


enielse (approved):
Good catch!

PR #191, opened 2021-06-10
Better seeding for RBDesign generation.

sserita:
Replace all global instances of np.random calls with a RandomState.

Fixes a bug introduced in PR #179 where parallel RBDesign construction seeded each worker the same
(resulting in multiple copies of the same circuit from each worker).

Also adds seeding capabilities for Mirror RB, simultaneous direct RB,
and basic unit tests for RBDesign construction (serial and parallel).


enielse (approved):
This all looks good.  The only concern I have is about the seed=1234 default option, which seems like it could inadvertently lead to the seed never being set by the user when maybe it should be.  Not a big deal.

sserita:
Yes I thought about that... Would you prefer `seed=None` and falling back to like `time.time()` or something?

PR #194, opened 2021-07-07
Matthew's fix for new Cython compiling on Linux

sserita:


enielse (approved):
Looks fine!

enielse (commented):
This looks pretty innocuous, but we're getting a test failure that maybe (?) is related.  Let's try to get the unit tests to pass before merging this in.

sserita:
Just following up on this: Seems like the cython just wasn't compiling before. Now that it is, there are some linking errors (not only for statevec evotype). Still looking into it.

PR #199, opened 2021-08-05
Feature vb analysis

tjproct:
Code for analyzing VB data, designed primarily for replicating the data analysis presented in https://arxiv.org/abs/2008.11294.  Code has been tested on the data from that paper and is working. However, there are no unit tests and not all docstrings are detailed. This will be fixed on a new branch in the near future.


enielse (approved):
Looks good!  The only potentially disruptive change that I see is the change in the way subdata outcome labels are computed for a combined experiment design - but it seems ok for now (no tests fail).

PR #202, opened 2021-08-12
WIP: Merge branch 'develop' into feature-tutorial-evotype-fixes

sserita:
This branch will eventually fix the tutorials post-refactor, and should be completed before the next release.

Here's a few other small things that should be discussed/resolved that may impact how the tutorials are updated.

- [ ] create_processor_spec() for ExplicitOpModels doesn't work with create_*_model because the unitary is on the whole space but the label sslbls are usually subsystems. One possible solution is not using sslbls but the whole state_space for explicit models, but then name would have to be adjusted to differentiate between same name, different sslbls?
- [ ] The behavior for create_crosstalk_free_model with independent_gates=False seems inconsistent. If I specify Gxpi2 in my pspec, and specify errors on only (Gxpi2, 0), I would have expected all Gxpi2 gates to share the same StaticStandardOp and have different appended noise ops. This doesn't seem to be the case, i.e. _setup_local_gates is not taking this into account. Maybe it doesn't matter since it's a Static op so there are no shared parameters, but one could imagine wanting to reparameterize the ideal gate later or specify nonstandard gates that do have parameters.

### DIFF
--- a/pygsti/forwardsims/termforwardsim_calc_generic.py
+++ b/pygsti/forwardsims/termforwardsim_calc_generic.py
@@ -244,7 +244,7 @@ def prs_directly(fwdsim, rholabel, elabels, circuit, repcache, comm=None, mem_li
 
 
 def refresh_magnitudes_in_repcache(repcache, paramvec):
-    from ..opcalc import bulk_eval_compact_polynomials_complex as _bulk_eval_compact_polynomials_complex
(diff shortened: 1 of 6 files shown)


sserita:
Both of the above issues will be handled in other PRs, first may already be fixed and second will be fixed in the model serialization branch. This WIP PR is not needed to track these issues anymore - similar issues should be tracked in Issue #204 .

PR #203, opened 2021-08-13
Feature nonclifford mirror rb

tjproct:
This pull request should update develop so that the RB code is sufficiently fixed/improved for the next release. Changes:

- Adds non-clifford mirror RB, and non-clifford circuit mirroring code (written by Jordan Hines).
- Updates random circuit sampling and RB code to integrate that non-Clifford mirror RB code.
- Changes the "edgegrab" sampler so that it's argument is the two-qubit gate density, not the mean number of two-qubit gates. This change can silently alter the behavior of old RB code (although it will typically break that code), but this update in convenient makes the code easier to use (and matches our published "edgegrab" algorithm).
- Removes (comments out) a bunch of random circuit sampling code that has not been updated and is not, to my knowledge, in use by anyone.
- Adds tests for the RB and random circuit sampling code.
- Updates the RB tutorials so that they work.
- Adds a tutorial that demonstrates how to run mirror circuit benchmarks similar to those in "Measuring the Capabilities of Quantum Computers".
- Misc bug fixes and additional functionality throughout pyGSTi.

No doc strings have been written/updated, and ideally this will be done on a new branch before the next release.


enielse (approved):
Looks good.  There are a number of functions without docstrings that we should probably document before the next release - in particular in randomcircuit.py

PR #221, opened 2021-12-08
Feature comprules enhancements

sserita:
This branch contains several additions to the base CompilationRules object.

- Initialization from a dictionary is now hooked up
- A `retrieve_compilation_of` function mirroring CliffordCompilationRules, which allows construction from both specific compilations and local templates
- Convenience function to apply CompilationRules to a circuit list

Also, there is a separate and potentially questionable change to `Circuit.change_gate_library`, where it is now possible to provide compilations for the empty layer. The plan was allow me to turn implicit idles in an empty layer into an explicit idle layer. These seems to work, but I can revert if needed.


pygsti/processors/compilationrules.py
```diff
@@ -227,7 +298,8 @@ def apply_to_processorspec(self, processor_spec, action="replace"):
                 # create boolean oracle function for availability
                 def _fn(sslbls):
                     try:
-                        self.function_templates[gn](sslbls)  # (returns a circuit)
+                        lbl = _Label(gn, sslbls)
```
enielse, line 301: This changes the format for a "function template" within a set of compilation rules -- previously it took just state space labels as an argument; now it takes an entire label.  This is noted in the comment above, but is 
this necessary?  If seems like it would be necessary iff a single template could be used to construct compilations for multiple gates, but the logic here seems counter to this.

enielse (commented):
All good except change in function template call signature (see line comment).  Maybe this is fine, but I don't understand why the additional complexity (a Label is more complex than just a set of state space labels) is needed.

pygsti/processors/compilationrules.py
sserita: It's been a while but I want to say this was because I needed access to the Label.args? E.g. compiling Gu3 gates down to ZXZXZ, basically each Z rotation gets something from each arg in the Gu3, and just having sslbls was not enough.

enielse: Ah, I see.  But at this location there's no way for additional args to make it in, so it must be somewhere else this is needed?  In any case, maybe we want to have the function template take an additional `args` argument instead of a label, so its signature would be `fn(sslbls, args)` or `fn(sslbls, args, time)`?

sserita: So I think I need the arg capabilities up above on line 212, where I actually pass the label (and am just matching label based on name).
This change is just to maintain compability with the new function template (that takes a Label, so I construct one from the sslbl).

sserita: I've updated the function prototype to be `fn(sslbls, args, time)` now. Note that this requires the user to specify a function that takes all 3. I can technically drop the second two args in this case and assume the user has kwarg'ed them to default to `None`, but it needs to take all three args in the earlier usecase. (or we could try to infer the signature from the contents on the Label, but I'm not sure I like that).

enielse (approved):
Looks great!

PR #222, opened 2021-12-09
Update interpygate regression tests

sserita:
Adjust the numerical comparisons in  interpygate regression tests for the combination of new interpolation + Brandon's bugfix.


enielse (approved):
Great!

PR #224, opened 2022-01-06
use unicode to specify quotes in css file

eendebakpt:
The quotes tag `<q>` is modified in the file `pygsti_dashboard.css`. In that file special characters are used for the quotes, which render incorrectly in the browser:

image

This PR defines the quotes using unicode, which renders the quotes as intended:

image

@enielse

### DIFF
--- a/pygsti/report/templates/offline/pygsti_dashboard.css
+++ b/pygsti/report/templates/offline/pygsti_dashboard.css
@@ -53,16 +53,16 @@ code {
   display: inline-block;
 }
 q:before {
-  content: "“";
+  content: "\201c";
 }
 q:after {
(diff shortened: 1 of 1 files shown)


enielse:
@eendebakpt, thanks for this!  I had noticed the quotes were getting misrendered for some time now but never figured this out.  Unfortunately, in order for me to accept your PR you'd need to fill out a contributor license agreement (CLI).  Since this PR only changes several lines, and to save us both paperwork, I've just incorporated your suggested edits into commit 88d74a771031b8da668d6600a39f43250b52ab19 (and credited you there).  I hope this is satisfactory to you.

PR #225, opened 2022-01-11
Feature windows actions

sserita:
Runs unit tests on Windows in GitHub Actions and fixes a number of int32/64 bugs for Windows.

After merging this PR, the bugfix-windows-integers branch should also be closed.

Note that this also turns on Windows testing for the integration tests that run on the beta branch. We should expect possible failure of those once this is pushed to beta.


enielse (approved):
Looks good!

PR #227, opened 2022-02-02
Bugfix lsgst ckt gen

coreyostrove:
This adds a two line fix to the lsgst circuit generation routine that resulted in the empty-germ along with a full set of corresponding fiducials being added to the experiment design even when the include_LGST flag was turned off. I also discovered today another strange side-effect of this bug that happened even when the include_LGST flag was on which resulted in the plaquette for the empty germ being labeled with the wrong germ attribute.


pygsti/circuits/gstcircuits.py
```diff
@@ -560,25 +567,38 @@ def add_to_plaquettes(pkey_dict, plaquette_dict, base_circuit, maxlen, germ, pow
...
+                #print('Germ Power: ', germ_power)
                 power = len(germ_power) // len(germ)  # this *could* be the germ power
+                #print('Germ x Power: ', germ*power)
                 if germ_power != germ * power:
                     power = None  # Signals there is no well-defined power
-
+                #print('Power: ', power)
+                
+                if power==0 and len(germ) !=0:
+                    continue
```
sserita, line 589: Been a while since we thought about this. I do recall something like this being the solution we settled on, and I don't see how it can introduce issues to default LSGST construction. If all tests pass, I say this is fine.

sserita (commented):
Assuming tests pass, I think this is fine. Will let Erik give final approval.

enielse (approved):
Tests pass and everything looks good.  We should probably remove the commented out print statements in gstcircuits.py sometime in the future.

PR #236, opened 2022-02-21
Feature imbq job limit

sserita:
This branch has a set of features I found useful while running LCU mirror experiments on IBMQ.

The major features include:

- Bugfix for IBMQExperiment dump/load
- Exiting from submit() when the concurrent job limit is reached
- U3 gate decompositions available in the openqasm translation (required nontrivial rework)
- An extra check for Labels without args/time in ErrorRatesModels, i.e. ErrorRatesModels with Gu3 gates now work


tjproct:
Stefan, as we briefly discussed, can you (or you and me) first merge "feature-mirror-benchmarks-bugfixes" into this branch? I made some similar changes previously on that branch, and did misc other things that I only half remember. I'm happy to help coordinate that merge, e.g., perhaps you could first take a look and then we can discuss if needed.

sserita:
Yes I will merge that branch into this before we complete the PR. Will reach out if I have any difficulties.

sserita:
feature-mirror-benchmarks-bugfixes is now merged into this branch, only one conflict (where Tim and I had decided on the same fix).

We should delete that branch when we accept this PR.

pygsti/forwardsims/forwardsim.py
```diff
@@ -886,10 +886,40 @@ def _bulk_fill_probs_block(self, array_to_fill, layout):
                                                                    outcomes, layout.resource_alloc(), cache, time=None)
 
     def _bulk_fill_dprobs_block(self, array_to_fill, dest_param_slice, layout, param_slice):
-        for element_indices, circuit, outcomes, cache in layout.iter_unique_circuits_with_cache():
-            self._compute_circuit_outcome_probability_derivatives_with_cache(
-                array_to_fill[element_indices, dest_param_slice], circuit, outcomes, param_slice,
-                layout.resource_alloc(), cache)
+
```
enielse, line 889: This finite differencing alternative should be placed elsewhere.  The MapForwardSimulator does finite differencing, but maybe this was placed here for an op-less model (in which case it might go in SuccessFailForwardSimulator).  In any case, this is ugly and finite differencing is already implemented elsewhere.  We should either call an existing implementation or decide to move all finite differencing to the ForwardSimulator base class, which I don't expect we'll want to do.

enielse (changes requested):
Everything looks good except hacked fallback to finite differencing in the ForwardSimulator class.  I'm not sure exactly what use case this is needed for, but we should discuss an alternative solution.

pygsti/forwardsims/forwardsim.py
sserita: I want to say this hacked in for the opless model, so should probably go to the SuccessFailForwardSimulator. I'll go ahead and make that change.

enielse: Ok, cool.  I expect the opless model might have finite differencing somewhere in it already -- so maybe check first :)

pygsti/algorithms/randomcircuit.py
```diff
@@ -72,8 +81,37 @@ def sample_compiled_haar_random_one_qubit_gates_zxzxz_circuit(pspec, zname='Gzr'
     circ.done_editing()
     return circ
 
+def sample_compiled_random_clifford_one_qubit_gates_zxzxz_circuit(pspec, zname='Gzr', xname='Gxpi2', qubit_labels=None):
+    """
+    TODO: docstring  #generate layer of random unitaries and make a series of circuit layers with the compiled versions
+     of these
+    """
+    if qubit_labels is not None:
```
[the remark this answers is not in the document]
sserita: We may actually want to use a subset of the qubits in this routine. Using pspec.qubit_labels is the fallback if this is None.

pygsti/forwardsims/forwardsim.py
sserita: Actually SuccessFailForwardSimulator actually implements _compute_circuit_outcome_probability_derivatives_with_cache(), and down the stack the default is to use finite difference. So maybe my recollection is wrong...

Either way I'm reverting this change and if it breaks something in an RB workflow somewhere, we can find it and fix in the correct spot later.

enielse (approved):
Looks great!

PR #243, opened 2022-05-04
Fixed bug in circuit mirroring for cz(theta) circuits

tjproct:
Fixed a bug in circuit mirroring of circuits containing Gczr(theta) gates, which created incorrect mirror circuits whenever there were parallel two-qubit gates in a layer. Created non-definite-outcome circuits, so will have created obviously faulty data for anyone who used it.


sserita (approved):
Just a minor fix, looks good!

PR #259, opened 2022-08-23
Update Deprecated Jinja Calls in Report Generation

coreyostrove:
This set of commits updates the call signature for some jinja function decorators which were deprecated and removed from the latest releases. Also updates package names and package requirements accordingly.


sserita (approved):
Looks good to me!

PR #261, opened 2022-08-30
FPR Bugfixes and Feature Enhancements

coreyostrove:
The changes in this branch include the following:

1. Fixes to the EigenvalueParamDenseOp to include as parameters all of the off-diagonal elements of the kite. This fix is of independent interest beyond FPR.
2. Modifications and feature enhancements for per-germ and per-germ power FPR.
3. Some minor bugfixes to the fisher information calculations in edesigntools (probably should have been made on a separate branch, sorry!).

Since github has indicated that fully automatic merging won't work, my apologies in advance if this pull request is a pain to incorporate into develop.


coreyostrove:
I just took a look at the logs to see which unit tests were failing and it looks like it is the test for verifying a certain set of fiducials is generated by the FPR algorithm. It looks like this is hardcoded in at the moment and I'm guessing that one of my changes made it so a different set of fiducials is returned now. Should be an easy thing to fix.

sserita:
Ok, thanks for looking into the tests. I'll take a look at this more closely when everything passes.

pygsti/algorithms/fiducialpairreduction.py
```diff
@@ -680,97 +945,374 @@ def _get_number_amplified(m0, m1, len0, len1):
...
+                                 gsGerm, power, mem_limit, printer, search_mode, seed, n_random, dof_per_povm, 
+                                 min_iterations=1, lowest_eigenval_tol=1e-1,
+                                 candidate_set_seed=None, num_soln_returned=1, type_soln_returned='best'):
     #Get dP-matrix for full set of fiducials, where
     # P_ij = <E_i|germ^exp|rho_j>, i = composite EVec & fiducial index,
     #   j is similar, and derivs are wrt the "eigenvalues" of the germ
     #  (i.e. the parameters of the gsGerm model).
+    
+    #debugging
+    print('Entered helper function _get_per_germ_power_fidpairs')
```
sserita, line 958: Maybe comment out these debug statements? In this whole function actually

pygsti/algorithms/fiducialpairreduction.py
```diff
@@ -290,11 +298,12 @@ def _get_number_amplified(m0, m1, len0, len1, verb):
                       for iEStr in range(nEStrs)]
     return listOfAllPairs
 
-
+#@_deprecated_fn('find_sufficient_fiducial_pairs_per_germ_power')
```
sserita, line 301: I thought we did end up deprecating per germ power? Or are we keeping it for the per-germ power random "robustness" feature

pygsti/tools/edesigntools.py
```diff
@@ -288,6 +290,23 @@ def calculate_fisher_information_matrices_by_L(model, circuits, num_shots=1, ter
 
     fisher_information_by_L = {}
     prev_L = None
+    
+    #Check whether we need to add in addtional circuits from _additonal_circuit by checking if it is empty.
+    if circuits._additional_circuits:
```
sserita, line 295: I forgot about this bugfix, thanks for catching it!

sserita (changes requested):
Just take out some debug print statements and I think we're basically there.

pygsti/algorithms/fiducialpairreduction.py
coreyostrove: Some of these are debug statements and others are for progress logging. I'll go ahead and comment out the debug ones and swap the progress logging with calls to one of pygsti's VerbosityPrinter objects with an appropriate verbosity flag so they only show up when desired.

pygsti/algorithms/fiducialpairreduction.py
coreyostrove: Good question, I'm not sure if the right move is deprecation or not here. The justification for keeping it around would be the robustness conjecture, but even that claim hasn't actually been tested yet.

pygsti/algorithms/fiducialpairreduction.py
coreyostrove: I've made the aforementioned logging changes.

coreyostrove:
Following up again on this endless saga of unit testing. As of now all of the unit tests are passing on ubuntu, and so are all of the unit tests related to FPR on windows . The remaining failing tests look to be related to Qibo integration, and the errors only show up on python 3.8+ for some reason. (See the actions logs for commit  3bf5e48 and 1654a70) I can look into this if you'd like (though, @enielse would probably be better suited to do so since it involves Qibo). Either way any unit test fixes related to Qibo are probably out of scope for this PR, and I'd suggest we stick those on a different bugfix branch.

sserita:
Thanks for your effort on this Corey. I'll do a full review of this PR again later this week, but my 30 second skim does leave me mildly concerned about the Qibo tests. At a high-level, it just seems like a kwarg mismatch - however, we have actions post-Qibo merge where these don't occur ([first example](https://github.com/pyGSTio/pyGSTi/runs/7813439521)).
What this means to me is maybe there was an issue merging develop into this branch? Or at least, I would like to double check that this was not the case before final approval.

coreyostrove:
I was re-running the CI jobs individually to confirm my assertion that the ubuntu jobs were still passing after the latest commits, but they failed on the qibo tests now which I found strange since there were no changes in those commits that could've possibly affected these tests (a kwarg update on a tutorial notebook and an update to the FPR unit tests). Long story short, the qibo devs pushed a new release to pypi on 9/3, and in this new release the seed kwarg has been removed from all of the Channel objects. There wasn't any explanation on the corresponding commit (SHA d182b9b if you're curious) on the rationale for this change, unfortunately.

On our end I'm not familiar with the qibo evotype code, so I'm not sure how essential this functionality was. Either way we'll need to have a separate discussion on how to proceed. I'm guessing we can either update the evotypes on our end to work around this or else update the requirements to use the previous release at the expense of whatever other new features/enhancements were included in this new one. Either way this is outside the scope of this PR so I'll open up an issue thread so we can migrate this discussion there.

sserita:
I see. I didn't realize it was a Qibo-side change. Thanks for looking into it more in-depth than I did.

Yes, in that case, I agree that it should be moved to a separate thread and either Erik or I will handle it. Incidentally, the proper location of RNG seeding is a problem we also have with other evotypes (e.g. CHP), so it's interesting to see that Qibo changed their interface...

jupyter_notebooks/Tutorials/algorithms/advanced/GST-FiducialPairReduction.ipynb
```diff
@@ -11,9 +11,11 @@
...
     "\n",
-    "- **Global fiducial pair reduction (GFPR)** removes the same intelligently-selected set of fiducial pairs for all germ-powers.  This is a conceptually simple method of reducing the operation sequences, but it is the most computationally intensive since it repeatedly evaluates the number of amplified parameters for en *entire germ set*.  In practice, while it can give very large sequence reductions, its long run can make it prohibitive, and the \"per-germ\" reduction discussed next is used instead.\n",
+    "- **Global fiducial pair reduction (GFPR)** removes the same intelligently-selected set of fiducial pairs for all germ-powers.  This is a conceptually simple method of reducing the operation sequences, but it is the most computationally intensive since it repeatedly evaluates the number of amplified parameters for en *entire germ set*.  In practice, while it can give very large sequence reductions, its long run can make it prohibitive, and the \"per-germ\" reduction discussed next is used instead. \n",
```
sserita, line 14: Just bookmarking for us, but we may want to deprecate/underemphasize global FPR... It's slow and not any better than per-germ.

pygsti/algorithms/fiducialpairreduction.py
sserita: We can keep it for now, and consider testing robustness vs deprecation later.

sserita (approved):

enielse (commented):
Looks great!

Issue #265, opened 2022-10-04
QubitProcessorSpec no longer encoding `availability` correctly


tjproct:
On develop (and probably master), when a QubitProcessorSpec is created with `availability` specified rather than `geometry`, the QubitGraph stored doesn't contain the connectivity information -- it's an empty graph.

pspec = pygsti.processors.QubitProcessorSpec(2, ['Gcnot',], qubit_labels=['Q0','Q1'],
                                             availability={'Gcnot':[('Q0','Q1'),]})
pspec.qubit_graph.edges()
# This will show an empty list

However, the processor's QubitGraph should inherit from the `availability` dict --- or at least this used to be what the intended behavior was, and some code in pyGSTi (the CRB and DRB code at least) relies on this and now fails with an obscure failure message at the Clifford compilation stage. I could fix this, but we should agree on what's meant to happen.

coreyostrove:
Thanks for reporting this, Tim. I'm fairly certain that this problem is essentially the same as was reported in #251. It doesn't look like we have a fix for that yet though.

tjproct:
Ah, I think you're correct Corey. #251 is a long and confusing thread, but I think its conclusion more-or-less amounts to this.

enielse:
This a more complicated issue than it might seem.  The `geometry` argument, which can be either a graph object or a string identifying a known graph (like `"line"`) was originally intended to provide an alternative and *more limited* way of initializing the `availability` dictionary within a processor spec, but not actually being a public attribute of the processor spec.  As such, the availability dictionary should be considered the authoritative store of the "which gates are available on which qubits" information, and the qubit graph as optional side-information whose sole purpose is to help define the `availability` attribute.  In this picture, the behavior noted above is expected: if you specify the `availability` directly then there's no need to initialize the processor spec's `qubit_graph` attribute.

Ok, so that was the original idea.  It seems reasonable and useful to me for the processor spec to hold some kind of qubit graph inside it, but we'll need to first figure out what this object should be in general.  When the processor spec has just a single 2-qubit gate it seems intuitive (to me) that the edges of this graph would match the availability of this gate.  But what if there are multiple 2Q gates with different availability?  Would there be a graph for each 2Q gate?  And what if there are 3Q gates (e.g. Toffoli)?  It seems like it's worth doing something sensible that reduces to the single-2Q-gate case nicely, but I'd advocate for discussing and documenting what this is before we try to patch this issue as stated.

[...]

PR #266, opened 2022-10-04
Update to devices.devcore to use a QubitGraph rather than an availability dict

tjproct:
This circumvents a bug, currently in develop, whereby `availability` dicts aren't propagated to a QubitProcessorSpecs's QubitGraph. Although this won't need to change if that bug is fixed, this is probably a more robust way to do things.


coreyostrove:
This looks good to me at first glance, but could you try merging develop into this branch? I took a skim through the unit testing logs and I think all of these errors are things that have been fixed in develop (most of them are due to calls to a deprecated numpy function which Erik fixed a couple months ago), but would want to make certain of that.

tjproct:
Ah yes, hadn't pulled develop in a long time. Latest develop merged in. Thanks Corey!

coreyostrove:
Thanks for merging that in. It looks like almost all of those errors were indeed fixed already on develop. The only remaining one I can see is TestDirectRBDesign on Windows with python 3.6 (3.7 works and I can't re-run the tests on 3.8 or 3.9 but they presumably work too). I'm guessing you may have written this unit test, so normally I'd ask that you look into it, but I've run into these sorts of OS/python version combination incompatibilities before and they are usually for obscure and tricky to track down reasons. So I don't mind looking into it first and pinging you if I can't figure it out.  

That said, given the discussion in #265 do you want to table this PR temporarily in order to incorporate any related refactors?

sserita (approved):

PR #267, opened 2022-11-07
Feature ddist wildcard integration

coreyostrove:
This integrates the ability to generate the single-parameter diamond distance wildcard models introduced in this paper (https://arxiv.org/pdf/2207.08786.pdf) and include visualizations of these models in pygsti reports. 

I have updated the docstring for GSTBadFitOptions to include an explanation of this new badfit action. This docstring still needs some TLC though, so a hopefully low-effort request for @enielse: any chance you could take a crack at filling in the rest of the blanks on that docstring before this gets merged into develop? I added spots for all of the undocumented options, but I don't actually know what most of them do.


coreyostrove:
RE: failed unit tests. Took a look at the logs and the four failed tests are all due to that qibo interface change we fixed. This branch was forked off of master so doesn't have that fix integrated. I'd merge that in and let the tests re-run, but since we deleted the bugfix branch I think we should just assume we'll be good to go once this is merged into develop where the fix exists.

coreyostrove:
Looks like everything is passing now.

pygsti/protocols/gst.py
```diff
@@ -1869,6 +1914,182 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
...
+    return _ddist_wildcard_model(alpha, ddists)
+    
+def _bisect_alpha(guess, tol, mdc_objfn, two_dlogl_threshold, redbox_threshold, 
+                  ddists, critical_percircuit_budgets):
+    
+    left = None
+    right = None
+    
+    while left is None or right is None:
+        print(f'Searching for interval [{left}, {right}] with guess {guess}')
```
sserita, line 1997: We probably want to make this just debugger/info logging rather than print all the time.

pygsti/protocols/gst.py
```diff
@@ -1869,6 +1914,182 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
...
+    ds_dof = ds.degrees_of_freedom(global_circuits_to_use)
+    nparams = model.num_modeltest_params  # just use total number of params
+    percentile = 0.025
+    nboxes = len(global_circuits_to_use)
+
+    two_dlogl_threshold = _chi2.ppf(1 - percentile, max(ds_dof - nparams, 1))
+    redbox_threshold  = _chi2.ppf(1 - percentile / nboxes, 1)
+    
+    # Not in cache, recompute
+    ddists = _compute_ddists(estimate)
```
sserita, line 1979: Just point for discussion for tomorrow's dev meeting: I could imagine the user wanting to do this sort of scaling single-param wildcard with something other than the diamond distance. Is it worth plumbing that in now? Could be as simple as adding a kwarg to this function, and we can leave the rest of the plumbing for when we need it?

sserita (approved):
I'm approving this because I think it looks fine, but there are still some minor things we could clean up before merging it in (print statements and docstrings).

pygsti/protocols/gst.py
coreyostrove: Good point, just pushed a commit implementing that.

pygsti/protocols/gst.py
```diff
@@ -556,7 +610,7 @@ def __init__(self, threshold=DEFAULT_BAD_FIT_THRESHOLD, actions=(),
                  wildcard_L1_weights=None, wildcard_primitive_op_labels=None,
                  wildcard_initial_budget=None, wildcard_methods=('neldermead',),
                  wildcard_inadmissable_action='print'):
-        valid_actions = ('wildcard', 'Robust+', 'Robust', 'robust+', 'robust', 'do nothing')
+        valid_actions = ('wildcard', 'ddist_wildcard', 'Robust+', 'Robust', 'robust+', 'robust', 'do nothing')
```
enielse, line 613: Maybe '1d_wildcard'?

pygsti/protocols/gst.py
```diff
@@ -1869,6 +1949,181 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
...
+    
+def _test_feasible_alphas(alphas, mdc_objfn, two_dlogl_threshold, redbox_threshold, ddists, critical_percircuit_budgets):
+    
+    if not isinstance(alphas, list):
+        alphas = [alphas]
+    
+    # Iterate through alphas and compute wildcard feasibilities
+    feasible = []
+    for alpha in alphas:
+        wcm = _ddist_wildcard_model(alpha, ddists)
```
enielse, line 2075: Just a note: we could avoid creating a new budget object on each iteration here if we wanted/needed the performance.  I think this is fine for now - just an opportunity for later on if performance becomes an issue.

enielse (approved):
Looks good to me on the whole.  We still may want to change the `ddist` name in various places and hardcoded diamond distance to be more flexible before merging (diamond distance requires CVXPY in particular, which can be grumpy).

coreyostrove:
Thanks for adding that additional docstring information, @enielse. I think I will work on generalizing this to work with more error metric in preparation for the next release, so I'll wait to merge this into develop until then.

coreyostrove:
Awesome work! This all looks great to me and will vastly simplify the addition of new error metrics down the line.

pygsti/optimize/wildcardopt.py
```diff
@@ -423,6 +423,75 @@ def _proxy_agg_dlogl_hessian(x, tvds, fn0s, percircuit_budget_deriv):
     return H
 
 
+def _get_percircuit_budget_deriv(budget, layout):
```
sserita, line 426: Thanks for refactoring this :)

pygsti/protocols/gst.py
```diff
@@ -2195,7 +2370,7 @@ def _evaluate_constraints(wv):
         # Note: active_constraints_list is typically stored in parameters['unmodeled_error active constraints']
         # of the relevant Estimate object.
         primOp_labels = _collections.defaultdict(list)
-        for lbl, i in budget.primOpLookup.items(): primOp_labels[i].append(str(lbl))
+        for lbl, i in budget.primitive_op_vecindex.items(): primOp_labels[i].append(str(lbl))
```
sserita, line 2373: This line causes tests to fail. Should it just be `primitive_op_index`, to match the new member variable in the base class?
```
======================================================================
1711) ERROR: test_add_badfit_estimates (test.unit.protocols.test_gst.GSTUtilTester)
----------------------------------------------------------------------
   Traceback (most recent call last):
    protocols\test_gst.py line 71 in test_add_badfit_estimates
      gst._add_badfit_estimates(res, 'test-estimate', badfit_opts, opt)
    C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pygsti\protocols\gst.py line 1874 in _add_badfit_estimates
      budget_dict = _compute_wildcard_budget(objfn_cache, mdc_objfn, parameters, badfit_options, printer - 1)
    C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pygsti\protocols\gst.py line 2373 in _compute_wildcard_budget
      for lbl, i in budget.primitive_op_vecindex.items(): primOp_labels[i].append(str(lbl))
   AttributeError: 'NoneType' object has no attribute 'items'
   ```

sserita: Checked this locally and the misnaming appears to be the case. We'll see if the latest set of tests pass.

PR #268, opened 2022-11-07
Bugfix matrixmod2 determinism

enielse:
**Adds the ability to seed Clifford compilations used in RB so deterministic behavior results**

Plumbs a `RandomState` object through functions in `compilers.py` and `matrixmod2.py` so that the randomness used therein can be seeded properly.  This PR also removes a duplicate `done_editing` call in `randomcircuit.py` that I stumbled upon as I was making the main updates.

I think (& hope) that this PR may fix a long-standing issue of the pyGSTi unit tests occasionally failing because of an RB circuit mismatch.  This erroring could be "fixed" by simply rerunning the unit tests again, which was suspicious and unnerving.


coreyostrove:
Took a look through the changes and this all looks good to me, fwiw.

sserita (approved):
All looks good to me too.

PR #270, opened 2022-11-08
Feature germsel enhancements

coreyostrove:
First, and primary, PR for a bunch of branches that revamp the germ selection code. This update adds a number of additional performance enhancements:

- Better deduping routines for decreasing the size of the search space.
- A new hybrid caching scheme that leverages the low-rank nature of the Jacobians for individual germs.
- An alternative implementation of the greedy search that uses this hybrid cache to very quickly evaluate the cost function for candidate germ set additions using low-rank psuedoinverse updates.

Because I practiced terrible branch hygiene (but also in part because some changes to the germ sel code were in service of other algorithms), there are a number of additional updates in the pipeline, specifically on the branches 'feature-fpr-algorithm-performance' and 'feature-fidsel-performance'. This includes better docstrings, logging/exception messages and the removal of a bunch of debugging print statements.


pygsti/algorithms/germselection.py
```diff
@@ -1529,15 +1744,41 @@ def find_germs_breadthfirst(model_list, germs_list, randomize=True,
...
+            for j, idx in enumerate(nonzero_weight_indices):
+                if j==0:
+                    temp_DDD = derivDaggerDeriv[0][idx] @ derivDaggerDeriv[2][idx]
+                else:
+                    temp_DDD += derivDaggerDeriv[0][idx] @ derivDaggerDeriv[2][idx]
+                    
+                #print('temp_DDD shape= ',temp_DDD.shape) 
+            currentDDDList.append(temp_DDD)
 
     else:  # should be unreachable since we set 'mode' internally above
```
enielse, line 1783: This is no longer unreachable - should remove this comment.

pygsti/algorithms/germselection.py
```diff
@@ -2219,3 +2472,1515 @@ def rcl_fn(x): return _scoring.filter_composite_rcl(x, alpha)
...
+    
+        
+    return cleaned_availableGermList, cleaned_circuit_cache_1
+    
+
+#new function for taking a list of available fiducials and generating a cache of the PTMs
+#this will also be useful trimming the list of effective identities and fiducials with
+#duplicated effects.
+
+def create_circuit_cache(model, circuit_list):
```
enielse, line 2532: I think is function is a duplicate of one in fiducialselection.py -- if so we should just use that one.

pygsti/algorithms/germselection.py
```diff
@@ -169,19 +244,31 @@ def find_germs(target_model, randomize=True, randomization_strength=1e-2,
             'comm': comm,
             'mem_limit': mem_limit,
             'profiler': profiler,
-            'num_nongauge_params': num_nongauge_params
+            'num_nongauge_params': num_nongauge_params,
+            'num_gauge_params': num_gauge_params,
```
enielse, line 248: Does this function need both num_nongauge_params and num_gauge_params as parameters?  It seems like the latter may not be used, in which case it'd probably be good to remove it?  I may be missing something here though.

enielse (approved):
Looks great - if possible, and before you forget, perhaps throw into the docstrings a few more references to the low-rank-update algorithms/formula, i.e., in the literature or on the web?

pygsti/algorithms/germselection.py
coreyostrove: Probably not. I think the value of num_gauge_params is used exactly once and just for convience (i.e. it could just as easily be replaced by num_params-num_nongauge_params). I must've had a reason for wanting this at some point, but that looks to be lost to time. I'll add refactoring this to remove that argument to my to-do list.

coreyostrove: Just pushed a refactor for this.

pygsti/algorithms/germselection.py
coreyostrove: Done

enielse (approved):

pygsti/algorithms/germselection.py
enielse: Ah, I see.  You could just leave this until later, but it'd probably be useful to at least add a comment now mentioning that this is essentially a duplicate of what's in `fiducialselection.py`.

PR #279, opened 2022-11-18
Feature wildcard serialization

enielse:
**Adds ability to serialize results objects to disk or MongoDB that contain wildcard budgets**

This seems useful for obvious reasons.  As a side effect, `Estimate` objects are updated to hold *serialized* wildcard budget objects instead of the objects themselves.  This could break external code that extracts wildcard budgets from result objects.

### DIFF
--- a/pygsti/protocols/gst.py
+++ b/pygsti/protocols/gst.py
@@ -1897,11 +1897,11 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
                     budget = _compute_wildcard_budget_1d_model(base_estimate, objfn_cache, mdc_objfn, parameters,
                                                                badfit_options, printer - 1)
 
-                    base_estimate.extra_parameters['wildcard1d' + "_unmodeled_error"] = budget
+                    base_estimate.extra_parameters['wildcard1d' + "_unmodeled_error"] = budget.to_nice_serialization()
                     base_estimate.extra_parameters['wildcard1d' + "_unmodeled_active_constraints"] \
                         = None
 
-                    base_estimate.extra_parameters["unmodeled_error"] = budget
+                    base_estimate.extra_parameters["unmodeled_error"] = budget.to_nice_serialization()
                     base_estimate.extra_parameters["unmodeled_active_constraints"] = None
                 except NotImplementedError as e:
                     printer.warning("Failed to get wildcard budget - continuing anyway.  Error was:\n" + str(e))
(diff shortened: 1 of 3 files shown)


pygsti/protocols/gst.py
```diff
@@ -1897,11 +1897,11 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
                     budget = _compute_wildcard_budget_1d_model(base_estimate, objfn_cache, mdc_objfn, parameters,
                                                                badfit_options, printer - 1)
 
-                    base_estimate.extra_parameters['wildcard1d' + "_unmodeled_error"] = budget
+                    base_estimate.extra_parameters['wildcard1d' + "_unmodeled_error"] = budget.to_nice_serialization()
```
sserita, line 1900: Why do we have to store the serialized form? Can we not convert to/from serialized form just when the estimate gets serialized?
Just thinking about whether we can be backwards compatible with things that expected these to be WildcardBudgets.

sserita (commented):
This looks good, I was just curious about the decision to store serialized version. It's possible that this might be accepted as-is after a quick discussion.

pygsti/protocols/gst.py
enielse: Yeah, this is definitely a debatable choice.  The issue is that the `Estimate.extra_parameters` dictionary gets serialized using general code that assumes the dictionary is JSON-able.  In order to keep the budget un-serialized, we'd need to either 
1. rework the general code considerably so that it can handle this case.  This would require making that code more complex and it's not clear it's worth doing this just for this case -- but it would be possible.  Depending on how it's done, it may change the format of a serialized Estimate object, causing backward compatibility issues with previously serialized Estimates unless we add version-specific code to handle this.
2. move the budgets from being held within `extra_parameters` to their own separate attribute, like `wildcard_error_models` for instance.  This seems like it'd be cleaner, but this would be a structural change to the Estimate and would require a cascade of changes as well as alter the serialization format.

I think I prefer option 2 over 1, but didn't want to make any significant structural changes to `Estimate` without more discussion.  The way it is not, the structure is the basically the same, which requires less cascading changes, but we end up storing a serialized version of an object which isn't ideal and can be confusing.

sserita: We talked about this, and something like Option 2 will be done in the future if it ends up being a problem. At the moment, the heavy lift to fix this and the limited use case for it makes it not worth fixing at the moment.

sserita (approved):

PR #280, opened 2022-11-21
Feature fidsel performance

coreyostrove:
This branch implements a new greedy search algorithm that leverages the same low-rank update tools now available for germ selection and FPR. This is specifically useful for accelerating fiducial selection for many-qubit systems in the 4+ qubit range (for 3 or less the existing algorithms seem to run quick enough for most applications). This new greedy search algorithm only works with an objective function based on the psuedoinverse-trace. Other notable changes in this branch:

- Added additional flexibility to the specification of candidate fiducial sets. Before we could only specify a maximum fiducial length and then all circuits up to that length would be used in the search space. Now the API matches the one used for germ selection were we can specify a dictionary with keys corresponding to lengths and the ability to either provide special keywords like 'all upto' or a value in which case we sample a certain number of circuits at random. We still have compatibility with the old kwarg though.
- There is a new deduping routine that can be used when the gate set members are generators of the clifford group (or some subset thereof).  So gate sets like XYI or XYCPHASE etc... In this case a flag can be set which allows deduping the circuit list in amortized O(N) time.
- Cleaned up a bunch of extraneous debug and print statements throughout various parts of the experiment design codebase (not just in fidsel)
- Fleshed out incomplete germ selection docstring.

### DIFF
--- a/pygsti/algorithms/fiducialselection.py
+++ b/pygsti/algorithms/fiducialselection.py
@@ -1815,10 +1815,9 @@ def _find_fiducials_greedy(model, fids_list, prep_or_meas, op_penalty=0.0,
                 #    fidArrayList = create_prep_mxs(model, [fiducial], fid_cache)
                 #elif prep_or_meas == 'meas':
                 #    fidArrayList = create_meas_mxs(model, [fiducial], fid_cache)
-                #current_score_mx= _np.concatenate(fidArrayList, axis=1)
-                #current_score_gramian = current_score_mx@current_score_mx.conj().T
+                current_score_mx= fiducial_compact_EVD_cache[fiducial]
                 current_score_gramian= fiducial_compact_EVD_cache[fiducial]@fiducial_compact_EVD_cache[fiducial].T
-                current_inv_trace = _np.trace(_np.linalg.pinv(current_score_gramian))
+                current_inv_trace = _np.trace(_np.linalg.pinv(current_score_gramian, hermitian=True))
                 #Don't add penalties for first fiducial when forcing it to be the empty germ.
                 current_rank= _np.linalg.matrix_rank(current_score_mx)
                 penalized_inv_trace= add_penalties_greedy(current_inv_trace, [fiducial], l1_penalty, op_penalty, gate_penalty)
--- a/pygsti/algorithms/fiducialselection.py
+++ b/pygsti/algorithms/fiducialselection.py
@@ -154,7 +154,6 @@ def find_fiducials(target_model, omit_identity=True, eq_thresh=1e-6,
                 availableFidList.extend(_circuits.list_random_circuits_onelen(
                     fidOps, fidLength, count, seed=candidate_seed))
         
-         = 
         printer.log('Initial Length Available Fiducial List: '+ str(len(availableFidList)), 1)
         
         
--- a/pygsti/algorithms/fiducialselection.py
+++ b/pygsti/algorithms/fiducialselection.py
@@ -14,6 +14,7 @@ import numpy as _np
 import scipy
 import random
 import itertools
+from warnings import warn
 from math import floor
 from pygsti.algorithms import grasp as _grasp
 from pygsti.algorithms import scoring as _scoring
@@ -30,7 +31,8 @@ def find_fiducials(target_model, omit_identity=True, eq_thresh=1e-6,
                    algorithm='grasp', algorithm_kwargs=None, verbosity=1,
                    prep_fids=True, meas_fids=True, candidate_list=None,
                    return_candidate_list=False, final_test= False, 
-                   assume_clifford=False, candidate_seed=None):
+                   assume_clifford=False, candidate_seed=None, 
+                   max_fid_length=None):
     """
     Generate prep and measurement fiducials for a given target model.
 
@@ -115,6 +117,12 @@ def find_fiducials(target_model, omit_identity=True, eq_thresh=1e-6,
         which allows us to use a faster deduping routine exploiting the properties
         of clifford circuits.
         
+    max_fid_length : int, optional (deprecated)
+        The maximum number of gates to include in a fiducial. The default is
+        not guaranteed to work for arbitrary models (particularly for quantum
+        systems larger than a single qubit). This keyword is now deprecated.
+        The behavior of the keyword is now equivalent to passing in an int
+        for the candidate_fid_counts argument.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


pygsti/algorithms/fiducialselection.py
```diff
@@ -365,7 +552,83 @@ def clean_fid_list(model, circuit_cache, available_fid_list,drop_identities=True
         if fid_string in cleaned_circuit_cache_2:
             cleaned_availableFidList.append(available_fid_list[i])
     
+    return cleaned_availableFidList, cleaned_circuit_cache_2
+    
+#function for cleaning up the available fiducial list to drop identities and circuits with duplicate effects
+#Specialized for the case where our fiducials are promised to correspond to clifford operations
+#in which case we know that the matrices must correspond to permutation matrices which
+#gives us some additional structure to exploit.
+def clean_fid_list_clifford(model, circuit_cache, available_fid_list,drop_identities=True, drop_duplicates=True, eq_thresh= 1e-6):
```
sserita, line 561: This might look cleaner if clean_fid_list just took the assume_clifford flag, and switched on it inside the drop_duplicates block.

pygsti/algorithms/fiducialselection.py
```diff
@@ -115,20 +147,48 @@ def find_fiducials(target_model, omit_identity=True, eq_thresh=1e-6,
...
         
         #Now that we have a cache of PTMs as numpy arrays for the initial list of available fiducials
         #we can clean this list up to remove any effective identities and circuits with duplicate effects.
-        
-        cleaned_availableFidList, cleaned_circuit_cache = clean_fid_list(target_model, circuit_cache, availableFidList,
-                                                                        drop_identities=True, drop_duplicates=True,
-                                                                        eq_thresh=eq_thresh)
+        #Use a flag to check if we can assume these are clifford circuits in which case
+        #we can use a more efficient deduping routine.
+        if assume_clifford:
```
sserita, line 184: Could be cleaned up IMO - see below comment in clean_fid_list_clifford.

sserita (changes requested):
Haven't checked the math for the new greedy code path closely, given it's the same process as the recent FPR and germ sel low rank updates.
I think there is one place where code can be consolidated, i.e. we can dedup the dedup code.

pygsti/algorithms/fiducialselection.py
```diff
@@ -91,6 +106,23 @@ def find_fiducials(target_model, omit_identity=True, eq_thresh=1e-6,
...
+        prep and measurement lists using the function test_fiducial_list to verify
+        we have an acceptable candidate set (this uses a different code path in some cases so
+        can be used to detect errors). 
+        
+    assume_clifford : bool, optional (default False)
+        When true then we assume that all of the circuits are clifford circuits,
+        which allows us to use a faster deduping routine exploiting the properties
+        of clifford circuits.
+        
+    max_fid_length : int, optional (deprecated)
```
enielse, line 120: Maybe make this into a warning that get's issued whenever `max_fid_length` is specified?  This would make the update more visible, just in case people don't read the docstring ;)

pygsti/algorithms/fiducialselection.py
```diff
@@ -185,6 +244,16 @@ def find_fiducials(target_model, omit_identity=True, eq_thresh=1e-6,
...
+            #if requested do a final check with test_fiducial_list
+            #to verify the algorithm succeeds
+            if final_test:
+                final_test_fiducial_list = test_fiducial_list(target_model, prepFidList, 'prep',
+                                                    score_func=algorithm_kwargs['score_func'], return_all=False,
+                                                    threshold=algorithm_kwargs.get('threshold', 1e-6), fid_cache=prep_cache)
+                if final_test_fiducial_list:
+                    printer.log('Final test of the candidate prep fiducial lists passed.', 1)
+                else:
+                    printer.log('Final test of the candidate prep fiducial lists failed.', 1)
```
enielse, line 256: The above block is repeated 8 (I think) times within this function, which makes me think it'd be worth consolidating it to a nested function.

pygsti/algorithms/fiducialselection.py
```diff
@@ -1429,3 +1692,269 @@ def rcl_fn(x): return _scoring.filter_composite_rcl(x, alpha)
...
+    current_best_inv_trace= None
+    current_best_composite_score= None
+    
+    #Continue iterating until acceptable solution is found.
+
+    while not acceptable_candidate_found:
+        #if the best fiducial set is not yet initialized
+        #and we have force_empty==True, then we will
+        #initialize it with the index of the empty fiducial.
+        if (best_fiducial_set is None) and force_empty:
```
enielse, line 1795: Maybe, for clarity, change this to `if best_fiducial_set is None:` and place nested `if force_empty: ... else: ...` blocks that over the first two cases here.

pygsti/algorithms/germselection.py
```diff
@@ -147,6 +145,42 @@ def find_germs(target_model, randomize=True, randomization_strength=1e-2,
     toss_random_frac : float, optional
         If specified this is a number between 0 and 1 that indicates the random fraction of candidate
         germs to drop randomly following the deduping procedure.
+        
+    mode : str, optional (default 'allJac')
```
enielse, line 149: Should change this doc line to:
`mode : {'allJac', 'singleJac', 'compactEVD'}, optional`
This shows the options clearly while also communicating that it must be a string.  No need to put default on this line, as it's in the function definition and it adds one more place to get things out of sync.

enielse (approved):
Great stuff!  I added a few suggestions that you can take or leave.

pygsti/algorithms/fiducialselection.py
coreyostrove: Sounds good. I refactored the deduping code implementing this change.

pygsti/algorithms/fiducialselection.py
coreyostrove: Sounds good, I refactored this in line with your suggestions.

pygsti/algorithms/fiducialselection.py
coreyostrove: Cool, added an additional warning about this.

pygsti/algorithms/fiducialselection.py
coreyostrove: Done.

pygsti/algorithms/germselection.py
coreyostrove: Done

pygsti/algorithms/fiducialselection.py
coreyostrove: I refactored this as suggested.

coreyostrove:
Thanks for the comments and feedback, @sserita and @enielse . I have refactored and updated the code in accordance with your suggestions. I'll plan on merging this in once I get a clean set of unit tests back on the latest commit (@sserita, I think you'll also need to approve the PR before merging is unblocked).

sserita (approved):

PR #284, opened 2022-12-20
Feature mongodb upgrade

enielse:
**Upgrades MongoDB serialization framework to be more powerful and efficient**

Updates the structure of pyGSTi objects stored in a MongoDB database so that now:
1.  objects are identified by a normal `bson.objectid.ObjectId` identifier, rather than a filesystem-like path.
2.  a parent object holds the IDs of its child or attribute documents within its "main" document.

This results in a more typical data model and allows documents to serve in a many-to-one capacity, improving space efficiency.  For example, an experiment design can be linked to multiple protocol data objects and a circuit document can be referenced by multiple experiment designs and/or data sets.  This could result in massive data saving by not having to duplicate, e.g., circuits shared by many different experiment designs.

Internally, the `MongoSerializable` base class has been elevated in importance and now serves as the base class for both large (experiment design, protocol data, and results) objects and small `NicelySerializable`-derived objects.  This allows nearly all of pyGSTi's objects to be serialized to a MongoDB using the same interface.

This PR marks a point where the MongoDB support is much more complete and established than it was.  I expect we'll still want to tweak some things before the format stabilizes completely, but this seems to be almost there.  The biggest issue regards removing documents, which should be done with care as they can now be shared, e.g. have multiple parent objects.

Unit tests exist and have been updated, but are not very extensive at this point.

### DIFF
--- a/pygsti/algorithms/core.py
+++ b/pygsti/algorithms/core.py
@@ -13,6 +13,7 @@ Core GST algorithms
 
 import collections as _collections
 import time as _time
+import copy as _copy
 
 import numpy as _np
 import scipy.optimize as _spo
@@ -797,8 +798,12 @@ def run_iterative_gst(dataset, start_model, circuit_lists,
 
             for j, obj_fn_builder in enumerate(iteration_objfn_builders):
                 tNxt = _time.time()
-                optimizer.fditer = optimizer.first_fditer if (i == 0 and j == 0) else 0
-                opt_result, mdc_store = run_gst_fit(mdc_store, optimizer, obj_fn_builder, printer - 1)
+                if i == 0 and j == 0:  # special case: in first optimization run, use "first_fditer"
+                    first_iter_optimizer = _copy.deepcopy(optimizer)  # use a separate copy of optimizer, as it
+                    first_iter_optimizer.fditer = optimizer.first_fditer  # is a persistent object (so don't modify!)
+                    opt_result, mdc_store = run_gst_fit(mdc_store, first_iter_optimizer, obj_fn_builder, printer - 1)
+                else:
+                    opt_result, mdc_store = run_gst_fit(mdc_store, optimizer, obj_fn_builder, printer - 1)
                 profiler.add_time('run_iterative_gst: iter %d %s-opt' % (i + 1, obj_fn_builder.name), tNxt)
 
             tNxt = _time.time()
--- a/pygsti/baseobjs/nicelyserializable.py
+++ b/pygsti/baseobjs/nicelyserializable.py
@@ -15,9 +15,10 @@ import pathlib as _pathlib
 import json as _json
 import numpy as _np
 import scipy.sparse as _sps
+from pygsti.baseobjs.mongoserializable import MongoSerializable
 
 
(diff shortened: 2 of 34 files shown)


pygsti/baseobjs/nicelyserializable.py
```diff
@@ -15,9 +15,10 @@
 import json as _json
 import numpy as _np
 import scipy.sparse as _sps
+from pygsti.baseobjs.mongoserializable import MongoSerializable
 
 
-class NicelySerializable(object):
+class NicelySerializable(MongoSerializable):
```
sserita, line 21: Just to confirm the swap of inheritance order. The intention is that everything that is filesystem serializable is also mongo serializable, but there are some objects (e.g. DataSet,  Estimate, Protocol) which are currently only mongo serializable (as of this PR)? I assume the goal is to enable the storing/reuse of these subcomponents, which before were only dumped as part of a larger tree structure?

sserita (commented):
Everything seems good but there's a lot of code here, so I asked a couple of clarifying questions to make sure I understood the goal of the PR a little better.

Also I wasn't sure if you were ready for this to go in, since you've been pushing bugfixes.

pygsti/algorithms/core.py
```diff
@@ -797,8 +798,12 @@ def _max_array_types(artypes_list):  # get the maximum number of each array type
 
             for j, obj_fn_builder in enumerate(iteration_objfn_builders):
                 tNxt = _time.time()
-                optimizer.fditer = optimizer.first_fditer if (i == 0 and j == 0) else 0
-                opt_result, mdc_store = run_gst_fit(mdc_store, optimizer, obj_fn_builder, printer - 1)
+                if i == 0 and j == 0:  # special case: in first optimization run, use "first_fditer"
```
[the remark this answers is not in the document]
enielse: Yes, and it's related to mongo serialization :). The "bug" was that we were modifying the optimizer object (by setting its `.fditer` attribute on old line 800) within this GST-running routine.  This modification is unexpected by the caller, and is caught by the MongoDB serialization which includes checks to make sure serializable objects like the optimizer (`Optimizer` is a subclass of `NicelySerializable`) don't change when we don't expect them to.

pygsti/baseobjs/nicelyserializable.py
enielse: Yes, this is correct, and I agree it's a bit counterintuitive at first.  The thinking is that "everything can be rendered as JSON can also be serialized to a MongoDB" which implies that anything that is `NicelySerializable` must also be `MongoSerializable`.  Another way of thinking about this is that something that is `NicelySerializable` is able to render itself as simple JSON, whereas something that is `MongoSerializable` is able to insert itself into a Mongo database, perhaps using multiple records.  In principle we could take all the records a `MongoSerializable` object generates and cram them together into one big piece of JSON, but in practice there are certain more complex objects (like `DataSet`s, `Estimate`s, and `Protocol`s) that naturally can save themselves as multiple records in a database but would be a bit unwieldy as a huge piece of JSON.

sserita (approved):

PR #285, opened 2022-12-21
Bugfix numpy deprecations

coreyostrove:
I took an initial pass at fixing the calls to the newly deprecated dtype aliases. I think I got everything in our codebase, but there are some broken calls in qibo that cause the import to die, and related unit tests to fail, so these have been temporarily disabled. There are a couple more unit tests that are failing, but these are related to germ selection and seem to be broken for a different reason (I suspect for a disjoint numpy change given the error messages). For the sake of having a clean commit history I am going to fix those problems on another branch where I also look into the edge case @enielse showed me about when the initial set of germs in germ selection is empty. So I'm requesting we merge this in even though there are still some failing tests.


sserita (approved):
All straightforward changes, thanks for tracking those down!

PR #286, opened 2022-12-21
Bugfix germsel initialization

coreyostrove:
This should fix the last two failing unit tests, which I believe were failing due to a numpy 1.24.0 related change (though not one easily ascertained from the changelog). These lines of codes haven't been touched in a very long time, so there isn't any other reason I can think of for them suddenly breaking.

Anyhow, I haven't yet resolved the other problem @enielse noticed when the list of forced germs is empty, but I don't think I'll get around to that until after we push the release (I can aim for whenever the first hotfix we push to master is).


sserita (approved):
Straightforward bugfix.

PR #293, opened 2023-01-18
Replace deprecated function calls to inspect module

coreyostrove:
This PR patches some deprecated function calls from the python inspect module that have been deprecated as of python 3.11. Patch courtesy of @sserita.


sserita (approved):
I think this patch was more of a team effort, @coreyostrove... Thanks for making the PR!

PR #294, opened 2023-01-18
Patch wildcard gauge optimization bugs

coreyostrove:
This patch addresses a bug in the 1D wildcard models that resulted from a mismatch in the gauge optimization settings used in different places in the code. We now have some basic plumbing for the GSTGaugeOptSuite that pipes the different gauge optimizations through to the badfit estimate code and makes sure that we are consistent. Also makes the wildcard models switchable on the gauge optimization in the report generation for cases where we have more than one gauge optimization used.

This patch follows up on the initial patch @enielse made for this issue in commit 283803466ab92fd2cd6c838a906625ef96c20565
(I actually hadn't noticed there was a recent commit targeting this until putting the PR together, funny that we both ran into this in such close succession!)


sserita (approved):
Thanks for piping the gaugeopt through, all looks good.

Issue #297, opened 2023-04-18
 pygsti.serialization.json.dump fails on pygsti.protocols.gst.ModelEstimateResults instance


eendebakpt:
**Describe the bug**
Serialization of an `pygsti.protocols.gst.ModelEstimateResults` instance fails if some instance methods have been called.

**To Reproduce**
A minimal example:
```
import tempfile
from pathlib import Path
import pygsti
from pygsti.modelpacks import smq1Q_XYI

report_folder = Path(tempfile.mkdtemp())

max_sequence_lengths=[1]
model_ideal = smq1Q_XYI.target_model()  # pylint: disable=no-member
prep_fiducials = smq1Q_XYI.prep_fiducials()  # pylint: disable=no-member
meas_fiducials = smq1Q_XYI.meas_fiducials()  # pylint: disable=no-member
germs = smq1Q_XYI.germs()  # pylint: disable=no-member

experiment_design=exp_design = pygsti.protocols.StandardGSTDesign(
    model_ideal, prep_fiducials, meas_fiducials, germs, max_sequence_lengths)

experiment_list = exp_design.all_circuits_needing_data

shots = 1000
results_counts  = [{'0': 1000},
 {'1': 520, '0': 480},
 {'1': 504, '0': 496},
 {'1': 1000},
 {'0': 488, '1': 512},
 {'1': 532, '0': 468},
 {'1': 507, '0': 493},
 {'0': 1000},
 {'1': 503, '0': 497},
 {'0': 504, '1': 496},
 {'1': 1000},
 {'1': 485, '0': 515},
 {'1': 507, '0': 493},
 {'0': 1000},
 {'1': 517, '0': 483},
 {'0': 482, '1': 518},
 {'1': 503, '0': 497},
 {'1': 509, '0': 491},
 {'1': 1000},
 {'0': 499, '1': 501},
 {'1': 495, '0': 505},
 {'1': 506, '0': 494},
 {'0': 493, '1': 507},
 {'1': 1000},
 {'0': 1000},
 {'0': 482, '1': 518},
 {'1': 503, '0': 497},
 {'1': 1000},
 {'1': 490, '0': 510},
 {'1': 503, '0': 497},
 {'0': 497, '1': 503},
 {'1': 1000},
 {'1': 491, '0': 509},
 {'0': 482, '1': 518},
 {'0': 1000},
 {'1': 498, '0': 502},
 {'1': 505, '0': 495},
 {'0': 503, '1': 497},
[...]

PR #298, opened 2023-04-18
Fix serialization of Estimate

eendebakpt:
See #297


sserita:
Closed because no CLA for contributor; however, bugfix will be included with others in upcoming release.

PR #300, opened 2023-04-26
CI workflow updates

coreyostrove:
This PR includes a number of updates to the CI workflow, as well as a number of bugfixes.

1. Migration from nosetests to pytest. Nose is no longer under active development and as of python 3.10 breaks on our existing test suite.
2. Addition of automatic notebook regression testing using the pytest module nbval. These are configured to run on push to beta or master (to save on excessive CI time). Currently configured to use 'lax' testing wherein we only check that the notebooks run to completion successfully. The module does support verifying the output of notebooks as well, so down the line we could consider doing this.
3. Removes testing for python 3.6 and adds testing for python 3.10.
4. Switches ubuntu tests from 18.04 to 20.04. Add testing for MacOS.
5. A bunch of bug fixes and unit test fixes 
6. Deprecation warnings added for older serialization functionality using the json and msgpack modules.


sserita (approved):
All the CI changes look good to me, thanks for taking care of this. I didn't check through every notebook change, but given that they are being run through the tests now, it should be fine.
I've also merged in my branch that got rid of a bunch of deprecation warnings that were exposed by pytest. I will merge as soon as that comes back clean.

PR #334, opened 2023-08-02
Switch from cirq to cirq-core in setup.py

coreyostrove:
This commit switches the cirq dependency to cirq-core from cirq. Recently we've begun having pip installation issues due to unresolvable dependency conflicts (both locally and on the github runners). I traced this back to cirq, and while I haven't definitively proven this I think that it is specifically a dependency issue with the cirq-rigetti submodule. When installing cirq it will by default include all of the optional submodules, most of which are specific hardware platform interface code which isn't needed for our purposes (which is just the ability to output cirq circuit objects). So we should be fine switching to cirq-core (which includes all of the core functionality we need) and avoiding installation of all of the extra submodules altogether. If a user wants that they can figure out the dependency issues for whichever hardware backend they want to connect to with cirq on their end.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ extras = {
         'seaborn',
         'ply',
         'qibo<=0.1.7',
-        'cirq',
+        'cirq-core',
         'notebook',
         'ipython'
     ]


coreyostrove:
Quick follow-up on the unit tests. I took a look at the logs and the unit tests that we have failing are entirely unrelated to this change and are due to some deprecation warnings introduced in numpy 1.25. A couple of our unit tests related to the forward simulator (which looks like it runs into this deprecation) have assertions that they encounter no warnings, so this is what is failing here. We can fix this but it should probably be a separate PR.

sserita:
It looks like the cirq -> cirq-core change snuck into the numpy125 PR as well. I'm closing this one and will accept that one which includes this change and passes the tests.

PR #345, opened 2023-09-06
Make more efficient use of CVXPY when computing the diamond norm

### LINKED ISSUE #159: Some inefficiency in SDP construction
Hi pyGSTI developers. This is just a note that the following use of CVXPY can be made more efficient.

https://github.com/pyGSTio/pyGSTi/blob/dd4ad669931c7f75e026456470cf33ac5b682d0d/pygsti/tools/optools.py#L326-L348

Specifically, the constraints ``sig0 == sig0.T, sig1 == sig1.T`` can be removed if the lines
```
        sig0 = _cvxpy.Variable(shape=(smallDim, smallDim))  # rho0.real
        sig1 = _cvxpy.Variable(shape=(smallDim, smallDim))  # rho1.real
```
are changed to
```
        sig0 = _cvxpy.Variable(shape=(smallDim, smallDim), symmetric=True)  # rho0.real
        sig1 = _cvxpy.Variable(shape=(smallDim, smallDim), symmetric=True)  # rho1.real
```
Using full matrices for the antisymmetric variables ``tau0`` and ``tau1`` is also inefficient, but that doesn't have a one-line fix.

rileyjmurray:
This PR means to resolve #159. It removes support for CVXPY 0.4 and updates the CVXPY 1.x code path to make efficient use of CVXPY's complex variables and hermitian matrices. Ready for review, @coreyostrove and @sserita!

#### Old PR description below. Keeping here so discussions make sense.
This PR is currently a work-in-progress. I'm making this it now to get input on whether it's worth supporting CVXPY 0.4 (which received its last update in 2017) in addition to the now-current CVXPY 1.x series. If we want to support both, then there will be some code duplication. I suspect that it's better to remove the code for CVXPY 0.4.

### DIFF
--- a/pygsti/tools/optools.py
+++ b/pygsti/tools/optools.py
@@ -392,38 +392,43 @@ def _diamond_norm_cvxpy_04(dim, smallDim, K, L):
     return prob, [Y, Z, sig0, sig1, tau0, tau1]
 
 
-def _diamond_norm_cvxpy_1x(dim, smallDim, jamiolkowski_matrix, use_herm=True):
-    K = jamiolkowski_matrix.real  # J.real
-    L = jamiolkowski_matrix.imag  # J.imag
+def _diamond_norm_cvxpy_1x(dim, smallDim, jamiolkowski_matrix):
+    # Compute the diamond norm
+    #
+    # Uses the primal SDP from arXiv:1207.5726v2, Sec 3.2
+    #
+    # Maximize 1/2 ( < J(phi), X > + < J(phi).dag, X.dag > )
+    # Subject to  [[ I otimes rho0,       X        ],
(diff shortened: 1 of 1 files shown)


coreyostrove:
I'm on board with removing cvxpy 0.4 support. Given the age of this release I can imagine it being incredibly unlikely that any of our users are still using 0.4, so there is probably little harm in simply axing it immediately. If we were squeamish about the slight possibility of breaking something without warning we could add a deprecation warning now and then slate it for removal in 0.9.13.

rileyjmurray:
Excellent. If you're not opposed to removing 0.4 support, then let's absolutely remove it. I've pushed changes accordingly!

sserita (approved):
Looks good (as far as I followed the change in constraint specification).

Issue #346, opened 2023-09-07
Evotype dependent serialization bugs


coreyostrove:
Some failing unit tests on the GST checkpointing branch revealed a serialization bug (really a bug that becomes apparent when trying to do serialization) that seems to be related to the use of the 'statevec' evotype. In particular, it looks like when using the 'statevec' evotype the _basis attribute is not always set properly remains initialized to None. We then encounter an error in serialization which expects there to be a basis object present and fails when trying to call the expected basis object's to_nice_serialization method.

Minimal working example for reproducing the bug:

```
from pygsti.modelpacks import smq1Q_XYI as std
test_model = std.target_model("static unitary", evotype='statevec')
test_model.to_nice_serialization()
```

Which returns the traceback:

```
 ---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 test_model.to_nice_serialization()

File ~\Documents\pyGSTi_master_branch\pygsti\baseobjs\nicelyserializable.py:164, in NicelySerializable.to_nice_serialization(self)
    154 def to_nice_serialization(self):
    155     """
    156     Serialize this object in a way that adheres to "niceness" rules of common text file formats.
    157 
   (...)
    162         Python type, e.g. a string or list.
    163     """
--> 164     state = self._to_nice_serialization()
    165     if self._dbcoordinates is not None and isinstance(state, dict):
    166         # HACK: the code below adds the special dbcoordinates attribute to `state` when we're serializing to
    167         # something other than a MongoDB (which holds the dbcoordinates intrinsically in the document's
   (...)
    170         # a flag or something else that determines this, but for now we add this fragile HACK that checks the
    171         # caller names for 'add_mongodb_write_ops' (only in the call stack when we're serializing to MongoDB).
    172         import inspect

File ~\Documents\pyGSTi_master_branch\pygsti\models\explicitmodel.py:1603, in ExplicitOpModel._to_nice_serialization(self)
   1584 state.update({'basis': self.basis.to_nice_serialization(),
   1585               'default_gate_type': self.operations.default_param,
   1586               'default_prep_type': self.preps.default_param,
   (...)
   1599                                        if (self._param_interposer is not None) else None)
   1600               })
   1602 mmgraph = self.create_modelmember_graph()
-> 1603 state['modelmembers'] = mmgraph.create_serialization_dict()
   1604 return state

[...]

PR #347, opened 2023-09-07
Feature gst checkpointing

coreyostrove:
This PR gives an initial implementation for checkpointing iterative GST fits and introduces some basic and extensible infrastructure for checkpointing protocols more broadly going forward. For this initial implementation there is currently only full support for checkpointing `ModelTest`, ` GateSetTomography` and 'StandardGST' protocols. Some highlights:

1. Adds ProtocolCheckpoint, GateSetTomographyCheckpoint, ModelTestCheckpoint and StandardGSTCheckpoint objects, each of which can be serialized natively to json.
2. Protocol run methods (and related driver functions) now take as input kwargs allowing one to give a previously generated checkpoint object if one exists, a path to write checkpoint files to during the course of the fit and finally an option to disable checkpointing entirely (checkpointing is on by default).

This should hopefully reduce one of the common pain points users encounter wherein their fits fail for some mundane reason (like running out of memory or cluster allocation time etc.) and then they need to restart an expensive calculation from scratch.

I have added some unit tests (these currently live in the test_drivers module in test_packages for convenience), which all seem to be passing at present. We do have some tests failing in the extras, but these look to be unrelated to new code beyond the fact that the new code made these bugs visible (discussed further in #346). So given that I'd advocate for merging this in despite those failing tests. Ultimately we'll want to disable checkpointing in these particular failing unit tests---I had done a pass through the tests to disable most of these to avoid needing to deal with file hygiene on the runners for tests that aren't related to checkpointing but missed these by accident---but for now I'd argue is useful to keep it enabled since that is what made these serialization bugs apparent.

Unrelatedly this includes a deprecation of the comma-separated string syntax for modes in StandardGST and some fixes to unit tests to account for that. 

Feedback and questions welcome.


pygsti/algorithms/core.py
sserita, line 1: Is it worth refactoring this now to avoid duplicating the iterative_gst function?

sserita (commented):
This all looks really good. I appreciate both your commitment to updating examples and tests to avoid deprecation warnings and the new unit tests. Only have one comment for us to touch base on before approving.

pygsti/algorithms/core.py
coreyostrove: Ah, good catch. It had slipped my mind that we had discussed merging these different code paths together. I'll refactor the run_iterative_gst function to use the generator version under the hood and push that change.

coreyostrove: Refactor done and I've also added a new test targeting a new option specific to use of the generator version. Thanks again for the feedback.

sserita (approved):
Thanks Corey! Looks great.

PR #349, opened 2023-09-13
Bugfix pspec instrument stdgst

sserita:
Several small fixes to allow StandardGST protocols and reports to work with models that use instruments. Includes several fixes for converting between explicit models and processor specs to no longer drop instruments.


coreyostrove (approved):
This all looks good to me!

PR #359, opened 2023-10-27
Feature ibm device

sserita:
This branch deprecates the majority of `pygsti.extras.devices.devcore` in favor of the new `pygsti.extras.devices.ExperimentalDevice`.

An `ExperimentalDevice` can be created in two ways:

1. By passing in a device name, which uses the now legacy device files in `pygsti.extras.devices`. This is provided mainly for backwards compatibility and in case you are working offline (or with a retired IBMQ device).
2. By passing in an active IBMQ backend (which is the intended default workflow).

I still have several bugfixes that I will push on this branch before it is ready, including:

- [x] IBMQ.write() bug
- [x] 'id' instruction deprecation
- [x] Deprecation of `IBMQ` to `qiskit-ibm-provider`

But I'm putting the PR up early to get feedback from IBMQ users. The tutorial should be up to date on usage if you don't feel like reading code


sserita:
Pending no other major feedback this week, I think this is ready to be merged in.

PR #371, opened 2023-11-14
WIP: Tagging for code level of user exposure

### LINKED ISSUE #354: Developer infrastructure: mark code with level of user-exposure
pyGSTi is both research software and a tool that's used in "production" QCVV. Its research flavor makes it impractical for us to spend a lot of effort differentiating between a public and a private API. At the same, its role in production code means we need to be judicious about changing pyGSTi in a way that would break users' workflows. These factors combined make it hard for newcomers (like myself) to contribute to pyGSTi. Here's a proposal to help get new contributors un-stuck.

### Proposed policy
I suggest the following taxonomy for marking a piece of code's level of "user exposure."
* High exposure: API changes are likely to break workflows for some users, including "casual" users.
* Low exposure: API changes are unlikely to affect casual users of pyGSTi but might affect some of pyGSTI's advanced users.
* Minimal exposure: while it's conceivable that API changes could break user code, the possibility of this is sufficiently remote that we can change the API at will.

I think everyone is on board with a taxonomy along those lines. However, we didn't discuss how these labels would be used to affect pyGSTi's development. I suggest we use it in the following way.
* API changes to high-exposure code require the support of all pyGSTi maintainers.
* API changes to low-exposure code require the support of a majority of pyGSTI's maintainers.
* API changes to minimal-exposure code can be made with the support of any single maintainer.

For now, I'd take "pyGSTi maintainer" to mean @sserita, @coreyostrove, or @enielse.

Note: In the call, we joked a bit about marking certain code as "godforsaken," to signify that we'd like to change it substantially, but we don't quite know how or when to do so. While I think that would be helpful, I don't think we need to figure that out now.

### Proposed adoption plan
Over the coming months, the pyGSTi maintainers mark all "high exposure" code. Low- and minimal-exposure code is marked along the way, to the extent possible.

The adoption process will effectively complete when we're comfortable with implicitly designating unlabeled code as "minimal exposure." API changes to any such code only require the approval of one maintainer, but the approving maintainer has a responsibility to think about whether or not the code _actually_ has minimal user exposure.

### Request to pyGSTi developers
Comment below indicating if you agree with this general direction, or if you'd like changes. Once we converge on what we want to do, we can figure out how exactly we'll do it.

rileyjmurray:
This work-in-progress PR means to address #354.

The plan is to keep this PR open and for each of us to periodically update this branch. @enielse has gotten us started with tagging a few modules. Although the process has just begun, we've already found a module that has minimal user exposure _and_ that's "forsaken"! (Ripe for the removing!)

Tips for working on this:
* Pull from GitHub before making changes. There's a chance someone else might have already pushed some work you don't see on your local repo.
* If you're about to spend a while on this (say, more than 15 minutes) then it would probably be a good idea to post here and explain what (sub) folders you're looking at. That way we don't duplicate effort.

CC: @sserita and @coreyostrove.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ name: Build and run tests
 
 on:
   push:
-    branches-ignore: [ "beta" ]
+    branches-ignore: [ "beta", "user-exposure-tagging" ]
   # Hacky way to only run pull requests from forked repositories (assumes : is not used in branch names unless forked)
   # https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/10
   pull_request:
--- a/pygsti/algorithms/compilers.py
+++ b/pygsti/algorithms/compilers.py
@@ -1,3 +1,4 @@
+## user-exposure: low (EGN)
 """
 Clifford circuit, CNOT circuit, and stabilizer state/measurement generation compilation routines
 """
--- a/pygsti/algorithms/contract.py
+++ b/pygsti/algorithms/contract.py
@@ -1,3 +1,4 @@
+## user-exposure: minimal, forsaken (EGN)
 """
 GST contraction algorithms
 """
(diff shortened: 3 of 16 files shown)


sserita:
This is obsolete in favor of #385, from the ionq fork.

Issue #369, opened 2023-11-13
Nondeterministic Unit Test Failures


coreyostrove:
Following the merge into develop of PR #350 we started getting a failure on one of the notebook regression tests. This failure was apparently caused by one of the additional lines added to the tutorial on germ selection in GST-FiducialAndGermSelection.ipynb with this latest update. At first glance this appeared to be restricted to Windows and on python 3.8 and 3.10, but subsequent re-runs of the workflow revealed that the failure was actually apparently non-deterministic. See notebook regression workflow [run 53](https://github.com/pyGSTio/pyGSTi/actions/runs/6818802500) for more on this. In particular, there are multiple attempts, and for attempts 2-4 only python 3.8 on windows fails these tests (always for the last cell of GST-FiducialAndGermSelection.ipynb with the same ValueError exception raised in the vicinity of line 1125 of germselection.py).

The exception that is being raised here is related to the construction of a projector that should be real-valued, but is apparently failing to be (and by quite a lot, this is not a matter of being just over the testing threshold, as I learned later on in testing). Given that the regression tests take forever to run, and that a bunch of the standard debugging output from pytest doesn't get produced when using nbval for the notebook testing, I turned this cell into a standard unit test. This provided some additional output, but hasn't really provided any insight into the nondeterminism. It is worth noting that when this exception is raised that it appears (based on some spot checking) that the values of the large imaginary projector components detected seem to be the same across the failed runs, independent of python version.

Here is the summary of what I've seen so-far from these unit testing runs (these runs were restricted only to windows to speed things up since I haven't seen strangeness on any other platform yet):

#1166 - python 3.9 failed
#1167 - python 3.10 failed
#1168 - python 3.8 failed
#1169 (attempts 1 and 2 pass for mysterious reasons), then [Python 3.9 fails](https://github.com/pyGSTio/pyGSTi/actions/runs/6845236256/job/18610970938#step:10:184) on attempt 3.

It is worth noting that I did make some minor changes to the source code in-between these runs, but nothing that should, at least in principle have made any difference to the testing (changes summarized below for reference).

* Between #1166 and #1167 I realized that we should really have been testing the magnitude of the complex components of the matrix being tested (to avoid missing large negative terms), so that change was made. In principle, if the behavior were deterministic I would expect this change to have strictly added more failures to the list, not shift them as was observed.
[...]

Issue #360, opened 2023-11-03
test_fiducialpairreduction takes a very long time


rileyjmurray:
The tests defined at
https://github.com/pyGSTio/pyGSTi/blob/d158976317310156316db8922804bbc2a4dbbf73/test/unit/algorithms/test_fiducialpairreduction.py
take several minutes to run on my Macbook Pro with an M2 Max CPU. This makes it impractical to casually run ``pytest test/unit/algorithms`` and see if some small changes to code broke anything or fix anything.

I'm filing this issue so that I (or something else) remembers to make this test more efficient at some point.

coreyostrove:
Thanks for raising this issue, @rileyjmurray. I'm familiar with the relevant knobs here and should be able to get the runtime trimmed down. I clocked this set of tests at 9:33 on a single thread on my x86-based machine. Some low-hanging fruit I noticed right away:

1. FPR typically doesn't do anything to the idle gate, so testing FPR with a gate set that includes the idle is a waste of computation.
2. For the randome FPR routines we can significantly reduce the number of per-iteration samples taken, as the resulting reduced sets don't actually need to be good for the sake of testing.

I'll make those changes and report back on the updated testing times (we can decide if we need any more pruning at that point).

coreyostrove:
Some updates:

- I've streamlined the tests in test_fiducialpairreduction.py and have got the runtime for that set of tests down to ~15 seconds on a single thread on my machine (down from 9:33 originally).
- As part of those updates I did some refactoring of the test fixtures to use a non-legacy modelpack. That broke a bunch of stuff so I had to refactor a good chunk of the remainder of the algorithms test module and implement some bugfixes in the actual codebase to fix that.
- I also noticed some coverage lapses in test_fiducialselection, which plugged a bit.

All in all the algorithms module now takes ~1:45 on a single thread on my laptop. Current time outliers are the gauge optimization tests in test_gaugeopt.py. So I'll take a quick look at those before pushing up a branch with these changes.

rileyjmurray:
That sounds great! Getting tests down below 2 minutes is very reasonable.

rileyjmurray:
@coreyostrove can you push these changes to a branch so I can pull them down?

coreyostrove:
These changes are live on the branch 'feature-faster-algorithm-tests' (same branch I've been doing my debugging for #369, fwiw). Sorry for not following up to indicate that I had pushed this.

sserita:
[...]

PR #350, opened 2023-09-16
Feature globally germ aware fpr

coreyostrove:
This PR includes a significant number of changes, primarily with regards to the experiment design code, as suggested by the name of the branch, but also with regards to a number of miscellaneous changes that were made in other parts of the codebase to serve projects that had been worked on while using this branch. Here is my best attempt at a summary, broken into categories:

- Per-Germ Global Fiducial pair reduction: implementation of a new version of fiducial pair reduction outlined in [this paper](https://arxiv.org/abs/2308.08781).
- Fiducial selection updates
    - Performance improvements for deduping candidate fiducials for non-clifford gate sets
    - Refactors for candidate fiducial set generation
    - Clean-up line-label handling on empty fiducial
    - Various bugfixes
- Germ selection bugfixes and minor updates:
    - Gate sets with statically parameterized gates
    - Memory efficient implementation of germ list completeness check
    - Better implicit model support/bugfixes
- Fisher information
    - MPI-support
    - Addition of approximate fisher information
    - Option for more memory efficient implementation
    - Additional options for pre-computation/caching
- Miscellaneous
    - Better reporting support for ModelTest protocols w/o gauge-optimization
    - Bugfix in distributed memory implementation of customsolve
    - Error generator plot bugfix when number of qubits>2
    - Add product method to MapForwardSimulator
    - Front load COPA layout construction and associated memory checks
    - Add a new method to StateSpace base class for retreiving a list of state space labels (implemented to address a bug that was in one of the fiducial selection routines, but should be useful more generally).

Unit tests have been added for new experiment design code and some existing coverage gaps filled. Updated Fidsel+Germsel and FPR tutorials to reflect these changes and highlight new capabilities.

PR #372, opened 2023-11-14
Failing Unit Test Bugfixes and Performance Related Testing Refactors

coreyostrove:
This PR includes two main changes:

1. Bugfixes for the failing notebook regression tests uncovered after merging in PR #350. The details of exactly what broke is complicated, but in a nutshell related to some new numerical stability problems with some of the older parts of the germ selection code (why this started randomly breaking now, who knows?). For more details see the thread at #369. 
2. A combination of unit test performance tweaks and refactors to address issue #360. Speedups are primarily due to tweaks to the algorithms and drivers testing modules. This includes switching from legacy modelpacks to modern ones in a number of places.

It looks like all of the unit tests, including the extras and the notebook regression tests (which I manually dispatched on this branch) are passing now. Let me know if you have any questions.

One thing worth noting, I had originally merged in the branch `bugfix-for-0.9.12.0`, but found that this resulted in a number of unit tests in the extras package failing. Check out 'build and run test extras' 211 for more details on that. So I opted to revert that merge on this branch to keep the changes separate (and save myself the effort of tracking that down).

### DIFF
--- a/pygsti/algorithms/directx.py
+++ b/pygsti/algorithms/directx.py
@@ -13,6 +13,7 @@ Functions for generating Direct-(LGST, MC2GST, MLGST) models
 
 from pygsti.algorithms import core as _core
 from pygsti import baseobjs as _baseobjs
+from pygsti.baseobjs import Label
 from pygsti import circuits as _circuits
 from pygsti import objectivefns as _objfns
 from pygsti.modelmembers.operations import FullArbitraryOp as _FullArbitraryOp
@@ -394,7 +395,9 @@ def direct_mc2gst_models(circuits, dataset, prep_fiducials, meas_fiducials,
         for i, sigma in enumerate(circuits):
             printer.show_progress(i, len(circuits), prefix="--- Computing model for string-", suffix='---')
             directLSGSTmodels[sigma] = direct_mc2gst_model(
-                sigma, "GsigmaLbl", dataset, prep_fiducials, meas_fiducials, target_model,
+                sigma, 
+                Label('GsigmaLbl') if sigma.line_labels == ('*',) else Label('GsigmaLbl', sigma.line_labels), 
+                dataset, prep_fiducials, meas_fiducials, target_model,
                 op_label_aliases, svd_truncate_to, min_prob_clip_for_weighting,
                 prob_clip_interval, verbosity)
 
@@ -560,7 +563,9 @@ def direct_mlgst_models(circuits, dataset, prep_fiducials, meas_fiducials, targe
         for i, sigma in enumerate(circuits):
             printer.show_progress(i, len(circuits), prefix="--- Computing model for string ", suffix="---")
             directMLEGSTmodels[sigma] = direct_mlgst_model(
-                sigma, "GsigmaLbl", dataset, prep_fiducials, meas_fiducials, target_model,
+                sigma, 
+                Label('GsigmaLbl') if sigma.line_labels == ('*',) else Label('GsigmaLbl', sigma.line_labels),
+                dataset, prep_fiducials, meas_fiducials, target_model,
                 op_label_aliases, svd_truncate_to, min_prob_clip,
                 prob_clip_interval, verbosity)
 
@@ -697,6 +702,8 @@ def focused_mc2gst_models(circuits, dataset, prep_fiducials, meas_fiducials,
         for i, sigma in enumerate(circuits):
             printer.show_progress(i, len(circuits), prefix="--- Computing model for string", suffix='---')
             focusedLSGSTmodels[sigma] = focused_mc2gst_model(
-                sigma, "GsigmaLbl", dataset, prep_fiducials, meas_fiducials, start_model,
+                sigma, 
+                Label('GsigmaLbl') if sigma.line_labels == ('*',) else Label('GsigmaLbl', sigma.line_labels),
+                dataset, prep_fiducials, meas_fiducials, start_model,
                 op_label_aliases, min_prob_clip_for_weighting, prob_clip_interval, verbosity)
     return focusedLSGSTmodels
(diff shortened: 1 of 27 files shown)


pygsti/algorithms/directx.py
```diff
@@ -394,7 +395,9 @@ def direct_mc2gst_models(circuits, dataset, prep_fiducials, meas_fiducials,
         for i, sigma in enumerate(circuits):
             printer.show_progress(i, len(circuits), prefix="--- Computing model for string-", suffix='---')
             directLSGSTmodels[sigma] = direct_mc2gst_model(
-                sigma, "GsigmaLbl", dataset, prep_fiducials, meas_fiducials, target_model,
+                sigma, 
+                Label('GsigmaLbl') if sigma.line_labels == ('*',) else Label('GsigmaLbl', sigma.line_labels), 
+                dataset, prep_fiducials, meas_fiducials, target_model,
```
rileyjmurray, line 400: Interesting that you've made a change to this file, since Erik just marked it as minimal-exposure. @coreyostrove any chance you'd disagree with Erik's take on this file's user exposure? Here's the conventions we agreed to, for reference:
* High exposure: API changes are likely to break workflows for some users, including "casual" users.
* Low exposure: API changes are unlikely to affect casual users of pyGSTi but might affect some of pyGSTI's advanced users.
* Minimal exposure: while it's conceivable that API changes could break user code, the possibility of this is sufficiently remote that we can change the API at will.

test/unit/algorithms/test_fiducialselection.py
```diff
@@ -126,18 +123,6 @@ def setUp(self):
             prep_or_meas="meas"
         )
 
-
-# LOL explicit is better than implicit, right?
-class PrepOptimizeIntegerFiducialsExtendedModelTester(
-        PrepOptimizeIntegerFiducialsStdModelTester, FiducialSelectionExtendedModel):
-    pass
```
rileyjmurray, line 133: Good call killing this ...

pygsti/algorithms/germselection.py
```diff
@@ -1045,16 +1045,31 @@ def _super_op_for_perfect_twirl(wrt, eps, float_type=_np.cdouble):
...
-    wrtEvecsInv = _np.linalg.inv(wrtEvecs)
     
+    #Test the wrt matrix to see if it is normal. If so use the schur
+    #decomposition, otherwise use the general eigenvalue decomposition.
+    wrt_conj_transpose = wrt.conj().T
+    wrt_commutator = wrt@wrt_conj_transpose - wrt_conj_transpose@wrt
+    
+    # Get spectrum and eigenvectors of wrt
+    # May as well use the same eps here too.
+    if _np.linalg.norm(wrt_commutator) < eps:
```
rileyjmurray, line 1061: Two comments.

1. I think np.linalg.norm defaults to Frobenius. If that's the case then we should check that the norm is ``<= (dim**0.5) * tol`` for some value of ``tol``.

2. I'd like for ``tol`` to be some fractional power of the working precision's unit roundoff. I think a safe value for the power in this context is 0.75. For double precision arithmetic the roundoff is about 1e-16, so I'd suggest ``tol = (1e-16)**0.75 = 1e-12``. (When I was debugging this code I saw ``eps`` as large as 1e-6.)

rileyjmurray (commented):
Mostly looks good! I have two substantive comments for the eig vs Schur issue. I also have a question about the level of user-exposure for code in ``pygsti/algorithms/directx.py``.

Somewhat tangentially, I stumbled across [this comment block](https://github.com/pyGSTio/pyGSTi/blob/84044313e6262c63431f8444db87e55965853a2a/pygsti/baseobjs/label.py#L30-L33) for the Label class, which says that it's supposed to be a base class only. Based on how the Label class is used in practice, it seems like this comment is wrong. Corey, what's your take?

pygsti/algorithms/germselection.py
coreyostrove: Thanks for the great suggestion, @rileyjmurray. I have just pushed a new commit implementing this change.

pygsti/algorithms/directx.py
coreyostrove: Good question. I agree with Erik's marking this as minimal-exposure. This change arose as a bugfix for some failures encountered in the corresponding test package (test_directx.py) as part of the refactor moving from legacy modelpacks to the modern implementation in testing. This particular line of code turned out to be fragile when it came to circuit line labeling, and so broke when using circuits from the new modelpacks that have different default circuit line labeling behavior than they do in the legacy modelpacks. 

In reality I suspect this is likely due to a disjoint bug in the circuit line label handling code itself in the circuits module that I've run into in other contexts and mentioned to @sserita before, but haven't ever properly documented. (This is a bug related to properly reconciling line labels between circuits with a  placeholder '*' and those with fully specified labels). In the moment it was easier to patch in a workaround in directx than to context switch and try to track the circuit bug. That said, I am in the correct headspace for this sort of thing now, so if nothing else I should be able to get a minimal reproduction of the aforementioned bug documented pretty soon.

While I was working in the directx code I did wonder to myself if we even need any of this anymore? I've personally never interacted with this code in practical usage, and don't know of anyone who does use this code. Given that, maybe this would be a decent candidate for a module we can retire to legacy as part of out spring cleaning process. @enielse and @sserita, thoughts on this last question?

coreyostrove:
RE: Label. I use the base Label class directly pretty regularly, for what that is worth. I suspect that I am not alone in that regard, and that the comment block you mentioned probably accurately describes the initial design intentions, but not present behavior.

pygsti/algorithms/directx.py
coreyostrove: Update: I just read through @enielse's latest commit on the code marking branch and saw that he has marked directx as:

> ## user-exposure: minimal, forsaken (EGN - we don't use this anymore and never really did)

So that is two votes in favor of decommissioning this part of the codebase. @sserita, how do you feel about formally marking this as deprecated in 0.9.12.0 and then removing it (remove = move to the legacy folder) either in 0.9.13.0 or some point thereafter?

rileyjmurray (approved):

pygsti/algorithms/directx.py
sserita: I agree with the minimal tag and we can formally mark this as deprecated. I will do so in the bugfix-for-0.9.12 branch before I merge that in.

sserita (approved):
This all looks good, thanks for your hard work on this!

PR #384, opened 2023-12-11
Implement CODEOWNERS

sserita:
Implement a CODEOWNERS file for pyGSTi to modernize our PR pipeline.

This branch/PR will stay open as we get volunteers in. Edit: We now have all the volunteers and subteams set up.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,15 +14,15 @@ pygsti/report/templates/drift_html_report/ @tjproct @sandialabs/pygsti-maintaine
 pygsti/modelmembers/instruments/ @pcwysoc @sandialabs/pygsti-maintainers
 
 ## RB owners ##
-pygsti/algorithms/compilers.py @jordanh6 @tjproct @sandialabs/pygsti-maintainers
-pygsti/algorithms/mirroring.py @jordanh6 @tjproct @sandialabs/pygsti-maintainers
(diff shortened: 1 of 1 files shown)


sserita:
So one annoyance I've been reading about with code owners is email spam because every PR auto-assigns a review to all code owners. As far as I can tell, there is only ONE exception to this: When a code owner is a team, that team can be set to auto-assign to a member within the team and NOT email everyone else.

With this in mind, I have settled on the following approach:

- When the code owner group is a single person, they are just added to the code owners as normal and will get notified/assigned to every PR that touches that part of the code. I believe this is probably intended behavior.
- When the code owner group is multiple people, we create a subteam of the QPL team which will actually be the code owner. Then we set that group to auto-assign round robin and not email everyone else.

We currently have this happening twice: once for the maintainers and once for RB.
subteams

Here are the relevant settings for pygsti-maintainers, with the important boxes surrounded in red:
pygsti-maintainers

And here we have pygsti-rb, which is similar. However, in this case, I have to be part of the group to maintain it but I don't want to count towards the RB reviewer quota, so we tell GitHub to skip me:
pygsti-rb

sserita:
I'll note that my previous solution has the added benefit that code owners for subteams can be changed out without requiring a modification of CODEOWNERS in the future.

rileyjmurray:
@sserita I like the idea of creating subteams

coreyostrove (commented):
This all looks good to me. Merge at will.

coreyostrove (approved):
Try that again, but hit the approve button this time...

PR #385, opened 2023-12-19
User exposure tagging

sserita:
This work-in-progress PR means to address https://github.com/sandialabs/pyGSTi/issues/354.

The plan is to keep this PR open and for each of us to periodically update this branch. @enielse has gotten us started with tagging a few modules. Although the process has just begun, we've already found a module that has minimal user exposure and that's "forsaken"! (Ripe for the removing!)

Tips for working on this:

Pull from GitHub before making changes. There's a chance someone else might have already pushed some work you don't see on your local repo.
If you're about to spend a while on this (say, more than 15 minutes) then it would probably be a good idea to post here and explain what (sub) folders you're looking at. That way we don't duplicate effort.
CC: @sserita and @coreyostrove.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ name: Build and run tests
 
 on:
   push:
-    branches-ignore: [ "beta" ]
+    branches-ignore: [ "beta", "user-exposure-tagging" ]
   # Hacky way to only run pull requests from forked repositories (assumes : is not used in branch names unless forked)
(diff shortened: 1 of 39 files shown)


sserita:
An update on this: We've decided to also try to do this piecewise as we touch various parts of the code for other fixes. This means that user-exposure tags will not be limited to this branch/PR. We may have merge conflicts as different people tag different parts with exposure, but this will either serve to a) validate the exposure level if they match or b) cause discussion if they don't match.

coreyostrove:
Update: We all more or less collectively agreed that there was little chance of following this through to completion. So we're officially abandoning this.

PR #386, opened 2024-01-08
New AffineShiftOp Modelmember and Re-implementation of ProtectedArray

coreyostrove:
This PR introduces a new operation modelmember called AffineShiftOp. As the name implies, this operation is one that induces an affine shift when applied to a vector. The general structure of these operations is all ones along the diagonal and non-zero entries in the first column. This operation is useful in the context of constructing FOGI representations for SPAM. The only free/mutable parameters of this operation are the non-diagonal elements of the first column of the process matrix.

The implementation of AffineShiftOp is similar to that of FullTPOp in that it uses ProtectedArray to impose read-only access on elements whose values are fixed according to the general structure of the operation class. In this case that restriction is on all elements other than the non-diagonal entries of the first column. The original implementation of ProtectedArray wasn't able to handle this case, however, as it required that the protected indices correspond to a sliceable sub-block of the original array. Generalizing the original implementation in order to handle more general protected index structures proved quite challenging, so in the end I decided to re-implement much of ProtectedArray from scratch to enable this support with much less difficulty. The new version uses a mask-based implementation to allow arbitrary index protections to be set. For what it's worth, I think the new implementation is also much easier to understand and simpler overall. That said, messing with something in baseobjs should be done with care, so having some extra attention on these changes would be appreciated.

Finally, I've added in some new unit tests for AffineShiftOp and for the new functionality in ProtectedArray.

Update 1/31/24: I've manually run the `test_packages` workflow and can confirm that these tests are all presently passing as well.


pygsti/baseobjs/protectedarray.py
sserita, line 1: Much cleaner!

sserita (approved):
Looks good, and with test updates too! @coreyostrove awesome work :)

Issue #387, opened 2024-01-09
Idea: Switch macOS tests to run against beta branch


coreyostrove:
Recently it has been taking forever (hours) for the standard unit test suite to get picked up by the macOS github runners. We also recently learned that the macOS runners use up our free computational credits at 10X the linux rate and 5X the Windows rate. Both of these things are kind of annoying. Question for discussion:

Should we change our CI workflow so that the tests in `unit` only get run on the macOS runners upon pushing to beta (i.e. when merged into develop), rather than against all pushes to any branch?

sserita:
I would be for this. The Mac tests rarely fail due to OS specific issues, so only checking on beta seems like a completely good solution to me.

A follow-up idea: We sometimes but rarely have Python version problems. Thoughts on only testing the lowest and highest versions of Python we support on most branches, and then testing all versions on either develop and/or beta?

coreyostrove:
Testing only the newest and oldest supported versions of python for most branches sounds good to me as well. I'll add these changes to my to-do list and update the workflows.

sserita:
Closed as of #388.

kgantchev:
@coreyostrove Hi, I'm the CEO and co-founder of FlyCI. If you're concerned about your GitHub runners eating up your free tier minutes, then definitely try the [FlyCI](https://www.flyci.net/)'s M1 and M2 runners which are on average about 2x faster and 2x cheaper than GitHub's runners. We also offer a free tier of 500 mins/month on the M1 runner so you don't have to use up your GitHub actions free tier minutes.

Easily replace your M1 runners: 
```diff
jobs:
 ci:
-    runs-on: macos-latest
+    runs-on: flyci-macos-large-latest-m1
   steps:
   - name: 👀 Checkout repo
     uses: actions/checkout@v4
```

Exclusive M2 runners: 
```diff
jobs:
  ci:
-    runs-on: macos-latest
+    runs-on: flyci-macos-large-latest-m2
    steps:
      - name: 👀 Checkout repo
        uses: actions/checkout@v4
 ```
 
We'll be very happy to have you on board.

PR #388, opened 2024-01-11
GitHub runner updates

sserita:
This includes several quality-of-life and deprecation runner updates and addresses #387.

- MacOS runners are oversubscribed and 10x more expensive, and rarely break due to OS specific issues not caught by Linux tests. Therefore, MacOS tests are now only run on `beta` and `master`.
- Bugfix/feature branches now only test the lowest and highest supported Python versions - this should be sufficient to flag any deprecation or missing language feature problems and cuts down the runtime on these branches by 2x. (Combined with no MacOS tests, this takes the number of runners needed from 12 to 4).
- The runner versions were updated to `-latest`, bumping our testing environment to Ubuntu 22.04 from 20.04, Windows 2022 from 2019, and MacOS 12 from 11, and will now track updated environments.
- The GitHub-provided actions were updated from `v2` to `v4` (or `v3` in the case of `cache`) for node16 updates, which should remove the node12 deprecation warning we've been getting.

As an example, this cuts the current runtime from nearly 2 hours (with the MacOS oversubscription problem) to 13 minutes on this feature branch.

### DIFF
--- a/.github/workflows/autodeploy.yml
+++ b/.github/workflows/autodeploy.yml
@@ -23,11 +23,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0  # to fetch all branches and *tags* (needed to get version number correctly)
   
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
         name: Install Python
         with:
           python-version: '3.10'
@@ -39,7 +39,7 @@ jobs:
           CIBW_BUILD_VERBOSITY: 1
           CIBW_BEFORE_ALL_LINUX: ./.github/ci-scripts/before_install.sh
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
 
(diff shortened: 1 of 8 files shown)


coreyostrove:
Question for you before I approve this. If I understand everything here correctly the new workflow structure is such that:

1.  main-minimal gets run upon pushes to any branch.
2. main, which includes python 3.9 and 3.10 on linux and windows, gets run only upon pushes to develop and master (and external PRs).
3. main-mac which includes python 3.8 through 3.11 on macOS only gets run upon pushes to beta and master (and external PRs).

Is splitting off main and main-mac necessary? In principle beta should track develop in which case there is little difference and these workflows could be merged. In practice this is only true when `test_extras` passes, though, so I wasn't sure if this was the reason behind that difference.

sserita:
Thanks for taking a look Corey - you understand the proposed workflow correctly.

The reason I split main and main-mac was to avoid the Mac runners stuffing up tests on develop, which I thought was the original intent of #387. Although rereading now, I think you may have just wanted `main-minimal` and `main`. The Mac tests rarely catch bugs that the linux tests do not, but we do have cases where develop fails and does not push to beta - in these cases, I thought having the more streamlined no Mac tests would be worthwhile. We do want to test the Mac environment eventually, thus we have `main-mac` which runs on beta and master just so we have the full complement of tests on those branches.

I don't feel too strongly about this. If you think it is cleaner to merge the workflows and don't mind waiting for the Mac tests when we push to develop, I will change that and remerge those.

coreyostrove (approved):
Thanks for the explanation. I'm sold, and this all looks good to me. Merge away!

PR #390, opened 2024-01-18
PyTorch-backed forward simulation

rileyjmurray:
This PR introduces ``TorchForwardSimulator``, a forward simulator (for computing circuit outcome probabilities) based on PyTorch. It uses automatic differentiation to compute the Jacobian of the map from model parameters to circuit outcome probabilities. In the future we could extend it to do computations on a system's GPU, or to use PyTorch-based optimization algorithms instead of pyGSTi's custom algorithms for MLE.

#### Approach

My approach required creating a new ModelMember subclass called ``Torchable``. This subclass adds two required functions, called ``stateless_data`` and ``torch_base``. Their meanings are given below:
https://github.com/sandialabs/pyGSTi/blob/1ec6909a7f9149c4f6758f9cd46256dbbff984c1/pygsti/modelmembers/torchable.py#L18-L43
In principle, TorchForwardSimulator can handle all models for which constituent parameterized ModelMembers are Torchable. So far I've only extended TPState, FullTPOp, and TPPOVM to be Torchable; these are the classes used in "full TP" GST.

The Python file that contains TorchForwardSimulator also defines two helper classes: StatelessCircuit and StatelessModel. I think it's fine to keep these classes as purely internal implementation-specific constructs for now. Depending on future performance optimizations of TorchForwardSimulator we might want to put them elsewhere in pyGSTi.

#### What should come after this PR
We should compare performance of TorchForwardSimulator to MapForwardSimulator on problems of interest. There's a chance that the former isn't faster than the latter with the current implementation. If that's the case then I should look at possible performance optimizations specifically inside TorchForwardSimulator.

We should add implementations of ``stateless_data`` and ``torch_base`` to GST models beyond "Full TP" (in particular I'd like to try CPTP).

#### Incidental changes

My implementation originally interacted with the following evotype classes
```
    <class 'pygsti.evotypes.densitymx[_slow].statereps.StateRepDense'>
    <class 'pygsti.evotypes.densitymx[_slow].opreps.OpRepDenseSuperop'>
    <class 'pygsti.evotypes.densitymx[_slow].effectreps.EffectRepConjugatedState'>
```
When I write ``[_slow]`` in the class names above you can put the empty string or just ``_slow``, depending on the default evotype specified in ``evotypes.py``. 

To my surprise, I found that interacting with evotypes was neither necessary nor sufficient for what I wanted to accomplish. So while I did make changes in ``evotypes/densitymx_slow/`` to remove unnecessary class inheritances and to add documentation, those changes were only to make life a little easier for future pyGSTi contributors.


pygsti/tools/basistools.py
rileyjmurray, line 1: This change is to resolve a deprecation warning.

pygsti/modelmembers/states/densestate.py
rileyjmurray, line 1: This change is just to improve readability.

sserita:
I'm giving a few quick comments here because I think this PR will actually take me some time to get through.

The main thing I am concerned and thinking about is the choice to extend the ModelMember class as opposed to adjusting the evotypes. The general purpose of the split between modelmembers and evotypes is so that we don't have to go through and implement these abstract methods in all modelmembers - we can make the change in the evotype and then any modelmember works.
It is totally possible that extending the API is the best way to do this. The TermSimulator follows a similar pattern where the API extensions are the cleanest way to do it. But I'll probably spend some time thinking about whether this is way we want to implement this. Probably a point of discussion for us in our dev meetings in the upcoming weeks.

rileyjmurray:
@sserita, regarding
> The main thing I am concerned and thinking about is the choice to extend the ModelMember class as opposed to adjusting the evotypes. The general purpose of the split between modelmembers and evotypes is so that we don't have to go through and implement these abstract methods in all modelmembers - we can make the change in the evotype and then any modelmember works.

Unfortunately there's no way to do this just through evotypes. Using pytorch's AD capabilities requires knowing the free parameters in a modelmember _and_ how those free parameters map to the common parameterization-agnostic representation (i.e., representations in evotypes).

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -0,0 +1,238 @@
...
+
+from pygsti.forwardsims.forwardsim import ForwardSimulator
+
+# Below: variables for type annotations.
+#   We have to create variable aliases rather than importing the types
+#   directly, since importing the types would cause circular imports.
+Label = TypeVar('Label')
+ExplicitOpModel = TypeVar('ExplicitOpModel')
+SeparatePOVMCircuit = TypeVar('SeparatePOVMCircuit')
+CircuitOutcomeProbabilityArrayLayout = TypeVar('CircuitOutcomeProbabilityArrayLayout')
```
rileyjmurray, line 32: Try to use ``from __future__ import annotations`` to avoid use of TypeVars.

rileyjmurray:
Notes from today's meeting:
* I removed empty base classes for OpRep, EffectRep, StateRep. I should make sure we aren't doing type checking elsewhere in pyGSTi that would be broken by this. (There's a chance we are and it simply isn't showing up in current tests.)

rileyjmurray:
@coreyostrove, @sserita, @enielse: this is ready for review.

pygsti/modelmembers/povms/tppovm.py
```diff
@@ -56,3 +59,35 @@ def __reduce__(self):
...
+    @staticmethod
+    def torch_base(sd: Tuple[int, int], t_param: _Torchable.Tensor) -> _Torchable.Tensor:
+        torch = _Torchable.torch_handle
+        num_effects, dim = sd
+        first_basis_vec = torch.zeros(size=(1, dim), dtype=torch.double)
+        first_basis_vec[0,0] = dim ** 0.25
+        t_param_mat = t_param.reshape((num_effects - 1, dim))
+        t_func = first_basis_vec - t_param_mat.sum(axis=0, keepdim=True)
+        t = torch.row_stack((t_param_mat, t_func))
+        return t
```
rileyjmurray, line 93: @coreyostrove, @sserita:  is the ``dim ** 0.25`` scale appropriate when the underlying vector space is something other than "tensor product of qubit space"?

pygsti/modelmembers/states/tpstate.py
```diff
@@ -158,6 +160,17 @@ def from_vector(self, v, close=False, dirty_value=True):
...
         self.dirty = dirty_value
 
+    def stateless_data(self) -> Tuple[int]:
+        return (self.dim,)
+
+    @staticmethod
+    def torch_base(sd: Tuple[int], t_param: _Torchable.Tensor) -> _Torchable.Tensor:
+        torch = _Torchable.torch_handle
+        dim = sd[0]
+        t_const = (dim ** -0.25) * torch.ones(1, dtype=torch.double)
```
rileyjmurray, line 170: @coreyostrove, @sserita: same question as above. Is the scale factor of ``dim ** -0.25`` appropriate for general spaces?

rileyjmurray: From dev meeting:

Docs for the ``__init__`` function mention an assumption on the nature of the basis associated to this TPState. That function should be changed to explicitly assert that the identity matrix is an element of the basis and is in fact the first element of the basis.

https://github.com/sandialabs/pyGSTi/blob/1ec6909a7f9149c4f6758f9cd46256dbbff984c1/pygsti/modelmembers/states/tpstate.py#L60-L68

Make the same change for TPPOVM and FullTPOp.

pygsti/modelmembers/povms/tppovm.py
rileyjmurray: The verdict is that TPPOVM already restricted to the Pauli-product or Gell-Mann bases, in which situation this scaling is appropriate.

rileyjmurray:
@coreyostrove and @sserita this is ready for review.

coreyostrove (changes requested):
A true tour de force, @rileyjmurray!

I have left some comments and questions for you above.

One additional piece of feedback is that it would be great to add in some additional unit tests for the new `Torchable` methods, as well as those for the `StatelessModel` class and the `TorchForwardSimulator`. These are covered indirectly by the integration test, but including some correctness checks for the individual function calls would go a long way. Additionally, adding an integration test that confirms the various ForwardSimulator classes give approximately the same probability distributions (within some reasonable tolerance) for a few select circuits would be valuable. (All of that said, I am viewing this functionality as still in beta testing, so I likely won't gate keep based on adding these so long as we commit to doing so eventually).

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -0,0 +1,237 @@
...
+        The Tensors correspond to the current values of an object's parameters.
+        For the purposes of forward simulation, we intend that the following 
+        equivalence holds:
+        
+            model == (self, [dict returned by this function]).
+            
+        That said, the values in this function's returned dict need to be
+        formatted by get_torch_cache BEFORE being used in forward simulation.
+        """
+        free_params = OrderedDict()
```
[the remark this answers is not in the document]
rileyjmurray: The preservation of order does matter in this context, given how it's used in ``get_torch_cache``:
image


But I agree that it's fine to rely on the order-preservation properties of the standard ``dict`` class.

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -0,0 +1,237 @@
...
+            metadata = self.param_metadata[i]
+
+            fp_label = metadata[0]
+            fp_type  = metadata[1]
+            param_t = fp_type.torch_base(metadata[2], fp_val)
+            torch_cache[fp_label] = param_t
+        
+        return torch_cache
+
+    def circuit_probs(self, torch_cache: Dict[Label, torch.Tensor]):
```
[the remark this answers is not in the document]
rileyjmurray: I guess it could be in ``TorchForwardSimulator``. My inclination was to keep it in ``StatelessModel`` since its logic is tightly coupled with ``StatelessModel::get_torch_cache``.

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -0,0 +1,237 @@
...
+    def _bulk_fill_probs(self, array_to_fill, layout, stripped_abstractions: Optional[tuple] = None):
+        if stripped_abstractions is None:
+            slm, torch_cache = TorchForwardSimulator.separate_state(self.model, layout)
+        else:
+            slm, torch_cache = stripped_abstractions
+
+        layout_len = TorchForwardSimulator._check_copa_layout(layout)
+        probs = slm.circuit_probs(torch_cache)
+        array_to_fill[:layout_len] = probs.cpu().detach().numpy().flatten()
+        pass
```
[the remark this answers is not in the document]
rileyjmurray: No good reason. I'll switch to a return.

pygsti/modelmembers/torchable.py
```diff
@@ -0,0 +1,43 @@
...
+try:
+    import torch
+    torch_handle = torch
+    Tensor = torch.Tensor
+except ImportError:
+    torch_handle = None
+    Tensor = TypeVar('Tensor')  # we'll access this for type annotations elsewhere.
+
+
+class Torchable(ModelMember):
```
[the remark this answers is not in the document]
rileyjmurray: The specs of ``torch_base`` refer to properties and functions in the ModelMember API. Although the specs are given in documentation (not something that would necessarily result in Exceptions if not followed), they're important enough that I'd like to keep the link in Torchable's definition.

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -0,0 +1,237 @@
...
+        Returns a dict mapping pyGSTi Labels to PyTorch tensors. The dict makes it easy
+        to simulate a stateful model implied by (self, free_params). It is obtained by
+        applying invertible transformations --- defined in various ModelMember subclasses
+        --- on the tensors stored in free_params.
+
+        If ``grad`` is True, then the values in the returned dict are preparred for use
+        in PyTorch's backpropogation functionality. If we want to compute a Jacobian of
+        circuit outcome probabilities then such functionality is actually NOT needed.
+        Therefore for purposes of computing Jacobians this should be set to False.
+        """
```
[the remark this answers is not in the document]
coreyostrove: In general I think the answer is yes. Looking at a single circuit, the jacobian itself will have dimensions that are $d \times N_p$, where d is the Hilbert space dimension, and $d=2^n$ where $n$ is the number of qubits, and $N_p$ is the number of model parameters. The number of model parameters grows much faster than the dimension of the Hilbert space ($\mathcal(O)(16^n)$ vs $\mathcal(O)(2^n)$), so this will generically be much wider than it is tall. Of course, that is assuming fully parameterized models. Reduced models (like weight 1+2 error generator models) scale differently. There is also the confounding factor that we only need to compute the cost function for outcomes we actually observe, in the context of MLE. Anyhow, all of this is to say that while I suspect we will find reverse-mode advantageous, figuring out exactly the parameter regimes where that is true will indeed require experimentation. 

Thoughts on adding a handle on this behavior?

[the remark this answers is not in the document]
rileyjmurray: @coreyostrove, the Jacobian calculation isn't done one circuit at a time; it's done with consideration to all circuits at once. I agree that if we're looking at an individual circuit then the map from model parameters to circuit outcome probabilities would be wide.

That said, I've modified the code to automatically select between forward-mode AD and reverse-mode AD depending on the aspect ratio of the Jacobian :)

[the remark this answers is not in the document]
coreyostrove: Awesome, great work!

pygsti/forwardsims/torchfwdsim.py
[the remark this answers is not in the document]
coreyostrove: Good enough for me

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -0,0 +1,237 @@
...
+        else:
+            slm, torch_cache = stripped_abstractions
+
+        layout_len = TorchForwardSimulator._check_copa_layout(layout)
+        probs = slm.circuit_probs(torch_cache)
+        array_to_fill[:layout_len] = probs.cpu().detach().numpy().flatten()
+        pass
+
+    def _bulk_fill_dprobs(self, array_to_fill, layout, pr_array_to_fill):
+        slm = StatelessModel(self.model, layout)
```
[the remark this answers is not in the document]
rileyjmurray: Tiny note up front: I've renamed the variable that you're referencing as``stripped_abstractions``. It's now called ``split_model``. Now to answer your question ...

For _bulk_fill_probs (no "d") I was able to add an optional argument since there's a codepath in _bulk_fill_dprobs (with "d") that requires calling _bulk_fill_probs. In order to make use of an extra argument here we'd need to override the functions in the ForwardSimulator base class that call this function. I've added a TODO for us to explore this later.

pygsti/forwardsims/torchfwdsim.py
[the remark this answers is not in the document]
rileyjmurray: I started writing documentation but then I realized the code was too messy. So I went back and did a general refactor, and _then_ I added docs. Please take another look and LMK if it's good!

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -19,206 +25,240 @@
...
 
     def __init__(self, spc: SeparatePOVMCircuit):
         self.prep_label = spc.circuit_without_povm[0]
-        self.op_labels  = spc.circuit_without_povm[1:]
+        if len(spc.circuit_without_povm) > 1:
+            self.op_labels  = spc.circuit_without_povm[1:]
+        else:
+            # Importing this at the top of the file would create a circular dependency.
+            from pygsti.circuits.circuit import Circuit
+            self.op_labels = Circuit(tuple())
```
coreyostrove, line 57: I have learned a bit more since I originally commented on this line, and it turns out that what you had actually does work. It turns out that, while if you were to `spc.circuit_without_povm[1]` on a length 1 Circuit you'd indeed get an attribute error, when grabbing the slice `spc.circuit_without_povm[1:]` this instead returns the empty circuit. We can debate how sensible that is on a design level (it turns out a lot of the code knows about and uses this fact, however, so changing this behavior would be nontrivial), but it does work here. There is also an argument to be made that this might be slightly preferable in this case, as the sliced circuit inherits the `line_labels` value directly from the parent, which you'd otherwise need to add some logic to grab in creating it directly as above.

So my recommendation would be to revert this back to how you originally had it. Apologies for the misdirection.

rileyjmurray: Man, you've really gotten into the weeds on this stuff, lol. Happy to revert, and no worries!

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -19,206 +25,240 @@
...
         self.povm_label = spc.povm_label
+        self.outcome_probs_dim = len(spc.effect_labels)
```
coreyostrove, line 54: Note for future development. I think this should be fine for circuits with only POVMS (as is currently supported), but I don't believe this will give you the correct value when instruments are present. OpModels have a method called `circuit_outcomes` which should handle the legwork of figuring this all out when instruments are present, though (it is more complicated to implement this than it seems, so it is always good when someone else figured it out already). Not action items now, just something to keep in mind whenever we decide to implement torchable instruments.

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -19,206 +25,240 @@
...
         return
+    
+    def outcome_probs(self, torch_bases: Dict[Label, torch.Tensor]) ->  torch.Tensor:
```
coreyostrove, line 57: I am not a fan of this change, and am coincidentally presently advocating for removing a comparable method from the `Circuit` class as part of PR #445. TLDR: In the abstraction hierarchy we typically use nowadays `Circuits` are used for description of a quantum process in terms of elementary operations. Models on the other hand, among other things, are responsible for taking as input that description of a quantum process and returning a probability distribution (part of which it itself farms out to a child ForwardSimulator).

rileyjmurray: Understood. I've just pushed a commit that goes back to the old approach. Everything now happens directly in ``StatelessModel.circuit_probs_from_torch_bases``.

pygsti/evotypes/densitymx_slow/effectreps.py
```diff
@@ -13,12 +13,13 @@
 import numpy as _np
 
 # import functools as _functools
-from .. import basereps as _basereps
 from pygsti.baseobjs.statespace import StateSpace as _StateSpace
 from ...tools import matrixtools as _mt
 
 
-class EffectRep(_basereps.EffectRep):
+class EffectRep:
```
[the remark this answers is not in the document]
sserita: My two cents: It is a relatively common (IMO) strategy to have a common base class to make checking via things like isinstance() straightforward. In this case, one could do isinstance(obj, _basereps.EffectRep) to check for an EffectRep from any evotype. Unless there is a performance reason to not do this inheritance, I'm somewhat opposed to just removing it - that seems to be no benefit for actively losing a feature.

rileyjmurray:
@coreyostrove I still plan on adding proper unit tests (in contrast to the existing tests, which are really integration tests).

pygsti/evotypes/densitymx_slow/opreps.py
```diff
@@ -26,7 +25,11 @@
 from ...tools import optools as _ot
 
 
-class OpRep(_basereps.OpRep):
+class OpRep:
```
sserita, line 28: Same thing here, I don't see a compelling reason to remove this

pygsti/forwardsims/forwardsim.py
```diff
@@ -373,7 +373,7 @@ def create_layout(self, circuits, dataset=None, resource_alloc=None,
             if 'epp' in array_types:
                 derivative_dimensions = (self.model.num_params, self.model.num_params)
             elif 'ep' in array_types:
-                derivative_dimensions = (self.model.num_params)
+                derivative_dimensions = (self.model.num_params,)
```
sserita, line 376: I thought we had already fixed this at some point? If not, glad we are getting it in now. :)

pygsti/modelmembers/states/tpstate.py
```diff
@@ -11,16 +11,25 @@
 #***************************************************************************************************
 
 
-import numpy as _np
+from __future__ import annotations
+from typing import Tuple, TYPE_CHECKING
+if TYPE_CHECKING:
+    import torch as _torch
```
sserita, line 17: Could this raise ImportError that would not be caught by the block below and be a problem?

pygsti/modelmembers/torchable.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def stateless_data(self) -> Tuple:
+        """
+        Return this ModelMember's data that is considered constant for purposes of model fitting.
+
+        Note: the word "stateless" here is used in the sense of object-oriented programming.
+        """
+        raise NotImplementedError()
```
sserita, line 31: No action items, but what is your stance on this strategy of raising NotImplemented over ABC/@abstractmethod and pass?

sserita (commented):
I think everything looks good so far. I'd prefer that we roll back the change removing inheritance from basereps, but otherwise my other comments are more questions than explicit action items.
I'm withholding explicit approval until 0.9.13 goes out just as a reminder that we don't want this to merge in before then, and it sounds like you are still working on getting in a few tests as well.

coreyostrove:
> I'm withholding explicit approval until 0.9.13 goes out just as a reminder that we don't want this to merge in before then, and it sounds like you are still working on getting in a few tests as well.

Minor clarification, did you mean 0.9.12.3?

pygsti/forwardsims/forwardsim.py
rileyjmurray: I also thought we fixed it. Not sure what happened with the original intended fix.

pygsti/modelmembers/states/tpstate.py
rileyjmurray: Nope! It looks weird, but ``TYPE_CHECKING`` is a constant that's equal to ``False`` at runtime.

pygsti/modelmembers/torchable.py
rileyjmurray: I don't have much experience with ABC/abstractmethod myself. I just looked into it and it seems the main difference is when an error would be raised.

* If you use ABC/abstractmethod then an error will be raised as soon as Python detects that a subclass fails to provide its own implementation of the method.
* If you use the NotImplementedError approach then an error would be raised if (and only if) the function was called at runtime.

sserita:
Yes I absolutely meant 0.9.12.3. Speaking of which, Tim just approved that so it will be going live as soon as all the tests pass and I get the release drafted up.

So just let me know when the tests are good on this branch and I'll give my approval also.

pygsti/evotypes/densitymx_slow/effectreps.py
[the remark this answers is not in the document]
sserita: Mmmm fine. It seems I'm being overruled 2 to 1 on this, so we'll go with your approach and I will just grumble off on the side, and keep an eye out on if there is ever a point at which there is a common API which would justify a base class :P

pygsti/modelmembers/torchable.py
sserita: Yea that's my understanding too, I've been playing around with ABC more recently so was just farming for more opinions :)

pygsti/evotypes/densitymx_slow/effectreps.py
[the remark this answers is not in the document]
coreyostrove: Should @rileyjmurray and I be worried about DM retribution in our next session? 🙃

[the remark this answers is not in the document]
sserita: Never :)

rileyjmurray:
@sserita, @coreyostrove all tests pass, including new tests that I wrote in ``test_forwardsim.py``. Merge when ready!

sserita (approved):
Thanks for all this great work @rileyjmurray! This looks good to me, will merge as soon as beta is clear/0.9.12.3 is out.

PR #392, opened 2024-01-22
Fix convolve import for scipy 3.12

### LINKED ISSUE #391: Pygsti is incompatible with scipy 1.12
**Describe the bug**
 Using pygsti fails on the import of colvolve from scipy (which is no longer available from the main scipy namespace in version 1.12)

**To Reproduce**
1. Install scipy 1.12 (latest on pypi) and pygsti
2. Using pygsti fails on the import of colvolve from scipy at line

https://github.com/sandialabs/pyGSTi/blob/7411a00ec81e67f50932b705e46426fc2d4701d8/pygsti/extras/drift/signal.py#L13

 - pyGSTi version [e.g. v0.9.7.4] '0.9.10.1.post449'
 - python version [e.g. 3.7, 2.7] 3.11
 - OS [e.g. OSX 10.14, Ubuntu 16.04 LTS] Windows

eendebakpt:
Replace the import of convolve from scipy to numpy. Fixes #391

### DIFF
--- a/pygsti/extras/drift/signal.py
+++ b/pygsti/extras/drift/signal.py
@@ -10,7 +10,7 @@
 
 import numpy as _np
 import numpy.random as _rnd
-from scipy import convolve as _convolve
+from numpy import convolve as _convolve
 from scipy.fftpack import dct as _dct
 from scipy.fftpack import fft as _fft
 from scipy.fftpack import idct as _idct


sserita (approved):
This looks fine, thanks for catching this @eendebakpt!
For posterity, there is also `scipy.signal.convolve` if we ever need to generalize to N-dimensional arrays instead of just the 1D we are using now.

PR #393, opened 2024-01-23
Bugfix germsel correctness tests

coreyostrove:
This PR updates the germ selection correctness/integration checks added to the unit tests in `test_packages` in a recent update to develop. These tests can be sensitive to various difference among OS and python version and so were failing on some runners. Additional solutions have been added to the tests to account for these local variations, and we now have all of the tests in the `test_packages` workflow passing. With the exception of python 3.11 on macOS which is failing because of an unrelated build error with CVXOPT.

~Note: New commits are 52a793ddf1ccab2f71fd0df365caa4fde828b2aa and onward. The other commits are from the bugfix-stricter-line... branch which has already been merged in, but they still seem to show in the history as new in this branch.~ It is because I had accidentally selected to merge into master, fixed now.

### DIFF
--- a/test/test_packages/algorithms/test_germselection.py
+++ b/test/test_packages/algorithms/test_germselection.py
@@ -7,20 +7,36 @@ from ..algorithms.algorithmsTestCase import AlgorithmTestCase
 
 class GermSelectionTestData(object):
     germs_greedy = {Circuit([Label('Gxpi2',0)]), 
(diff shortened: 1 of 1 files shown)


sserita (approved):

sserita:
I merged this in now because these tests were now the failing ones on beta. I'd like to get a clean set of tests before merging in other branches in for the release.

Issue #380, opened 2023-12-02
Faster unit tests/notebooks


sserita:
**Is your feature request related to a problem? Please describe.**
The unit and notebook regression tests take too long.

**Describe the solution you'd like**
Faster unit tests. This could be achieved by using different inputs or options which don't affect coverage but do affect runtime. Simple examples for GST related tests include using lower germ lengths, smaller gate sets, etc. This was already done for the `test/test_packages/algorithms` tests in #372, but we should do this across the whole codebase.

**Task list**
Below are tasklists made by running pytest with `--durations 0` and taking the most expensive tests.

__Everything in `test/unit` around 5s or above__:

- [ ] 37.93s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptAllPenaltyTester::test_gaugeopt_with_none_gauge_group                                                                                 
- [ ] 37.87s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptAllPenaltyTester::test_gaugeopt                                                                                                                                                                                                                             
- [ ] 19.16s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptSPAMPenaltyTester::test_gaugeopt_with_none_gauge_group                                                                                
- [ ] 19.09s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptSPAMPenaltyTester::test_gaugeopt                                                                                                      
- [ ] 17.01s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptAutoMethodTester::test_gaugeopt_gates_metrics                                                                                         
- [ ] 11.64s call     test/unit/algorithms/test_gaugeopt.py::CPTPGaugeOptSPAMPenaltyTester::test_gaugeopt                                                                                                      
- [ ] 11.05s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptAllPenaltyTester::test_gaugeopt_no_target                                                                                             
- [ ] 8.91s call     test/unit/algorithms/test_gaugeopt.py::LGSTGaugeOptAutoMethodTester::test_gaugeopt_spam_metrics                                                                                           
- [ ] 8.47s call     test/unit/tools/test_optools.py::ProjectModelTester::test_logTiG_model_projection                                                                                                         
- [ ] 8.19s call     test/unit/tools/test_optools.py::ProjectModelTester::test_logGTi_model_projection                                                                                                         
[...]

PR #403, opened 2024-02-23
Feature faster testing stuff

coreyostrove:
This PR contains primarily a set of updates for the unit tests and for the tutorial/example notebooks aimed at reducing their runtime in the CI setting. This also includes some changes modernizing tests/notebooks moving away from use of legacy modelpacks. See issue #380 for a checklist summarizing the notebooks and tests which were identified as bottlenecks which have been addressed. In all cases the changes made are ones that reduce computational requirements without meaningfully impacting either the test coverage, or the message conveyance of a tutorial. E.g. switching from the XYI to the XY gate set in an example to reduce parameter counts (making stuff run more quickly) while still demoing the intended feature/analysis. With all of these changes in place the notebook regression checks now run in almost exactly half the time (at least from spot checking a couple ubuntu runners comparing two-weeks ago to today).

In the mix are also a couple of minor bugfixes/codebase changes. These are:

- Set the default maximum iteration count for the optimizer to 100 when performing GST using the driver methods. Down from 100000 (essentially unlimited). This brings the default behavior inline with the defaults when using the OO interface using Protocol objects.
- Some modernizations and tweaks to the qutrit model building helpers to conform with new line labeling behavior, and a switch to using TP constrained modelmembers by default.
- Bugfix for StandardGST that ensures we make a copy of the target model rather than optimizing it in-place. I swear we had already done this fix well over a year ago, but perhaps this was inadvertently overwritten at some point.


sserita (approved):
Thanks for all your hard work on optimizing these tests Corey. I feel pretty good about our runtime on the workers now.

PR #407, opened 2024-03-05
Fixes bug where depolarizing noise strength differs with parameterization

enielse:
Depolarizing noise constructed with `depolarization_parameterization="lindblad"` doesn't match the same constructio when `depolarization_parameterization="depolarize"`, as demonstrated by the below code.  This is due to the use of the normalized Pauli basis (`"pp"`) where we should be using the un-normalized basis (`"PP"`).  This PR fixes this issue and adds a unit test.

To see the bug, run this (printed output should be the same but is not):
```
import pygsti
from pygsti.modelmembers.operations.opfactory import ComposedOpFactory
from pygsti.modelmembers.operations.depolarizeop import DepolarizeOp

pspec = pygsti.processors.QubitProcessorSpec(1, ["Gi"], geometry="line")

mdl1 = pygsti.models.create_crosstalk_free_model(
        pspec,
        depolarization_parameterization="lindblad",
        depolarization_strengths={'Gi': 0.02}
)

mdl2 = pygsti.models.create_crosstalk_free_model(
        pspec,
        depolarization_parameterization="depolarize",
        depolarization_strengths={'Gi': 0.02}
)

c = pygsti.circuits.Circuit("Gi:0@(0)")

print(dict(mdl1.probabilities(c)))
print(dict(mdl2.probabilities(c)))
```

### DIFF
--- a/pygsti/models/modelnoise.py
+++ b/pygsti/models/modelnoise.py
@@ -799,7 +799,7 @@ class DepolarizationNoise(OpNoise):
 
         # LindbladErrorgen with "depol" or "diagonal" param
         basis_size = state_space.dim  # e.g. 4 for a single qubit
-        basis = _BuiltinBasis('pp', basis_size)
+        basis = _BuiltinBasis('PP', basis_size)
         rate_per_pauli = self.depolarization_rate / (basis_size - 1)
         errdict = {('S', bl): rate_per_pauli for bl in basis.labels[1:]}
         return _op.LindbladErrorgen.from_elementary_errorgens(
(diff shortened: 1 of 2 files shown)


coreyostrove (approved):
This all looks good to me, thanks @enielse! One downside I notice about the manner external PRs are handled is that we don't get the results of the github actions passes merged into the PR as is typically convenient. Nonetheless I can see from the github action history on the IonQ fork that the standard test suite is all passing. So this all looks good to me.

sserita (approved):

PR #409, opened 2024-03-19
Check if outcome_labels is of type numbers.Integral instead of np.int64

OliverGorman:
`pygsti.data.Dataset` constructor has the `outcome_labels` argument which may be a `list` or `int` according to the docstring. However, passing a Python `int` did not work as the `isinstance` check was using `np.int64`. This pull request generalises the check to also work with Python `int`s using the `numbers` standard library module.

```
>>> import numbers, numpy as np
>>> isinstance(1, np.int64)
False
>>> isinstance(1, int)
True
>>> isinstance(np.int64(1), int)
False
>>> isinstance(np.int64(1), numbers.Integral)
True
>>> isinstance(1, numbers.Integral)
True
```

... after writing this up I see that pull requests by authors outside of SNL cannot be merged. I'm more than happy for a maintainer to take this change and put it in under their name.

### DIFF
--- a/pygsti/data/dataset.py
+++ b/pygsti/data/dataset.py
@@ -1031,7 +1031,7 @@ class DataSet(_MongoSerializable):
             self.olIndex = outcome_label_indices
             self.olIndex_max = max(self.olIndex.values()) if len(self.olIndex) > 0 else -1
         elif outcome_labels is not None:
-            if isinstance(outcome_labels, _np.int64):
+            if isinstance(outcome_labels, _numbers.Integral):
                 nqubits = outcome_labels
                 tup_outcomeLabels = [("".join(x),) for x in _itertools.product(*([('0', '1')] * nqubits))]
             else:


sserita:
Hey Oliver,
Thanks for the fix! If you think you will be making more contributions in the future, we can set you up with a contributor license agreement (CLA) which would allow us to accept your PRs. If not, I'll just make this bugfix independently.
Please email us at pygsti@sandia.gov for a CLA if you decide to go down that route.

sserita:
Hi Oliver,
Thanks again for your contribution! We've included an equivalent bugfix in [ef9def6](https://github.com/sandialabs/pyGSTi/commit/ef9def6205543f08a9047b155d761a500bf8c577) that we think will be more cross-platform friendly, so I'm going to close this PR. Again, if you find yourself wanting to contribute in the future, please feel free to reach out for a CLA.

PR #410, opened 2024-03-19
Leakage-aware GST

rileyjmurray:
This PR significantly expands pyGSTi's leakage modeling capabilities. 

*Note.* This PR description was written after [Corey's approval](https://github.com/sandialabs/pyGSTi/pull/410#pullrequestreview-3132748543) and a few follow-up commits from me.

### Important stuff in new files

jupyter_notebooks/Examples/Leakage_automagic.ipynb. This illustrates how data from a 1-qubit GST experiment can be used to fit a 1-qutrit model, and how to generate reports that include leakage analysis of that 1-qutrit model.

pygsti/tools/leakage.py. I've divided this file into four regions.
 * _Report generation_. Right now there's one function here,``construct_leakage_report``, which handles gauge optimization and report generation. It returns a Report object and an updated ModelEstimateResults object.
 * _Metrics_. We have several new gate metrics of use when we don't care about how gates differ on "leakage space." These functions are called with ``n_leak=1`` when a user calls ``construct_leakage_report``. Users are also free to call these functions themselves.
 * _Model construction_. Right now this consists of one function, ``leaky_qubit_model_from_pspec``, which creates a three-level ExplicitOpModel by lifting a qubit processor spec.
 * _Gauge optimization_. Of the three functions here, only ``add_lago_models`` is particularly user-facing. It adds additional "leakage-aware gauge optimized" or "LAGO" models to the provided ModelEstimateResults object. It's called automatically in ``construct_leakage_report``.

pygsti/tools/sdptools.py. This has pyGSTi's semidefinite programming modeling code. Some of its contents were pulled from pygsti/tools/optools.py. Much of its contents are new. The ``diamond_distance_projection_model`` function  lets us compute a superoperator's diamond-distance projection onto certain convex sets of superoperators (namely, any intersection of the sets of CPTP operators, leakage-free operators, and seepage-free operators); this same function lets us compute the projection with respect to the standard diamond norm and a "subspace" diamond norm.

#### Things with higher user-exposure

1. Calling ``b = BuiltinBasis('l2p1', 9)`` gets you a special leakage basis for nine-dimensional Hilbert-Schmidt space.
The basis reflects a canonical separation of its underlying Hilbert space into a direct sum of a 2-dimensional space and a 1-dimensional space. In the future we extend this to ``l{x}p{y}`` for integers ``x >= 2`` and ``y >= 1``.

2. pygsti/models/explicitmodel.py has a new free function:
   ```python
    transform_composed_model(mdl : ExplicitOpModel, s: GaugeGroupElement) -> ExplicitOpModel
    ```
    which uses ComposedOp, ComposedState, and ComposedPOVM to build a gauge-transformed copy of ``mdl`` that has the same parameterization as ``mdl``. 

3. The ``jamiolkowski_iso`` and ``inv_jamiolkowski_iso`` functions in pygsti/tools/jamiolkowski.py can now operate on CVXPY Expression objects. Separately, all Jamiolkowski functions in this file new accept an optional ``normalized`` argument that defaults to True, indicating if we're working with the normalized (inverse) Choi matrix.

4. The LazyBasis class (which is a parent of BuiltinBasis) has a new member called ``elindlookup``. This is a lazily-constructed dict where 
    ```python
    element = basis.elements[basis.elindlookup[element_label]]
    ```
     is the basis element corresponding to ``element_label``.

5. Our LinearOperator class (a subclass of ModelMember) now has a ``.shape`` parameter, in order to provide API similarity with common Python array types.

#### Things with low user-exposure that could be higher

I added several new functions to pygsti/report/reportables.py. Nearly all of these functions also has a class representation generated by the ``opfn_factory`` function from pygsti/report/modelfunction.py. The descriptions of these functions can be found in the docstring and implementation of ``info_of_opfn_by_name``. Functions that appear in pygsti/tools/leakage.py have their own documentation. I'm leaving it to future work to include all leakage-related metrics in pygsti/tools/leakage.py.

#### Things with low exposure that should stay that way

The ConfidenceRegionFactoryView class has a new instance method called ``compute_grad_f``. This is for computing the gradient of a ModelFunction object at a given point. Evaluating ``compute_grad_f`` constitutes most of the work in a call to ConfidenceRegionFactoryView.compute_confidence_interval.

pygsti/tools/matrixtools.py has a new class called ``IdentityOperator``. Objects of this class implement ``@``, ``.T``, and ``.conj()`` as though they were numpy ndarrays. However, these operations are "no-ops," and so cost basically nothing. This class is used in new gauge optimization code. It comes with a free function called ``is_operatorlike`` for checking if something can plausibly be used as a linear operator with a numpy-like API.

The gaugeopt_to_target function in pygsti/algorithms/gaugeopt.py accepts an ``n_leak`` argument that's a nonnegative integer and that defaults to 0. Choices of ``n_leak > 0`` cannot be used with the default optimizer (``"ls"``); one should request ``"L-BFGS-B"`` instead.

The instance method ``ExplicitOpModelCalc.frobeniusdist`` now has a different behavior for the ``transform_mx`` argument. You can pass in None or an ndarray (as in previous code) or you can pass in a pair of objects where ``matrixtools.is_operatorlike(obj)`` returns True.

### Other notes

We now test for correctness of gauge optimization code in test/unit/algorithms/test_gaugeopt_correctness.py. Old tests (which were just smoke tests) were moved to test/unit/algorithms/test_gaugeopt_noerrors.py.

The code in the "Leakage-automagic" example notebook is run as a smoke test in test/unit/tools/test_leakage_gst_pipeline.py.

Some tests were added to test/unit/tools/test_optools.py.

Unresolved comments:
 * https://github.com/sandialabs/pyGSTi/pull/410/files#r2255468471
 * https://github.com/sandialabs/pyGSTi/pull/410/files#r2255477604 (spelling)
 * https://github.com/sandialabs/pyGSTi/pull/410/files#r2255482265 (spelling)
 *https://github.com/sandialabs/pyGSTi/pull/410/files#r2255546335 (request for test to trigger a new error handling codepath).

Future work:
 * systems with > 1 qubit.
 * systems with > 1 leakage level.
 * CPTP kite models


pygsti/tools/matrixtools.py
rileyjmurray, line 1: These changes are largely incidental to this PR. What we should really do with this functions is find everywhere they're called and then just in-line their one-line implementations.

pygsti/algorithms/gaugeopt.py
```diff
@@ -456,31 +453,33 @@ def _jacobian_fn(gauge_group_el):
...
+            objective: rho_term = (rho - S * target_rho)
+            jac:       d(rho_term) = - dS * target_rho
 
-            # objective: ET_term = (E.T - target_E.T * S_inv)
-            # jac:       d(ET_term) = - target_E.T * -(S_inv * dS * S_inv)
-            #            d(ET_term) = target_E.T * (S_inv * dS * S_inv)
+            objective: ET_term = (E.T - target_E.T * S_inv)
+            jac:       d(ET_term) = - target_E.T * -(S_inv * dS * S_inv)
+                       d(ET_term) = target_E.T * (S_inv * dS * S_inv)
+            """
```
rileyjmurray, line 482: This is just formatting changes.

pygsti/protocols/gst.py
rileyjmurray, line 1: Changes needed for leakage analysis of the kind I run in existing notebooks.

pygsti/report/factory.py
rileyjmurray, line 1: Changes core to leakage analysis (and which have benefits in other contexts).

pygsti/report/reportables.py
rileyjmurray, line 1: Changes core to leakage analysis. (Plus error handling of value more broadly.)

rileyjmurray: Need to resolve some "TO WRITE" notes in this file.

pygsti/report/section/gauge.py
rileyjmurray, line 1: Changes core to leakage analysis.

pygsti/report/section/summary.py
rileyjmurray, line 1: Changes core to leakage analysis.

pygsti/report/templates/standard_html_report/tabs/GaugeVariants.html
rileyjmurray, line 1: Need to discuss if we want to keep this change; see also pygsti/report/templates/standard_pdf_report.tex.

pygsti/tools/jamiolkowski.py
rileyjmurray, line 1: Not leakage-specific, but necessary for leakage analysis.

test/unit/tools/test_optools.py
rileyjmurray, line 1: Modest tests for code used in leakage analysis

rileyjmurray (commented):
Leaving file-level comments so @coreyostrove and I can figure out what to split out from this PR.

pygsti/modelmembers/povms/tppovm.py
rileyjmurray, line 1: readability

pygsti/modelmembers/states/densestate.py
rileyjmurray, line 1: Figuring this one out was a doozy. I'm certain that I could make a test which caused the existing code to break without doing anything leakage-specific.

pygsti/algorithms/gaugeopt.py
```diff
@@ -594,18 +591,34 @@ def _mock_objective_fn(v):
...
+        else:
+            transform_mx_arg = (_np.eye(mxBasis.dim), None)
+
+        assert gates_metric != "frobeniustt"
+        assert spam_metric  != "frobeniustt"
+        # assert spam_metric == gates_metric
+        # ^ Erik and Corey said these are rarely used. I've removed support for
+        # them in this codepath (non-LS optimizer) in order to make it easier to
+        # read my updated code for leakage-aware metrics. It wouldn't be hard to
+        # add support back, but I just want to keep things simple. -- Riley
```
rileyjmurray: @coreyostrove, the current changes remove support for the 
transform-target Frobenius norm loss when ``_create_objective_fn``(in gaugeopt.py) is passed a value for ``method`` other than ``"ls"``. Does this change sound okay to you? If so, I'll adjust the comment accordingly.

pygsti/circuits/circuit.py
```diff
@@ -977,11 +977,10 @@ def __eq__(self, x):
...
             if len(self) != len(x):
                 return False
+            elif self._static and x._static and self._hash != x._hash:
+                return False
             else:
-                if self._static and x._static:
-                    return self._hash == x._hash
-                else:
-                    return self.tup == x.tup
+                return self.tup == x.tup
```
rileyjmurray, line 983: This change is necessary for correctness.

pygsti/data/dataset.py
```diff
@@ -1292,11 +1292,28 @@ def gate_labels(self, prefix='G'):
...
+                        opLabels.add(layer.name)
+                else:
+                    raise ValueError()
+
+        bracket_idle = '[]' in opLabels
+        opLabels = [op for op in opLabels if op != '[]']
+        opLabels.sort()
+        if bracket_idle:
+            opLabels = ['[]'] + opLabels
+
```
rileyjmurray, line 1316: TODO: create a test case showing why this is needed, or revert this change.

pygsti/modelmembers/modelmembergraph.py
```diff
@@ -19,8 +19,8 @@
 class ModelMemberGraph(object):
     """A directed acyclic graph of dependencies of ModelMembers"""
 
-    @classmethod
-    def load_modelmembers_from_serialization_dict(cls, sdict, parent_model):
+    @staticmethod
+    def load_modelmembers_from_serialization_dict(sdict, parent_model):
```
rileyjmurray, line 23: There are no child classes of ModelMemberGraph, so there's no benefit to a classmethod over a staticmethod.

pygsti/models/explicitcalc.py
```diff
@@ -241,47 +259,30 @@ def residuals(self, other_calc, transform_mx=None, item_weights=None):
...
+            resids.append(resid)
+            nSummands += wt**2 * rhoV.dim
+
+        for lbl, Evec in self.effects.items():
+            wt = sqrt_itemWeights.get(lbl, spamWeight)
+            other_effect = other_calc.effects[lbl]
+            resid = wt * Evec.residuals(other_effect, T, Ti)
+            resids.append(resid)
+
+            nSummands += wt**2 * Evec.dim
```
rileyjmurray, line 264: I collapsed this ``if-else`` branching to always use the ``if`` branch. All I needed to do was to set Ti to None if T is None.

pygsti/models/explicitcalc.py
```diff
@@ -609,17 +610,13 @@ def nongauge_and_gauge_spaces(self, item_weights=None, non_gauge_mix_mx=None):
...
-        nongauge_space = _mt.nullspace_qr(orthog_to.T)  # cols are non-gauge directions
-        # print("DB: nullspace of gen_dG (shape = %s, rank=%d) = %s" \
-        #       % (str(gen_dG.shape),_np.linalg.matrix_rank(gen_dG),str(gen_ndG.shape)))
-
-        #REMOVE
-        ## reduce gen_dG if it doesn't have full rank
-        #u, s, vh = _np.linalg.svd(gen_dG, full_matrices=False)
-        #rank = _np.count_nonzero(s > P_RANK_TOL)
-        #if rank < gen_dG.shape[1]:
-        #    gen_dG = u[:, 0:rank]
```
rileyjmurray: A change along this lines is needed to satisfy the assertion at the end of this function.  The problematic case is when we hit the ``item_weights is not None`` codepath and the matrix ``metric = _np.diag(metric_diag) `` ends up being singular (for example, because all SPAM weight was zero). 

This would be the first change to this function since it was introduced in [this commit](https://github.com/sandialabs/pyGSTi/commit/159efbf6b19579ac9be618839bea99f8913dc044#diff-2cde56589c9c701a958943ff86c4f7d5c7d2c8d6b8eba42ee081b6d535a22b9f).

This function is called twice in this file. One call is a diagnostic/testing function that was only written for development purposes. The other call is in ExplicitOpModelCalc.nongauge_projector. If you look at how the output of this function is used in that function, you'll see that there's no harm in this function returning orthonormal bases for the two subspaces. In fact, ExplicitOpModelCalc.nongauge_projector is just doing extra work given an assumption of orthonormality. I'd prefer to _not_ change that function in this PR because we can take care of that in the audit for np.linalg.inv (see #454).

This function is also called six times in confidenceregionfactory.py. Some of those calls provide values for item_weights or non_gauge_mix_mx. As far as I understand, there's no harm in us returning orthonormal bases from this function in those contexts.

pygsti/report/reportables.py
```diff
@@ -2443,6 +2545,8 @@ def info_of_opfn_by_name(name):
     info = {
         "inf": ("Entanglement|Infidelity",
                 "1.0 - <psi| 1 x Lambda(psi) |psi>"),
+        "la-inf": ("Entanglement|Infidelity (subspace)",
+                   "TO-WRITE"),
```
rileyjmurray, line 2549: Write me.

pygsti/report/reportables.py
```diff
@@ -2475,6 +2589,8 @@ def info_of_opfn_by_name(name):
                         "where (a_i,b_i) are corresponding eigenvalues of A and B."),
         "frob": ("Frobenius|Distance",
                  "sqrt( sum( (A_ij - B_ij)^2 ) )"),
+        "la-frob": ("Frobenius|Distance (subspace)",
+                    "TO WRITE"),
```
rileyjmurray, line 2593: Write me.

pygsti/report/templates/standard_html_report/tabs/GaugeVariants.html
```diff
@@ -7,8 +7,8 @@ <h1>Gauge Variant Error Metrics</h1>
...
-	<figcaption><span class="captiontitle">Individual gate error metrics</span> <span class="captiondetail">This table presents various (gauge-variant) metrics that quantify errors in each individual estimated logic gate, with respect to the ideal target gates.  Note that "Entanglement infidelity" and "Average gate infidelity" are two common definitions of process fidelity, and related by a constant dimensional factor.  A description of each metric can be found by  hovering the pointer over the column header.</span></figcaption>
-    {{ final_gates_vs_target_table_gauge_var|render }}
+	<figcaption><span class="captiontitle">Individual gate error metrics</span> <span class="captiondetail">This table presents various (gauge-variant) metrics that quantify errors in each individual estimated logic gate, with respect to the ideal target gates. Note that "Entanglement infidelity" and "Average gate infidelity" are two common definitions of process fidelity, and related by a constant dimensional factor. A description of each metric can be found by hovering the pointer over the column header.</span></figcaption>
+	{{ final_gates_vs_target_table_gauge_var|render }}
```
rileyjmurray, line 11: Typesetting.

pygsti/report/templates/standard_pdf_report.tex
```diff
@@ -409,7 +409,7 @@ \section{Gauge-dependent Error Metrics}
     \begin{adjustbox}{max width=\textwidth}
       \putfield{final_gates_vs_target_table_gauge_var}{}
     \end{adjustbox}
-    \caption{\textbf{Individual gate error metrics} This table presents various (gauge-variant) metrics that quantify errors in each individual estimated logic gate, with respect to the ideal target gates.  Note that "Entanglement infidelity" and "Average gate infidelity" are two common definitions of process fidelity, and related by a constant dimensional factor.  A description of each metric can be found by  hovering the pointer over the column header.}
+    \caption{\textbf{Individual gate error metrics} This table presents various (gauge-variant) metrics that quantify errors in each individual estimated logic gate, with respect to the ideal target gates. Note that "Entanglement infidelity" and "Average gate infidelity" are two common definitions of process fidelity, and related by a constant dimensional factor. A description of each metric can be found by hovering the pointer over the column header.}
```
rileyjmurray, line 412: Typesetting.

pygsti/baseobjs/basis.py
```diff
@@ -1607,6 +1607,7 @@ def __init__(self, component_bases, name=None, longname=None):
         # Note: this assertion also means dim == product(component_elsizes) == elsize, so basis is *simple*
```
nkoskelo: There are some misspellings within the file which would be nice to correct.
- "natrual" is misspelled on line 1601 and line 75.
- "woudl" is misspelled on line 1742.
- "interpeted" is misspelled on line 78.

pygsti/baseobjs/basisconstructors.py
```diff
@@ -697,6 +697,36 @@ def gm_labels(matrix_dim):
     lblList.extend(["Z_{%d}" % (k) for k in range(1, d)])
     return lblList
 
+def lf_labels(matrix_dim):
+    if matrix_dim != 3:
+        raise NotImplementedError()
+    return ("I", "X", "Y", "Z", "LX0", "LX1", "LY0", "LY1", "L")
```
nkoskelo: I assume that, "I", "X", "Y", and "Z" should be the standard leakage free Clifford gates and that "LXt" is a leakage X gate where the data leaks into the "t" component of the Bloch sphere. Or is this a leakage into another qubit? A comment clarifying this in the code would be useful.

pygsti/circuits/circuit.py
```diff
@@ -977,11 +977,10 @@ def __eq__(self, x):
         if isinstance(x, Circuit):
```
nkoskelo, line 977: Since you are changing this file anyway, there are some misspellings which could be corrected.
- We have numerous cases were we raise an error with ("Error intializing Circuit: ")
- "arbirary" on line 897.
- "suject" on line 3548.
- "diration" on line 3844.
- "whcih" on line 4120.
- "repetative" on line 4627.

pygsti/data/dataset.py
```diff
@@ -1292,11 +1292,28 @@ def gate_labels(self, prefix='G'):
...
         """
-        opLabels = []
-        for opLabelString in self:
-            for opLabel in opLabelString:
-                if not prefix or opLabel.name.startswith(prefix):
-                    if opLabel not in opLabels: opLabels.append(opLabel)
+        opLabels = set()
+        empty_label = _Label(())
+        for circuit in self:
+            for layer in circuit.layertup:
```
nkoskelo, line 1298: Instead of needing to repeat the same logic of iterating over the layers of a circuit to find the distinct label names, why not have a function within circuit or circuit tools which converts a circuit to a set of gate names used? Perhaps we could also have a function which maps a circuit to a dictionary of gate names whose values are the number of times that specific gate appears within the circuit?

Then, here we just call the appropriate API.

pygsti/extras/idletomography/idtcore.py
```diff
@@ -601,7 +601,8 @@ def extract_action(g, cur_sslbls, ql):
         # StaticArbitraryOp, LindbladDenseOp, other gates...
         if len(cur_sslbls) == 1 and cur_sslbls[0] == ql:
             mx = g.to_dense()
-            assert(mx.shape == (4, 4))
```
nkoskelo, line 604: This could also be added by using ``assert (item_to_check), msg_on_failure``. This way when you optimize the python code with python -o the assert statement will be removed. Currently, the if block may still exist after the optimization pass. If you specifically want this to be a ValueError and not an AssertionError then please ignore this.

pygsti/extras/idletomography/idtcore.py
```diff
@@ -601,7 +601,8 @@ def extract_action(g, cur_sslbls, ql):
         # StaticArbitraryOp, LindbladDenseOp, other gates...
```
nkoskelo, line 601: We also have some spelling errors that could be fixed in this file.
- "conventient" on line 34.
- "corresponging" on line 161 and 240.
- "repectively" on lines 323, 702, and 787.

pygsti/modelmembers/modelmembergraph.py
```diff
@@ -19,8 +19,8 @@
 class ModelMemberGraph(object):
     """A directed acyclic graph of dependencies of ModelMembers"""
```
nkoskelo, line 20: Misspelling on line 176 for "derefernced".

pygsti/modelmembers/operations/linearop.py
nkoskelo, line 1: Misspellings:
- "transfomed" on lines 398, 429.
- "accomodate" on line 30.

pygsti/modelmembers/operations/linearop.py
```diff
@@ -92,6 +92,14 @@ def hilbert_schmidt_size(self):
         """
         return (self.dim)**2
 
+    @property
```
nkoskelo, line 97: Should we add warnings to suggest that people use the ```shape``` parameter as opposed to relying on ```._rep.shape```?

pygsti/modelmembers/operations/linearop.py
```diff
@@ -92,6 +92,14 @@ def hilbert_schmidt_size(self):
...
+    def shape(self):
```
nkoskelo: A type signature for this function would be nice.

pygsti/modelmembers/povms/tppovm.py
nkoskelo, line 1: Some type signatures for the functions would improve the readability of the code.

pygsti/modelmembers/states/densestate.py
```diff
@@ -242,9 +242,33 @@ def to_memoized_dict(self, mmg_memo):
     @classmethod
     def _from_memoized_dict(cls, mm_dict, serial_memo):
         vec = cls._decodemx(mm_dict['dense_superket_vector'])
-        state_space = _statespace.StateSpace.from_nice_serialization(mm_dict['state_space'])
-        basis = _Basis.from_nice_serialization(mm_dict['basis']) if (mm_dict['basis'] is not None) else None
-        return cls(vec, basis, mm_dict['evotype'], state_space)
+        try:
```
nkoskelo, line 246: Please add a regression test which can trigger this code path.

pygsti/modelmembers/states/densestate.py
nkoskelo, line 1: Misspellings:
- "mimicing" on line 28.

pygsti/models/explicitcalc.py
```diff
@@ -16,6 +16,7 @@
 import warnings as _warnings
 
 import numpy as _np
+import scipy.linalg as _la
```
nkoskelo, line 19: Maybe import as ```_sla``` instead of ```_la```. It can be common practice as well to ```import numpy.linalg as la``` which may result in confusion in the future.

pygsti/models/explicitcalc.py
nkoskelo, line 1: Misspellings:
- "prefereable" on line 807.
- "psuedo" on line 800.
- "substitue" on line 796.
- "tranformations" on line 752.
- "whic" on line 715.
- "multipy" on line 594.
- "coefficents" on line 594.
- "bookeeping" on line 536.
- "subtituted"  on line 135.
- "dimenstion" on line 69 and 45.
- "abount" on line 36.

pygsti/models/explicitcalc.py
```diff
@@ -146,52 +153,63 @@ def frobeniusdist(self, other_calc, transform_mx=None,
...
                 wt = item_weights.get(opLabel, opWeight)
-                d += wt * gate.frobeniusdist_squared(other_calc.operations[opLabel])
+                gate_mx = gate.to_dense()
+                other_mx = other_calc.operations[opLabel].to_dense()
+                delta = gate_mx - other_mx
+                if P is not None:
+                    delta = delta @ P
+                if invP is not None:
+                    delta = invP @ delta
+                val = _np.linalg.norm(delta.flatten())
```
nkoskelo: The Frobenius norm on the 2D array delta would return the same. So we do not need to explicitly call ```flatten()```.

pygsti/models/explicitcalc.py
```diff
@@ -241,47 +259,30 @@ def residuals(self, other_calc, transform_mx=None, item_weights=None):
...
-                resids.append(
-                    wt * rhoV.residuals(other_calc.preps[lbl]))
-                nSummands += wt**2 * rhoV.dim
-
-            for lbl, Evec in self.effects.items():
-                wt = sqrt_itemWeights.get(lbl, spamWeight)
-                resids.append(
-                    wt * Evec.residuals(other_calc.effects[lbl]))
-                nSummands += wt**2 * Evec.dim
+        Ti = None if T is None else _np.linalg.pinv(T)
```
nkoskelo, line 240: Do we know any information about T? Or are we trying to keep this for general matrices?

pygsti/models/explicitcalc.py
```diff
@@ -609,17 +610,13 @@ def nongauge_and_gauge_spaces(self, item_weights=None, non_gauge_mix_mx=None):
         else:
             orthog_to = gauge_space
 
+        u,s,_ = _la.svd(orthog_to, full_matrices=True, compute_uv=True)
+        TOL = 1e-7
+        r = _np.count_nonzero(s >= TOL*s[0])
```
nkoskelo: Can we label this as numerical rank instead of ```r``` so that we have a search term in the future?

pygsti/tools/jamiolkowski.py
```diff
@@ -187,9 +214,10 @@ def jamiolkowski_iso_inv(choi_mx, choi_mx_basis='pp', op_mx_basis='pp'):
...
 
-def fast_jamiolkowski_iso_std(operation_mx, op_mx_basis):
+def fast_jamiolkowski_iso_std(operation_mx, op_mx_basis, normalized=True):
     """
-    The corresponding Choi matrix in the standard basis that is normalized to have trace == 1.
+    Returns the standard-basis representation of the Choi matrix for operation_mx,
+    where operation_mx is interpreted in op_mx_basis.
 
     This routine *only* computes the case of the Choi matrix being in the
     standard (matrix unit) basis, but does so more quickly than
```
nkoskelo, line 231: Do we have a test that proves that this routine is better than ```jamiolkowski_iso``` in the case of standard (matrix unit) basis? It would be nice to link the test / paper / theorem here.

pygsti/tools/jamiolkowski.py
nkoskelo, line 1: Misspellings:
- "particuarly" on line 224.
- "arbitarily" on line 345, and 376.
- "matricies" on line 405.

nkoskelo, line 1: We are not using the ```from pygsti.tools import matrixtools as _mt```. Therefore, we should remove the import.

nkoskelo, line 1: For the function signatures that you modified please add typing information.

pygsti/tools/optools.py
```diff
@@ -23,7 +24,10 @@
 from pygsti.tools import jamiolkowski as _jam
 from pygsti.tools import lindbladtools as _lt
 from pygsti.tools import matrixtools as _mt
-from pygsti.baseobjs.basis import Basis as _Basis, ExplicitBasis as _ExplicitBasis, DirectSumBasis as _DirectSumBasis
+from pygsti.tools import sdptools as _sdps
+from pygsti.baseobjs import basis as _pgb
+from pygsti.baseobjs.basis import Basis as _Basis, DirectSumBasis as _DirectSumBasis, \
```
nkoskelo: I believe you can import these on separate lines if you use ```(``` and ```)``` around the imports. It should also still respect the ```as``` keyword.

pygsti/tools/optools.py
```diff
@@ -518,6 +458,20 @@ def is_tp_fn(x):
     return fidelity(JA, JB)
 
 
+def tensorized_with_eye(op, basis, ten_basis=None, std_basis=None, ten_std_basis=None):
```
nkoskelo: Needs a docstring and typing information.

pygsti/tools/optools.py
```diff
@@ -660,7 +614,7 @@ def entanglement_infidelity(a, b, mx_basis='pp', is_tp=None, is_unitary=None):
     EI : float
         The EI of a to b.
     """
-    return 1 - float(entanglement_fidelity(a, b, mx_basis, is_tp, is_unitary))
+    return 1 - entanglement_fidelity(a, b, mx_basis, is_tp, is_unitary)
```
nkoskelo, line 638: To maintain backwards compatibility in the event that ```entanglement_fidelity``` does not return a float, use ```1.0 - entanglement_fidelity```.

pygsti/tools/optools.py
nkoskelo, line 1: Misspellings:
- "paramterization" on line 2660.
- "coeffiecients" on line 2162.
- "requrested" on line 2316.
- "matricies" on line 1845.
- "multply" on line 1831.
- "fromt" on line 1713 (looks like should be "from the")
- "funtion" on line 1054.
- "Decompse" on line 1052.

pygsti/tools/sdptools.py
```diff
@@ -0,0 +1,231 @@
...
+except:
+    CVXPY_ENABLED = False
+
+
+def diamond_norm_model_jamiolkowski(J):
+    # return a model for computing the diamond norm.
+    #
+    # Uses the primal SDP from arXiv:1207.5726v2, Sec 3.2
+    #
+    # Maximize 1/2 ( < J, X > + < J.dag, X.dag > )
```
nkoskelo, line 46: I would maybe recommend "^{\dagger}" in that case. Or does the comment indicate a specification that cvxpy will use to generate appropriate code?

pygsti/tools/sdptools.py
nkoskelo, line 1: There are unused imports in this file.

nkoskelo, line 1: Since this is a new file, it would also be nice to have a type signature for each of the functions.

pygsti/baseobjs/basis.py
rileyjmurray: I'll defer spelling fixes until we resolve all other comments. (I'll only leave a comment to that effect here.)

pygsti/baseobjs/basisconstructors.py
rileyjmurray: Done

pygsti/extras/idletomography/idtcore.py
rileyjmurray: I didn't know about the -o flag in Python! Very cool.

That said, in this case I definitely want a value error.

pygsti/modelmembers/operations/linearop.py
rileyjmurray: I like that idea, but it's hard to implement. There may well be contexts in which someone's working just with a rep and not a LinearOperator. In those cases it's fine to access ``rep.shape``. 

We _can_ CTRL+F for ``._rep.shape``, though.

rileyjmurray: Looks like the only hit for ``._rep.shape`` is in this file. So I'll mark this as resolved.

pygsti/modelmembers/operations/linearop.py
rileyjmurray: Excellent catch

pygsti/modelmembers/povms/tppovm.py
rileyjmurray: I agree. Since my changes to this file were only for readability in the first place, it's fair that I should be more comprehensive.

rileyjmurray: Uhh... actually, I'm going to punt on this. All the complexity of the types is inherited from basepovm.py.

In the spirit of my original agreement I _will_ roll-back the readability change. It was extremely minor anyway.

pygsti/models/explicitcalc.py
rileyjmurray: Looks like we only use numpy.linalg as nla.
image

So I'll keep it as _la here.

pygsti/models/explicitcalc.py
rileyjmurray: Fair!

pygsti/models/explicitcalc.py
rileyjmurray: Turns out there is harm in always returning orthonormal bases, at least for the function calls in confidenceregionfactory.py. The issue is that for a function $x\mapsto f(x)$ with Hessian $H(x)$, the Hessian of the function $x\mapsto f(A^{-1} x)$ is $A^T H(x) A$. If $A$ is non-orthogonal then $A^T H(x) A$ will differ from $H(x)$ by more then just a unitary transformation (obvious check: take the determinant of both matrices).

I'll revert this functional change and just leave comments for the future person who finds themselves with unhandled exceptions here.

rileyjmurray: UGGGGHHH but nullspace_qr was already only returning an orthonormal basis. 

I hate this. I guess I'll keep my change after all.

pygsti/models/explicitcalc.py
rileyjmurray: Keeping it for general matrices. So there's no special structure to exploit in the pinv calculation.

pygsti/models/explicitcalc.py
rileyjmurray: Done.

pygsti/tools/jamiolkowski.py
rileyjmurray: I looked into the function calls that this implementation triggers. It seems this performs all calculations as memory movement plus two big calls to GEMM. Compare to the "not fast" implementation, which needs a ton of explicit Kronecker products and AXPYs. So I don't think we need tests showing this function is faster than the "not fast" one.

pygsti/tools/jamiolkowski.py
rileyjmurray: Done.

pygsti/tools/optools.py
rileyjmurray: Done!

pygsti/tools/optools.py
rileyjmurray: I'm fairly confident that the float coercion was in case entanglement_fidelity returned a complex number. Looking at entanglement_fidelity (and then fidelity) shows that we'll always get real numbers, so I'll keep this change as-is.

pygsti/tools/sdptools.py
rileyjmurray: This is just comment for us humans. Will switch to dagger.

pygsti/tools/jamiolkowski.py
rileyjmurray: Done.

pygsti/tools/optools.py
rileyjmurray: Done.

pygsti/tools/sdptools.py
rileyjmurray: Done.

coreyostrove:
> Text from an email Piper sent (not showing plots since they contained real data):
> 
> > I was looking at the model violation tab, trying to work out why the smaller L models weren’t fitting the data as well as might be expected as quantified by Nsigma. I think this maybe be a function of how we calculate Nsigma = (2delta logl -k)/sqrt(2k). The number of non-gauge model parameters N_p is reported as 351 in the leakage-enhanced GST report (I don’t have any particular sense of whether this is correct). The number of data points N_s is 193 for L=1. This means that k (the degrees of freedom) is negative if you just calculate it as N_s-N_p. In cases like this, pyGSTi chooses k=1. k remains small until L is fairly large, which means two delta logl must be quite small for Nsigma to be small. This explains why we don’t see any red boxes in the per sequence detail for the L=1 iteration (and Nsigma is large enough that I would expect to see some or at least more dark gray squares than can be seen).
> > If you directly compare two delta logl between the two fits, you can see that the two delta logl is in fact smaller in the leakage-enhanced report than in the standard report.
> 
> I think the solution here is just to make sure that the first iteration of leakage-aware GST uses L=2 instead of L=1. (Or, more generally, we use the smallest L where N_s > N_p.)

This isn’t a uniquely leakage-related shortcoming, I’ve encountered this in other settings too (with really aggressive FPR schemes when not forcing inclusion of LGST circuits, for example) and I suggest we open up a separate github issue for this and give some additional thought to how we should be approaching the statistics in this case.

pygsti/algorithms/gaugeopt.py
coreyostrove: I don’t even know what these do, and have never used them, so I think removing them is fine.

pygsti/algorithms/gaugeopt.py
```diff
@@ -594,18 +591,34 @@ def _mock_objective_fn(v):
     else:
         # non-least-squares case where objective function returns a single float
         # and (currently) there's no analytic jacobian
+        dim = int(_np.sqrt(mxBasis.dim))
+        if n_leak > 0:
+            B = _tools.leading_dxd_submatrix_basis_vectors(dim - n_leak, dim, mxBasis)
+            P = B @ B.T.conj()
+            if _np.linalg.norm(P.imag) > 1e-12:
+                raise ValueError()
```
coreyostrove: Please add a descriptive error message for this case.

pygsti/algorithms/gaugeopt.py
coreyostrove: Please add an error message for the AssertionError indicating that these metrics are not supported when doing leakage-aware gauge optimization (i.e. n_leak>0).

pygsti/algorithms/gaugeopt.py
```diff
@@ -617,81 +630,86 @@ def _objective_fn(gauge_group_el, oob_check):
...
-                        ret += mdl.frobeniusdist(target_model, None, wts)
-
-                elif gates_metric == "frobeniustt":
-                    if spam_metric == "frobeniustt":
-                        ret += transformed_target.frobeniusdist(model, None, item_weights)
+                """
+                Leakage-aware metric supported, per implementation in mdl.frobeniusdist.
+                Refer to how mdl.frobeniusdist handles the case when transform_mx_arg
+                is a tuple in order to understand how the leakage-aware metric is defined.  
+                """
```
coreyostrove: I think there might be a (likely preexisting) bug here related to instruments. For the fidelity and tracedistance metrics I’m not seeing the instrument elements being accessed (except in passing in the spam metric computation where it is just used to filter the item weights). For frobenius this is harder to see, but I took a look at the implementation of the ‘frobeniusdist’ method for ExplicitOpModels [here](https://github.com/sandialabs/pyGSTi/blob/822148ab0f2712aa7fb0c680ebc9fa048ad76fb0/pygsti/models/explicitcalc.py#L109) and I’m not seeing instruments there either.

pygsti/circuits/circuit.py
coreyostrove: These look functionally equivalent to me. I’ll take your word for it, but for my own edification what about the original version was incorrect?

pygsti/data/dataset.py
coreyostrove: This is a good suggestion and I think it would be an independently useful primitive for circuits to return their unique gate labels, probably with some flags to control matching behavior (e.g. just label names versus full labels with sslbls).

pygsti/models/explicitcalc.py
```diff
@@ -146,52 +153,63 @@ def frobeniusdist(self, other_calc, transform_mx=None,
...
                 nSummands += wt * rhoV.dim
 
             for lbl, Evec in self.effects.items():
                 wt = item_weights.get(lbl, spamWeight)
-                d += wt * Evec.frobeniusdist_squared(other_calc.effects[lbl], T, Ti)
+                d += wt * Evec.frobeniusdist_squared(other_calc.effects[lbl], P, invP)
                 nSummands += wt * Evec.dim
-
         else:
+            # we're in the special case that I'm creating.
```
coreyostrove: What is the special case?

pygsti/protocols/confidenceregionfactory.py
```diff
@@ -654,16 +655,21 @@ def _project_hessian(self, hessian, nongauge_space, gauge_space, gradient=None):
...
         # v.T @ H @ v = some 2nd deriv = v.T @ B.T @ H' @ B @ v in another basis
         # so H' = invB.T @ H @ invB
-        assert(_np.allclose(hessian, hessian.T))
+        TOL = 1e-7
+        assert(_la.norm(hessian.imag) == 0)
+        sym_err_abs = _la.norm(hessian - hessian.T)
+        sym_err_rel = sym_err_abs / _la.norm(hessian)
+        assert(sym_err_rel < TOL)
+        hessian += hessian.T
+        hessian /= 2
```
coreyostrove, line 664: This is just using the following factoid about quadratic forms, right? https://en.wikipedia.org/wiki/Quadratic_form#General_case

pygsti/report/reportables.py
coreyostrove, line 1: It looks like most/all of these reportable functions hard code in that n_leak is one. I don't think there is anything to flag a user specifying something different in `construct_standard_report`, though. That is fine for now, but maybe add an assertion for this to avoid unexpected behavior? (i.e. a user being confused by why this was quietly changed under the hood).

pygsti/tools/leakage.py
coreyostrove, line 1: A general comment, a lot of these new functions lack doctrings explaining how/what they do. Some of these also look like they might not be user facing with their API. So please add docstrings where you can for functions that you expect to be user-facing, and mark others as part of the internal API (e.g. with a leading underscore).

pygsti/tools/sdptools.py
coreyostrove, line 1: Similar comment to the one I left on the leakage model regarding docstrings and API-level.

coreyostrove (changes requested):
Just finished my first pass through. This is a ton of effort and these contributions are sincerely appreciated. Thank you as always for the excellent work, @rileyjmurray!

Of the comments I've left, the only one I am deeply concerned about is the potential bug that I've pointed out in the cost function for gauge optimization as it pertains to instruments. I have good reason to suspect (I mean, the git history is pretty clear and this present on develop so it isn't really a suspicion) that in the event there is a bug here it is very likely you had nothing to do with introducing it, but since you're already knee deep in this code I'd ask that you take a look and if necessary clean this up on our behalf.

Rest of my comments are either minor clarifying questions or otherwise nice-to-have but non-blocking requests. I've also added my concurrence with a couple comments/suggestions from @nkoskelo.

pygsti/report/reportables.py
rileyjmurray: Good point. I'll update the construct_standard_report docstring and raise an exception if the user passes in n_leak other than 0 or 1.

pygsti/tools/leakage.py
rileyjmurray: You got me! Will do so.

pygsti/algorithms/gaugeopt.py
rileyjmurray: Done!

pygsti/models/explicitcalc.py
rileyjmurray: Good catch. That's a terrible comment for me to have left in place.

Rather than answer your question, I've refactored this function in order to remove the distinction between the two cases.

pygsti/tools/leakage.py
```diff
@@ -0,0 +1,484 @@
...
+        superop_std = pgot.unitary_to_std_process_mx(u3x3)
+        superop = pgbt.change_basis(superop_std, 'std', basis)
+        return superop
+
+    for gatename, unitary in Us.items():
+        gatekey = (gatename, 0) if gatename != '{idle}' else ('Gi',0)
+        tm_3level[gatekey] = u2x2_to_9x9_superoperator(unitary)
+
+    tm_3level.sim = 'map'  # can use 'matrix', if that's preferred for whatever reason.
+    return tm_3level
```
rileyjmurray: Needs docstring.

pygsti/tools/leakage.py
```diff
@@ -0,0 +1,484 @@
...
+        gop_params = {'LAGO': lago_gos_lods}
+        gos = GSTGaugeOptSuite(gaugeopt_argument_dicts=gop_params)
+    if isinstance(est_key, str):
+        add_param_preserving_gauge_opt(results, est_key, gos, verbosity)
+    elif est_key is None:
+        for est_key in results.estimates.keys():
+            add_lago_models(results, est_key, gos, verbosity)
+    else:
+        raise ValueError()
+    return
```
rileyjmurray: Need documentation.

pygsti/tools/sdptools.py
rileyjmurray: I'm going to say that everything here is in the private API for now.

rileyjmurray:
@coreyostrove I think I've addressed your comments. Just pushed what should be the last round of changes.

coreyostrove (approved):
Fantastic work, @rileyjmurray! Thank you for the significant effort this work represents and your contributions advancing our leakage modeling capabilities. Merge at will.

One last request as you merge things in, can you flesh out the summary in the PR? That’d be helpful as I prepare release notes (and for posterity).

pygsti/protocols/confidenceregionfactory.py
rileyjmurray: Essentially, yes. But if your function is twice continuously differentiable you're guaranteed that the Hessian is symmetric. This symmetrization is to eliminate issues from rounding errors in finite-difference methods.

rileyjmurray:
The single failing test is caused by implementation details of /test/unit/mpi/test_mpi.py::MPITester::test_all. Merging as-is.

PR #411, opened 2024-03-19
Update/New Cirq Parsing Capabilities

coreyostrove:
This PR updates the existing parser for conversion of pyGSTi Circuit objects to Cirq Circuit objects. Also introduced is a new parser for converting from Cirq Circuits objects into pyGSTi Circuit objects. A few other minor changes include:

1. Updates to pyGSTi's built-in gate definitions to include the following: 

- 'Giswap' and 'Gsqrtiswap' (2-qubit gates commonly used on superconducting platforms)
- 'Gcres' and 'Gecres' (2-qubit cross-resonance and echoed cross-resonance gates, respectively, native on many platforms including IBM's)
-  'Gn' (one-qubit N gate, a pi/2 rotation about the (np.sqrt(3)/2, 0, -1/2) axis of the Bloch sphere, commonly used native gate in some silicon qubit platforms, the N referred to in the smq1Q_ZN modelpack), 
- 'Gxpi4', 'Gypi4', 'Gzpi4' (pi/4 rotations above the various Bloch sphere axes, Gxpi4 appeared in one of the modelpacks, and I figured it best to add the Y and Z counterparts at the same time).

2. Updated the function `unitary_to_standard_gatename` in `tools.internalgates` to allow the option for checking for equivalence of unitaries to a built-in gate up to an overall global phase.

3. New unit tests (including one for qasm conversion, which was previously not unit tested) and updates to the cirq integration demo notebook to include the new functionality. 

Note: I branched this off of #403, which is why there is so much testing update stuff in the commit log (once that gets merged in it should go away from this PR), stuff relevant to this update begins 3/13/24 w/ 185a59ad759bbb2d7a494e7f48f8ab36773aaa52.

### DIFF
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -43,6 +43,11 @@ from pygsti.tools import slicetools as _slct
 # c[1:3,'Q0'] = ('Gx','Gy') # assigns to a part of the Q0 line
 
 
+#Add warning filter
+msg = 'Could not find matching standard gate name in provided dictionary. Falling back to try and find a'\
+     +' unitary from standard_gatename_unitaries which matches up to a global phase.'
+_warnings.filterwarnings('module', message=msg, category=UserWarning)
+
 def _np_to_quil_def_str(name, input_array):
     """
     Write a DEFGATE block for RQC quil for an arbitrary one- or two-qubit unitary gate.
@@ -3764,7 +3769,7 @@ class Circuit(object):
         return cirq.Circuit(moments)
     
     @classmethod
-    def from_cirq(cls, circuit, qubit_conversion=None, implied_idles = False, global_idle = False):
+    def from_cirq(cls, circuit, qubit_conversion=None, cirq_gate_conversion= None,implied_idles = False, global_idle = False):
         """
         Converts and instantiates a pyGSTi Circuit object from a Cirq Circuit object.
 
@@ -3777,6 +3782,11 @@ class Circuit(object):
             A dictionary specifying a mapping between cirq qubit objects and 
             pyGSTi qubit labels (either integers or strings).
             If None, then a default mapping is created.
+
+        cirq_gate_conversion : dict, optional (default None)
+            If specified a dictionary with keys given by cirq gate objects,
+            and values given by pygsti gate names which overrides the built-in
+            conversion dictionary used by default.
 
         implied_idles : bool, optional (default False)
             A flag indicating whether to explicitly include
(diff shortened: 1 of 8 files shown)


pygsti/circuits/circuit.py
```diff
@@ -3744,22 +3749,200 @@ def convert_to_cirq(self,
...
+            If specified a dictionary with keys given by cirq gate objects,
+            and values given by pygsti gate names which overrides the built-in
+            conversion dictionary used by default.
+
+        implied_idles : bool, optional (default False)
+            A flag indicating whether to explicitly include
+            implied idles as part of a circuit layer containing
+            other explicitly specified gates.
+
+        global_idle : bool or string or Label, optional (default False)
```
sserita, line 3796: A question/clarification on implicit vs global idle behavior. The default is False for both, but this results in Gi's being dropped for implicit idles everywhere EXCEPT for global idles - that is, in cell 17 of the example, we have Gi labels in partial idle layers, but the global idle layer does have Gis. I kinda expected no Gis anywhere in this case.

Maybe more succintly put, I expected the "True" behavior of global idle to be the default (i.e. cell 20), and to recover the current (i.e. cell 17) behavior, a user would have to explicitly put "Gi" for the global_idle string here. Another advantage of this approach is then the global_idle argument only takes one type of argument (a label-like one) rather than switching behavior on boolean/label-like arguments.

I am open to being convinced that current behavior is most desired though!

P.S. Another place where Jupytext examples/tutorials would probably make this discussion easier...

sserita (commented):
Minor discussion on default global idle behavior, otherwise close to merge!

pygsti/circuits/circuit.py
coreyostrove: Yeah, I was conflicted on the default behavior here. The original decision to make replacement of the global idle layers non-default is that unlike with implicit idles where the convention is pretty universally accepted (i.e. not including them), the convention for global idles is less so. I know we have Label(()), but I've always kind of disliked this convention (from a dev perspective it is kind of annoying and creates lots of edge cases to check for), so I initially wanted to force the choice of substitution to be an affirmative one. As you can see from the final implementation I eventually flip-flopped on this and did add a default label replacement choice of Label(()), so I think making this default would be alright. I'll probably change the names of these kwargs to be more descriptive while I'm at it.

coreyostrove:
Thanks for the feedback, @sserita. I have just pushed some updates which rename the kwargs of implied and global idle handling to `remove_implied_idles` and `global_idle_replacement_label`, respectively, which should hopefully be more self explanatory for end users. I have also changed the default behavior for the global idle (implied handling behavior is the same, but the default value is now instead True to match the renaming of the kwarg) so that they are replaced. Let me know if you have any other concerns.

sserita (approved):
I made a minor fix to the remove_implied_idles docstring, and think this is good to go as soon as the tests pass!

PR #413, opened 2024-03-26
Replace deprecated numpy.product, numpy.NaN and others

eendebakpt:
Numpy 2.0 is about to be released and several deprecated attributes will be removed. In this PR we replace them with the preferred alternatives.

Also see: https://github.com/numpy/numpy/issues/24300, https://numpy.org/devdocs/numpy_2_0_migration_guide.html


sserita (approved):
Thanks for these Pieter!

Issue #406, opened 2024-03-01
TypeError when performing gauge optimization with respect to fidelity or trace distance


rileyjmurray:
Running gauge optimization with respect to fidelity hits the following codeblock:
https://github.com/sandialabs/pyGSTi/blob/0299e1d13222658ab385a17f6e16b9c7ce566163/pygsti/algorithms/gaugeopt.py#L636-L640

The documentation of ``entanglement_fidelity`` says its arguments need to be numpy arrays. However, the code above passes in pyGSTi modelmember objects of one kind or another. If you take a quick look at [the implementation of ``entanglement_fidelity``](https://github.com/sandialabs/pyGSTi/blob/0299e1d13222658ab385a17f6e16b9c7ce566163/pygsti/tools/optools.py#L433-L461) then you'll find that this can easily result in an error. (For example, if either operation is a CompsedOp object.) 

This _specific_ issue could be fixed by replacing 638 -- 640 in the gaugeopt.py snippet to the following:
```python
op_a = target_model.operations[opLbl].to_dense()
op_b = mdl.operations[opLbl].to_dense()
infidelity = 1.0 - _tools.entanglement_fidelity(op_a, op_b, mxBasis)
ret += wt*(infidelity ** 2)
```

That said, _the issue isn't isolated to the snippet above._ Similar bugs exist throughout the definition of the [the objective function for non-LS gauge optimization models](https://github.com/sandialabs/pyGSTi/blob/0299e1d13222658ab385a17f6e16b9c7ce566163/pygsti/algorithms/gaugeopt.py#L594).

coreyostrove:
Great job catching this, @rileyjmurray!

I'm not positive what the history is behind this, but I suspect this is vestigial from the time before more complicated modelmembers for operations existed when pretty much all of the operations inherited from DenseOperatorInterface or similar, and as such behaved in many contexts as if they were numpy arrays. This is clearly no longer true, and your proposed solution makes sense to me.

enielse:
I agree with Corey's assessment - we haven't tried gauge optimizing to
fidelity in a long time and the initial implementation was more
experimental research and there are likely no unit tests.  So yeah,
converting operators to dense matrices first is I think what you want to do.

sserita:
I think this particular bug is basically solved by Timo's PR #414. By performing the `to_dense` cast inside `entanglement_fidelity`, we avoid having to find all the places where the function is called.

[...]

PR #414, opened 2024-03-29
Cast gates to dense matrices in `optools`

Timo1104:
### Description of changes

Minor change: in `optools.entanglement_fidelity`, accept both arrays and gates for the `a` and `b` arguments. The arguments are cast to (dense) arrays within the function scope.

Perform similar casting in other functions, that call `a.shape`. 

### Motivation

See also #406.

Since some update (cannot pinpoint which exactly), this casting was not done anymore, neither within `optools.entanglement_fidelity`, nor in any of the functions calling this one.

I've chosen to cast to dense array within `optools.entanglement_fidelity`, rather than before calling this function, for two reasons:

- It keeps the number of changes as small as possible;
- It provides a more consistent interface across `optools`. Other functions (such as `average_gate_fidelity`, which calls `entanglement_fidelity`) accept arrays or gates. Hence, I'm of the opinion that casting to array within `entanglement_fidelity` is the more consistent choice. Alternatively, casting can be done before calling it, but (IMO) it's more consistent to then also do that for the other metrics functions defined in `optools`. I'm okay with either way, but this seems the simpler and less backwards-breaking solution, but I'm happy to implemetn otherwise if required.
- It resolves an error in for example `average_gate_fidelity`, which states in the docstring that it accepts arrays or gates for `a` and `b`, but directly passes `a` and `b` to `entanglement_fidelity`, leading to an error.
- Prevents errors on calling `a.shape` on gates that do not have the `shape` attribute (such as `ComposedOp`s)

### DIFF
--- a/pygsti/tools/optools.py
+++ b/pygsti/tools/optools.py
@@ -514,6 +514,10 @@ def average_gate_fidelity(a, b, mx_basis='pp', is_tp=None, is_unitary=None):
     AGI : float
         The AGI of a to b.
     """
+    # Cast to dense to ensure we can extract the shape.
+    with _contextlib.suppress(AttributeError):
+        a = a.to_dense()
+
     d = int(round(_np.sqrt(a.shape[0])))
     PF = entanglement_fidelity(a, b, mx_basis, is_tp, is_unitary)
     AGF = (d * PF + 1) / (1 + d)
@@ -720,6 +724,10 @@ def unitarity(a, mx_basis="gm"):
     -------
     float
     """
+    # Cast to dense to ensure we can extract the shape.
+    with _contextlib.suppress(AttributeError):
+        a = a.to_dense()
+        
     d = int(round(_np.sqrt(a.shape[0])))
     basisMxs = _bt.basis_matrices(mx_basis, a.shape[0])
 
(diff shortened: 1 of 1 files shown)


pygsti/tools/optools.py
```diff
@@ -430,6 +433,13 @@ def entanglement_fidelity(a, b, mx_basis='pp', is_tp=None, is_unitary=None):
     -------
     float
     """
+    # Attempt to cast to dense array. If this is already an array, the AttributeError
+    # will be suppressed.
+    with _contextlib.suppress(AttributeError):
```
sserita, line 438: Can we change the contextlib.suppress calls to an isinstance(_, LinearOperator) instead?

It's more pyGSTi style to check for the base class that provides the used API than a blanket error suppression.

sserita (changes requested):
Will approve once the contextlib calls are swapped out for LinearOperator checks.

Thanks for your PR!

pygsti/tools/optools.py
Timo1104: Of course! I'm more used to EAFP-style programming, but happy to accomodate LBYL, too :)  Just pushed the changes!

sserita (approved):
Thanks Timo. There's definitely arguments to be made that EAFP is a nicer way to go, but it's appreciated that you're willing to conform to current style :)

Looks good, thanks for the contribution!

Issue #412, opened 2024-03-25
HTML Reports for Non-Gauge Optimized Models 


pcwysoc:
**Describe the bug**
When gauge optimization is not performed during GST the HTML report does not generate the gauge-dependent figures. There is a similar, less critical, problem for notebook reports. 

**To Reproduce**

[GaugeOptReportBug.ipynb.zip](https://github.com/sandialabs/pyGSTi/files/14749681/GaugeOptReportBug.ipynb.zip)

**Expected behavior**
I believe that the raw estimates and error metrics should still generate, although they won't be as informative. Otherwise, there should be some flag to remove the gauge-dependent sections of the report. 

**Environment (please complete the following information):**
 - pyGSTi version qiskit-feature-mcm
 - python version 3.10
 - OS 13.6.5

coreyostrove:
Thanks for the bug report, @pcwysoc! And for the example code reproducing the error. I have implemented a fix for this (along with some other issues I discovered along the way) on the branch 'feature-circuitlistsdesign-reporting'. These should be merged into develop sometime soon (there is already a PR for these changes), but in the meantime you can merge that branch into your own if you need this working right away.

sserita:
Closing as this is now merged in with #415

Issue #320, opened 2023-05-19
Model Testing and Report Generation for More General Experiment Designs


coreyostrove:
At present pygsti report generation fails pretty badly when encountering non-GST experiment designs. In some cases we have specialized reporting pipelines for other classes of experiment design. But we also have support in pyGSTi for incredibly general experiment designs call CircuitListsDesigns, which require no more structure than to have a list of circuits associated with them. While many aspects of a GST report clearly won't work in this setting, such as that color box plots which require the regular structure of a GST experiment, more general model violation statistics certainly should. Likewise figures for error generators, tables for fidelities, eigenvalue and process matrix information etc. all should work.

I have implemented a series of patches to the report generation code that mostly plug in these gaps and they live on the branch feature-circuitlists-designs. We should be able to migrate these changes into pyGSTi for the next release.

PR #415, opened 2024-04-01
Fixes for CircuitListsDesign Reporting and Reporting w/o Gauge Optimization

coreyostrove:
This PR covers changes intended to improve reporting support for two scenarios:

1. Report generation when using a CircuitListsDesign (addressing #320)
2. Report generation when performing GST without using gauge optimization. (addressing #412)

For item 1, previously it was the case that many plots, especially those related to model violation summarization had the assumption that the design has a PlaquetteGridStructure format hardcoded in. This of course makes sense for the model violation and TVD plaquette plots, but is an unnecessary assumption for the per-circuit histogram plots and other model violation statistics plots which would fail to generate. With the changes in this PR we can now generate all of the model violation plots that don't explicitly require this structure.

For item 2, this is in response to the bug report from @pcwysoc in #412. As reported there, when performing GST without gauge optimization the report generation would not create plots/tables related to gauge-variant quantities including distance metric tables, error generator plots, etc... This report reminded me that I ran into this exact same issue over a year ago in the context of generating reports for ModelTest protocols and fixed it on this branch (where I allowed it to subsequently languish after getting distracted). So I've now ported the fix I made for ModelTest over to the GateSetTomography protocol (and pre-emptively LinearGateSetTomography since this would have the same issue if anyone were to use it this w/o gauge optimization).

In the process of working on item 2 I originally implemented some changes to deprecate 'none' as an option, before reverting these changes upon the realization that the utility of having this option (beyond what is available using an empty GSTGaugeOptSuite) is that it allows for including no gauge optimization as an option among other gauge optimization suites (including multiple gauge optimizations in an estimate isn't something we do super often, but is fully supported). In going through this back and forth, however, it occurred to me (spurred on by the intuition about how the code worked that I'd gained from the other bugfixes) that there was going to be an edge case when using multiple gauge optimization suites when including 'none' as one of them wherein this would be skipped/not included in the report. Some subsequent testing confirmed this was indeed the case, with reports produced including all of the selected suites except for None. TLDR, I knew if I didn't fix this now I may never have done so, so I figured out a patch and that fix is included in this PR as well.

Unrelated change: While going through the gst protocol module I happened to noticed that the LinearGateSetTomography protocol was missing the auxfile specifications needed for directory-based serialization of the sort we have for GateSetTomography and StandardGST. So I added those (pattern matching off of the other two), which means that this protocol should be serializable now. Probably not a big deal either way, since I don't think many folks use this with any regularity, but figured I'd do it now since while it was fresh. I also added some new unit tests for all three of these protocols testing the directory-based serialization for each them.


pygsti/report/workspaceplots.py
```diff
@@ -662,42 +697,82 @@ def _circuit_color_scatterplot(circuit_structure, sub_mxs, colormap,
...
+        yvals = g.used_ys
 
     if addl_hover_submxs is None:
         addl_hover_submxs = {}
 
     if hover_info:
-        hover_info = _create_hover_info_fn(circuit_structure, xvals, yvals, sum_up, addl_hover_submxs)
-
+        if isinstance(g, _PlaquetteGridCircuitStructure):
+            hover_info = _create_hover_info_fn(circuit_structure, xvals, yvals, sum_up, addl_hover_submxs)
```
sserita, line 710: If we use `g.used_xs` and `g.used_ys` here, we could probably remove the `xvals`/`yvals` definition above. Minor thing, but probably helpful to developers to not have those variables defined when thinking about the CircuitList codepath instead of the PlaqGrid one.

pygsti/report/workspaceplots.py
```diff
@@ -662,42 +697,82 @@ def _circuit_color_scatterplot(circuit_structure, sub_mxs, colormap,
...
+                continue
             else:
-                for iiy, iix, opstr in plaq:
-                    if opstr in gstrs: continue  # skip duplicates
-                    xs.append(len(opstr))
-                    ys.append(sub_mxs[iy][ix][iiy][iix])
-                    gstrs.add(opstr)
+                if sum_up:
+                    pass
+                    #TODO: Implement sum_up behavior mirroring that above.
```
sserita, line 750: Is this something we should implement now before it leaves mental cache? I can't imagine we want to touch this code very frequently.

I think this is something I can take a stab at also - we just need to iterate through the circuit lists, collect/sum the information, and store it as a summary_label attribute to be used in the hover_fn construction function?

pygsti/report/workspaceplots.py
```diff
@@ -1851,6 +2025,8 @@ def _create(self, plottypes, circuits, dataset, model, prec, sum_up, box_labels,
                                                     colorbar, hover_info, sum_up, ytitle,
                                                     scale, addl_hover_info)
             elif typ == "histogram":
+                #print(subMxs)
```
sserita, line 2028: If making any changes, maybe just remove these debug statements also?

pygsti/report/workspaceplots.py
```diff
@@ -1899,6 +2075,33 @@ def _create(self, plottypes, circuits, dataset, model, prec, sum_up, box_labels,
 def _mx_fn_from_elements(plaq, x, y, extra):
     return plaq.elementvec_to_matrix(extra[0], extra[1], mergeop=extra[2])
 
+#modified version of the above meant for working with circuit lists
+def _mx_fn_from_elements_circuit_list(circuit_list, extra):
```
sserita, line 2079: I don't see a way around basically having this duplicate function, but what are your thoughts on including this as an elementvec_to_matrix function in CircuitList so that it is more "compatible" stylistically with how the plaq function is done?
I mostly want to avoid someone updating the elementvec_to_matrix in plaq and forgetting to do something similar here - I know that moving this into CircuitList would not enforce it, but at least a quick search for elementvec would have both classes pop up instead of remembering to change the second instance here.

pygsti/report/workspaceplots.py
```diff
@@ -1841,7 +2007,15 @@ def _create(self, plottypes, circuits, dataset, model, prec, sum_up, box_labels,
             else: assert(False), "Internal logic error"  # pragma: no cover
 
             if typ == "boxes":
-                newfig = _circuit_color_boxplot(circuit_struct, subMxs, colormap,
+                if not isinstance(circuit_struct, _PlaquetteGridCircuitStructure):
+                    #if not a plaquette structure then maybe try returning a NotApplicable object
+                    #for the figure?
+                    return NotApplicable(self.ws)
+                else:
+                    #I am expecting this cast won't do anything at the moment, but
```
pcwysoc, line 2015: I don't understand this section. Does it not do anything now? What is it intended to do in the future then? Why doesn't it work now?

pcwysoc (approved):
Workspaceplots.py largely looks good--I agree with Stefan's comments. There will be some areas that will require modification in order to integrate my model violation fixes for the per sequence box plot, but that's mostly a problem for me merging into my branch for right now.

sserita (approved):
Oops...

sserita (changes requested):
Oops I approved the wrong PR, this one is still in progress.

pygsti/report/workspaceplots.py
coreyostrove: Good catch, I hadn't noticed when updating this that these variables are only used in one place in this function, and that is already wrapped in an isinstance check. Updated.

pygsti/report/workspaceplots.py
coreyostrove: Good question. As far as I can recall what I was noting in that comment is that in the original implementation of this method there are a number of places where we cast circuits to a PlaquetteGridCircuitStructure even though we've already checked for whether the object being cast was an instance of one earlier (as here). In a nutshell the comment can be translated as: 'I'm not sure why it is exactly we're doing these casts, but @enielse probably had a reason (maybe some future code extensions) and I don't see any harm in keeping them just to be safe'.

pygsti/report/workspaceplots.py
coreyostrove: The ship has long since sailed on this leaving my mental cache, unfortunately. However, in starting the process of trying to implement this I realized why it was I didn't do so earlier. TLDR: I haven't figured out exactly what the `sum_up` option should actually do for circuit lists (or lists of circuit lists). For plaquettes what this does is it sums over the constituents of a single plaquette and then adds a scatter point to the figure with a y value given by the sum over the plaquette, and an x-value corresponding to the length of the base circuit (germ raised to some power). So the meaning of this action is fairly clear. I'm not sure how we'd meaningfully port this to CircuitList or lists of CircuitList objects though, since I don't know what x would be. If we had some additional structural guarantees/metadata we could do so, but presently we don't. (Incidentally, the thought did cross my mind last year that it may be useful to add a new circuit structure as a wrapper around CircuitList objects that provided this sort of additional metadata, but obviously I never did so).

sserita: Fair enough, I also didn't think it through that far. Let's leave it as is in that case, and we can revisit this if we ever do something with a container of CircuitList objects (which is an interesting idea, but agreed that it's a relatively low priority).

pygsti/report/workspaceplots.py
coreyostrove: Just pushed a commit which does this refactor, and it looks like everything is still working with these changes given some simple tests.

sserita (approved):
Thanks @coreyostrove! This looks good now :)

Issue #396, opened 2024-02-02
Fiducial selection for prep fiducials generates wrong qubit labels 


pcwysoc:
**Describe the bug**
During fiducial selection prep fiducials get incomplete qubit labels 
<img width="418" alt="Screenshot 2024-02-01 at 7 23 27 PM" src="https://github.com/sandialabs/pyGSTi/assets/144378483/eb6c9d48-428c-4e23-b1f8-117894edb83f">

**To Reproduce**
1. Create a 2Q target model with qubit labels ('Q0', 'Q1')
2. Do fiducial selection using: 
prepFiducials, measFiducials = fidsel.find_fiducials(target_model, algorithm='greedy',candidate_fid_counts=3)
3. Look at the prepfiducials 

**Expected behavior**
Prep fiducials should all have qubit labels @('Q0', 'Q1') 

**Environment (please complete the following information):**
 - pyGSTi version feature-qiskit-mcm
 - python version 3.10
 - OS Ventura 13.6.4

coreyostrove:
Hey @pcwysoc, thanks for the report. Question for you, when encountering this behavior did you subsequently encounter any related errors caused by this? Or what this simply unexpected behavior?

I ask because this is something that was on our radar due to some errors that we have previously run into in the context of GST on systems with native parity readout. In that context an error arises because the action of measuring a single qubit is not defined. I ask mostly to see if there are other ways this interacts with the codebase we might not be aware of, but should be cognizant of when addressing it.

That leaves open the question on the right way to address this, which I’d be open to feedback on. The _easy_ ‘fix’ would be to simply have the edesign code always include line-labels for every qubit in all circuits. This feels suboptimal in certain contexts though, particularly for many-qubit circuits. I imagine part of the motivation for the current (though computational basis centric) behavior is that it quickly gets unruly to list 100+ line labels for a single-qubit circuit. Introspecting the modelmembers and inferring this seems tricky though.

Spitballing ideas here, but one thing we are (to my knowledge) currently lacking is a way to specify availability for POVMs and instruments. Perhaps we could use this to implicitly specify line-labeling requirements? Or otherwise enable a cleaner inference of what is required? (Independent of the line-labeling issue patching this missing part of processor specs is probably something to prioritize regardless given the increasing use of instruments in general, though I think this listed in the instrument omnibus issue already).

pcwysoc:
[...]

PR #418, opened 2024-04-09
Bugfix for fiducial circuit line labels

coreyostrove:
This is a quick bugfix for issue #396. The fix here is to simply force each of the line labels for candidate fiducial circuits to match the state space labels of the input target model. This is a suboptimal fix looking forward to many-qubit systems (as I mentioned in a comment on #396), but we don't presently have the infrastructure needed for doing this a better way and this fix should be fine for 99% of use cases with ExplicitOpModels in the few-qubit setting.


sserita (approved):
Agree that this is a short-term fix but better to have it for now. Thanks!

PR #420, opened 2024-04-16
Bugfix Cirq Datasets and LGST Bug

coreyostrove:
This PR covers two bugfixes. The first is in the method for adding cirq trials to pygsti datasets which did not behave as expected for 2+ qubits. The default behavior of the histogram method of the cirq `Result` object, which is used for gathering aggregated counts for each measurement results is to index these results using integers corresponding to the binary string for an outcome. In pygsti we more commonly use the binary strings themselves as the key. To address this we have added new options to `add_cirq_trial_result` which controls conversion back to a bit string before adding it to the count dictionary.

The second bugfix was to the LGST routine in order to patch in support for datasets with sparse formats wherein we only include outcomes with non-zero counts in the count dict for a circuit. Previously the implementation assumed that the data set was in a dense format with zero entries listed.

### DIFF
--- a/pygsti/algorithms/core.py
+++ b/pygsti/algorithms/core.py
@@ -247,7 +247,10 @@ def run_lgst(dataset, prep_fiducials, effect_fiducials, target_model, op_labels=
                     circuit = rhostr
                 dsRow_fractions = dataset[circuit].fractions
                 # outcome labels should just be effect labels (no instruments!)
-                EVec[0, i] = dsRow_fractions[(effectLabel,)]
+                # when using a sparse data set format it might not be the case
+                # that all effect labels are present (only ones with non-zero counts are)
+                # so return 0 for the fraction in that case.
+                EVec[0, i] = dsRow_fractions.get((effectLabel,), 0)
             EVec_p = _np.dot(_np.dot(EVec, Vd), Pj)  # truncate Evec => Evec', shape (1,trunc)
             povm_effects.append((effectLabel, _np.transpose(EVec_p)))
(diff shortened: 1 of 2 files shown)


pygsti/algorithms/core.py
```diff
@@ -247,7 +247,10 @@ def run_lgst(dataset, prep_fiducials, effect_fiducials, target_model, op_labels=
                     circuit = rhostr
                 dsRow_fractions = dataset[circuit].fractions
                 # outcome labels should just be effect labels (no instruments!)
-                EVec[0, i] = dsRow_fractions[(effectLabel,)]
+                # when using a sparse data set format it might not be the case
+                # that all effect labels are present (only ones with non-zero counts are)
+                # so return 0 for the fraction in that case.
+                EVec[0, i] = dsRow_fractions.get((effectLabel,), 0)
```
sserita, line 253: Perfect, exactly the bugfix I was thinking of!

pygsti/data/dataset.py
```diff
@@ -1619,6 +1619,16 @@ def add_cirq_trial_result(self, circuit, trial_result, key):
         key : str
             The string key of the measurement. Set by cirq.measure.
 
+        convert_int_to_binary : bool, optional (defaut True)
+            By default the keys in the cirq Results object are the integers representing
+            the bitstrings of the measurements on a set of qubits, in big-endian convention.
+            If True this uses the cirq function `cirq.big_endian_int_to_bits` to convert back
```
sserita, line 1625: I was wondering if we were going to get bit by endian-ness... Nice job covering that edge case!

sserita (approved):
Excellent work @coreyostrove!

PR #422, opened 2024-04-16
Bugfix kraus decomp with degeneracies

enielse:
**Fixes a bug in `DenseOperator.kraus_operators` when Choi matrix has degenerate spectrum.**
    
Switches a `numpy.linalg.eig` call to `numpy.linalg.eigh` to ensure that the matrix of eigenvectors is unitary.  This is not always the case for `eig` (in particular for matrices with degenerate spectra), and is assumed by the code in `kraus_operators`.  The `eig` routine has had a history of not always giving a correct diagonalization, and so this PR also adds an assertion to ensure this is the case.

Also adds a unit test that verifies Kraus decomposition works as expected for a depolarizing channel (one example of a channel with a degenerate Choi matrix spectrum).

### DIFF
--- a/pygsti/modelmembers/operations/denseop.py
+++ b/pygsti/modelmembers/operations/denseop.py
@@ -416,11 +416,11 @@ class DenseOperator(DenseOperatorInterface, _KrausOperatorInterface, _LinearOper
         #CHECK 1 (to unit test?) REMOVE
         #tmp_std = _bt.change_basis(superop_mx, self._basis, 'std')
         #B = _bt.basis_matrices('std', superop_mx.shape[0])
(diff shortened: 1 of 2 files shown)


rileyjmurray (approved):
Looks good to me!

rileyjmurray:
@enielse, this PR initially had pygsti's master branch as the target. This made it look like your changes were much farther reaching than they actually are, and that caused GitHub to unnecessarily add a bunch of people as required PR reviewers. You should probably close this PR and open a new PR that targets develop (rather than master) from the beginning.

enielse:
Ah, thanks Riley -- I must have missed that.  I'll close this and reopen another.

PR #432, opened 2024-04-25
Bugfixes and efficiency improvements for basis conversion, inner products, and computing Frobenius distance

### LINKED ISSUE #430: Replace instances of ``np.trace(np.dot(...`` when computing the trace inner product or Frobenius norm.
(On the latest commit of develop, at time of writing. pyGSTi 0.9.12.)

The standard inner product on the space of real $m \times n$ matrices is $(A, B) \mapsto \text{trace}(A^T B)$. There are two candidates for the standard inner products for complex matrices; either $(A, B) \mapsto \text{trace}(A^\dagger B)$ or $(A, B) \mapsto \text{trace}(B^\dagger A)$, which differ up to a complex conjugate.

Several places in pyGSTi require computing these inner products, and explicitly compute a matrix-matrix product before taking a trace. Some places even do this as a way to compute (squared) Frobenius norms. The easy way to find these instances are to search for the text "``_np.trace(_np.dot(``" (I prepend with underscores since that's pyGSTi's import convention, but there are hits without the underscores as well). 

We should replace these instances with a more efficient and consistent pattern. One way to do this is with ``np.einsum``, along with a manual conjugation of one of the matrices.  I suggest ``np.vdot``, since that will handle conjugation consistently for us.

Thoughts, @coreyostrove, @sserita?

rileyjmurray:
Changes:
* Remove frobeniusnorm and frobeniusnorm_squared functions (which were only correct for real inputs). Inline equivalent numpy functions in the few places they were used.
* Replace the np.trace(np.dot(...)) pattern for computing the trace inner product with np.vdot(...). This resolves #430.
* Note that use of np.vdot has a secondary affect of conjugating the first argument when dealing with complex inputs. This resolves a limitation of basis conversion discussed in an email thread titled _Re: Possible pyGSTi bugs w/rt converting between matrix bases?_ that Robin started on Feb 23 of this year.
* I rewrote our implementation for computing state fidelity. It no longer needs the ``_hack_sqrtm`` function. It also has $O(n^2)$ time checks for the fidelity formulas when one of the arguments is rank-1. The overhead of these checks might actually degrade performance for 2-by-2 matrices, but for anything bigger we should be fine. If the overhead in the 2-by-2 case becomes an issue down the line then we can special-case that codepath.

### Old text
Some tests are failing. I'm not sure why. Maybe they're related to the conjugation that's applied when using ``vdot`` in the basis conversion? If so, I would guess that the failing tests are comparing to a nominal value which actually was not correct to begin with. Thoughts, @coreyostrove, @sserita, @enielse?
```shell
FAILED test_packages/extras/test_interpygate.py::InterpygateTestCase::test_timedep_factory - AssertionError: 0.005780277171757022 != 0 within 7 places (0.005780277171757022 difference)
FAILED test_packages/extras/test_interpygate.py::InterpygateTestCase::test_timedep_op - AssertionError: 0.00048145808272016756 != 0 within 7 places (0.00048145808272016756 difference)
FAILED test_packages/extras/test_interpygate.py::InterpygateTestCase::test_timeindep_factory - AssertionError: 0.005780277171757022 != 0 within 7 places (0.005780277171757022 difference)
FAILED test_packages/extras/test_interpygate.py::InterpygateTestCase::test_timeindep_op - AssertionError: 0.00048145808272016756 != 0 within 7 places (0.00048145808272016756 difference)
```

-- No one was able to reproduce these errors (including me, later on!).

### DIFF
--- a/pygsti/baseobjs/basis.py
+++ b/pygsti/baseobjs/basis.py
@@ -547,7 +547,7 @@ class Basis(_NicelySerializable):
         """
         if self.elndim == 2:
             for i, mx in enumerate(self.elements):
-                t = _np.linalg.norm(mx.ravel(), 2) # == sqrt(tr(mx mx))
+                t = _np.linalg.norm(mx) # == sqrt(tr(mx mx))
                 if not _np.isclose(t, 1.0): return False
             return True
         elif self.elndim == 1:
--- a/pygsti/baseobjs/basis.py
+++ b/pygsti/baseobjs/basis.py
@@ -547,8 +547,7 @@ class Basis(_NicelySerializable):
         """
         if self.elndim == 2:
             for i, mx in enumerate(self.elements):
-                t = _np.trace(_np.dot(mx, mx))
-                t = _np.real(t)
+                t = _np.linalg.norm(mx.ravel(), 2) # == sqrt(tr(mx mx))
                 if not _np.isclose(t, 1.0): return False
             return True
         elif self.elndim == 1:
--- a/pygsti/modelmembers/operations/fullcptpop.py
+++ b/pygsti/modelmembers/operations/fullcptpop.py
@@ -93,7 +93,7 @@ class FullCPTPOp(_KrausOperatorInterface, _LinearOperator):
             Lmx = _np.linalg.cholesky(choi_mx)
 
         #check TP condition: that diagonal els of Lmx squared add to 1.0
-        Lmx_norm = _np.linalg.norm(Lmx.ravel())  # = sqrt(tr(Lmx' Lmx))
+        Lmx_norm = _np.linalg.norm(Lmx)  # = sqrt(tr(Lmx' Lmx))
         assert(_np.isclose(Lmx_norm, 1.0)), "Cholesky decomp didn't preserve trace=1!"
 
         self.params = _np.empty(dim**2, 'd')
--- a/pygsti/modelmembers/operations/fullcptpop.py
+++ b/pygsti/modelmembers/operations/fullcptpop.py
@@ -93,7 +93,7 @@ class FullCPTPOp(_KrausOperatorInterface, _LinearOperator):
             Lmx = _np.linalg.cholesky(choi_mx)
 
         #check TP condition: that diagonal els of Lmx squared add to 1.0
-        Lmx_norm = _np.trace(_np.dot(Lmx.T.conjugate(), Lmx))  # sum of magnitude^2 of all els
+        Lmx_norm = _np.linalg.norm(Lmx.ravel())  # = sqrt(tr(Lmx' Lmx))
         assert(_np.isclose(Lmx_norm, 1.0)), "Cholesky decomp didn't preserve trace=1!"
 
         self.params = _np.empty(dim**2, 'd')
--- a/pygsti/report/reportables.py
+++ b/pygsti/report/reportables.py
@@ -538,10 +538,6 @@ if _CVXPY_AVAILABLE:
                 nearby_model.sim.product(self.circuit), mxBasis)
             JBstd = self.d * _tools.fast_jamiolkowski_iso_std(self.B, mxBasis)
             J = JBstd - JAstd
-            # Old code. Keep for now, to make it easier to see correctness of new code.
-            #   Jt = J.T
-            #   val = 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
-            # New code.
             val = 0.5 * (_np.vdot(J.real, self.W.real) + _np.vdot(J.imag, self.W.imag))
             return val
 
@@ -1248,10 +1244,6 @@ if _CVXPY_AVAILABLE:
             JAstd = self.d * _tools.fast_jamiolkowski_iso_std(A, mxBasis)
             JBstd = self.d * _tools.fast_jamiolkowski_iso_std(self.B, mxBasis)
             J = JBstd - JAstd
-            # Old code. Keep for now, to make it easier to see correctness of new code.
-            #   Jt = J.T
-            #   val = 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
-            # New code.
             val = 0.5 * (_np.vdot(J.real, self.W.real) + _np.vdot(J.imag, self.W.imag))
             return val
 
@@ -2031,20 +2023,7 @@ def robust_log_gti_and_projections(model_a, model_b, synthetic_idle_circuits):
                 noise = first_order_noise(opstr, errOnGate, gl)
                 jac[:, i * nSuperOps + k] = [_np.vdot(errOut.flatten(), noise.flatten()) for errOut in error_superops]
 
-                # DEBUG CHECK  # keep old code for now, to show correctness of new code.
-                #
-                # ------------------------------ original code ----------------------------
-                # check = [_np.trace(_np.dot(
-                #     _tools.jamiolkowski_iso(errOut, mxBasis, mxBasis).conj().T,
-                #     _tools.jamiolkowski_iso(noise,  mxBasis, mxBasis))) * 4  # for 1-qubit...
-                #     for errOut in error_superops]
-                # --------------------------- more readable code --------------------------
-                # check = []
-                # for errOut in error_superops:
-                #     arg1 = _tools.jamiolkowski_iso(errOut, mxBasis, mxBasis).conj().T
-                #     arg2 = _tools.jamiolkowski_iso(noise,  mxBasis, mxBasis)
-                #     check.append(_np.trace(_np.dot(arg1, arg2)) * 4)
-                # ---------------------------- efficient code -----------------------------
+                # DEBUG CHECK
                 check = []
                 for errOut in error_superops:
                     arg1 = _tools.jamiolkowski_iso(errOut, mxBasis, mxBasis)
--- a/pygsti/report/reportables.py
+++ b/pygsti/report/reportables.py
@@ -537,8 +537,13 @@ if _CVXPY_AVAILABLE:
             JAstd = self.d * _tools.fast_jamiolkowski_iso_std(
                 nearby_model.sim.product(self.circuit), mxBasis)
             JBstd = self.d * _tools.fast_jamiolkowski_iso_std(self.B, mxBasis)
-            Jt = (JBstd - JAstd).T
-            return 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
+            J = JBstd - JAstd
+            # Old code. Keep for now, to make it easier to see correctness of new code.
+            #   Jt = J.T
+            #   val = 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
+            # New code.
+            val = 0.5 * (_np.vdot(J.real, self.W.real) + _np.vdot(J.imag, self.W.imag))
+            return val
 
     #def circuit_half_diamond_norm(model_a, model_b, circuit):
     #    A = model_a.sim.product(circuit) # "gate"
@@ -1238,12 +1243,17 @@ if _CVXPY_AVAILABLE:
             -------
             float
             """
-            gl = self.oplabel; mxBasis = nearby_model.basis
-            JAstd = self.d * _tools.fast_jamiolkowski_iso_std(
-                nearby_model.operations[gl].to_dense(on_space='HilbertSchmidt'), mxBasis)
+            mxBasis = nearby_model.basis
+            A = nearby_model.operations[self.oplabel].to_dense(on_space='HilbertSchmidt')
+            JAstd = self.d * _tools.fast_jamiolkowski_iso_std(A, mxBasis)
             JBstd = self.d * _tools.fast_jamiolkowski_iso_std(self.B, mxBasis)
-            Jt = (JBstd - JAstd).T
-            return 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
+            J = JBstd - JAstd
+            # Old code. Keep for now, to make it easier to see correctness of new code.
+            #   Jt = J.T
+            #   val = 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
+            # New code.
+            val = 0.5 * (_np.vdot(J.real, self.W.real) + _np.vdot(J.imag, self.W.imag))
+            return val
 
     def half_diamond_norm(a, b, mx_basis):
         """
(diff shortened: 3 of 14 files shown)


pygsti/tools/basistools.py
```diff
@@ -549,9 +549,9 @@ def stdmx_to_vec(m, basis):
     v = _np.empty((basis.size, 1))
     for i, mx in enumerate(basis.elements):
         if basis.real:
-            v[i, 0] = _np.real(_mt.trace(_np.dot(mx, m)))
+            v[i, 0] = _np.real(_np.vdot(mx, m))
         else:
-            v[i, 0] = _np.real_if_close(_mt.trace(_np.dot(mx, m)))
+            v[i, 0] = _np.real_if_close(_np.vdot(mx, m))
```
rileyjmurray, line 554: There are lots of places in this PR where I replaced ``np.trace(np.dot(`` with equivalent calls to ``np.vdot``. The change I've made here is specifically _not_ equivalent. It differs from the existing code by (implicitly) taking the adjoint of ``mx``. That is, this new code is equivalent to ``np.trace(np.dot(mx.T.conj() @ m))``.

pygsti/tools/jamiolkowski.py
```diff
@@ -117,9 +117,14 @@ def jamiolkowski_iso(operation_mx, op_mx_basis='pp', choi_mx_basis='pp'):
...
-            choiMx[i, j] = _mt.trace(_np.dot(opMxInStdBasis, BiBj_dag)) \
-                / _mt.trace(_np.dot(BiBj, BiBj_dag))
+            # BiBj_dag = _np.transpose(_np.conjugate(BiBj))
+            # num = _np.trace(_np.dot(opMxInStdBasis, BiBj_dag))  # original code
+            #     = _np.trace(_np.dot(BiBj_dag, opMxInStdBasis))  # cycle the trace
+            #     = _np.vdot(BiBj, opMxInStdBasis)                # efficient version
+            # den = _np.trace(_np.dot(BiBj, BiBj_dag))
+            num = _np.vdot(BiBj, opMxInStdBasis)
+            den = _np.linalg.norm(BiBj.ravel()) ** 2
+            choiMx[i, j] = num / den
```
rileyjmurray, line 127: If folks are happy with the correctness of this change then we can remove the commented-out code.

pygsti/tools/matrixtools.py
```diff
@@ -31,37 +31,6 @@
...
-#    with warnings.catch_warnings():
-#        warnings.filterwarnings('error')
-#        try:
-#            ret =
-#        except Warning:
-#            print "BAD trace from:\n"
-#            for i in range(M.shape[0]):
-#                print M[i,i]
-#            raise ValueError("STOP")
-#    return ret
```
rileyjmurray, line 62: This function is now inlined with ``np.trace``. The reported memory leak was fixed a long time ago.

pygsti/tools/matrixtools.py
```diff
@@ -125,47 +94,7 @@ def is_valid_density_mx(mx, tol=1e-9):
     bool
         True if mx is a valid density matrix, otherwise False.
     """
-    return is_hermitian(mx, tol) and is_pos_def(mx, tol) and abs(trace(mx) - 1.0) < tol
```
rileyjmurray, line 128: I'm not sure why GitHub thinks I removed this line. If you scroll farther down you'll see a "new" line of code that's identical to this old line of code.

test/unit/tools/test_optools.py
```diff
@@ -387,13 +387,6 @@ def setUp(self):
             [-0.35432747-0.27939404j, -0.02266757+0.71502652j, -0.27452307+0.07511567j,  0.35432747+0.27939404j],
             [ 0.71538573+0.j,  0.2680266 +0.36300238j, 0.2680266 -0.36300238j,  0.28461427+0.j]])
 
-    def test_frobenius_distance(self):
-        self.assertAlmostEqual(ot.frobeniusdist(self.A, self.A), 0.0)
-        self.assertAlmostEqual(ot.frobeniusdist(self.A, self.B), (0.430116263352+0j))
-
-        self.assertAlmostEqual(ot.frobeniusdist_squared(self.A, self.A), 0.0)
-        self.assertAlmostEqual(ot.frobeniusdist_squared(self.A, self.B), (0.185+0j))
-
```
rileyjmurray, line 396: Our implementation is trivial and now no longer needs testing code.

test/unit/tools/test_optools.py
```diff
@@ -404,11 +397,6 @@ def test_diamond_distance(self):
         self.assertAlmostEqual(ot.diamonddist(self.A, self.A, mx_basis="std"), 0.0)
         self.assertAlmostEqual(ot.diamonddist(self.A, self.B, mx_basis="std"), 0.614258836298)
 
-    def test_frobenius_norm_equiv(self):
-        from pygsti.tools import matrixtools as mt
-        self.assertAlmostEqual(ot.frobeniusdist(self.A, self.B), mt.frobeniusnorm(self.A - self.B))
-        self.assertAlmostEqual(ot.frobeniusdist(self.A, self.B), np.sqrt(mt.frobeniusnorm_squared(self.A - self.B)))
-
```
rileyjmurray, line 411: I removed the ``frobeniusnorm`` and ``frobeniusnorm_squared`` functions.

pygsti/report/reportables.py
```diff
@@ -1238,12 +1243,17 @@ def evaluate_nearby(self, nearby_model):
...
+            JAstd = self.d * _tools.fast_jamiolkowski_iso_std(A, mxBasis)
             JBstd = self.d * _tools.fast_jamiolkowski_iso_std(self.B, mxBasis)
-            Jt = (JBstd - JAstd).T
-            return 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
+            J = JBstd - JAstd
+            # Old code. Keep for now, to make it easier to see correctness of new code.
+            #   Jt = J.T
+            #   val = 0.5 * _np.trace(_np.dot(Jt.real, self.W.real) + _np.dot(Jt.imag, self.W.imag))
+            # New code.
+            val = 0.5 * (_np.vdot(J.real, self.W.real) + _np.vdot(J.imag, self.W.imag))
```
rileyjmurray, line 1255: If people are okay with the new implementation then I'll remove the comments.

pygsti/report/reportables.py
```diff
@@ -537,8 +537,13 @@ def evaluate_nearby(self, nearby_model):
...
+            return val
```
rileyjmurray, line 546: If people are okay with the new implementation then I'll remove the comments.

pygsti/report/reportables.py
```diff
@@ -1238,12 +1243,17 @@ def evaluate_nearby(self, nearby_model):
...
```
rileyjmurray, line 1256: This change is identical to one above. If folks agree with the change, I'll remove the comments.

sserita:
Haven't done a full review yet, but I think I have the most familiarity with the interpygate code so I will put that on my todo list. The interpygate tests are much closer to integration tests than true unit tests, so it is possible the nominal value was wrong, but needs more investigation.

coreyostrove (changes requested):
Thanks for the great work, @rileyjmurray! I've left a few minor questions for you as part of my review, but by and large these changes all look good, and should bring some performance gains to boot!

coreyostrove:
As for the failing unit tests you've encountered, I pulled the latest version of this branch and tested this locally on my machine and was unable to reproduce this behavior. That would likely point to this either being a non-deterministic error, or else something configuration dependent in which case we'd need to track this down a bit more. One bit of low-hanging fruit we can try to eliminate is whether this is somehow macOS specific, so I'll manually launch the mac runners on github to see if we can identify anything there.

coreyostrove:
Update: I ran the unit tests in question on the runners, and with the exception of the MacOS runner for python 3.8 (which failed because of perennial problems with compiling CVXOPT that typically fix themselves after a while) these all passed. So whatever is going on it seems to be localized to the particular combination of packages/configuration on your machine, @rileyjmurray.

Concerns about potential testing issues was the main reason why I didn't mark this as approved when doing my review, so I'm going to switch that designation now that I'm seeing these passing. The other question I had I am still interested in, but these aren't blocking concerns/questions in any way.

coreyostrove (approved):

pygsti/baseobjs/basis.py
```diff
@@ -547,8 +547,7 @@ def is_normalized(self):
         """
         if self.elndim == 2:
             for i, mx in enumerate(self.elements):
-                t = _np.trace(_np.dot(mx, mx))
-                t = _np.real(t)
+                t = _np.linalg.norm(mx.ravel(), 2) # == sqrt(tr(mx mx))
```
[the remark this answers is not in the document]
rileyjmurray: I suppose the ravel isn't needed. I only added it because I prefer to be explicit when specifying the norm, and ``norm(arg.ravel(), 2)`` will always work for the Frobenius norm, while ``norm(arg, 'fro')`` will only work when ``arg.ndim == 2``. 

I'll change to ``norm(arg)`` here and elsewhere.

test/unit/tools/test_basisconstructors.py
```diff
@@ -40,8 +40,7 @@ def test_orthogonality(self):
         gm_trMx = np.zeros((N, N), 'complex')
         for i in range(N):
             for j in range(N):
-                gm_trMx[i, j] = np.trace(np.dot(np.conjugate(np.transpose(mxs[i])), mxs[j]))
-                #Note: conjugate transpose not needed since mxs are Hermitian
+                gm_trMx[i, j] = np.vdot(mxs[i], mxs[j])
```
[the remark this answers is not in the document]
rileyjmurray: I'll leave it as-is.

pygsti/tools/optools.py
```diff
@@ -105,17 +105,15 @@ def fidelity(a, b):
         evals = _np.linalg.eigvals(a)
         _warnings.warn(("sqrtm(a) failure when computing fidelity - beware result. "
                         "Maybe due to rank defficiency - eigenvalues of a are: %s") % evals)
-    F = (_mt.trace(_hack_sqrtm(_np.dot(sqrtA, _np.dot(b, sqrtA)))).real)**2  # Tr( sqrt{ sqrt(a) * b * sqrt(a) } )^2
+    F = (_np.trace(_hack_sqrtm(_np.dot(sqrtA, _np.dot(b, sqrtA)))).real)**2  # Tr( sqrt{ sqrt(a) * b * sqrt(a) } )^2
```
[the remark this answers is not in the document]
rileyjmurray: I actually changed the fidelity implementation on my branch for leakage-aware error metrics and gauge optimization. I think it'll be a while before that branch is merged, so I've added that change to this PR. Take a look and LMK if it looks okay to you? The most important change is that we now raise an error if ``a`` or ``b`` isn't Hermitian up to a certain tolerance.

pygsti/tools/optools.py
```diff
@@ -81,41 +68,92 @@ def fidelity(a, b):
...
+            message = f"""
+                Input matrix 'mat' is not Hermitian, up to tolerance {__SCALAR_TOL__}.
+                The absolute values of entries in (mat - mat^H) are \n{hermiticity_error}. 
+            """
+            raise ValueError(message)
+    
+    assert_hermitian(a)
+    assert_hermitian(b)
+
+    def check_rank_one_density(mat):
```
coreyostrove, line 87: Could you add a sentence or two explaining the logic to the comments for this function (or a reference to the algorithm if it is well known). I'm going to try and parrot back what I think the logic is here, along with some stuff I am confused by, to see if I understood correctly (I'm going to do so with bra-ket notation, so I apologize). Assume mat is a rank 1 matrix. If so we can write it as a rank 1 projector |v><v|. Let |w> be a random test vector. Applying mat to |w> we get  <v|w>*|v>, which can can renormalize to get what you called candidate_v. I.e. if our assumption is correct then candidate_v = |v>. alpha is the (candidate_v, ) matrix element of mat: <cand_v|mat|cand_v>.  Finally, take the outer product of |cand_v> with itself and compare to mat. If these are sufficiently close it must've been the case that mat was indeed a rank 1 projector, otherwise we have a proof by contradiction that mat was not a rank 1 projector. 

Questions: 
1. Is there any reason why the test vector is chosen randomly? I would have thought that any test vector with full support would have worked, e.g. the all-ones vector.
2. Why do we need alpha here? If I described this correctly then shouldn't alpha always be 1 in the case where mat is a rank 1 projector?

pygsti/tools/optools.py
```diff
@@ -81,41 +68,92 @@ def fidelity(a, b):
...
+            The PSD part of the input matrix is not trace-1 up to tolerance {__VECTOR_TOL__}.
+            Beware result!
+            """
+            _warnings.warn(message)
+        sqrt_mat = U @ (_np.sqrt(evals).reshape((-1, 1)) * U.T.conj())
+        return sqrt_mat
+    
+    sqrt_a = psd_square_root(a)
+    tr_arg = psd_square_root(sqrt_a @ b @ sqrt_a)
+    f = _mt.trace(tr_arg).real ** 2  # Tr( sqrt{ sqrt(a) * b * sqrt(a) } )^2
```
coreyostrove: Inline numpy, this function was deleted.

coreyostrove (commented):
Added a few more comments and questions related to your most recent changes.

test/unit/tools/test_optools.py
```diff
@@ -388,36 +380,35 @@ def setUp(self):
...
         if SKIP_DIAMONDIST_ON_WIN and sys.platform.startswith('win'): return
-        self.assertAlmostEqual(ot.diamonddist(self.A, self.A, mx_basis="std"), 0.0)
-        self.assertAlmostEqual(ot.diamonddist(self.A, self.B, mx_basis="std"), 0.614258836298)
+        val = ot.diamonddist(self.A_TP, self.A_TP, mx_basis="pp")
+        self.assertAlmostEqual(val, 0.0)
+        val = ot.diamonddist(self.A_TP, self.B_unitary, mx_basis="pp")
+        self.assertGreaterEqual(val, 0.7)
 
     def test_entanglement_fidelity(self):
-        fidelity = ot.entanglement_fidelity(self.A, self.B)
```
coreyostrove, line 413: Why was this test removed?

pygsti/baseobjs/basis.py
```diff
@@ -547,8 +547,7 @@ def is_normalized(self):
         """
         if self.elndim == 2:
             for i, mx in enumerate(self.elements):
-                t = _np.trace(_np.dot(mx, mx))
-                t = _np.real(t)
+                t = _np.linalg.norm(mx) # == sqrt(tr(mx mx))
```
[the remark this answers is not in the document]
rileyjmurray: Good point! Yes, this definitely would have been a bug, lol.

pygsti/modelmembers/operations/fullcptpop.py
```diff
@@ -93,7 +93,7 @@ def _set_params_from_choi_mx(self, choi_mx, truncate):
             Lmx = _np.linalg.cholesky(choi_mx)
 
         #check TP condition: that diagonal els of Lmx squared add to 1.0
-        Lmx_norm = _np.trace(_np.dot(Lmx.T.conjugate(), Lmx))  # sum of magnitude^2 of all els
+        Lmx_norm = _np.linalg.norm(Lmx)  # = sqrt(tr(Lmx' Lmx))
```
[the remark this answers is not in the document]
rileyjmurray: The square root is just because of using the norm as opposed to the squared norm. It's fine to use either in this context since we're checking that a quantity is near one. In general, if I have an expected value for a norm or squared norm, then I prefer to state it in terms of the norm.

test/unit/tools/test_optools.py
rileyjmurray: ``A`` and ``B`` are psd matrices, not operation matrices. If you try to get their Choi matrix representations then you get a non-Hermitian matrix as a result, which leads to an error (correctly) getting thrown by the new implementation of fidelity.

pygsti/tools/optools.py
rileyjmurray: @coreyostrove I've updated this code to be more readable and to include comments. Take another look?

pygsti/tools/optools.py
rileyjmurray: Fixed!

pygsti/tools/optools.py
coreyostrove: Thanks, I had thought this looked a lot like you were doing a power-iterationesque algorithm here. I'm admittedly still not sure why randomness is necessary here, as all for the power iteration all you need is non-trivial overlap with the dominant eigenspace. But it is clear to me this is sufficient, if not necessary, so I'm cool with suspending disbelief. Thanks for adding the additional explanation (future us will thank you!).

pygsti/report/reportables.py
```diff
@@ -2021,11 +2031,26 @@ def error_generator_jacobian(opstr):
...
+                #     _tools.jamiolkowski_iso(errOut, mxBasis, mxBasis).conj().T,
+                #     _tools.jamiolkowski_iso(noise,  mxBasis, mxBasis))) * 4  # for 1-qubit...
+                #     for errOut in error_superops]
+                # --------------------------- more readable code --------------------------
+                # check = []
+                # for errOut in error_superops:
+                #     arg1 = _tools.jamiolkowski_iso(errOut, mxBasis, mxBasis).conj().T
+                #     arg2 = _tools.jamiolkowski_iso(noise,  mxBasis, mxBasis)
+                #     check.append(_np.trace(_np.dot(arg1, arg2)) * 4)
+                # ---------------------------- efficient code -----------------------------
```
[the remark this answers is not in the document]
rileyjmurray: Taking Corey's approval as an OK for the new implementation; removing these comments and marking as resolved.

pygsti/report/reportables.py
rileyjmurray: Taking Corey's approval as an OK for the new implementation; removing these comments and marking as resolved.

pygsti/report/reportables.py
rileyjmurray: Taking Corey's approval as an OK for the new implementation; removing these comments and marking as resolved.

pygsti/tools/jamiolkowski.py
rileyjmurray: Taking Corey's approval as an OK for the new implementation; removing these comments and marking as resolved.

rileyjmurray:
@sserita, I've gotten Corey's sign-off. Can you approve and merge?

sserita (approved):
Thanks @coreyostrove and @rileyjmurray for the great work :)

PR #439, opened 2024-05-14
Fix densitymx slow bug.

### LINKED ISSUE #438: densitymx_slow does not work
**Describe the bug**
The densitymx_slow implementation does not work. At `pygsti/evotypes/densitymx_slow/opreps.py:177`,

```python
class OpRepStandard(OpRepDenseSuperop):
    def __init__(self, name, basis, state_space):
        ...
        super(OpRepStandard, self).__init__(superop, state_space)
```

The issue is that `OpRepDenseSuperof.__init__` doesn't have that signature;

```python
class OpRepDenseSuperop(OpRep):
    def __init__(self, mx, basis, state_space):
        ...
```

**To Reproduce**
1. Install pygsti without cython
2. Try and run a circuit
3. :(

**Expected behavior**
Ideally we would get the fast cython implementation working. However, while setting up a new user it would be nice for the slow fallback to work so we can get _something_ out.

sserita:
Primarily fixes #438.

Also cleans up Github Actions for exactly the same commands between OSes for running tests, creates a [testing_no_cython] option for installing, adds a no Cython environment for testing via Actions, and deletes the manualdeploy Action as the autodeploy is now working reliably (to minimize maintenance overhead of actions).


coreyostrove (approved):
Everything looks good to me, thanks for taking care of this!

PR #442, opened 2024-05-22
Refactor functions in matrixtools.py

### LINKED ISSUE #429: Audit implementations of functions in ``matrixtools.py``.
The [matrixtools.py](https://github.com/sandialabs/pyGSTi/blob/ad0c1a5cbcdf081419d8b68ded0ae0da067da66b/pygsti/tools/matrixtools.py) file is almost 3000 lines long. Some functions in this file are much less efficient than they could be. For example, the check for a valid density matrix is implemented as follows: https://github.com/sandialabs/pyGSTi/blob/ad0c1a5cbcdf081419d8b68ded0ae0da067da66b/pygsti/tools/matrixtools.py#L128

This is inefficient because Python will evaluate the boolean conditions from left to right, so we should start with the cheapest check and end with the most expensive check. What we actually see is that the cheapest check happens at the end and the most expensive check (requiring an eigendecomposition) happens in the middle.

I suspect there are many functions with subtle inefficiencies like this. I'm creating this issue so that I can make sure to sit down and do a proper audit of these functions, rather than identifying inefficiencies and fixing them piecemeal.

rileyjmurray:
This PR resolves #429. I've provided detailed comments explaining my changes to ``matrixtools.py``. I've also removed some commented-out code.


pygsti/modelmembers/povms/composedeffect.py
rileyjmurray, line 1: These changes just removed commented-out code.

pygsti/modelmembers/states/composedstate.py
rileyjmurray, line 1: These changes just removed commented-out code.

test/unit/tools/test_matrixtools.py
```diff
@@ -17,8 +17,8 @@ def test_is_hermitian(self):
         self.assertFalse(mt.is_hermitian(non_herm_mx))
 
     def test_is_pos_def(self):
-        pos_mx = np.array([[ 4, 0.2],
-                           [0.1, 3]], 'complex')
+        pos_mx = np.array([[ 4.0, 0.2],
+                            [0.2, 3.0]], 'complex')
```
rileyjmurray, line 21: This matrix was not positive semidefinite, since it wasn't Hermitian.

pygsti/modelmembers/operations/lindbladerrorgen.py
rileyjmurray, line 1: Changes to this file were just to remove commented-out code.

test/unit/objects/test_fogi.py
```diff
@@ -170,15 +170,16 @@ def test_crosstalk_free_fogi(self):
...
+        temp = mdl.fogi_errorgen_components_array(include_fogv=False, normalized_elem_gens=True)
+        self.assertArraysAlmostEqual(temp, mdl.to_vector()[nprefix:])
 
         v = mdl.to_vector()  # just test this works
 
         # Test if from_vector works too
         w = np.random.rand(mdl.num_params)
         w[0:nprefix] = 0 # zero out all unused params (these can be SPAM and can't be any value?)
         mdl.from_vector(w)
+        pass
```
rileyjmurray, line 182: These changes are just for readability and debugging.

pygsti/tools/matrixtools.py
```diff
@@ -49,14 +82,13 @@ def is_hermitian(mx, tol=1e-9):
...
     (m, n) = mx.shape
-    for i in range(m):
-        if abs(mx[i, i].imag) > tol: return False
-        for j in range(i + 1, n):
-            if abs(mx[i, j] - mx[j, i].conjugate()) > tol: return False
-    return True
+    if m != n:
+        return False
+    else:
+        return _np.all(_np.abs(mx - mx.T.conj()) <= tol)
```
rileyjmurray, line 88: The old implementation used O(1) memory. This implementation uses O(mx.size) memory, but it avoids any Python for-loops. We could implement is_hermitian in Cython if we wanted to keep the memory requirement at O(1).

pygsti/tools/matrixtools.py
```diff
@@ -73,7 +105,15 @@ def is_pos_def(mx, tol=1e-9):
...
-    evals = _np.linalg.eigvals(mx)
+    if not is_hermitian(mx, tol):
+        return False
+    if attempt_cholesky:
+        try:
+            _ = _spl.cholesky(mx)
+            return True # Cholesky succeeded
+        except _spl.LinAlgError:
+            pass # we fall back on eigenvalue decomposition
+    evals = _np.linalg.eigvalsh(mx)
```
rileyjmurray, line 116: The property of being positive semidefinite is generally ascribed to Hermitian matrices, since the value of the quadratic form $x^\dagger A x = x^\dagger((A + A^{\dagger})/2)x$ only depends on the Hermitian part of $A$. The original implementation didn't check that, and so it sometimes reported that asymmetric matrices with positive eigenvalues were positive definite.

The new implementation starts with a check that the matrix is Hermitian. If we do have a Hermitian matrix then the fastest way to check if it's positive definite is a Cholesky decomposition. However, Cholesky will fail on matrices that are only positive semidefinite, so there is no point in trying Cholesky if we expect ``is_pos_def`` will regularly handle singular matrices. I've set the default behavior for ``attempt_cholesky`` to False.

pygsti/tools/matrixtools.py
```diff
@@ -177,21 +215,22 @@ def nice_nullspace(m, tol=1e-7, orthogonalize=False):
...
+        return nullsp  # empty 0-by-N array
+    _, _, p = _spl.qr(nullsp.T.conj(), mode='raw', pivoting=True)
+    ret = nullsp @ (nullsp.T[:, p[:dim_ker]]).conj()
+    # ^ That's equivalent to, but faster than:
+    #     nullsp_projector = nullsp @ nullsp.T.conj()
+    #     _, _, p = _spl.qr(nullsp_projector mode='raw', pivoting=True)
+    #     ret = nullsp_projector[:, p[:dim_ker]]
+
+    if orthogonalize:
+        ret, _ = _spl.qr(ret, mode='economic')
```
rileyjmurray, line 229: Let's say ``m`` is $M \times N$ and $K =$``dim_ker`` is positive. Generically, we'll have $K = N - M$.

Computing the initial nullspace for the old implementation and the new implementation costs $O(M^2 N)$. In the new implementation that's the only operation which costs $O(M^2 N)$.

The original implementation had other expensive steps: SVDs of sizes at least $M \times \ell$ for $\ell \in \\{1,\ldots,\min(M,K)\\}$ and as large as $M \times \ell$ for $\ell \in \\{N-K+1,\ldots,N\\}$. The total complexity of these SVDs was around $O(M^4)$, if not larger.

pygsti/tools/matrixtools.py
```diff
@@ -248,14 +287,16 @@ def column_norms(m, ord=None):
...
-        #this could be done more efficiently, e.g. by converting to csc and taking column norms directly
+        ord_list = ord if isinstance(ord, (list, _np.ndarray)) else [ord] * m.shape[1]
+        assert(len(ord_list) == m.shape[1])
         norms = _np.array([_np.linalg.norm(m[:, j].toarray(), ord=o) for j, o in enumerate(ord_list)])
+    elif isinstance(ord, (list, _np.ndarray)):
+        assert(len(ord) == m.shape[1])
+        norms = _np.array([_np.linalg.norm(m[:, j], ord=o) for j, o in enumerate(ord)])
     else:
-        norms = _np.array([_np.linalg.norm(m[:, j], ord=o) for j, o in enumerate(ord_list)])
+        norms = _np.linalg.norm(m, axis=0, ord=ord)
```
rileyjmurray, line 298: The main change here was to use vectorization when possible.

pygsti/tools/matrixtools.py
```diff
@@ -337,9 +379,11 @@ def columns_are_orthonormal(m, tol=1e-7):
...
     bool
     """
-    if m.size == 0: return True  # boundary case
-    check = _np.dot(m.conj().T, m)
-    return bool(_np.allclose(check, _np.identity(check.shape[0], 'd'), atol=tol))
+    if m.size == 0:
+        return True  # boundary case
+    check = gram_matrix(m)
+    check[_np.diag_indices_from(check)] -= 1.0
+    return bool(_np.linalg.norm(check) / check.size < tol)
```
rileyjmurray, line 386: This has half the memory requirement of the old implementation.

pygsti/tools/matrixtools.py
```diff
@@ -369,27 +413,28 @@ def independent_columns(m, initial_independent_cols=None, tol=1e-7):
...
-                num_indep_cols += 1
+        if initial_independent_cols is None:
+            proj_m = m.copy()
+        else:
+            assert initial_independent_cols.shape[0] == m.shape[0]
+            q = _spl.qr(initial_independent_cols, mode='econ')[0]
+            # proj_m = (I - qq')m
+            temp1 = q.T.conj() @ m
+            temp2 = q @ temp1
+            proj_m = m - temp2
```
rileyjmurray, line 425: The changes here are comparable to those I made in ``nice_nullspace``. The runtime has improved to $O(MN^2)$ from the original $O(MN^3)$.

pygsti/tools/matrixtools.py
```diff
@@ -408,15 +453,33 @@ def independent_columns(m, initial_independent_cols=None, tol=1e-7):
...
     """
-    The "sign" matrix of `m`
+    Compute the matrix s = sign(m). The eigenvectors of s are the same as those of m.
+    The eigenvalues of s are +/- 1, corresponding to the signs of m's eigenvalues.
+
+    It's straightforward to compute s when m is a normal operator. If m is not normal,
+    then the definition of s can be given in terms of m's Jordan form, and s
+    can be computed by (suitably post-processing) the Schur decomposition of m.
+
+    See https://nhigham.com/2020/12/15/what-is-the-matrix-sign-function/ for background.
```
rileyjmurray, line 482: I made new codepaths that are more efficient for Hermitian matrices and normal matrices. (The asymptotics are the same, but they avoid running nested for-loops in Python.)

pygsti/tools/matrixtools.py
```diff
@@ -904,9 +955,13 @@ def norm1(m):
     -------
     numpy.ndarray
     """
-    return float(_np.real(_np.trace(_sqrtm(_np.dot(m.conj().T, m)))))
+    s = _spl.svdvals(m)
+    nrm = _np.sum(s)
+    return nrm
```
rileyjmurray, line 960: The new implementation is asymptotically faster if ``m`` is wide. In all other situations the asymptotic runtime is unchanged, but the constant factors are better for computation and memory use.

pygsti/tools/matrixtools.py
```diff
@@ -925,12 +980,13 @@ def random_hermitian(dim):
...
         b = _np.random.random(size=[dim, dim])
-        c = a + 1.j * b + (a + 1.j * b).conj().T
+        c = a + 1.j * b
+        c += c.conj().T
         my_norm = norm1(c)
     return c / my_norm
 
 
-def norm1to1(operator, num_samples=10000, mx_basis="gm", return_list=False):
+def norm1to1(operator, num_samples=10000, mx_basis="gm"):
```
rileyjmurray, line 989: This function was never called with ``return_list=True``, so I removed that option. The efficiency improvements are very minor.

pygsti/tools/matrixtools.py
```diff
@@ -1398,78 +1454,6 @@ def safe_dot(a, b):
...
-                ret = _sps.csr_matrix((_np.imag(a.data).copy(), a.indices.copy(),
-                                       a.indptr.copy()), shape=a.shape, dtype='d')
-            ret.eliminate_zeros()
-            return ret
-        else:
-            raise NotImplementedError("safe_real() doesn't work with %s matrices yet" % str(type(a)))
-    else:
-        return _np.imag(a)
-
-
```
rileyjmurray, line 1472: I looked at where these functions were called in pygsti. The ``check`` was never set to True and sparse matrices were never passed. So I removed these functions and replaced with inlined code as needed.

pygsti/tools/matrixtools.py
```diff
@@ -2243,30 +2227,6 @@ def project_onto_antikite(mx, kite):
...
-    for j in range(mx.shape[1]):
-        rnk = _np.linalg.matrix_rank(mx[:, 0:j + 1], tol)
-        if rnk == last_rank:
-            cols_to_remove.append(j)
-        else:
-            last_rank = rnk
-    #print("Removing %d cols" % len(cols_to_remove))
-    return _np.delete(mx, cols_to_remove, axis=1)
-
-
```
rileyjmurray, line 2269: This was called in one function (below), which I reimplemented to rely on ``independent_columns``.

test/unit/tools/test_matrixtools.py
```diff
@@ -160,42 +160,6 @@ def test_fancy_assignment(self):
...
-        M = mx.copy(); M = mt.safe_real(M, inplace=True)
-        self.assertArraysAlmostEqual(M, np.real(mx))
-        M = mx.copy(); M = mt.safe_imag(M, inplace=True)
-        self.assertArraysAlmostEqual(M, np.imag(mx))
-
-        M = smx.copy(); M = mt.safe_real(M, inplace=True)
-        self.assertArraysAlmostEqual(M.toarray(), np.real(mx))
-        M = smx.copy(); M = mt.safe_imag(M, inplace=True)
-        self.assertArraysAlmostEqual(M.toarray(), np.imag(mx))
-
```
rileyjmurray, line 198: I removed the functions considered in this test.

rileyjmurray:
@sserita and @enielse, please take a look when you can.

pygsti/extras/interpygate/__init__.py
```diff
@@ -10,3 +10,10 @@
 
 from .core import PhysicalProcess, InterpolatedDenseOp, InterpolatedOpFactory
 from .process_tomography import vec, unvec, run_process_tomography
+
+# Note from Riley on May 22, 2024:
+#
+#   I wanted to remove the implementations of vec and unvec and just in-line equivalent 
+#   code in the few places they were used. However, the fact that they're included in this
+#   __init__.py file suggests that they might be used outside of pyGSTi itself.
+#
```
sserita, line 19: We can check with @kevincyoung and @bruzic3, but I think I'm OK with inlining these in our source, deprecating the functions in pyGSTi, and then assisting/enforcing the relevant changes in the downstream code so that they can be removed if they are inefficient/not best practice.

pygsti/tools/matrixtools.py
```diff
@@ -30,6 +30,53 @@
...
+BLAS_FUNCS = {
+    'herk': {
+        's' : _spl.blas.ssyrk,
+        'd' : _spl.blas.dsyrk,
+        'c' : _spl.blas.cherk,
+        'z': _spl.blas.zherk
+    }
+}
+
+def gram_matrix(m, adjoint=False):
```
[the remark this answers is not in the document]
sserita: Knowing all these BLAS settings seems like a... Herk-ulean task.

pygsti/tools/matrixtools.py
sserita: A couple of things:

1. Should we ensure the docstring is consistent? It says PD in the first line, but return says PSD.
2. Both implementations are clearly for PSD. Should the function name be `is_pos_semidef` instead? Pro is that it is more accurate, con is that it is an API change. While this is in matrixtools and therefore could have moderate/high user exposure, it seems like the only use is in `is_valid_density_mx` below (which also seems like it has low usage), so maybe the API change would go mostly unnoticed?

pygsti/tools/matrixtools.py
sserita: Is it the `pivoting=True` that essentially retains the same behavior as the old implementation of only taking the columns that increase the rank? I.e. pivoting will sort it/give us the indices such that taking the first `dim_ker` columns will return the matrix of the expected rank?

pygsti/tools/matrixtools.py
```diff
@@ -904,9 +954,13 @@ def norm1(m):
     -------
     numpy.ndarray
     """
-    return float(_np.real(_np.trace(_sqrtm(_np.dot(m.conj().T, m)))))
+    s = _spl.svdvals(m)
+    nrm = _np.sum(s)
+    return nrm
 
 
+# Riley note: I'd like to rewrite this, but I don't want to mess with reproducibility
```
sserita, line 962: Lol what reproducibility? We aren't seeding the RNG here so this is essentially impossible to reproduce without being very careful about calls to np.random in all scopes...
I'm generally in favor of a new implementation, or at least one where we can seed it such that we make reproducing it possible. Doesn't have to be in this PR, but just a future to-do.

pygsti/tools/matrixtools.py
```diff
@@ -1372,6 +1424,9 @@ def _findx(a, inds, always_copy=False):
         return a_inds
 
 
+# TODO: reevaluate the need for this function. It seems like we could just in-line @
```
sserita, line 1427: Generally in favor of this. Not a requirement for approval for this PR though, can punt to future.

test/unit/tools/test_matrixtools.py
sserita: Makes me wonder if it was actually intended for it to be PD, not PSD in that case... Although the implementation was clearly PSD. Either way, I'm happy for the test to actually test the PSD edge case, although perhaps a version of the test with a PD matrix hitting the `attempt_cholesky` codepath could be worth adding?

sserita (changes requested):
Thanks for the great effort @rileyjmurray, as always! Also, I greatly appreciate your preemptive comments on some of the more technical changes, it makes it much easier to follow. :)

There is only one real action item here, for an additional unit test for the `attempt_cholesky` codepath in `is_pos_def` (and a small docstring update). However, I've also made some non-blocking comments/questions. I 👍'd any of yoiur comments that I read and agreed with but had no active comments on.

pygsti/tools/matrixtools.py
rileyjmurray: In the context of quantum information I find people are lax about positive definite versus positive semidefinite, so I think the function name is fine. But I'll definitely change the docstring to be consistent about really checking that a matrix is PSD.

pygsti/tools/matrixtools.py
rileyjmurray: Yup! Pivoting essentially sorts columns according to how they contribute to the numerical rank of the matrix. So if you know in advance that the matrix is rank $r$ (say, by having computed an SVD, or knowing that it's full-rank), then you can identify a set of columns that span its range by taking the first $r$ columns from pivoted QR decomposition.

The pivots chosen by pivoted QR are also invariant under left-multiplication by (partial) isometries, so I could pivot on ``nullsp.T.conj()`` instead of ``nullsp @ nullsp.T.conj()``.

pygsti/tools/matrixtools.py
rileyjmurray: Fair point! I have a slight preference to defer these discussions till a new PR.

pygsti/tools/matrixtools.py
rileyjmurray: Cool. I'll try to make this change then.

test/unit/tools/test_matrixtools.py
rileyjmurray: I'm not worried about testing the various PSD vs PD codepaths. I've written functions like ``is_pos_def`` a dozen times. It comes up so often with cvxpy that I actually wrote all this code:
https://github.com/cvxpy/cvxpy/blob/19bd785ae430d5e81aaf1cea3b52812edc21cc56/cvxpy/utilities/linalg.py#L50-L257.

The test matrix was neither PSD nor PD simply because being Hermitian is part of the (far and away most common) definition of PSD/PD-ness.

sserita (approved):
Looks good, thanks @rileyjmurray!

Issue #408, opened 2024-03-07
RB edesigns' `truncate_to_available_data` fails to truncate auxiliary information


jordanh6:
MirrorRBDesign.truncate_to_available_data truncates the the edesign's circuit_lists, but it does not truncate idealout_lists. This means that the mapping between circuits and target bit strings is lost. 

Example notebook attached. 

Expected behavior: Returned edesign should have idealout_lists that contain exactly the idealouts of the circuits remaining in the truncated design. 

Based on skimming the code I suspect other RB protocol edesigns have an analogous issue. 

Environment:
 - pyGSTi v0.9.12.1
 - python 3.8

[edesign-truncate-bug.zip](https://github.com/sandialabs/pyGSTi/files/14519065/edesign-truncate-bug.zip)

sserita:
Closing as this is merged in develop with #443

Issue #315, opened 2023-05-18
Method for Clifford RB to return gates-per-Clifford


coreyostrove:
There currently is not a method in the clifford RB implementation that returns the number of gates per clifford operation, which is relevant for mapping the per-clifford error rate estimated in RB to an approximate per-gate error rate.

sserita:
@coreyostrove Do you remember exactly what your use case/intended solution for this looked like?

A recap of the relevant code path: The `CliffordRBDesign` takes a list of `depths` that passes those depths on to `create_clifford_rb_circuit`, which samples the correct number of Cliffords and then uses a `CompilationRules` object to convert it into native gates.
Given that, it seems like there are two things you could want:

1. The `CliffordRBDesign` knows the number of Cliffords sampled, as well as the length of the circuit in the terms of native gates. This could give an average number of native gates per Clifford.
2. We could look up the number of native gates per Clifford from the `CompilationRules` object directly (which again are given as lengths of "compiled" circuits).

Depending on whether we wanted some sort of member function for `CliffordRBDesign` or just a utility function may determine which strategy we want to take.

coreyostrove:
I was just the messenger for this one (iirc from a list of suggestions/requests brought up in an IRL roadmapping session). I believe @jordanh6 was the brainchild behind this (or else may recall who was) and may be able to provide some context.

For what is is worth, I think having both items 1 and 2 would likely be useful, but I don't personally have an articulated need readily in mind.

jordanh6:
I don't recall suggesting this, so it was probably @tjproct.

In my opinion both of your solutions are reasonable, but I also don't have a use case in mind beyond what's in the initial comment (rescaling the RB error rate). I have personally used approximately solution (1) to rescale RB error rates before on 3+ qubits, where you'd really rather not check the compilation length of each n-qubit Clifford gate. However, I think the utility function described in (2) could be nice to have also.

sserita:
Closing as this is merged in develop with #443.

Issue #314, opened 2023-05-18
Custom compilation for 1-qubit Clifford RB


coreyostrove:
Add the ability to pick your own deterministic compiler when doing 1-qubit Clifford RB (could be done for 2-qubit too, but probably not worth the bother).

sserita:
Closing as this is merged in develop with #443

PR #443, opened 2024-05-22
RBDesign bugfixes and improvements

sserita:
Fixes several RB issues. I still plan to write a few tests for these, but it is at least ready to review from the RB codeowners to ensure that these are good solutions to the features.

### Deterministic Clifford compilation (#314)

The desire was to be able to do deterministic Clifford compilation when generating `CliffordRBDesign` objects. We actually already had a partial way to do this because the `CliffordCompilationRules` are more or less exactly this - a rule for how to build up a circuit from native gates that implement a Clifford. So my implementation is just to add a `exact_compilation_key` kwarg which uses the compilation rules to do this deterministic compilation when provided, and otherwise to do the previous behavior of synthesizing using the random algorithms.

An example:

```
import pygsti
from pygsti.processors import QubitProcessorSpec as QPS
from pygsti.processors import CliffordCompilationRules as CCR

n_qubits = 1
qubit_labels = ['Q'+str(i) for i in range(n_qubits)] 
gate_names = ['Gi', 'Gxpi2', 'Gxpi', 'Gxmpi2', 'Gypi2', 'Gypi', 'Gympi2', 
              'Gzpi2', 'Gzpi', 'Gzmpi2', 'Gcphase']
availability = {'Gcphase':[('Q'+str(i),'Q'+str((i+1) % n_qubits)) for i in range(n_qubits)]}
pspec = QPS(n_qubits, gate_names, availability=availability, qubit_labels=qubit_labels)
compilations = {'absolute': CCR.create_standard(pspec, 'absolute', ('paulis', '1Qcliffords'), verbosity=0),            
                'paulieq': CCR.create_standard(pspec, 'paulieq', ('1Qcliffords', 'allcnots'), verbosity=0)}
depths = [0, 2]
k = 10
qubits = qubit_labels
```

This will create the design as before, with random Clifford synthesis:
```
design = pygsti.protocols.CliffordRBDesign(pspec, compilations, depths, k, qubit_labels=qubits, 
                                           randomizeout=False, seed=20240522)
```

This will create a design using the `absolute` compilation of each Clifford instead:
```
design = pygsti.protocols.CliffordRBDesign(pspec, compilations, depths, k, qubit_labels=qubits, 
                                           randomizeout=False, seed=20240522, exact_compilation_key="absolute")
```

Or similarly, we could do a compilation up to Pauli equivalence with the `paulieq` compilation:

```
design = pygsti.protocols.CliffordRBDesign(pspec, compilations, depths, k, qubit_labels=qubits, 
                                           randomizeout=False, seed=20240522, exact_compilation_key="paulieq")
```

A non-standard compilation rule could also be done:
```
compilations["nonstd"] = pygsti.processors.CompilationRules({some dict of clifford names -> circuit templates})
design = pygsti.protocols.CliffordRBDesign(pspec, compilations, depths, k, qubit_labels=qubits, 
                                           randomizeout=False, seed=20240522, exact_compilation_key="nonstd")
```

### Native gate per Clifford statistics (#315)

The `CliffordRBDesign` now keeps track of how many native gates are in the first `length`+1 Cliffords, i.e. excluding the inversion Clifford, stores them in a `native_gate_count_lists` member variable, and provides convenience functions to compute the averages.

```
design = pygsti.protocols.CliffordRBDesign(pspec, compilations, [0, 2], 5, qubit_labels=qubits, 
                                           randomizeout=False, citerations=10)
design.native_gate_count_lists # A list of lists of the number of native gates for the compiled Cliffords (without inversion)
design.average_native_gates_per_clifford()  # Compute average over all circuits
```

As an aside, I also added a `num_gates` utility function to `Circuit` that was apparently missing. There was `num_nq_gates` and `num_multiq_gates` so you could have done `num_nq_gates(1) + num_multiq_gates`, but might as well provide it.

### Truncation bugfix for RBDesigns (#408)

The various `BenchmarkingDesign` classes often have lists that are intended to be paired with the `circuit_lists` member attribute. These paired attributes were not being truncated properly.

To fix this, the `BenchmarkingDesign` now has a `paired_with_circuit_attrs` member variable which lists the names of any attribute that is intended to correspond 1-to-1 with the `circuit_lists`. During truncation, these paired attributes are zipped up with the circuits during filtering, and then unzipped into the new truncated lists. Paired attributes are also serialized separately to JSON files, as we often want to look at them independently.

A currently exhaustive list of paired attributes (paired attributes propogate to derived classes):

- `BenchmarkingDesign`: `idealout_lists`
- `CliffordRBDesign`: `num_native_gate_lists` (due to the #315 fix above)
- `BinaryRBDesign`: `measurements`, `signs`

### DIFF
--- a/pygsti/algorithms/randomcircuit.py
+++ b/pygsti/algorithms/randomcircuit.py
@@ -2140,7 +2140,7 @@ def create_direct_rb_circuit(pspec, clifford_compilations, length, qubit_labels=
 
 def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_labels=None, randomizeout=False,
                                citerations=20, compilerargs=None, interleaved_circuit=None, seed=None,
-                               return_num_native_gates=False):
+                               return_num_native_gates=False, exact_compilation_key=None):
     """
     Generates a "Clifford randomized benchmarking" (CRB) circuit.
 
@@ -2166,9 +2166,10 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
         the gates in `pspec`, and will respect the connectivity encoded by `pspec`.
 
     clifford_compilations : dict
-        A dictionary with the potential keys `'absolute'` and `'paulieq'` and corresponding
+        A dictionary with at least the potential keys `'absolute'` and `'paulieq'` and corresponding
         :class:`CompilationRules` values.  These compilation rules specify how to compile the
-        "native" gates of `pspec` into Clifford gates.
+        "native" gates of `pspec` into Clifford gates. Additional :class:`CompilationRules` can be
+        provided, particularly for use with `exact_compilation_key`.
 
     length : int
         The "CRB length" of the circuit -- an integer >= 0 --  which is the number of Cliffords in the
@@ -2227,6 +2228,15 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
     
     return_num_native_gates: bool, optional
         Whether to return the number of native gates in the first `length`+1 compiled Cliffords
+    
+    exact_compilation_key: str, optional
+        The key into `clifford_compilations` to use for exact deterministic complation of Cliffords.
+        The underlying :class:`CompilationRules` object must provide compilations for all possible
+        n-qubit Cliffords that will be generated. This also requires the pspec is able to generate the
+        symplectic representations for all n-qubit Cliffords in :meth:`compute_clifford_symplectic_reps`.
+        This is currently generally intended for use out-of-the-box with 1-qubit Clifford RB;
+        however, larger number of qubits can be used so long as the user specifies the processor spec and
+        compilation rules properly.
 
     Returns
     -------
@@ -2253,6 +2263,11 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
     # The number of qubits the circuit is over.
     n = len(qubits)
 
+    if exact_compilation_key is not None:
+        # Precompute some of the symplectic reps if we are doing exact compilation
+        srep_cache = _symp.compute_internal_gate_symplectic_representations()
+        srep_cache.update(pspec.compute_clifford_symplectic_reps())
+
     rand_state = _np.random.RandomState(seed)  # OK if seed is None
 
     # Initialize the identity circuit rep.
@@ -2264,14 +2279,36 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
     # Sample length+1 uniformly random Cliffords (we want a circuit of length+2 Cliffords, in total), compile
     # them, and append them to the current circuit.
     num_native_gates = 0
-    for i in range(0, length + 1):
-
-        s, p = _symp.random_clifford(n, rand_state=rand_state)
-        circuit = _cmpl.compile_clifford(s, p, pspec,
-                                         clifford_compilations.get('absolute', None),
-                                         clifford_compilations.get('paulieq', None),
-                                         qubit_labels=qubit_labels, iterations=citerations, *compilerargs,
-                                         rand_state=rand_state)
+    for _ in range(0, length + 1):
+        if exact_compilation_key is not None:
+            # Deterministic compilation based on a provided clifford compilation
+            assert exact_compilation_key in clifford_compilations, \
+                    f"{exact_compilation_key} not provided in `clifford_compilations`"
+            
+            # Pick clifford
+            cidx = rand_state.randint(24**n)
+            lbl = _lbl.Label(f'C{cidx}', qubits)
+            
+            # Try to do deterministic compilation
+            try:
+                circuit = clifford_compilations[exact_compilation_key].retrieve_compilation_of(lbl)
+            except AssertionError:
+                raise ValueError(
+                    f"Failed to compile n-qubit Clifford 'C{cidx}'. Ensure this is provided in the " + \
+                    "compilation rules, or use a compilation algorithm to synthesize it by not " + \
+                    "specifying `exact_compilation_key`."
+                )
+        
+            # compute the symplectic rep of the chosen clifford
+            s, p = _symp.symplectic_rep_of_clifford_circuit(circuit, srep_cache)
+        else:
+            # Random compilation
+            s, p = _symp.random_clifford(n, rand_state=rand_state)
+            circuit = _cmpl.compile_clifford(s, p, pspec,
+                                            clifford_compilations.get('absolute', None),
+                                            clifford_compilations.get('paulieq', None),
+                                            qubit_labels=qubit_labels, iterations=citerations, *compilerargs,
+                                            rand_state=rand_state)
         num_native_gates += circuit.num_gates
 
         # Keeps track of the current composite Clifford
--- a/pygsti/algorithms/randomcircuit.py
+++ b/pygsti/algorithms/randomcircuit.py
@@ -2139,7 +2139,8 @@ def create_direct_rb_circuit(pspec, clifford_compilations, length, qubit_labels=
 
 
 def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_labels=None, randomizeout=False,
-                               citerations=20, compilerargs=None, interleaved_circuit=None, seed=None):
+                               citerations=20, compilerargs=None, interleaved_circuit=None, seed=None,
+                               return_num_native_gates=False):
     """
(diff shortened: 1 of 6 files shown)


pygsti/algorithms/randomcircuit.py
```diff
@@ -2255,14 +2278,39 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
...
+            s, p = _symp.symplectic_rep_of_clifford_circuit(circuit, srep_cache)
+        else:
+            # Random compilation
+            s, p = _symp.random_clifford(n, rand_state=rand_state)
+            circuit = _cmpl.compile_clifford(s, p, pspec,
+                                            clifford_compilations.get('absolute', None),
+                                            clifford_compilations.get('paulieq', None),
+                                            qubit_labels=qubit_labels, iterations=citerations, *compilerargs,
+                                            rand_state=rand_state)
+        num_native_gates += circuit.num_gates
```
rileyjmurray, line 2312: Can the contents of this loop be factored out into a new function? It looks semantically meaningful, and this ``create_clifford_rb_circuit`` function is quite long.

pygsti/protocols/vb.py
```diff
@@ -67,6 +68,25 @@ def map_qubit_labels(self, mapper):
...
+
+        Returns
+        -------
+        ByDepthDesign
+            The truncated experiment design.
+        """
+        ret = _copy.deepcopy(self) # Works for derived classes too
+        ret.depths = [self.depths[i] for i in list_indices_to_keep]
+        ret.circuit_lists = [self.circuit_lists[i] for i in list_indices_to_keep]
+        return ret
```
rileyjmurray, line 88: Might there be scenarios where the deepcopy isn't needed? If so, it might be good to have an optional function argument ``deepcopy=False`` that controls whether the copy is deep or shallow.

rileyjmurray (commented):
Mostly looks good. I made one suggestion to factor out some logic into its own function, and left one comment about a deepcopy.

pygsti/algorithms/randomcircuit.py
coreyostrove: +1

coreyostrove (commented):
This all looks good to me. Once we get the seal of approval from one of the RB codeowners I think this is good to go.

pygsti/algorithms/randomcircuit.py
sserita: Good point, will refactor that function out!

pygsti/protocols/vb.py
sserita: There could be cases where the deepcopy is not desired - for example, if we want to create a sort of "nested" edesign and want to point to the original circuit lists.
However, I'm inclined to keep this deepcopy as-is for now because it's actually indicative of a hack to make these functions work for derived classes. I think we could more cleanly do this with Generics/TypeVars, which I've been learning more about recently, but I want to keep this the same as the pattern used elsewhere in the code so I can clean it up all at once later and don't feel like plumbing a temporary kwarg in through all of this now.

pygsti/algorithms/randomcircuit.py
```diff
@@ -2255,14 +2278,39 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
...
-                                         rand_state=rand_state)
+    num_native_gates = 0
+    for _ in range(0, length + 1):
+        if exact_compilation_key is not None:
+            # Deterministic compilation based on a provided clifford compilation
+            assert exact_compilation_key in clifford_compilations, \
+                    f"{exact_compilation_key} not provided in `clifford_compilations`"
+            
+            # Pick clifford
+            cidx = rand_state.randint(24**n)
```
tjproct, line 2289: This is incorrect because there are not 24**n Cliffords. You can use tools.sympletic.compute_num_cliffords(n) to get the number of Cliffords

pygsti/algorithms/randomcircuit.py
```diff
@@ -2255,14 +2278,39 @@ def create_clifford_rb_circuit(pspec, clifford_compilations, length, qubit_label
...
+                circuit = clifford_compilations[exact_compilation_key].retrieve_compilation_of(lbl)
+            except AssertionError:
+                raise ValueError(
+                    f"Failed to compile n-qubit Clifford 'C{cidx}'. Ensure this is provided in the " + \
+                    "compilation rules, or use a compilation algorithm to synthesize it by not " + \
+                    "specifying `exact_compilation_key`."
+                )
+        
+            # compute the symplectic rep of the chosen clifford
+            s, p = _symp.symplectic_rep_of_clifford_circuit(circuit, srep_cache)
```
tjproct, line 2303: Note for if we ever want this to run quickly (we probably don't care) -- This is slightly absurdly way to do this, as given the index of the Clifford we should be able to easily compute s and p directly. However, there's only code right now to easily compute s (compute_symplectic_matrix(i, n)) and not p so there's no to-do I'm suggesting here.

pygsti/circuits/circuit.py
```diff
@@ -3405,6 +3405,30 @@ def two_q_gate_count(self):
...
+                    return sum([cnt(sublbl) for sublbl in lbl.components])
+        else:
+            def cnt(obj):  # obj is either a simple label or a list
+                if isinstance(obj, _Label):  # all Labels are simple labels
+                    return 1 if (obj.sslbls is not None) else 0
+                else:
+                    return sum([cnt(sub) for sub in obj])
+
+        return sum([cnt(layer_lbl) for layer_lbl in self._labels])
+
```
tjproct, line 3431: If I understand this correctly, this is computing the number of explicitly included (i.e., not implicit idle) gates in the circuit? I.e., this is different to self.depth * self.width right? If so, (1) this seems like a useful function to have, and (2) it's debatable whether this is what should be computed by the RB code. I suspect that what should be returned is this *and* the circuit size (width * depth)? Some users might also be interested in the 2-qubit gate count, so that could also be computed and all wrapped up in a dictionary?

pygsti/protocols/rb.py
```diff
@@ -214,40 +222,114 @@ def __init__(self, pspec, clifford_compilations, depths, circuits_per_depth, qub
...
+        float
+            The average number of native gates per Clifford
+        """
+        if self.num_native_gate_lists is None:
+            raise ValueError("Number of native gates not available, cannot compute average gates per Clifford")
+        num_native_gates = 0
+        num_clifford_gates = 0
+        for list_idx in range(len(self.depths)):
+            num_native_gates = sum(self.num_native_gate_lists[list_idx])
+            num_clifford_gates = len(self.num_native_gate_lists[list_idx]) * (self.depths[list_idx] + 1)
```
tjproct, line 330: Should these two lines be a += not a =?

tjproct (changes requested):
There is one error (the size of the Clifford group) that needs to be fixed. The bit about adding different statistics about the native gates in each Clifford is optional, and an alternative to making the suggested additions is to edit the doc-string to make it very clear what the number of native gates per Clifford means here.

pygsti/protocols/rb.py
sserita: Nice catch, that's a copy-paste error.

pygsti/algorithms/randomcircuit.py
sserita: Agreed that this is not the best way to do this. What I actually wanted to do (and might still be useful in the future) was reuse the generate s/p code and then be able to look up the rule for compiling that Clifford from the CompilationRules. The CliffordCompRule object has some support for symplectic reps already, although not this reverse lookup (and I didn't want to restrict this to only taking CliffordCompRules).

So either way, I think we can revisit this when we update the symplectic tools to have either functionality (and I've left a TODO comment in the code noting that).

pygsti/algorithms/randomcircuit.py
sserita: Nice catch!

pygsti/circuits/circuit.py
sserita: You are correct, this is explicitly included gates and not depth*width (which already exists as `circuit.size`). FYI, the 2-qubit gate count also exists in that you can do `circuit.num_nq_gates(2)`.

Computing all three and wrapping it in a dict is a pretty easy and self-documenting solution, so I'll implement that. :)

sserita: This is now what the output of the `average_native_gates_per_clifford()` looks like:

Screenshot 2024-05-30 at 1 04 16 PM

tjproct (approved):
Looks great! :)

PR #445, opened 2024-05-31
Faster Circuit Primitives

coreyostrove:
I have always been slightly perturbed by the amount of time that things such as COPA layout creation take when performing GST analysis. There are a number of reasons for this, but one family of reasons is that layout generation relies heavily on a number of Circuit primitives that are not as performant as they can be. This PR is a step toward rectifying this through re-implementation of a number of methods identified as bottlenecks through profiling in the context of actual end-to-end GST analysis. 

In terms of the top-level numbers, for 1-qubit GST fitting using the 'full TP' parameterization with an experiment design containing 1120 circuits (max depth 128), the changes on this branch collectively reduce the time required for layout creation by ~50%, and reduce the overall runtime of the end-to-end protocol by ~15%. Your mileage will of course vary, however, and I would anticipate higher relative savings for larger/deeper experiment designs. I have not done profiling for the two-qubit case yet, so I won't posit any guesses on the impact there (I would guess it to be less relatively speaking, but can't easily ballpark this without actually testing).

### Summary of Changes:
Here is a high-level summary of the changes. I will also be leaving some comments on particular bits of code to explain any more detailed reasoning for some changes (when in doubt the answer is likely that I tried a few different things and picked the fastest way to do something).

**Note on synthetic benchmarks:** For the quoted figures below these were mostly done using a randomly generated set of test circuits generated according to the following code snippet:

```
import pygsti
from pygsti.circuits.circuitconstruction import list_random_circuits_onelen
from pygsti.modelpacks import smq1Q_XYI
import numpy as np
from numpy.random import default_rng
from itertools import product
lens = [1,2,4,8,16,32,64]
op_labels = tuple(smq1Q_XYI.target_model().operations.keys())
random_circuits = []
for length in lens:
    random_circuits.extend(list_random_circuits_onelen(op_labels, length, 100, seed = 1234))
random_circuits_2 = []
for length in lens:
    random_circuits_2.extend(list_random_circuits_onelen(op_labels, length, 100, seed = 12345))
rng = default_rng(1234)
rng.shuffle(random_circuits)
rng = default_rng(12345)
rng.shuffle(random_circuits_2)
random_pairs = list(product(random_circuits,repeat=2))
```

For equality testing this was run against `random_pairs`, for inclusion testing against `random_circuits` and `random_circuits_2`, and for other tests against `random_circuits` unless stated otherwise.

1. Circuit equality testing: The primary change was adding a length check to allow quicker short circuiting. A somewhat faster codepath was also added for comparisons between static circuits, avoiding recomputation of a circuit's tuple representation. In synthetic testing this gave a ~3X speed up.
2. Inclusion testing:  Nothing specific was added here, but the change to circuit equality testing also sped this up by a slightly larger factor of ~3.4.
3. Hashing: Hashes are now precomputed and cached at the point at which a circuit is made static (this is an accepted practice in python for immutable objects, e.g. strings cache their hashes). Related to this change, at staticization time we precompute the tuple representation of the circuit and store this as well. This is the thing which actually gets hashed. Previously the `tup` property was called whenever this was needed and this property call regenenerated the tuple representation from scratch each time, which is expensive. In synthetic tests this resulted in a ~4.5X speedup, but in the context of the GST profiling described above this was closer to 10X (the computational savings should scale with circuit depth, so this isn't too shocking of a difference). As part of this change I have added static assertions to a few methods previously missing them for safety (we don't want to inadvertently invalidate the cached hash).
4. Copying: Previously copying a circuit went through the standard `__init__` pathway to create a copied circuit instance. With the new implementation we utilize a much faster specialized codepath inspired by the previously existing `_fastinit` one. But with a few more microptimizations to take advantage of some of the additional promises we can rely on in the context of copying (which we can't as easily in `_fastinit`). Speedups here range from 3-15X depending on the input and output circuit types. With the low end corresponding to switching static to editable (or vice versa), and the high end keeping the input type unchanged.

This is the end of the changes where I profiled these using synthetic tests. For the rest I did this in the context of GST optimization using cProfile/snakeviz, but didn't carefully record before/after relative changes.

5. Microoptimization galore: There are a bunch of microoptimizations of the sort that usually don't matter except for when  code is particularly hot, and the circuit code is. I won't give an exhaustive list, but this includes things like:
    - Attribute access is faster than method/property access/resolution (by 2X or more in some cases). So many trivial (and some non-trivial) property calls have had the corresponding attributes inlined.
    - Inlining functions to reduce function call overhead.
    - List comprehensions are faster than generator expressions. So, e.g. `tuple([i for i in iterable])` is measurably faster than `tuple(i for i in iterable)` in practice.
    - List comprehensions are faster than the 'build via appends in for loop' approach. Often by a notable amount.
    - Using falsy/truthiness directly is faster than the `if len(obj)>0` idiom if all you're doing is checking if an iterable is empty.

6. Rework SeparatePOVMCircuit internals: The `full_effect_labels` property was getting called a lot, and each time it was it was regenerating the result from scratch (with some non-trivial preprocessing). I've updated the internals to cache this, and also deal with updating the cached value when necessary. This gave a measurable speedup.
7. New `sandwich` method for circuits. This takes as input two tuples of Labels which are prepended and appended to a circuit. This is around twice as fast as doing this with a pair of calls to `__add__` (since we only need 1 call to `_fastinit`). This is not an uncommon task (I added as part of some changes to the circuit completion code in OpModel), so if you spot any places where this is being done it'd be worth switching (maybe @rileyjmurray can write a fancy regex to find these instances).

Additional changes (either outside of the Circuit class, or otherwise not directly performance focused):

8. Bulk versions of `complete_circuit` and `split_circuit`: There are now versions of the `complete_circuit` and `split_circuit` methods for the OpModel class, which respectively add prep and povm labels to a circuit or split them off, which operate on a complete list of circuits all at once. This allows for avoiding some model specific lookups (for values which don't change circuit-to-circuit) many times and is faster overall.
9. `expand_subcircuits_inplace`: The implementation of `expand_subcircuits_inplace` has been reworked so that the behavior is now to recursively expand CircuitLabel objects. Previously the expansion routine only took a single pass through the circuit, meaning that if the circuit contained any nested CircuitLabels (which are perfectly happy with being nested, an easy way to get something like this is to call `to_label` on a circuit containing CircuitLabels) we would only end up expanding out the top level. I noticed there were a few places in the layout creation/forward simulation code that implicitly assumed only a single-level of CircuitLabels would be present, which would likely have eventually led to some really annoying bugs to track down. This change covers that edge case, and should only be slightly more expensive than the previous routine in the typical case where there is only a single-level of CircuitLabels.
10. I've added a deprecation warning to the `simulate` method. If there is strong opposition to removing this functionality we can discuss this, of course. My reasoning here is that given the rest of pygsti's structure this functionality really isn't something we should be asking a Circuit to do, as the preferred pattern nowadays is delegate this job to the Model/ForwardSimulator, rather than asking the circuit itself. I don't know how widespread the use of this is (my guess is that this is largely vestigial), but I'd recommend slating this for removal in 0.9.14. Related note, I don't know if we have a running list of things we've deprecated and their expiry date, but if not we should assemble one.

Finally, there are a number of miscellaneous documentation fixes, bugfixes, and updates to unit tests.

P.S. There are some more layout creation performance enhancements inbound on a separate branch, but I've opted to keep these changes together in a separate PR to keep things thematic and make it easier to review without context switching.


pygsti/circuits/circuit.py
```diff
@@ -965,21 +974,39 @@ def __pow__(self, x):  # same as __mul__()
...
     def __eq__(self, x):
-        if x is None: return False
+        
         if isinstance(x, Circuit):
-            return self.tup.__eq__(x.tup)
+            if len(self) != len(x):
+                return False
+            else:
+                if self._static and x._static:
+                    return self._hashable_tup == x._hashable_tup
```
sserita, line 983: Can we not just check `_hash` directly in this case, since we store those?

pygsti/circuits/circuit.py
```diff
@@ -1535,13 +1602,12 @@ def insert_idling_lines_inplace(self, insert_before, line_labels):
...
-        # Actually, this is OK even for static circuits because it won't affect the hashed value (labels only)
+        assert(not self._static),"Cannot edit a read-only circuit!"
         if insert_before is None:
-            i = len(self.line_labels)
+            i = len(self._line_labels)
         else:
-            i = self.line_labels.index(insert_before)
-        self._line_labels = self.line_labels[0:i] + tuple(line_labels) + self.line_labels[i:]
+            i = self._line_labels.index(insert_before)
+        self.line_labels = self._line_labels[0:i] + tuple(line_labels) + self._line_labels[i:]
```
sserita, line 1610: Why did the LHS get moved to the property? Can we not just assign to `_line_labels` still?

pygsti/circuits/circuit.py
```diff
@@ -1789,7 +1857,7 @@ def delete_lines(self, lines, delete_straddlers=False):
                     raise ValueError(("Cannot remove a block that is straddled by "
                                       "%s when `delete_straddlers` == False!") % _Label(l))
             self._labels[i] = new_layer
-        self._line_labels = tuple([x for x in self.line_labels if x not in lines])
+        self.line_labels = tuple([x for x in self._line_labels if x not in lines])
```
sserita, line 1860: Another instance of LHS changing to property version. Keep as `_line_labels`?

pygsti/circuits/circuit.py
```diff
@@ -2316,11 +2399,11 @@ def tensor_circuit_inplace(self, circuit, line_order=None):
 
             new_line_labels = line_order
         else:
-            new_line_labels = self.line_labels + circuit.line_labels
+            new_line_labels = self._line_labels + circuit._line_labels
 
         #Add circuit's labels into this circuit
         self.insert_labels_as_lines_inplace(circuit._labels, line_labels=circuit.line_labels)
-        self._line_labels = new_line_labels  # essentially just reorders labels if needed
+        self.line_labels = new_line_labels  # essentially just reorders labels if needed
```
sserita, line 2406: Another instance of LHS not being `_line_labels` anymore

pygsti/circuits/circuit.py
```diff
@@ -4681,12 +4702,35 @@ class SeparatePOVMCircuit(object):
...
+    
+    @property 
+    def effect_labels(self):
+        return self._effect_labels
+    
+    @property 
+    def povm_label(self):
+        return self._povm_label
+
+    @effect_labels.setter
```
sserita, line 4721: Should setters have static checks?

sserita (changes requested):
Thanks for the great work as usual @coreyostrove!

It seems like the major parts of this PR are in the new initialization via `_copy_init`, the `tup` property, the `copy()` and `expand_subcircuits_inplace()` functions, as well as some of the dunder functions (`__eq__` most notably). These look pretty good to me reading through them, but I think it'll be good to get this into develop and pay attention to any issues as we get the rest of 0.9.13 stuff in.

I've also mentioned some other smaller things that will hopefully make it more readable or changes I wasn't sure were necessary.

pygsti/baseobjs/label.py
```diff
@@ -201,6 +198,9 @@ class LabelTup(Label, tuple):
     acted upon by the object this label refers to.
     """
 
+    #flag used in certain Circuit subroutines
+    _is_simple= True
```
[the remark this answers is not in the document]
coreyostrove: I didn't remove the old methods. I could do so, but perhaps I should just refactor this in the rest of the code as well (and make this a public class attribute to signal as much). Taking a quick search through the source code there look to be maybe another half dozen calls to these methods, so it should be pretty straightforward.

pygsti/circuits/circuit.py
```diff
@@ -573,51 +493,65 @@ def __init__(self, layer_labels=(), line_labels='auto', num_lines=None, editable
...
+        else:
+            self._str = None # can be None (lazy generation)
+            #only meant to be used in settings where we're explicitly checking for self._static.
+        #self._reps = reps # repetitions: default=1, which remains unless we initialize from a CircuitLabel...
+        self._name = name  # can be None
+        #self._times = None  # for FUTURE expansion
+        self.auxinfo = {}  # for FUTURE expansion / user metadata
+
+    #specialized codepath for copying
+    def _copy_init(self, labels, line_labels, editable, name='', stringrep=None, occurrence=None,
```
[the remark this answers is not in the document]
coreyostrove: I just gave refactoring `_bare_init` a shot, and there wasn't a particularly clean way to re-use `_copy_init` without needing set a number of the attributes twice (some attributes need to be set before the tuple representation and hash can be computed...). So while the code duplication is unfortunate I think it is the lesser of two evils. I'll add a note to the two methods warning the next person to work in this part of the code to check if they need to be resynced when updated.

pygsti/circuits/circuit.py
```diff
@@ -715,35 +651,81 @@ def tup(self):
         -------
         tuple
         """
+        comp_lbl_flag = ('__CMPLBL__',) if self._compilable_layer_indices_tup else ()
+        if self._static:
```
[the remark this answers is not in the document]
coreyostrove: Good catch, I've reworked this according to your suggestion. 👍

pygsti/circuits/circuit.py
```diff
@@ -844,7 +822,7 @@ def __contains__(self, x):
     def __radd__(self, x):
         if not isinstance(x, Circuit):
             assert(all([isinstance(l, _Label) for l in x])), "Only Circuits and Label-tuples can be added to Circuits!"
-            return Circuit._fastinit(x + self.layertup, self.line_labels, editable=False)
+            return Circuit._fastinit(x + self.layertup, self._line_labels, editable=False)
```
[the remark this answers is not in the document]
coreyostrove: Yup, in my profiling I found that property calls introduce about a 2-3X penalty (presumably method call overhead) over attribute access. E.g.

```
%%timeit
for ckt in random_circuits:
    ckt.line_labels

62.1 µs ± 2.72 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

```
%%timeit
for ckt in random_circuits:
    ckt._line_labels

23.7 µs ± 195 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

I was surprised at first, but aggregated over many calls the overall savings was meaningful.

pygsti/models/model.py
```diff
@@ -1210,8 +1210,126 @@ def split_circuit(self, circuit, erroron=('prep', 'povm'), split_prep=True, spli
             povm_lbl = None
 
         return prep_lbl, circuit, povm_lbl
+    
+    def split_circuits(self, circuits, erroron=('prep', 'povm'), split_prep=True, split_povm=True):
```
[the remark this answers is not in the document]
coreyostrove: We have an implementation of this already, I didn't make any modifications to it though which is why it isn't showing up in the diff. If I make the refactors of the single circuit versions you suggested then they will be appearing shortly.

pygsti/circuits/circuit.py
coreyostrove: SeparatePOVMCircuit doesn't presently have a static flag. I'd have to take a closer look at how these are used in practice elsewhere in the code (i.e. whether other parts of the code are interacting with these attributes. They are of course leading underscored, so maybe not? My main motivation for adding the setters was that the value of _full_effect_labels is derived from the values of _povm_label and _effect_labels, so I wanted to make it slightly more difficult to have a situation where they could get out of sync now that _full_effect_labels is being cached.

pygsti/circuits/circuit.py
coreyostrove: For reasons that I admittedly didn't fully understand at the time I had found that changing this and the similar one for tensoring broke some stuff in testing (i.e. in the unit tests). But, changing this again it looks like it isn't broken anymore. So I'm guessing I must have fixed whatever weird interaction was causing this at some point. 🤷‍♂️

pygsti/circuits/circuit.py
coreyostrove: Great point! That is indeed faster.

pygsti/circuits/circuit.py
```diff
@@ -927,6 +909,33 @@ def __add__(self, x):
 
         return Circuit._fastinit(self.layertup + x.layertup, new_line_labels, editable=False, name='',
                                  stringrep=s, occurrence=None)
+    
+
+    def sandwich(self, x, y):
```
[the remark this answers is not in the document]
coreyostrove: This is used in the implementation of the new `complete_circuits` method in `OpModel`. I got to a point in optimizing that where the runtime was dominated by the two circuit additions with prepend and postpend the prep and POVM labels (~80% of runtime was just that). It then occurred to me that since almost all of the time for `__add__` is the `_fastinit` call that we could do the sandwiching operation in nearly the same time as a single addition. Even with that change the `sandwich` calls still account for ~50% of that method's runtime.

pygsti/models/model.py
```diff
@@ -1222,40 +1340,136 @@ def complete_circuit(self, circuit):
...
+                raise ValueError(f"Missing POVM in {circuit.str} and there's no default!")
+                
+        if prep_lbl_to_prepend: 
+            circuit = (prep_lbl_to_prepend,) + circuit
+        if povm_lbl_to_append: 
+            circuit = circuit + (povm_lbl_to_append,)
 
         return circuit
 
+    def complete_circuits(self, circuits, prep_lbl_to_prepend=None, povm_lbl_to_append=None, return_split = False):
```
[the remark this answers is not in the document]
coreyostrove: We want to avoid calling the single case many times (the various model lookups for default POVMS and preps and whatnot genuinely takes a long time), but I think refactoring the single-use case to use the multi-circuit implementation should be fine.

pygsti/models/model.py
```diff
@@ -1210,8 +1210,126 @@ def split_circuit(self, circuit, erroron=('prep', 'povm'), split_prep=True, spli
...
+        primitive_povm_labels_set = set(primitive_povm_labels_tup)
+
+        #precompute unique default povm labels.
+        unique_sslbls = set([ckt._line_labels for ckt in circuits])
+        default_povm_labels = {sslbls:self._default_primitive_povm_layer_lbl(sslbls) for sslbls in unique_sslbls}
+        
+        if split_prep and split_povm: #can avoid some duplicated effort in this case.
+            split_circuits = []
+            for ckt in circuits:
+                if len(ckt) > 0 and ckt[0] in primitive_prep_labels_set:
```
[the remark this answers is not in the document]
coreyostrove: Stylistically I prefer avoiding nested/local function definitions where possible and I am not sure it would improve readability over using the sets directly. Your comment did make it occur to me that it would be (very slightly) more efficient to only construct the sets we need given the value of `split_prep` and `split_povm`, and doing so does improve code locality a bit (at the expense of a bit more repetition, but sometimes these two desiderata are at odds).

coreyostrove:
Thanks for the feedback, @sserita! I have responded to a number of your questions/comments above. I've also just pushed a commit incorporating some of your suggested changes.

pygsti/baseobjs/label.py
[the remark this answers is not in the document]
coreyostrove: I have just pushed a refactor of this to switch to using the class attribute throughout the code, and removing the old methods.

pygsti/baseobjs/label.py
```diff
@@ -1743,6 +1639,10 @@ class LabelTupWithArgs(Label, tuple):
     refers to.  This label type also supports having arguments and a time value.
     """
 
+    #flag used in certain Circuit subroutines
+    #Whether this is a "simple" (opaque w/a true name, from a circuit perspective) label or not.
+    is_simple= True
```
rileyjmurray, line 1644: Same comments as before

pygsti/baseobjs/label.py
```diff
@@ -2023,6 +1907,10 @@ class LabelTupTupWithArgs(Label, tuple):
     This label type also supports having arguments and a time value.
     """
 
+    #flag used in certain Circuit subroutines
+    #Whether this is a "simple" (opaque w/a true name, from a circuit perspective) label or not.
+    is_simple= False
```
rileyjmurray, line 1912: Same comment as before

pygsti/circuits/circuit.py
```diff
@@ -704,8 +636,9 @@ def layertup(self):
         if self._static:
             return self._labels
         else:
-            return tuple([to_label(layer_lbl) for layer_lbl in self._labels])
-
+            #return tuple([to_label(layer_lbl) for layer_lbl in self._labels])
+            return tuple([layer_lbl if isinstance(layer_lbl, _Label)
```
rileyjmurray, line 640: Use a generator instead of a list comprehension in ``tuple(...)``.

pygsti/circuits/circuit.py
```diff
@@ -867,24 +827,28 @@ def __add__(self, x):
 
         if not isinstance(x, Circuit):
             assert(all([isinstance(l, _Label) for l in x])), "Only Circuits and Label-tuples can be added to Circuits!"
-            return Circuit._fastinit(self.layertup + x, self.line_labels, editable=False)
+            new_line_labels = set(sum([l.sslbls for l in x if l.sslbls is not None],
```
rileyjmurray, line 830: Use a generator instead of a list comprehension in ``sum(...)``.

pygsti/circuits/circuit.py
```diff
@@ -927,6 +891,33 @@ def __add__(self, x):
...
+            Same as `x`, but appended instead.
+
+        Returns
+        -------
+        Circuit
+        """
+
+        assert(isinstance(x, tuple) and isinstance(y, tuple)), 'Only tuples of labels are currently supported by `sandwich` method.'
+        combined_sandwich_labels = x + y
+        assert(all([isinstance(l, _Label) for l in combined_sandwich_labels])), "Only Circuits and Label-tuples can be added to Circuits!"
```
rileyjmurray, line 916: Use a generator instead of a list comprehension in ``all(...)``.

pygsti/circuits/circuit.py
```diff
@@ -927,6 +891,33 @@ def __add__(self, x):
...
+        new_line_labels = set(sum([l.sslbls for l in combined_sandwich_labels if l.sslbls is not None],
```
rileyjmurray, line 917: Use a generator instead of a list comprehension in ``sum(...)``.

pygsti/circuits/circuit.py
```diff
@@ -927,6 +891,33 @@ def __add__(self, x):
...
+        -------
+        Circuit
+        """
+
+        assert(isinstance(x, tuple) and isinstance(y, tuple)), 'Only tuples of labels are currently supported by `sandwich` method.'
+        combined_sandwich_labels = x + y
+        assert(all([isinstance(l, _Label) for l in combined_sandwich_labels])), "Only Circuits and Label-tuples can be added to Circuits!"
+        new_line_labels = set(sum([l.sslbls for l in combined_sandwich_labels if l.sslbls is not None], 
+                                  self._line_labels)) #trick for concatenating multiple tuples
+        new_line_labels = sorted(list(new_line_labels))
```
rileyjmurray, line 919: There's no need to convert to a list before passing into ``sorted``. Examples:
```python
>>> sorted({'c','a','b'})
['a', 'b', 'c']
>>> sorted(('c', 'b', 'a'))
['a', 'b', 'c']
```

pygsti/circuits/circuit.py
```diff
@@ -1008,9 +1017,43 @@ def copy(self, editable="auto"):
...
+            if self._static:
+                #if presently static leverage precomputed hashable_tup and hash. 
+                #These values are only used by _copy_init if the circuit being 
+                #created is static, and are ignored otherwise.
+                return ret._copy_init(self._labels, self._line_labels, editable, 
+                                      self._name, self._str, self._occurrence_id, 
+                                      self._compilable_layer_indices_tup, 
+                                      self._hashable_tup, self._hash)
+            else:
+                static_labels = tuple([layer_lbl if isinstance(layer_lbl, _Label) else _Label(layer_lbl)
```
rileyjmurray, line 1050: Use a generator rather than a list comprehension in ``tuple(...)``.

pygsti/circuits/circuit.py
```diff
@@ -1752,12 +1802,12 @@ def delete_layers(self, layers=None):
             del self._labels[i]
 
         #Shift compilable layer indices as needed
-        if len(self._compilable_layer_indices_tup) > 0:  # begins with __CMPLBL__
+        if self._compilable_layer_indices_tup:
             deleted_indices = set(layers)
-            new_inds = list(filter(lambda x: x not in deleted_indices, self._compilable_layer_indices_tup[1:]))
+            new_inds = list(filter(lambda x: x not in deleted_indices, self._compilable_layer_indices_tup))
```
rileyjmurray, line 1807: Is this ``list(filter(...))`` pattern really needed? I imagine it would suffice to do
```python
new_inds = [x for x in self._compilable_layer_indices_tup if x not in deleted_indices]
```

pygsti/circuits/circuit.py
```diff
@@ -1789,7 +1839,7 @@ def delete_lines(self, lines, delete_straddlers=False):
                     raise ValueError(("Cannot remove a block that is straddled by "
                                       "%s when `delete_straddlers` == False!") % _Label(l))
             self._labels[i] = new_layer
-        self._line_labels = tuple([x for x in self.line_labels if x not in lines])
+        self._line_labels = tuple([x for x in self._line_labels if x not in lines])
```
rileyjmurray, line 1842: Use a generator instead of a list comprehension in ``tuple(...)``.

pygsti/circuits/circuit.py
```diff
@@ -3569,11 +3569,11 @@ def abbrev(lbl, k):  # assumes a simple label w/ name & qubits
                              for i in range(0, self.num_lines)])
                         for j in range(0, self.num_layers)]
 
-        max_linelabellen = max([len(str(llabel)) for llabel in self.line_labels])
+        max_linelabellen = max([len(str(llabel)) for llabel in self._line_labels])
```
rileyjmurray, line 3572: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4015,19 +4017,19 @@ def convert_to_quil(self,
             # To tell us whether we have found a standard qubit labelling type.
             standardtype = False
             # Must first check they are strings, because cannot query q[0] for int q.
-            if all([isinstance(q, str) for q in self.line_labels]):
-                if all([q[0] == 'Q' for q in self.line_labels]):
+            if all([isinstance(q, str) for q in self._line_labels]):
```
rileyjmurray, line 4020: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4015,19 +4017,19 @@ def convert_to_quil(self,
...
+                if all([q[0] == 'Q' for q in self._line_labels]):
```
rileyjmurray, line 4021: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4015,19 +4017,19 @@ def convert_to_quil(self,
...
                     standardtype = True
-                    qubit_conversion = {llabel: int(llabel[1:]) for llabel in self.line_labels}
-            if all([isinstance(q, int) for q in self.line_labels]):
-                qubit_conversion = {q: q for q in self.line_labels}
+                    qubit_conversion = {llabel: int(llabel[1:]) for llabel in self._line_labels}
+            if all([isinstance(q, int) for q in self._line_labels]):
```
rileyjmurray, line 4024: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4192,22 +4194,19 @@ def convert_to_openqasm(self, num_qubits=None,
             # To tell us whether we have found a standard qubit labelling type.
             standardtype = False
             # Must first check they are strings, because cannot query q[0] for int q.
-            if all([isinstance(q, str) for q in self.line_labels]):
-                if all([q[0] == 'Q' for q in self.line_labels]):
+            if all([isinstance(q, str) for q in self._line_labels]):
```
rileyjmurray, line 4197: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4192,22 +4194,19 @@ def convert_to_openqasm(self, num_qubits=None,
...
+                if all([q[0] == 'Q' for q in self._line_labels]):
```
rileyjmurray, line 4198: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4192,22 +4194,19 @@ def convert_to_openqasm(self, num_qubits=None,
...
                     standardtype = True
-                    qubit_conversion = {llabel: int(llabel[1:]) for llabel in self.line_labels}
-            if all([isinstance(q, int) for q in self.line_labels]):
-                qubit_conversion = {q: q for q in self.line_labels}
+                    qubit_conversion = {llabel: int(llabel[1:]) for llabel in self._line_labels}
+            if all([isinstance(q, int) for q in self._line_labels]):
```
rileyjmurray, line 4201: prefer generator to list comprehension

pygsti/circuits/circuit.py
```diff
@@ -4391,7 +4391,10 @@ def done_editing(self):
         """
         if not self._static:
             self._static = True
-            self._labels = tuple([_Label(layer_lbl) for layer_lbl in self._labels])
+            self._labels = tuple([layer_lbl if isinstance(layer_lbl, _Label) 
+                                  else _Label(layer_lbl) for layer_lbl in self._labels])
```
rileyjmurray, line 4395: Prefer generator to list comprehension. Example:
```python
>>> tuple(a if a > 0 else 0 for a in [1,2,-1,3])
(1, 2, 0, 3)
```

pygsti/circuits/circuit.py
```diff
@@ -4681,12 +4684,35 @@ class SeparatePOVMCircuit(object):
     """
     def __init__(self, circuit_without_povm, povm_label, effect_labels):
         self.circuit_without_povm = circuit_without_povm
-        self.povm_label = povm_label
-        self.effect_labels = effect_labels
+        self._povm_label = povm_label
+        self._effect_labels = effect_labels
+        self._full_effect_labels = tuple([(self.povm_label + "_" + el) for el in self._effect_labels])
```
rileyjmurray, line 4689: prefer generator to list comprehension

pygsti/models/explicitmodel.py
```diff
@@ -1521,11 +1521,11 @@ def extract_unitary(Umx, U_sslbls, extracted_sslbls):
             return U_extracted
 
         def add_availability(opkey, op):
-            if opkey == _Label(()) or opkey.is_simple():
+            if opkey == _Label(()) or opkey.is_simple:
```
rileyjmurray, line 1524: Consider swapping the order of these two boolean checks. The ``is_simple`` check will be faster and might be true more often anyway.

test/test_packages/cmp_chk_files/idt_nQsequenceCache.pkl
rileyjmurray, line 1: Did you intend for this to be part of the PR?

coreyostrove: Yes, this pickle is (for better or worse, see #446 ) used as part of the unit tests for idle tomography. The pickle contained serialized Circuit objects and broke due to them missing the new attributes added in this PR). Eventually we'll want to replace all such pickles with our native serialization (which is less prone to breaking), but that is outside of the current scope.

pygsti/models/explicitmodel.py
coreyostrove: Great observation!

pygsti/circuits/circuit.py
coreyostrove: I've posted a reply in the corresponding issue thread #450.

rileyjmurray:
@coreyostrove I've gone through all my PR comments and (tried to) close the ones that were just about generators vs list comprehensions. There are still some meaningful comments left, so here's direct links to them:

* https://github.com/sandialabs/pyGSTi/pull/445#discussion_r1627643676 (there are many copies of this one)
* https://github.com/sandialabs/pyGSTi/pull/445#discussion_r1627673976
* https://github.com/sandialabs/pyGSTi/pull/445#discussion_r1627678229

pygsti/baseobjs/label.py
```diff
@@ -201,6 +198,10 @@ class LabelTup(Label, tuple):
     acted upon by the object this label refers to.
     """
 
+    #flag used in certain Circuit subroutines
+    #Whether this is a "simple" (opaque w/a true name, from a circuit perspective) label or not.
+    is_simple= True
```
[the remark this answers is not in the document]
coreyostrove: I did consider making this an instance attribute instead of a class attribute (for presumably the same safety considerations that you have). I decided against that for two reasons. 1. the way this is set/used it really is conceptually a property of the class and not of any given instance, so semantically this makes more sense as a class attribute. 2. It added a small (but measurable) overhead to `Label` creation, and we do a *lot* of `Label` creation.

Thoughts on how best to signify this is an attribute that is meant to be read but not set? When I originally refactored this I had it 'private' with a leading underscore, but wasn't sure that signified what I wanted.

Your new thread on slots, #449 is actually quite timely here, as this is something I considered doing for the Label class (in part because it seemed it might be a way to make Label instantiation a bit faster). I ended up opting not to do so not because I thought it wasn't a good idea, but simply because I have never used slots before and was too lazy to learn how to do so properly. It sounds like it could be a real net win though in the right circumstances (and I think Labels are one one of them).

pygsti/circuits/circuit.py
coreyostrove: Fixed with the latest commit, thanks!

pygsti/circuits/circuit.py
coreyostrove: Good point, much more readable (one doesn't need to remember what filter does). I've updated this in the latest commit and refactored the remaining instances of filter in this code as well.

coreyostrove:
Thanks for the feedback, @sserita  and @rileyjmurray. I have addressed a number of your comments, so I have gone through and marked those as such to make it easier to see which conversation threads still need following up on. 

Right now it looks like the main open discussions remaining are:

1. Code reuse options for the new `_copy_init` codepath. (Incidentally, I've actually wound up using `_copy_init` for something other than direct copying on the 'feature-faster-layout-creation' branch, so it is proving useful for accelerating some other circuit methods besides copying).
2. The right way to refactor the `is_simple` attribute/method for `Label` classes.
3. Whether or not we want to add a static flag/static checks to the SeparatePOVMCircuit class.

Let me know if there is anything I have missed, or if there is something I have marked as resolved which you think still needs working on.

pygsti/baseobjs/label.py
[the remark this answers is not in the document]
sserita: The common way to do a read-only thing is to make it a property. I think I would be in favor of adding the leading underscore back in, having a property for the non-underscore version, and then internally using the underscore version if you want to avoid the property call overhead.

sserita:
@coreyostrove My current stance on your short list:

1. This would be something that would be great to sort out, but if you have already tried and it was unclear the best path for reuse, then your current solution of comments pointing people to keep the functions in sync is a reasonable one. It does make some sense that the restricted version of _copy_init has optimizations that makes it hard to use within _bare_init... It was just not obvious to me at first glance that those were truly limiting.
2. I think this is definitely something we should fix, and I left a suggestion on the relevant thread.
3. When I made my comment about this, I didn't realize we were in SeparatePOVMCircuit. I'm OK with shelving this for now - I just wanted to make sure that Circuit had consistent static logic, not necessarily force other classes to adopt the static mode.

pygsti/circuits/circuit.py
sserita: Didn't realize this was in SeparatePOVMCircuit, my comment is less relevant in this case and I don't think adding static to a new class is needed for this PR :)

pygsti/baseobjs/label.py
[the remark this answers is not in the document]
coreyostrove: Great, thanks for the advice. I have hybridized your suggestions, making the class attribute screaming snake case, but also adding a lower cased `is_simple` property (without a setter). I have updated the calls in `circuit.py` to use the renamed class attribute directly, but have left the rest of the half dozen or so instances in the remainder of the codebase pointing to the property (I don't anticipate the method hit mattering in those cases).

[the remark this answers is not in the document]
sserita: Lol I forgot that was called screaming snake case. Sounds like the perfect convex combination of the suggestions to me :)

coreyostrove:
Awesome, I just pushed the changes addressing the `Label`  attribute conventions, so with that I think we're in good shape. Thanks @sserita and @rileyjmurray for your careful reviews and valuable feedback!

sserita (approved):
Thanks for the great work as usual @coreyostrove!

I'm approving, but this is again going to wait on merge until we clear beta tests up...

enielse:
These updates all look great to me.

sserita:
Beta tests clear, merging!

PR #451, opened 2024-06-06
Replace calls with .flatten() to .ravel() when appropriate

### LINKED ISSUE #424: legacyio.py contains no executable code.
All of the code in this file has been commented out. I suggest we delete the file.

https://github.com/sandialabs/pyGSTi/blob/ee2158575f4a6a9cffa7465983054be13772e0f8/pygsti/io/legacyio.py#L1-L385

### LINKED ISSUE #431: Replace unnecessary calls to ``.flatten()`` with ``.ravel()``.
As of the latest commit on [develop](https://github.com/sandialabs/pyGSTi/tree/ad0c1a5cbcdf081419d8b68ded0ae0da067da66b) (between releases of pyGSTi 0.9.12 and 0.9.13)m there are 102 hits for ``.flatten()`` in the pyGSTi codebase.

The [documentation for flatten](https://numpy.org/doc/1.25/reference/generated/numpy.ndarray.flatten.html) says that this returns a flattened copy of the input array. In the vast majority of situations it's preferable to use [``.ravel()``](https://numpy.org/doc/1.25/reference/generated/numpy.ravel.html#numpy.ravel) instead, which only returns a copy when necessary. I'll make a PR in the near future for these changes.

This is spiritually similar to #429 and #430, but different enough to warrant its own tracking issue.

rileyjmurray:
This PR aims to resolve #431. It also resolves #424 because "why not."

More specific changes:
* removed a giant commented-out code block in algorithms/randomcircuit.py.
* replaced some calls to ``np.dot`` with ``@``.
* simplified some code branching in a handful of cases.
* reduced use of multiple statements per line. Example: rewrite ``x = []; y = []``  as ``x, y = [], []``.
* I've left a note asking if it's cool to remove commented-out code in a function function in ``matrixtools.py``.
* removed ``matrixtools.py::residuals(a, b)`` and inlined equivalent code where needed.
* replaced instances of ``np.vdot(a.flatten(), b.flatten())`` with ``np.vdot(a, b)``.
* ~~changed code like ``list(array.flatten())`` with ``list(array.flat)`` instead of ``list(array.ravel())``~~.
    * Apparently ravel is often faster than flat, so I'm actually sticking with ravel.


pcwysoc (approved):
Looks good to me for my files! One thing I'd keep in mind that you will have to do this file (workspacesplots.ipynb) again when Corey merges in his substantial reporting changes since I doubt the auto-merge will catch this.

pygsti/report/reportables.py
```diff
@@ -2161,10 +2160,10 @@ def general_decomposition(model_a, model_b):
...
                 decomp[str(gl) + "," + str(gl_other) + " axis angle"] = 10000.0  # sentinel for irrelevant angle
 
-            real_dot = _np.clip(
-                _np.real(_np.dot(decomp[str(gl) + ' axis'].flatten(),
-                                 decomp[str(gl_other) + ' axis'].flatten())),
-                -1.0, 1.0)
+            real_dot = _np.real(_np.dot(
+                decomp[str(gl) + ' axis'].ravel(), decomp[str(gl_other) + ' axis'].ravel()
+            ))  # Riley question: should this be vdot instead of dot?
+            real_dot = _np.clip(real_dot, -1.0, 1.0)
```
coreyostrove, line 2166: The short answer is no.

The longer answer:
I traced through this function, and what is being calculated here is the inner product between the vectors of Hamiltonian rates for two different operators. This is of interest because the hamiltonian error generator rates for a psuedovector whose direction defines an axis of rotation, and whose length gives the angle. Assuming I understand everything here correctly the following should be true.
1. `logG` is always real-valued (this is enforced on line 2155 and within `tools.matrixtools.approximate_matrix_log`.
2. The hamiltonian projections are hardcoded to be with respect to the pauli-product basis.
I believe that combining these two facts it should be the case that the hamiltonian projection vectors are always going to be real-valued. Ergo, there is no need to take a conjugate transpose. (as a corollary to that I also think the np.real call is unnecessary). 

Fwiw I also think the ravel itself is uneeded. These vectors are references to a `LindbladCoefficientBlock`'s `block_data` attribute, and for a Hamiltonian coefficient block this should already by a 1D array. 

All that said, I might be missing something here. If I were you I'd stick in a couple assert statements and run the unit test suites (which should include report generation which will hit this codepath) and see if the stated assumptions are violated in some context.

pygsti/report/workspaceplots.py
```diff
@@ -3033,13 +3034,13 @@ def _create(self, dataset, target, maxlen, fixed_lists, scale):
 
         xs = list(range(svals.size))
         trace1 = go.Bar(
-            x=xs, y=list(svals.flatten()),
+            x=xs, y=list(svals.flat),
```
coreyostrove, line 3037: the `flat` property for ndarrays returns an instance of a `flatiter` object, so I had a suspicion this might be a bit slower. TLDR: For small arrays (4x4) `flat` is about 40% faster. For larger arrays (16x16 and 64x64) `flat` is about 15% slower. May well not matter here (depends on the typical size of svals, which I don't know off hand), but worth keeping in mind.

pygsti/tools/matrixtools.py
```diff
@@ -682,11 +682,16 @@ def approximate_matrix_log(m, target_logm, target_weight=10.0, tol=1e-6):
     assert(_np.linalg.norm(m.imag) < 1e-8), "Argument `m` must be a *real* matrix!"
     mx_shape = m.shape
 
+    #
+    #   Riley note: I'd like to remove all commented-out code in this function.
+    #   @Corey or @Stefan -- you okay with that?
+    #
+
```
coreyostrove, line 689: If you're referring to that block on the bottom beginning with 'OLD DEBUG' then go for it.

coreyostrove (changes requested):
Great work, @rileyjmurray! I have left some comments answering the questions that you had asked and pointed out a few (minor) inefficiencies. @sserita, none of my feedback is blocking, so feel free to approve this whenever you and @rileyjmurray think it is ready.

robinbk:
Quick comment re: Corey's observation about Hamiltonian error rates. The
part about axis and angle of rotations is only true for d=2, ie a single
qubit. A Hamilton defines a unitary rotation, represented as a
superoperator that is an orthogonal rotation in SO(d^2-1).  For d=2, this
is SO(3) which is basically SU(2), and so every rotation corresponds to an
axis and an angle.

For d>2, we're in SO(>>3), and the rules are different. A rotation can be
around multiple axes (the entire algebra of operators that commute with the
rotation), and has multiple angles. So caveat emptor!!

I don't know if this has any effect on Corey's actual conjecture -- it may
still be true -- but if the intuition about axes and angles is critical,
there's a risk of GIGO.

-R

pygsti/report/workspaceplots.py
rileyjmurray: Man that's wild. 

I'll switch to ``.ravel()`` for consistency, since it'll always be at least as fast as ``.flatten()`` and should be faster.

pygsti/tools/matrixtools.py
rileyjmurray: Nah, that's in a different function. I've actually come to to realize that this ``approximate_matrix_log`` function is doing something quite weird, so I think keeping comments for posterity's sake is reasonable.

pygsti/report/reportables.py
rileyjmurray: Thanks for investigating! I've run the entire test suite under a debugger. The only tests that hit these lines are in ``pygsti/test/test_packages/reportb`` (note the ``b`` in that last folder name). All hits to these lines on those tests confirmed your suspicions, so I've changed in the following ways:
* No flatten() or ravel() (unneeded since they're already arrays with ``.ndim == 1``)
* No taking a real part (unneeded since the expression was already real)
* Use ``@`` instead of ``np.dot``

sserita (approved):
Everything looks good to me, thanks @rileyjmurray!

PR #457, opened 2024-06-26
Add ECR support to QASM translation

### LINKED ISSUE #440: ECR Gate Compatibility
**Is your feature request related to a problem? Please describe.**
Current code for interfacing with IBM devices is not compatible with the ECR gate. However, all but one of the available IBM devices uses the ECR gate. 

**Describe the solution you'd like**
I believe we should refactor the IBM code to convert directly from pyGSTi circuits to Qiskit circuits, as this would smoothly integrate the ECR gate. 

**Describe alternatives you've considered**
I've considered trying to convert ECR gates to OpenQASM 2.0 as we currently do. However, this leads to very fragile QASM code: 

OPENQASM 2.0;
include "qelib1.inc";
gate rzx(param0) q0,q1 { h q1; cx q0,q1; rz(param0) q1; cx q0,q1; h q1; }
gate ecr q0,q1 { rzx(pi/4) q0,q1; x q0; rzx(-pi/4) q0,q1; }
qreg q[3];
creg c[1];
ecr q[1],q[2];
ecr q[0],q[2];
measure q[2] -> c[0];
barrier q[0],q[1],q[2];

This works for jobs submitted through Runtime or the Quantum Composer on the IBM website, but seems to fail for jobs submitted through backend.run():

Screenshot 2024-05-15 at 12 41 16 PM

I'm concerned that if we pursue this avenue that the fix would be unstable.

sserita:
Should fix #440 by adding an ECR macro when ECR gates are detected.

This change should be primarily independent of anything in #379.

### DIFF
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -4250,6 +4250,9 @@ class Circuit(object):
             # Include a delay instruction
             openqasm += 'opaque delay(t) q;\n\n'
 
+        # Add a template for ECR commands that we will replace/remove later
+        openqasm += "ECRPLACEHOLDER"
+
         openqasm += 'qreg q[{0}];\n'.format(str(num_qubits))
         # openqasm += 'creg cr[{0}];\n'.format(str(num_qubits))
         openqasm += 'creg cr[{0}];\n'.format(str(num_qubits + num_IMs))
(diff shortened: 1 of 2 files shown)


coreyostrove (approved):
Looks good to me, thanks, @sserita!

PR #458, opened 2024-07-08
Remove unused isin 

eendebakpt:
The import of `isin` used is not availabe in numpy 2.0. If the import is required for backwards compatibility, then I suggest to replace it with `from numpy import isin`


coreyostrove (approved):
Thanks as always for your help, @eendebakpt!

I agree that this doesn't appear to be used at present, so I don't seen any reason not to remove the import.

sserita (approved):
Looks good, thanks @eendebakpt!

Thanks @coreyostrove for catching the branch.

PR #468, opened 2024-07-18
Interleaved RB

coreyostrove:
This PR adds first party support for interleaved randomized benchmarking (as described in the original 2012 paper from Magesan et al. https://arxiv.org/pdf/1203.4550) to the RB codebase. The analysis for IRB is incredibly simple, so from an implementation standpoint much of what has been added is aimed at streamlining the experiment design construction process, automating this analysis, and creating data structures for keeping the two relevant subexperiments stored together.

Specifically I have added the following:

1. `InterleavedRBDesign`: This class both performs the creation of the two subexperiment designs (a standard CRB experiment and one with a target gate interleaved between random cliffords) and acts as a container for these two experiment designs. This subclasses off of `CombinedExperimentDesign` which allows for use to index into each of the subdesigns as well as interact with them jointly (getting the combined `all_circuits_needing_data` value at data collection time, for example).
2. `InterleavedRandomizedBenchmarking`: This is a new protocol class which farms off the RB estimation for each of the subexperiments to `RandomizedBenchmarking` and then extracts the relevant results for calculating the IRB numbers and bounds. This returns a `ProtocolResultsDir` containing the IRB specific results along with the full results for each of the subexperiments.
3. `InterleavedRandomizedBenchmarkingResults`: This is a simple (mostly data) class for storing the IRB results at the top node of the aforementioned `ProtocolResultsDir`.

Aside from the code described above I have also added a new section to the Clifford RB tutorial notebook demonstrating the use of the new functionality. Also included are new unit tests, both for the IRB features added and for improving testing coverage on existing parts RB (particularly with regards to serialization). Lastly, there is a bug fix for a reported issue where serialization of CliffordRBDesign was failing when constructed using the (previously undocumented) `interleaved_circuit` kwarg.

Big thanks go out to @jordanh6 and @enielse for their valuable advice in the achitecting of these additions, and to @jordanh6 and @tjproct for their guidance on the IRB theory.

### DIFF
--- a/pygsti/protocols/rb.py
+++ b/pygsti/protocols/rb.py
@@ -1212,7 +1212,8 @@ class InterleavedRBDesign(_proto.CombinedExperimentDesign):
         print('Constructing Interleaved CRB Subdesign:')
         icrb_subdesign = CliffordRBDesign(pspec, clifford_compilations, depths, circuits_per_depth, qubit_labels, randomizeout,
                                               interleaved_circuit, citerations, compilerargs, exact_compilation_key,
-                                              descriptor + ' (Interleaved)', add_default_protocol, seed+1, verbosity, num_processes)
+                                              descriptor + ' (Interleaved)', add_default_protocol, seed+1 if seed is not None else None, 
+                                              verbosity, num_processes)
 
         self._init_foundation(crb_subdesign, icrb_subdesign, circuits_per_depth, interleaved_circuit, randomizeout,
                               citerations, compilerargs, exact_compilation_key, interleave)
--- a/pygsti/protocols/rb.py
+++ b/pygsti/protocols/rb.py
@@ -1188,7 +1188,8 @@ class InterleavedRBDesign(_proto.CombinedExperimentDesign):
 
     seed : int, optional
         A seed to initialize the random number generator used for creating random clifford
-        circuits.
+        circuits. The first of the two subdesigns will use the specified seed directly,
+        while the second will use seed+1.
 
     verbosity : int, optional
         If > 0 the number of circuits generated so far is shown.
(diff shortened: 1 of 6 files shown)


pygsti/protocols/rb.py
```diff
@@ -1273,17 +1410,9 @@ def _get_rb_fits(circuitdata_per_depth):
...
             for depth in depths:
                 percircuitdata = circuitdata_per_depth[depth]
-                #print(percircuitdata)
-                if self.square_mean_root:
-                    #print(percircuitdata)
-                    adj_sps.append(_np.nanmean(_np.sqrt(percircuitdata))**2)
-                    #print(adj_sps)
-                else:
-                    adj_sps.append(_np.nanmean(percircuitdata))  # average [adjusted] success probabilities or energies
+                adj_sps.append(_np.nanmean(percircuitdata))  # average [adjusted] success probabilities or energies
```
coreyostrove, line 1413: The removal of the `square_mean_root` kwarg (and related behavior shown here) is unrelated to the IRB stuff and was done at the recommendation of @tjproct. This was added at one point as an experiment and no longer makes sense to keep as an analysis option.

pygsti/tools/nameddict.py
```diff
@@ -88,7 +88,7 @@ def _serialize(x):
             #TODO: serialize via _to_memoized_dict once we have a base class
             if x is None or isinstance(x, (float, int, str)):
                 return x
-            elif isinstance(x, _np.int64):
+            elif isinstance(x, (_np.int64, _np.int32)):
```
coreyostrove, line 91: Unrelated serialization fix to address pickle fallback (I think this is a windows specific thing related to the whole 32-bit integer thing that gave us headaches a couple years ago) and a related warning when serializing some RB-related objects.

sserita: Thanks for catching these sorts of Windows oddities!

test/unit/protocols/test_rb.py
```diff
@@ -547,13 +645,74 @@ def setUp(self):
...
         
     def test_mirrorrb_protocol_noisy(self):
         proto = pygsti.protocols.rb.RandomizedBenchmarking(datatype='adjusted_success_probabilities', defaultfit='A-fixed', rtype='EI',
-                 seed=(0.8, 0.95), bootstrap_samples=200, depths='all', square_mean_root=False, name=None)
+                 seed=(0.8, 0.95), bootstrap_samples=200, depths='all', name=None)
+        
+        result = proto.run(self.data_noisy)
+
+
+class TestInterleavedRBProtocol(BaseCase):
```
sserita, line 660: Thanks as always for writing tests!

sserita (approved):
Looks great, thanks @coreyostrove!

PR #480, opened 2024-08-23
Bugfix reparameterize

juangmendoza19:
Addressing issue #425. 

convert() function has been changed for POVM elements and state preparation. They now support proper conversion from full TP to GLND parameterization. The optimization is now done over non-gauge directions, which are found through Jacobians.

This function still needs to be changed to appropriately handle other reparameterizations.

Lastly, this is not guaranteed to work for POVMs, just because the number of degrees of freedom of a POVM can be greater than the number of error generators that span TP maps for a given number of qubits. By pigeon hole principle, we may not be able to find a description for a POVM consisting of Ideal_POVM + error generators. There is an assertion to try to catch this.


sserita (commented):
Just bookkeeping: Not ready for review, rerequest from me when ready.

pygsti/baseobjs/errorgenbasis.py
```diff
@@ -460,6 +460,19 @@ def label_index(self, elemgen_label, ok_if_missing=False):
...
+            
+            if ok_if_missing:
+                return None
+            else:
+                raise error
+        """
+        TODO: 2 qubit labels returning None when label does exist in self.labels. 
+        '(support, left_support) not in self._offsets[eetype]' returning True incorrectly
+
+
```
juangmendoza19, line 475: Fancy code below to find the index of a label was not working for 2 qubits. Temporary easy solution implemented by just calling index(). This was creating problems during the creation of 2 qubit FOGI models

pygsti/modelmembers/povms/__init__.py
```diff
@@ -375,14 +386,110 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                    except RuntimeError:  # try static mixed states next:
+                        #if idl.get(lbl,None) is not None:
+                        
+                        base_items = []
+                        for lbl, vec in povm.items():
+                            ideal_effect = idl.get(lbl,None)
+                            if ideal_effect is not None:
+                                base_items.append((lbl, convert_effect(ideal_effect, 'static', basis, ideal_effect, flatten_structure)))
+                            else:
+                                base_items.append((lbl, convert_effect(vec, 'static', basis, idl.get(lbl, None), flatten_structure)))
```
juangmendoza19, line 398: This is code to properly collect the ideal POVM effects. Before, the ideal_effect passed in was being ignored

pygsti/modelmembers/povms/__init__.py
```diff
@@ -375,14 +386,110 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                
+                #def callback(x): print("callbk: ",_np.linalg.norm(x),_objfn(x))  # REMOVE
+                soln = _spo.minimize(_objfn, _np.zeros(len(phys_directions), 'd'), method="Nelder-Mead", options={},
+                                        tol=1e-13)  # , callback=callback)
+                #print("DEBUG: opt done: ",soln.success, soln.fun, soln.x)  # REMOVE
+                if not soln.success and soln.fun > 1e-6:  # not "or" because success is often not set correctly
+                    raise ValueError("Failed to find an errorgen such that <ideal|exp(errorgen) = <effect|")
+                errgen_vec = _np.linalg.pinv(phys_directions)  @ soln.x
+                errorgen.from_vector(errgen_vec)
+
```
juangmendoza19, line 492: Code added to find errgen parameterization for POVMs based on optimization, just like state prep was being done before, but for some reason was never implemented on POVMs. Note that the optimization variations are done over a POVM space that lacks the dumb-gauge/metagauge. To find that space we wrote the function calc_physical_subspace which does a numerical calculation of the Jacobian with respect to POVM parameters.

pygsti/modelmembers/states/__init__.py
```diff
@@ -256,20 +261,57 @@ def convert(state, to_type, basis, ideal_state=None, flatten_structure=False):
...
+                            exp_errgen.from_vector(new_vec)
+                            J[:,i] = (exp_errgen.to_dense() @ ideal_prep - ideal_prep)[1:]/epsilon
+
+                        _,S,V = _np.linalg.svd(J)
+                        return V[:len(S),]
+                        
+                    phys_directions = calc_physical_subspace(dense_state)
+    
+                    #We use optimization to find the best error generator representation
+                    #we only vary physical directions, not independent error generators
```
juangmendoza19, line 295: Just as explained in the POVM init file, we added a function to identify the dumb-gauge/metagauge and move orthogonally to it. This allows for better optimization.

pygsti/modelmembers/states/__init__.py
```diff
@@ -256,20 +261,57 @@ def convert(state, to_type, basis, ideal_state=None, flatten_structure=False):
...
+                        for coeff, phys_direction in zip(v,phys_directions):
+                            L_vec += coeff * phys_direction
+                        errorgen.from_vector(L_vec)
+                        proc_matrix = _spl.expm(errorgen.to_dense())
+                        #basis of to_dense() CHECK if always pp
+                        dense_basis = Basis.cast('pp', (2**num_qubits)**2)
+                        return _np.linalg.norm(proc_matrix @ dense_st - dense_state) + cptp_penalty * sum_of_negative_choi_eigenvalues_gate(proc_matrix, dense_basis)
+                    
+                    soln = _spo.minimize(_objfn, _np.zeros(len(phys_directions), 'd'), method="Nelder-Mead", options={},
+                                         tol=1e-13)  # , callback=callback)
```
juangmendoza19, line 308: the objective function to be optimized over was modified to work with linear combinations of error generators which lack these dumb-gauge degrees of freedom

pygsti/models/explicitmodel.py
```diff
@@ -340,7 +340,7 @@ def __getitem__(self, label):
             raise KeyError("Key %s has an invalid prefix" % label)
 
     def convert_members_inplace(self, to_type, categories_to_convert='all', labels_to_convert='all',
-                                ideal_model=None, flatten_structure=False, set_default_gauge_group=False, cptp_truncation_tol= 1e-6):
+                                ideal_model=None, flatten_structure=False, set_default_gauge_group=False, cptp_truncation_tol= 1e-7, spam_cptp_penalty = 1e-6):
```
juangmendoza19, line 343: added piping for the cptp penalty factor to be used in the optimization within the SPAM convert functions discussed above

pygsti/tools/jamiolkowski.py
```diff
@@ -273,6 +273,27 @@ def fast_jamiolkowski_iso_std_inv(choi_mx, op_mx_basis):
...
+    -------
+    float
+        the sum of the negative eigenvalues of the Choi representation of op_mx
+    """
+    sumOfNeg = 0
+    J = fast_jamiolkowski_iso_std(op_mx, op_mx_basis)  # Choi mx basis doesn't matter
+    evals = _np.linalg.eigvals(J)  # could use eigvalsh, but wary of this since eigh can be wrong...
+    for ev in evals:
+            if ev.real < 0: sumOfNeg -= ev.real
+    return sumOfNeg
```
juangmendoza19, line 296: Created a function to sum the negative choi eigenvalues of a single gate to be used in our optimization functions to measure CPTPness

juangmendoza19:
All the code that was changed which does not have a comment attached was a result of merging the develop branch.

coreyostrove:
The commits coming from merging in develop shouldn't be showing up, so something got glitched out here. I can confirm that when I just went into the new PR creation interface and looked at the diff that was created for merging in this branch into develop only the new commits related to these bugfixes showed up. This might be the sort of thing that fixes itself if you switch the merge target to something other than develop (e.g. master) and then back again. It doesn't look like I have the permissions to do that on my end, so could either @juangmendoza19  or someone else with the requisite permissions give that a shot?

pygsti/modelmembers/states/__init__.py
```diff
@@ -256,20 +261,57 @@ def convert(state, to_type, basis, ideal_state=None, flatten_structure=False):
...
-                    #def callback(x): print("callbk: ",_np.linalg.norm(x),_objfn(x))  # REMOVE
-                    soln = _spo.minimize(_objfn, _np.zeros(errorgen.num_params, 'd'), method="CG", options={},
-                                         tol=1e-8)  # , callback=callback)
+                        L_vec = _np.zeros(len(phys_directions[0]))
+                        for coeff, phys_direction in zip(v,phys_directions):
+                            L_vec += coeff * phys_direction
+                        errorgen.from_vector(L_vec)
+                        proc_matrix = _spl.expm(errorgen.to_dense())
+                        #basis of to_dense() CHECK if always pp
+                        dense_basis = Basis.cast('pp', (2**num_qubits)**2)
```
coreyostrove, line 304: This shouldn't assume a hardcoded pauli-product basis. I don't think the error generator modelmember itself has a basis attribute to grab, but it should have a reference to a parent model which would have this available. Can you update this to grab the appropriate basis? (And if you were doing something similar for the POVMs and I missed it add the change there too.)

pygsti/tools/optools.py
```diff
@@ -306,9 +306,6 @@ def diamonddist(a, b, mx_basis='pp', return_x=False):
 
     # currently cvxpy is only needed for this function, so don't import until here
     import cvxpy as _cvxpy
-    old_cvxpy = bool(tuple(map(int, _cvxpy.__version__.split('.'))) < (1, 0))
-    if old_cvxpy:
-        raise RuntimeError('CVXPY 0.4 is no longer supported. Please upgrade to CVXPY 1.0 or higher.')
```
coreyostrove, line 309: Is this the official @rileyjmurray endorsed fix for this? Do we no longer need this check?

coreyostrove (changes requested):
See my inline comments and questions below. Thanks!

test/unit/objects/test_model.py
```diff
@@ -715,7 +741,8 @@ def test_bulk_fill_hprobs_with_high_smallness_threshold(self):
     def test_iter_hprobs_by_rectangle(self):
         self.skipTest("TODO should probably warn user?")
 
-
+class LinbladModelTester(GLNDModelBase, StandardMethodBase, BaseCase):
```
sserita, line 744: Typo: Linblad -> Lindblad

pygsti/modelmembers/povms/__init__.py
```diff
@@ -327,13 +331,23 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+            within this gauge orbit which are CPTP.
 
     Returns
     -------
     POVM
         The converted POVM vector, usually a distinct
         object from the object passed as input.
     """
+
+    ##TEST CONVERSION BETWEEN LINBLAD TYPES
```
sserita, line 350: Remove comment?

sserita (changes requested):
One refactor requested, and a few minor fixes while you are at it.

pygsti/modelmembers/states/__init__.py
sserita: Is this function exactly the same as the other one (at first glance they at least look very similar)?.
If they are the same (or are the same up to some minor change that could be controlled with a function parameter), can you refactor this function out to avoid code duplication?

sserita (changes requested):
Forgot to hit Enter on my refactor request...

coreyostrove:
@juangmendoza19 and @sserita : Do either of you recall what still needs to get sorted out on this PR? I see that there is at least one merge conflict (probably from merging in develop recently) and some unresolved discussion threads above.

juangmendoza19:
> @juangmendoza19 and @sserita : Do either of you recall what still needs to get sorted out on this PR? I see that there is at least one merge conflict (probably from merging in develop recently) and some unresolved discussion threads above.

Stefan’s requests above have not been implemented yet.

pygsti/tools/optools.py
coreyostrove: Bump for @rileyjmurray

rileyjmurray: This was the official endorsed fix. It's no longer needed, so I support removing.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -418,7 +422,7 @@ def povm_type_from_op_type(op_type):
     return povm_type_preferences
 
 
-def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
+def convert(povm, to_type, basis, cp_penalty=1e-7, ideal_povm=None, flatten_structure=False):
```
rileyjmurray, line 425: When adding keyword arguments it's good to append them to the existing argument sequence. This way if someone uses positional arguments they don't get an error when using updated code. In this situation, placing the new cp_penalty keyword argument before the existing ideal_povm keyword argument could lead to errors for people specifying ideal_povm by its position.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -498,14 +512,102 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                    errgen = _LindbladErrorgen.from_error_generator(povm.state_space.dim, parameterization=to_type)
+
+                    if degrees_of_freedom > errgen.num_params:
+                        warnings.warn("POVM has more degrees of freedom than the available number of parameters, representation in this parameterization is not guaranteed")
+                    exp_errgen = _ExpErrorgenOp(errgen)
+                    
+                    num_errgens = errgen.num_params
+                    #TODO: Maybe we can use the num of params instead of number of matrix entries, as some of them are linearly dependent.
+                    #i.e E0 completely determines E1 if those are the only two povm elements (E0 + E1 = Identity)
+                    num_entries = dense_ideal_povm.shape[0]*dense_ideal_povm.shape[1]
```
rileyjmurray, line 562: If ``dense_ideal_povm`` is a 2d numpy array then it's equivalent to set ``num_entries = dense_ideal_povm.size``.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -498,14 +512,102 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                    for i in range(num_errgens):
+                        new_vec = _np.zeros(num_errgens)
+                        new_vec[i] = epsilon
+                        exp_errgen.from_vector(new_vec)
+                        vectorized_povm = _np.zeros(num_entries)
+                        perturbed_povm = (dense_ideal_povm @ exp_errgen.to_dense() - dense_ideal_povm)/epsilon 
+
+                        perturbed_povm_t = perturbed_povm.transpose()
+                        for j, column in enumerate(perturbed_povm_t):
+                            vectorized_povm[j*len(perturbed_povm_t[0]):(j+1)*len(perturbed_povm_t[0])] = column
```
rileyjmurray, line 577: Minimum change: assign ``len(perturbed_povm_t[0])`` to a variable before the for-loop. Then use that variable in the loop for improved readability.

Preferred change: replace the for loop with ``vectorized_povm = perturbed_povm.flatten(order='F')``. Note that you don't need to predefine ``vectorized_povm = _np.zeros(num_entries)`` if you make this change.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -498,14 +512,102 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                        
+                        J[:,i] = vectorized_povm.transpose()
```
rileyjmurray, line 579: You've defined ``vectorized_povm`` as a 1d numpy array. Calling ``.transpose()`` on such an array has no effect.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -498,14 +512,102 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                    #i.e E0 completely determines E1 if those are the only two povm elements (E0 + E1 = Identity)
+                    num_entries = dense_ideal_povm.shape[0]*dense_ideal_povm.shape[1]
+
+                    #Compute the jacobian with respect to the error generators. This will allow us to see which
+                    #error generators change the POVM entries
+                    J = _np.zeros((num_entries,num_errgens))
+                    
+                    for i in range(num_errgens):
+                        new_vec = _np.zeros(num_errgens)
+                        new_vec[i] = epsilon
```
rileyjmurray, line 570: Minor point: rather than defining ``new_vec`` at the start of every loop, you can define it before the loop, then at the start of a loop iteration you set ``new_vec[i] = epsilon`` and at the end of the iteration you set ``new_vec[i] = 0``.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -498,14 +512,102 @@ def convert(povm, to_type, basis, ideal_povm=None, flatten_structure=False):
...
+                        perturbed_povm = (dense_ideal_povm @ exp_errgen.to_dense() - dense_ideal_povm)/epsilon 
+
+                        perturbed_povm_t = perturbed_povm.transpose()
+                        for j, column in enumerate(perturbed_povm_t):
+                            vectorized_povm[j*len(perturbed_povm_t[0]):(j+1)*len(perturbed_povm_t[0])] = column
+                        
+                        J[:,i] = vectorized_povm.transpose()
+
+                    _,S,V = _np.linalg.svd(J)
+                    return V[:len(S),]
```
rileyjmurray, line 582: Three comments.
1. numpy and scipy SVD functions return the transpose of V, not V itself. So it's good naming convention to say ``Vt`` instead of ``V``. If you were dealing with complex matrices then you'd want to write ``Vh`` instead of ``V``.
2. If you only need the first "number of singular value" singular vectors, then you should call this as ``_, _, Vt = _np.linalg.svd(J, full_matrices=False)``.
3. This implementation isn't checking if any singular values in ``S`` are zero. If they are, then you probably want tor return fewer singular vectors. Consider how, in the current implementation, ``V[:len(S),]`` can be can arbitrary orthogonal matrix when ``J`` is the zero matrix.

pygsti/modelmembers/states/__init__.py
```diff
@@ -256,20 +261,54 @@ def convert(state, to_type, basis, ideal_state=None, flatten_structure=False):
...
+                        return _np.linalg.norm(proc_matrix @ dense_st - dense_state) + cp_penalty * sum_of_negative_choi_eigenvalues_gate(proc_matrix, basis)
+                    
+                    soln = _spo.minimize(_objfn, _np.zeros(len(phys_directions), 'd'), method="Nelder-Mead", options={},
+                                         tol=1e-13)  
+                    
                     if not soln.success and soln.fun > 1e-6:  # not "or" because success is often not set correctly
                         raise ValueError("Failed to find an errorgen such that exp(errorgen)|ideal> = |state>")
-                    errorgen.from_vector(soln.x)
+                    
+                    errgen_vec = _np.linalg.pinv(phys_directions)  @ soln.x
```
rileyjmurray, line 310: If you want to apply the pseudo-inverse of a matrix to a vector, it's better to do that implicitly by solving a least squares problem: ``errgen_vec = _np.linalg.lstsq(phys_directions, soln.x)[0]``.

pygsti/tools/jamiolkowski.py
```diff
@@ -273,6 +273,27 @@ def fast_jamiolkowski_iso_std_inv(choi_mx, op_mx_basis):
...
+    op_mx_basis : Basis
+
+    Returns
+    -------
+    float
+        the sum of the negative eigenvalues of the Choi representation of op_mx
+    """
+    sumOfNeg = 0
+    J = fast_jamiolkowski_iso_std(op_mx, op_mx_basis)  # Choi mx basis doesn't matter
+    evals = _np.linalg.eigvals(J)  # could use eigvalsh, but wary of this since eigh can be wrong...
```
rileyjmurray, line 293: Please elaborate on what you noticed when using ``eigh`` instead of ``eig``. I'd consider the former function more reliable than the latter.

Did you just copy over the comment from ``sums_of_negative_choi_eigenvalues`` which predates this PR?

test/unit/objects/test_model.py
rileyjmurray, line 1: If you're editing code that uses ``_np.dot``, please go ahead and change to use the infix operator ``@`` instead.

test/unit/objects/test_model.py
```diff
@@ -656,7 +682,8 @@ def test_bulk_fill_hprobs_with_high_smallness_threshold(self):
     def test_iter_hprobs_by_rectangle(self):
         self.skipTest("TODO should probably warn user?")
 
-
+class LindbladModelTester(GLNDModelBase, StandardMethodBase, BaseCase):
+    pass
```
rileyjmurray, line 686: Why is this needed / what is it accomplishing?

rileyjmurray (changes requested):
Left a few comments. Happy to iterate quickly to see them resolved!

pygsti/tools/jamiolkowski.py
coreyostrove: This one I can confirm is a carry over comment from Erik's original code, but I don't know what he meant by this.

test/unit/objects/test_model.py
coreyostrove: The testing code (a lot of which is architected like this) is kind of a pain to read. The way to parse this is that a set of generically defined testing methods (generic in the sense of they make sense independent of a model's parameterization) defined in `StandardMethodBase` should be run against a model with parameterization as defined in `GLNDModelBase`, with access to a number of useful testing utilities defined in `BaseCase`. So while it doesn't look like the class definition does anything, it is actually the case that the act of wrapping all of those parent classes as done here (and with particular keywords in the class name) results in a test suite which pytest knows how to parse and run (I think it looks for "Tester" in the class name, among other possible keywords).

rileyjmurray: Gotcha. Marking as resolved!

pygsti/tools/jamiolkowski.py
rileyjmurray: Ping @enielse. Do you remember why you preferred eig as opposed to eigh?

pygsti/modelmembers/povms/__init__.py
rileyjmurray: @juangmendoza19, can you take a look at points 2 and 3? My sense is that if ``S[i]`` is near zero then you wouldn't want to include ``Vt[i,:]`` among the returned vectors. Ping @coreyostrove in case you can provide insight on how (near-)zero singular values should be handled.

coreyostrove: Your intuition is correct, singular vectors associated with near-zero singular vectors shouldn't be included. It looks like Juan's most recent pair of commits accounts for this now.

rileyjmurray: Just looked at the change. The logic is fine with me now, but I did leave a comment asking that the three or four lines be reduced to a single line.

pygsti/tools/jamiolkowski.py
rileyjmurray: I opened issue #582 to investigate this. Marking this issue as resolved so we can get this merged.

rileyjmurray (approved):

test/unit/objects/test_output.txt
coreyostrove, line 1: This looks like it was accidentally added to one of your commits. Could you delete this?

coreyostrove:
> @coreyostrove this looks good to me. Other than resolving any merge conflicts or failing tests, do you still have unaddressed comments?
> 
> I see that Stefan has unresolved comments. Rather than try to resolve those I suggest we dismiss Stefan's review (see screenshot below) and make a GitHub issue about following up on Stefan's unresolved points later. image

One more minor change related to an errant file on my end, but after that I'm satisfied with all of this. I'll approve once that last change is made and we have confirmation that tests are still passing.

rileyjmurray:
@coreyostrove all tests pass. If you approve then we can merge.

coreyostrove (approved):
Great work, @juangmendoza19! And thanks for bearing with the many changes requested during the review process.

Issue #456, opened 2024-06-24
Method `parameter_labels()` returns an empty array


Pablovinas:
**Describe the bug**
After adding operations to an explicit model constructed via `pygsti.models.ExplicitOpModel()`, the method `parameter_labels()` returns an empty array and does not recognize the model's parameters. If the model has been initially imported from `pygsti.modelpacks`, the array is not updated with the new parameters.

**To Reproduce**
1.  Create a model, such as those defined in the the `ExplicitModel.ipynb` tutorial. We can additionally define custom operations as in the `CustomOperator.ipynb` tutorial and keep track of our parameters after naming them using the `parameter_labels(self)` property.
2. Once the model is created, use `mdl.parameter_labels` to get the vector of model member's parameters within the model. It returns an empty array, causing further issues such as the `collect_parameters()`method not working.

**Expected behavior**
I would expect the labels and parameters defined in the operations to be automatically recognised and updated in the model containing them.

**Environment:**
 - pyGSTi version 0.9.12.3
 - python version 3.12.3
 - OS: macOS Ventura 13.0.1

**Additional context**
I have been able to circumvent this issue by explicitly using `mdl._rebuild_paramvec()`. However, I believe this should be done automatically when defining the model.

coreyostrove:
Hi @Pablovinas, thanks for the detailed report!

This isn't a bug per se, but rather related to a fairly old design choice to make the parameter management for models update lazily. As you've identified above, as currently implemented simply assigning a new value to a model member in one of the `Model` object's member dictionaries doesn't actually force a rebuild (i.e. a call to `_rebuild_paramvec`) in and of itself. I wasn't around for the original implementation of this, but I imagine it was for performance related considerations (rebuilding the parameter vector is actually a pretty expensive operation). The downside, as you discovered, is unexpected behavior wherein the values returned for certain model attributes can fall out of sync with the correct values depending on when you make the queries. This is not a particularly great behavior to have and it has tripped me up a number of times as well, so I think this is a good opportunity to fix this.

There are a couple options I can think of on how to proceed, so I wanted to open up some discussion on this to see what other folks think. 

1. Keep things lazy, but audit `Model` and add additional rebuild checks for publicly facing properties where we identify a potential for post-assignment inconsistencies.
[...]

PR #482, opened 2024-09-05
Bugfix Model Parameter Desync

coreyostrove:
These changes are meant to address the concerns raised in issue #456. To summarize that issue, when manually adding model members of a model's member dictionary it was previously possible to run into scenarios where the values returned for certain public facing attributes, such as the list of parameter labels, had fallen out of sync with the current constituents of the model. In this PR this has been addressed by adding additional checks to various public facing model attributes which check whether we need to rebuild the model, and performs that rebuild if needed before returning the requested value.

This specifically adds checks to the `parameters_labels`, `set_parameter_label`, `parameter_bounds` and `num_modeltest_params` properties/methods. 

Also included is an unrelated bug in the `Circuit` class which got caught by unit testing. 

In the course of testing these changes I also discovered some unexpected and undocumented behavior related to the handling of parameter bounds during the model rebuild stage. When manually setting parameter bounds at the `Model` level and not at the `ModelMember` level these manually specified values get overwritten during the rebuild process by whatever values (if any) are present in the constituent model members. We now have a check in the rebuild process which detects when there is a non-trivial parameter bound vector for a model, and a warning is raised alerting users to this behavior and the possible need to reset any manually specified parameter bounds again at the Model level (if appropriate).


pygsti/models/model.py
```diff
@@ -610,6 +611,56 @@ def num_params(self):
...
         self._clean_paramvec()
         return len(self._paramvec)
+    
+    @property
+    def parameter_labels(self):
+        """
+        A list of labels, usually of the form `(op_label, string_description)` describing this model's parameters.
+        """
+        self._clean_paramvec()
+        return self._paramlbls
```
sserita, line 621: So I thought the merge would catch this but apparently not. If the merge didn't catch it, then I thought the tests would complain about a double property definition. But it looks like neither happened...
But #488 also introduced a parameter_labels, defined on 493 above. Should these be combined like so?

```
@property
    def parameter_labels(self):
        """
        A list of labels, usually of the form `(op_label, string_description)` describing this model's parameters.
        """
        self._clean_paramvec()
        return self._ops_paramlbls_to_model_paramlbls(self._paramlbls)
```

sserita (commented):
Potential merge conflict with #488 to resolve (which was not caught by git or the tests, apparently).

pygsti/models/model.py
coreyostrove: That is strange... but they should indeed be combined as you did above.

sserita: Mkay, I will do so and merge after that unless you had other blocking issues?

sserita (approved):

Issue #441, opened 2024-05-22
CloudNoiseModel needs better callable op logic


sserita:
**Describe the bug**
Passing in a `UnitaryGateFunction` to `create_cloud_crosstalk_model` fails because the `CloudNoiseModel` does not handle callable ops correctly.

**To Reproduce**
[CloudNoiseCallableOpBug.ipynb.zip](https://github.com/sandialabs/pyGSTi/files/15406570/CloudNoiseCallableOpBug.ipynb.zip)

**Expected behavior**
The callable op should be wrapped in a `UnitaryOpFactory` just like `LocalNoiseModel` does.

**Environment (please complete the following information):**
 - pyGSTi 0.9.12
 - python 3.11
 - OSX 10.14

rileyjmurray:
The following seems to execute correctly after applying the changes in PR #485.

```python
import numpy as np

from pygsti.processors import QubitProcessorSpec
from pygsti.baseobjs.statespace import QubitSpace
from pygsti.modelmembers.operations.opfactory import UnitaryOpFactory
from pygsti.models.modelconstruction import create_cloud_crosstalk_model_from_hops_and_weights

class CallableUnitaryOpFactory(UnitaryOpFactory):
    
    shape = (2, 2)
    
    def __call__(self, args=None, sslbls=None):
        return self.create_object(args=args, sslbls=sslbls)

def gu3_impl(arg):
    theta, phi, lamb = (float(arg[0]), float(arg[1]), float(arg[2]))
    return np.array([[np.cos(theta/2), -np.exp(1j*lamb)*np.sin(theta/2)],
                         [np.exp(1j*phi)*np.sin(theta/2), np.exp(1j*(phi + lamb))*np.cos(theta/2)]])

Gu3Factory = CallableUnitaryOpFactory(fn=gu3_impl, state_space=QubitSpace(1))

n_qubits = 4
pspec = QubitProcessorSpec(
    n_qubits, ['Gu3', 'Gcnot'], geometry='line',
    nonstd_gate_unitaries={'Gu3' : Gu3Factory},
    qubit_labels=[f'Q{i}' for i in range(n_qubits)]
)

model = create_cloud_crosstalk_model_from_hops_and_weights(
    pspec,
    # custom_gates={ 'Gu3' : Gu3Factory },  # doesn't matter if you have this line.
    max_idle_weight=0,
    max_spam_weight=1,
    maxhops=2,
    extra_weight_1_hops=0,
[...]

PR #485, opened 2024-09-16
Bugfix cloud noise model callable ops

sserita:
When completed, this PR will address the following issues:

- Labels with gate arguments not being handled properly in CloudNoiseModels.
- TBD issue, see comment below.
- #441


pygsti/models/modelnoise.py
```diff
@@ -537,7 +537,7 @@ def apply_errormap_stencil(self, stencil, evotype, state_space, target_labels=No
                 # `None` in list signals a non-present direction => skip these terms
                 sslbls_list = list(filter(lambda x: x is not None, sslbls_list))
             for sslbls in sslbls_list:
-                op_to_embed = local_errormap if (sslbls is None or state_space.is_entire_space(sslbls)) \
```
coreyostrove, line 540: It is not immediately obvious what bug would have been caused by the `is_entire_space` call. I'm sure it's fine, but could you explain what this fixes?

sserita: Currently, these two fixes come from Aidan and were edge cases he ran into while doing some CloudNoiseModel simulations. I understand the other fix, but forget what necessitated this fix. @AidanGauthier do you remember why we needed to do this?
@coreyostrove btw, I'll probably move this to a draft while updating the issue description. I want to fix #441 at the same time.

sserita: I talked to Aidan and here's the edge case this caused: `local_errorgen` comes from the stencil, so it has generic stencil state space labels. However, the `is_entire_space` call only checks _after_ the sslbls have been mapped  on line 481. In the case the mapped state space fully matches, then the still-stencil errorgen is passed through, causing downstream sslbl mismatch errors.

EmbeddedOps presumably fix this because it is already doing sslbl mapping, but it's not the most efficient thing to do in this case. @coreyostrove, unless you have another suggestion, I'll probably expand this line into an if/elif/else and handle the is_entire_space in its own block where we copy the op with the correct sslbls instead of relying on EmbeddedOp to do so.

coreyostrove: Sounds perfect, thanks to you and @AidanGauthier for tracking this down and explaining!

rileyjmurray (approved):
LGTM.

rileyjmurray:
I'd like to merge as-is. Will confer with Stefan when he gets to the office today.

PR #489, opened 2024-09-19
Bugfix guard signal

sserita:
Small bugfixes:

- Guarding setting SIGINT via `signal` in the customlm code. Needed in order to use pyGSTi with some external parallelization frameworks, such as Dask.
- Fixing unit tests by including a `Label.is_simple()` to `Label.IS_SIMPLE` that was missed.


jccalvojackson:
Why was this reverted?

Issue #262, opened 2022-09-07
Qibo API Change


coreyostrove:
The qibo devs pushed a new release to pypi on 9/3, and in this new release the seed kwarg has been removed from all of the Channel objects. There wasn't any explanation on the corresponding commit (SHA d182b9b if you're curious) on the rationale for this change, unfortunately.

As it stands the qibo evotype in pygsti won't work fully until we address this. I'm guessing we can either update the evotypes on our end to work around this or else update the requirements to use the previous release at the expense of whatever other new features/enhancements were included in this new one.

@enielse and @sserita, you both are the ones in the know on the qibo integration work, any thoughts on how to proceed?

sserita:
This has been currently avoided by limiting Qibo to version 0.1.7 in setup.py. Long-term, we probably want to reach out to the Qibo developers and understand why they made this change. It seems harder to get deterministic weak forward simulation now (since the RNG was important for sampling from operator "unravelings").

sserita:
We will likely deprecate the Qibo interface in the upcoming release, and remove it in a future release. Qibo is moving in a different direction (towards a full stack package rather than just a forward simulator) and is no longer a good fit for a third-party integration.

sserita:
Closed with #490.

PR #490, opened 2024-09-19
Feature qibo deprecation

sserita:
This PR addresses #262 by removing the deprecated Qibo evotype. This also removed the deprecated `pygsti.algorithms.directx` module, and a legacy `packages` folder that was for compatibility to pre-`pip install` installation methods.


coreyostrove (approved):
This all looks great, thanks for taking care of this!

PR #493, opened 2024-09-28
Maintenance of the Basis class

rileyjmurray:
This PR makes changes to the Basis class to help with my leakage PR (#410).

Specifics, at time of writing:
 * I split the mind-bogglingly-complicated Basis.cast function into four separate functions. Basis.cast dispatches the appropriate function just like before.
 * I removed the "classical_label" keyword argument from Basis.cast. It only appeared in a specific codepath, and it turns out that none of pyGSTI's existing code set this to anything other than its default value.
 
#### About EmbeddedBasis ...
It turns out that the EmbeddedBasis class isn't used anywhere in pyGSTi. (There are a few places that import it, but they still never use it.) I discovered this when I got an error trying to access ``eb.elements`` for an EmbeddedBasis ``eb``. That trigged evaluation of EmbeddedBasis._lazy_build_elements, which tried to call an undefined function. See [here](https://github.com/search?q=repo%3Asandialabs%2FpyGSTi%20reduce_dims_densitymx_to_state_inplace&type=code) for a repo-wide search for that function.

~I'd like to use this PR as an opportunity to either (a) make EmbeddedBasis work, possibly with new semantics, or (b) remove EmbeddedBasis altogether.~

After getting @enielse's blessing I've decided to remove EmbeddedBasis.


coreyostrove (approved):
Approved, great work!

pygsti/baseobjs/basis.py
```diff
@@ -1782,283 +1760,3 @@ def create_simple_equivalent(self, builtin_basis_name=None):
                 if all([c.name == first_comp_name for c in self.component_bases]):
                     builtin_basis_name = first_comp_name  # if all components have the same name
         return BuiltinBasis(builtin_basis_name, self.elsize, sparse=self.sparse)
-
-
-class EmbeddedBasis(LazyBasis):
```
sserita, line 1787: No action items, just a personal note that we are removing this so that I can include it in my spring cleaning section of the release notes.

pygsti/baseobjs/basis.py
```diff
@@ -154,126 +155,103 @@ class Basis(_NicelySerializable):
...
+        else:
+            return BuiltinBasis(name, dim, sparse)
+        
+    @classmethod
+    def cast_from_basis(cls, basis, dim=None, sparse=None):
+        #then just check to make sure consistent with `dim` & `sparse`
+        if dim is not None:
+            if isinstance(dim, _StateSpace):
+                state_space = dim
+                if hasattr(basis, 'state_space'):  # TODO - should *all* basis objects have a state_space?
```
sserita, line 209: Maybe. Is it just ExplicitBasis that does not have it now? ExplicitBasis implicitly does need all the information that a state_space has -- labels, consistent element size (i.e. dim), etc -- so we could make it an explicit requirement and pull state_space into the base class instead of just BuiltInBasis.

@rileyjmurray Can you spawn off an issue for us to discuss this?

pygsti/baseobjs/basis.py
```diff
@@ -154,126 +155,103 @@ class Basis(_NicelySerializable):
...
+            return cls.cast_from_name_and_dims(arg, dim, sparse)
+        if (arg is None) or (hasattr(arg,'__len__') and len(arg) == 0):
+            return ExplicitBasis([], [], "*Empty*", "Empty (0-element) basis", False, sparse)
+            # ^ The original implementation would return this value under two conditions.
+            #   Either arg was None, or isinstance(arg,(tuple,list,ndarray)) and len(arg) == 0.
+            #   We're just slightly relaxing the type requirement by using this check instead.
+        
+        # At this point, original behavior would check that arg is a tuple, list, or ndarray.
+        # Instead, we'll just require that arg[0] is well-defined. This is enough to discern
+        # between the two cases we can still support.
```
sserita, line 246: These comments are helpful while doing review, but will be increasingly less useful as time goes on and we forget what the "original" was doing. Thoughts on removing these comments (at minimum) or replacing them with something that is more self-contained and explain what the check is looking for and why we cast specific ways in that case (stretch)?

sserita (commented):
Thanks for this clean up @rileyjmurray, it's super helpful. I've made a few comments and will approve once you take a look at them, but there are no major changes being requested.

pygsti/baseobjs/basis.py
sserita: Discussion from dev meeting: Add create_state_space() to Basis and a compatibility function between Basis and StateSpace. Make .state_space into a private variable for BuiltInBasis and use the create_state_space() in places in the code where we want to use the state space from basis. create_state_space should take labels and tensor prod dims? Can be None for defaults.

@rileyjmurray I will split this out into a separate issue, consider this point handled. :)

sserita (approved):

PR #500, opened 2024-10-18
WIP: slimmed-down Levenberg-Marquardt for nonlinear least squares

rileyjmurray:
pyGSTi's nonlinear least squares solver is a custom implementation with Levenberg-Marquardt with a dizzying array of options. The core algorithm is over 800 lines long and has 10 levels of indentation at its deepest point. The complexity of this existing implementation makes it basically impossible to extend. 

This PR introduces slimmed-down infrastructure for nonlinear least squares. I started by copying customlm.py into a new file called simplerlm.py, then I removed one feature at a time in a series of commits. "Removal" of a feature meant making the default behavior for a given option the _only_ behavior. I haven't tested these changes yet but they should work just fine.

This PR has two goals. 
1. Have pyGSTi rely on simplerlm.py unless an option is requested that's only been requested that's only in customlm.py (in which scenario we dispatch the current code in customlm.py).
2. Formally deprecate the contents of customlm.py. I don't think we need a timeline for removal, but I would like to start raising deprecation warnings.

#### Notes

I tried to use ChatGPT to help split the monolithic custom_leastsq function in customlm.py into simpler functions. This turned out to not work for reasons that I'll discuss off-GitHub if people are interested.

For the curious, I'm doing this because I want to extend our nonlinear least squares solver in two ways. First, I want it to support pytorch Tensors; right now we only support numpy arrays and whatever home-cooked thing we have for distributed-memory computations. Second, I want to experiment with optimization algorithms that rarely (if ever!) require evaluating the full Jacobian of circuit outcome probabilities.

### DIFF
--- a/pygsti/optimize/__init__.py
+++ b/pygsti/optimize/__init__.py
@@ -12,6 +12,7 @@ pyGSTi Optimization Python Package
 
 from .arraysinterface import *
 from .customlm import *
+from .simplerlm import *
 from .customsolve import *
 # Import the most important/useful routines of each module into
 # the package namespace
--- a/pygsti/optimize/customlm.py
+++ b/pygsti/optimize/customlm.py
@@ -19,6 +19,7 @@ import scipy as _scipy
 
 from pygsti.optimize import arraysinterface as _ari
 from pygsti.optimize.customsolve import custom_solve as _custom_solve
+from pygsti.optimize.simplerlm import Optimizer, OptimizerResult
 from pygsti.baseobjs.verbosityprinter import VerbosityPrinter as _VerbosityPrinter
 from pygsti.baseobjs.resourceallocation import ResourceAllocation as _ResourceAllocation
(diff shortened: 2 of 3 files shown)


rileyjmurray:
Ping @enielse for review, when you get a chance. Now is a good time for us to document why options are worth keeping in the SimplerLMOptimizer class and the simplish_leastsq function. I'd be all for removing even more options if you think some are unnecessary.

rileyjmurray:
Notes for self:
Uphill may not be super important

Different damping nodes were needed for higher qubit GST. (e.g., 6-qubit).

Don’t deprecate CustomLM.

coreyostrove (approved):
Looks great! Thanks for the hard work, @rileyjmurray!

rileyjmurray:
Ping @sserita -- can you take a look at this? Once you approve we can merge this in.

sserita (approved):
Thanks for this @rileyjmurray!

rileyjmurray:
WOOO!

Issue #497, opened 2024-10-17
smq2Q_XY modelpack gates includes the CPhase gate


eendebakpt:
**Describe the bug**

The model pack `smq2Q_XY` includes the CPhase in the description and the gates, but it is not used in the germs or fiducials.

https://github.com/sandialabs/pyGSTi/blob/d27e4e688a64914a0b7aaec1de91aa56f7ce70c2/pygsti/modelpacks/smq2Q_XY.py#L20-L23

Also the `smq2Q_XYI` description contains a reference to the CPhase (and many some more modelpacks).

**Expected behavior**

I believe the `smq2Q_XY` should not contain any references to `CPhase`. And maybe the docstrings could be expanded to describe the model and its uses in more detail.

**Environment (please complete the following information):**
 - pyGSTi version [e.g. v0.9.7.4]: Current  master branch
 - python version [e.g. 3.7, 2.7]: Any
 - OS [e.g. OSX 10.14, Ubuntu 16.04 LTS]: Any

coreyostrove:
Closing as this has been addressed by #502, which has now been merged into develop.

PR #502, opened 2024-10-23
Fix docstrings and gate attribute

coreyostrove:
This PR fixes some mismatches between the documentation for some of the two-qubit modelpacks, as well as the gate list attribute, and their actual contents.

This issue was reported in #497. Thanks to @eendebakpt for bringing the issue to our attention!


sserita (approved):
Thanks for catching these!

Issue #464, opened 2024-07-10
Fisher Information w/ Experiment Designs Containing MCMs


pcwysoc:
It would be nice to be able to use Fisher information with experiment designs containing intermediate measurements, as part of our goal to get to feature parity between instruments and operations.

coreyostrove:
Closing as this has been merged in with #503.

PR #503, opened 2024-10-25
Fisher Information for Instruments

coreyostrove:
TLDR: This PR updates the implementation of the fisher information for experiment designs to support models/circuits with quantum instruments. This was requested ages ago in #464.

As part of updating the fisher information related code to support instruments this PR also changes the way that we do regularization. Previously this regularization was managed by adding a small amount of depolarizing error to the SPAM operations. Following a recommendation from @robinbk this has been changed to instead perform regularization by setting a minimum value for the probability p_min such that any probability below this threshold is clipped to this minimum value. As such, while in my testing the results looked nearly identical to previously calculated fisher information matrices for known experiment designs, the new code won't exactly reproduce the results of the older implementation.

### DIFF
--- a/pygsti/tools/edesigntools.py
+++ b/pygsti/tools/edesigntools.py
@@ -122,7 +122,7 @@ def calculate_edesign_estimated_runtime(edesign, gate_time_dict=None, gate_time_
     return total_circ_time + total_upload_time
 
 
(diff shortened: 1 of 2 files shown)


sserita (approved):
No changes needed, thanks for these updates. Also nice catch on the `cumulative` flag not being used properly...

PR #504, opened 2024-10-31
Fix invalid escape sequence warnings

eendebakpt:
We use raw docstrings to avoid some escape sequence warnings. Using double slashes is also an option, but makes the plain docstring a bit harder to read.


sserita (approved):
Thanks for fixing these 3.11 warnings :)

PR #511, opened 2024-12-10
egn-updates sync: wildcard and Hessians for non-ExplicitOpModels

enielse:
In this sync with Erik's ongoing back-burner branch, this PR primarily **adds support for computing Hessian-based  error bars and wildcard errors for non-`ExplicitOpModel` models.**

Specific updates relating to this include:
- models without gauge groups (e.g. non-`ExplicitOpModel`s) can now use `"none"` as a Hessian projection type without unnecessarily hitting an error from executing some `ExplicitOpModel`-specific code (`model.compute_nongauge_and_gauge_spaces`).
- framework now calls `_add_gaugeopt_and_badfit` even when the gauge optimization suite is empty or `None` when there are bad-fit options to process.
- adds if/else cases in several places along the code path for calling report/metric functions and ultimately computing wildcard error so that model members are properly accessed for `LocalNoiseModel`s as well as `ExplictOpModel` objects.  The code for the former may apply to more than just `LocalNoiseModel`s, but we should look at this in more detail.  This includes updates to "model functions", metric functions in `reportables.py`, and within the wildcard error computation code itself.

Other, less-related updates also included in this PR are:
- allowing `self.model` to be `None` in a `ModelDatasetCircuitsStore` object.  This facilitates constructing wildcard errors when you have predictions from a non-pyGSTi model (e.g. a time-dependent model simulated manually).
- minor bug fix in Hessian computation code related to the computation of `firsts` variables, which hold indices to the first outcome for a circuit that doesn't have data for all its possible outcomes.
- fixes a bug in `TreeNode` serialization due to json converting tuples -> lists when serializing & unserializing.  When a tuple is a key (e.g. for an experiment design node) then unserializing can cause an error because lists are mutable in Python and cannot be used as dictionary keys.  This PR adds a quick fix for this that patches this specific case.  A more general fix may be desired in the future.


sserita (approved):
Thanks @enielse!

PR #517, opened 2024-12-17
Fwdsim patches

coreyostrove:
This PR is a set of patches related to #496, which was just merged in. 

### Substantive changes:

- The casting behavior of `ForwardSimulator` has been updated so that the 'auto' keyword argument now defaults to the `MapForwardSimulator` in all cases (previously this was only true for qudit models or for two-or-more qubits).
- The germ selection algorithm now automatically converts to a `MatrixForwardSimulator` if needed.

### Unit Test Fixes

- Some unit tests on beta which started failing after #496 flagged the fact that there were a few tests in `test_packages` which were written assuming the default simulator was `MatrixForwardSimulator`. A whole bunch more broke for this reasone when I made the patch above changing the default casting behavior. Where appropriate I've updated these tests to be compatible with both simulators. In places where the `MatrixForwardSimulator` really was required (e.g. `dproduct` and `hproduct` methods were required for the thing being tested) I've made sure those tests have had the proper simulator specified explicitly.
- Spring cleaning: While doing these fixes I bumped into some tests which were making use of an old `Model` method which no longer exists. I went ahead and deleted these for the sake of cleaning things up. These tests were already being skipped or commented out, so there is no change to the tests being run.


rileyjmurray (approved):
Looks good to me!

sserita (approved):
All looks good, thanks @coreyostrove!

Issue #526, opened 2025-01-17
Reconcile/merge optools and reportables


coreyostrove:
At present we have a lot of code duplication between function in optools and functions in reportables, particularly with regards to distance metrics where we have a lot of near exact duplication. We also have some functionality that is in reportables and not in optools (e.g. instrument fidelity and instrument diamond distance), which makes this functionality harder to find. We should consider finding a better way of managing this overlapping code and reducing the amount of duplication/parity mismatch present.

sserita:
Thanks for making this issue, I was just thinking about this recently as well... My gut feeling is that I'd like most things to be in tools and the report code can use the tools to generate the tables as needed. But open to alternatives.

Issue #463, opened 2024-07-10
Wildcard w/ QILGST


pcwysoc:
See [WildcardQILGST.zip] (https://github.com/pyGSTio/pyGSTi/files/12667276/WildcardQILGST.zip)

sserita:
Closed with #527

Issue #273, opened 2022-11-11
Generalize diamond distance wildcard models


coreyostrove:
There is a new feature being develop that adds support for wildcard models where the weighting is based on the diamond distance between the target operations and the estimated ones. There is no reason this couldn't be generalized to other error metrics, so this feature would add support for some number of additional error metrics beyond diamond distance. It is also the case that cvxpy is sometimes unavailable on certain systems, so this would all use of this technique in those cases as well.

sserita:
The final commits of PR #267 laid the groundwork for this, although additional options have not yet been implemented.

PR #527, opened 2025-01-17
Wildcard Support for Instruments

coreyostrove:
This adds partial support for adding wildcard models to estimates containing quantum instruments. At the moment this only patches in support for 1D diamond distance wildcard models. This addresses issue #463.

An additional change is that a slightly different version of the instrument fidelity and instrument diamond distance functions in reportables have been copied over to optools for better discoverability. For a separate thread on eventually reconciling those two files see #526.

### DIFF
--- a/pygsti/protocols/gst.py
+++ b/pygsti/protocols/gst.py
@@ -2294,43 +2294,41 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
 def _compute_wildcard_budget_1d_model(estimate, objfn_cache, mdc_objfn, parameters, badfit_options, verbosity, gaugeopt_suite=None):
     """
     Create a wildcard budget for a model estimate. This version of the function produces a wildcard estimate
-    using the model introduced by Tim and Stefan in the RCSGST paper.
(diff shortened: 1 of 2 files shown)


pygsti/protocols/gst.py
```diff
@@ -2294,43 +2294,41 @@ def _add_badfit_estimates(results, base_estimate_label, badfit_options,
...
+        When gauge optimization is applied, this will ensure the gauge optimized models used are consistent 
+        in the wildcard model and the parent estimates.
 
     Returns
     -------
-    PrimitiveOpsWildcardBudget
+    PrimitiveOpsWildcardBudget or dict
+        The computed wildcard budget. If gauge optimization is applied, a dictionary of 
+        budgets keyed by gauge optimization labels is returned.
+
```
sserita, line 2331: As always, these docstring updates are very appreciated!

sserita (approved):
Straightforward addition. We may revisit this and clean it all up a bit when doing #273, but I see no reason to hold up this PR over any of those concerns.

PR #528, opened 2025-01-20
Bugfixes for GST with penalties in objective functions

rileyjmurray:
@coreyostrove reported bugs in fitting GST models where the objective function contains penalty terms. Quoting from an email:

> The functionality in question is the ability to add in modifications to the objective functions in the form of a penalty function. The one relevant to this discussion which we were trying to test out for gauge regularization purposes is an existing one set using the 'errorgen_penalty_factor' option. This dies somewhere in customsolve, as I recall. For the sake of testing whether the issue was isolated to this particular penalty function we also tried the built-in 'cptp_penalty_factor' and 'spam_penalty_factor' which also both failed (though in different ways for each option).

Running a notebook that Corey sent me, I was able to reproduce the error. I was also able to correct the error by adding input handling to vec_to_stdmx, casting non-numpy objects to ndarrays.

This PR sat stagnant for a year until Juan followed up with a request that I investigate how we penalize violation of complete positivity (of gates --- if we're talking SPAM then it's just violation of positivity). My first investigation into how we did things thought we were doing something crazy. It turns out that we were not. _HOWEVER_, what we were doing was awkward and unnecessarily complicated. This PR changes how we regularize to be decidedly _not awkward_ and to structure it in a way that cleanly isolates operations that apply to gates and SPAM.


rileyjmurray:
@pcwysoc you sent an email asking where the penalty fix(es) are. Here's the WIP PR!

juangmendoza19:
About a month ago, I ran into a use case that requires the use of the cptp_penalty argument for the objective functions. I noticed that the results coming out of GST were a bit strange.I did GST on experimental data for one qubit while varying the cptp_penalty in the following way:
```
cptp_penalties = np.logspace(1e-3,2,20)
for penalty in cptp_penalties:
	chi2_builder = pygsti.objectivefns.Chi2Function.builder(
		'chi2', regularization={'min_prob_clip_for_weighting': min_prob_clip}, penalties={'cptp_penalty_factor': penalty})
	mle_builder = pygsti.objectivefns.PoissonPicDeltaLogLFunction.builder(
		'logl', regularization={'min_prob_clip': min_prob_clip, 'radius': min_prob_clip}, penalties={'cptp_penalty_factor': penalty})
	iteration_builders = [chi2_builder] 
	final_builders = [mle_builder]
	builders = pygsti.protocols.GSTObjFnBuilders(iteration_builders, final_builders)
```
I ran this with a CPTPLND model (where the penalty should not matter, right?), and these are the results:
image
Here the blue line represents the sum of the negative choi eigenvalues for all the operations in the model, to quantify CPTPness.

As the penalty increases, the model found has a worse pygsti.tools.two_delta_logl(model, dataset).

This seems like unwanted behavior, as increasing the penalty in a scenario where they should not change the outcome of GST, it yields a worse model.

The change in likelihood above is small, but this unwanted behavior was first spotted in other scenarios like GLND model GST. For the same data, it yields:

image

In theory with an ideal optimizer I would expect the yellow line to never cross the red one, right? Beyond that, it looks like even after the penalty is high enough such that all the models are CPTP (the blue line is zero), the model gets worse by 7 units of delta log-L.

@rileyjmurray This is what we briefly talked about on a group meeting a couple of weeks ago.

rileyjmurray:
Curious. I'll investigate this week. Ordinarily I'd promise a fast turn-around for a nice report like this, but I'm on work travel M-Th.

rileyjmurray:
Thanks a million for reporting, Juan. It turns out that the implementation of CPTP penalty was ~~entirely unreasonable~~ very weird. Here it is:

https://github.com/sandialabs/pyGSTi/blob/37e7615aa60951da00d82ecd2f0d8f1116e87331/pygsti/objectivefns/objectivefns.py#L5797-L5810

Note that the penalty has little to do with CPTP violation. It's just the trace norm of the (unnormalized) Choi state (which, if a model is CPTP, is just equal to the dimension of the pure Hilbert space).

It's easy to fix this function. Fixing the Jacobian calculation may be harder. Will report back.

EDIT: We always have $\| \chi \|_ {\mathrm{Tr}} \geq \mathrm{Tr}(\chi)$. If we assume TP-ness and $\chi$ is normalized to trace-1, then $\| \chi \|_ {\mathrm{Tr}} - 1$ is nonnegative and twice the sum-of-magnitudes of negative eigenvalues of $\chi$. So the current metric is valid, just larger than it should be (ought to subtract 1 and then divide by 2).

pygsti/objectivefns/objectivefns.py
```diff
@@ -5805,162 +5834,104 @@ def _paramvec_norm_penalty(reg_factor, paramvec):
...
-        v = _np.tensordot(sgnchi, m_dgate_dp_std, ((0, 1), (1, 2)))
-        v *= prefactor * (0.5 / _np.sqrt(_tools.tracenorm(chi)))  # add 0.5/|chi|_Tr factor
-        assert(_np.linalg.norm(v.imag) < error_tol)
-        cp_penalty_vec_grad_to_fill[i, :] = 0.0
-
-        gate_gpinds_subset, within_wrtslice, within_gpinds = _slct.intersect_within(wrt_slice, gate.gpindices)
-        cp_penalty_vec_grad_to_fill[i, within_wrtslice] = v.real[within_gpinds]  # indexing w/array OR
-        #slice works as expected in this case
-        chi = sgnchi = dgate_dp = m_dgate_dp_std = v = None  # free mem
-
```
rileyjmurray, line 5871: The diff is correct that this content has been removed. It was only appropriate for a strange way of penalizing violation of complete positivity. This PR implements a better-motivated penalty for CP violation, and it does so in a more modular way.

pygsti/objectivefns/objectivefns.py
```diff
@@ -5805,162 +5834,104 @@ def _paramvec_norm_penalty(reg_factor, paramvec):
...
-        v = _np.tensordot(_np.tensordot(sgndm, ddenmx_dv, ((0, 1), (1, 2))), dv_dp, (0, 0))
-        v *= prefactor * (0.5 / _np.sqrt(_tools.tracenorm(denmx)))  # add 0.5/|denMx|_Tr factor
-        assert(_np.linalg.norm(v.imag) < error_tol)
-        spam_penalty_vec_grad_to_fill[i, :] = 0.0
-        gpinds_subset, within_wrtslice, within_gpinds = _slct.intersect_within(wrt_slice, prepvec.gpindices)
-        spam_penalty_vec_grad_to_fill[i, within_wrtslice] = v.real[within_gpinds]  # slice or array index works!
-        denmx = sgndm = dv_dp = v = None  # free mem
-
-    #Compute derivatives for effect terms
-    i = len(mdl.preps)
```
rileyjmurray, line 5924: The diff is correct that this code has been removed. It was needed for the old way of penalizing CP violation of state prep. The logic for the new type of penalty is contained in _spam_penalty_jac_fill (see below), in a much cleaner way.

pygsti/objectivefns/objectivefns.py
```diff
@@ -5805,162 +5834,104 @@ def _paramvec_norm_penalty(reg_factor, paramvec):
...
-            v = _np.tensordot(_np.tensordot(sgn_e, demx_dv, ((0, 1), (1, 2))), dv_dp, (0, 0))
-            v *= prefactor * (0.5 / _np.sqrt(_tools.tracenorm(emx)))  # add 0.5/|EMx|_Tr factor
-            assert(_np.linalg.norm(v.imag) < error_tol)
-
-            spam_penalty_vec_grad_to_fill[i, :] = 0.0
-            gpinds_subset, within_wrtslice, within_gpinds = _slct.intersect_within(wrt_slice, effectvec.gpindices)
-            spam_penalty_vec_grad_to_fill[i, within_wrtslice] = v.real[within_gpinds]
-            i += 1
-
-            sgn_e = dv_dp = v = None  # free mem
```
rileyjmurray, line 5960: This code has been replaced by 
```python
        for _, effect in povm.simplify_effects(povmlbl).items():
            handle_herm_modelmember(effect, count)
            count += 1
```
below. Using the new `handle_herm_modelmember` function removes logic that was duplicated for POVM effects and state prep.

pygsti/objectivefns/objectivefns.py
```diff
@@ -5805,162 +5834,104 @@ def _paramvec_norm_penalty(reg_factor, paramvec):
...
+        dmx_dp = _np.reshape(dmx_dp, (gate.num_params, mdl.dim, mdl.dim))
+        g      = _np.zeros((gate.num_params,))
+        for j, panel in enumerate(dmx_dp):
+            dJ_dpj = _tools.fast_jamiolkowski_iso_std(panel, op_basis, normalized=True)
+            val  = _np.vdot(dFdJ, dJ_dpj)
+            g[j] = val.real
+
+        # Step 3: embed into the row of the Jacobian.
+        _, within_wrtslice, within_gpinds = _slct.intersect_within(wrt_slice, gate.gpindices)
+        cp_penalty_vec_grad_to_fill[i, within_wrtslice] = prefactor * g[within_gpinds]
```
rileyjmurray, line 5894: This is the core of the revised `_cptp_penalty_jac_fill` function.

pygsti/objectivefns/objectivefns.py
```diff
@@ -5805,162 +5834,104 @@ def _paramvec_norm_penalty(reg_factor, paramvec):
...
+    U, v, _  = _tools.eigendecomposition(herm_mx, assume_hermitian=True)
+    selector = v < 0
+    num_neg  = _np.count_nonzero(selector)
+    if num_neg == 0:
+        return _np.empty((0,))
+    U =  U[:, selector].reshape((-1, num_neg))
+    G = -U @ U.T.conj()   # shape = (dim, dim)
+    if NEG_EIG_PENALTY_USE_SQRT:
+        G *= 0.5/_np.sqrt(NEG_EIG_PENALTY_SQRT_SHIFT + (_np.sum(-v[selector])))  # handles the sqrt() wrap.
+    return G
```
rileyjmurray: This is the body of a new function called ` _grad_of_neg_sum_neg_eigvals`. This new function is used in `_cptp_penalty_jac_fill` and `_spam_penalty_jac_fill`.

rileyjmurray (commented):
Just leaving comments to explain what's going on with the diff.

pygsti/objectivefns/objectivefns.py
```diff
@@ -5805,162 +5834,104 @@ def _paramvec_norm_penalty(reg_factor, paramvec):
     return out
 
 
-def _cptp_penalty_jac_fill(cp_penalty_vec_grad_to_fill, mdl, prefactor, op_basis, wrt_slice, error_tol=1e-4):
+def _grad_of_neg_sum_neg_eigvals(herm_mx: _np.ndarray) -> _np.ndarray:
```
coreyostrove, line 5843: Can you document that the expected input matrix in the context of computing the CPTP penalty are the choi matrices of each gate and the hermitian representations of the SPAM? I recognize it works for generic hermitian operators, but usage of this function not clear from the current documentation.

pygsti/tools/basistools.py
```diff
@@ -481,6 +481,10 @@ def vec_to_stdmx(v, basis, keep_complex=False):
     """
     if not isinstance(basis, _basis.Basis):
         basis = _basis.BuiltinBasis(basis, len(v))
+    if not isinstance(v, _np.ndarray):
+        v = v.to_dense()
+        # TODO: track down functions that hit this code path, then change
+        # then so this conversion isn't needed in vec_to_stdmx.
```
coreyostrove, line 487: Please create a tracking issue for this.

coreyostrove (approved):
No concerns on the technical end. Great work there. I have left two documentary requests for you. I'm approving now so that you can merge immediately upon adding these notes.

pygsti/objectivefns/objectivefns.py
rileyjmurray: Wrote a fuckton of documentation because I cannot control myself.

Will merge once tests pass.

pygsti/tools/basistools.py
rileyjmurray: done

PR #536, opened 2025-02-11
Fix DataSet.add_count_list() outcome label misordering

### LINKED ISSUE #535: Misordering of outcome labels with DataSet.add_count_list
**Describe the bug**
The outcome labels/counts are not properly ordered -- that is, the ordering of the outcome labels is ignored -- when adding counts via `add_count_lists`.

MWE notebook attached.

[02-11_DataSet-count-labels-bug.ipynb.zip](https://github.com/user-attachments/files/18757448/02-11_DataSet-count-labels-bug.ipynb.zip)

 - pyGSTi v0.9.13+
 - python 3.12
 - OSX 10.15

**Additional context**

sserita:
This fixes #535 by having `add_count_list` call `add_count_dict`, which handles the outcome label sorting properly. This has the advantage of minimizing code duplication before the `add_raw_series_data` call.

I have not migrated `add_count_arrays` because that uses `add_raw_arrays` instead, but in principle it can also be transition to use `add_count_dict` and therefore centralize all the `collisionAction == "aggregate"` code.

### DIFF
--- a/pygsti/data/dataset.py
+++ b/pygsti/data/dataset.py
@@ -1488,7 +1488,7 @@ class DataSet(_MongoSerializable):
         #unsafe=True OK b/c outcome_label_list contains the keys of an OutcomeLabelDict
 
     def add_count_list(self, circuit, outcome_labels, counts, record_zero_counts=True,
-                       aux=None, update_ol=True, unsafe=False):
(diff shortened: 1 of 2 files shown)


coreyostrove (approved):
Looks good to me!

PR #540, opened 2025-02-17
Bugfix Germ Selection and ProcessorSpec Documentation

coreyostrove:
This PR is a spin-off from #508. In that PR changes were introduced to address two issues. The first involved the line labels produced by the germ selection code, and the second some changes introduced to address an issue with instrument label handling in processor specs. Check out that PR for detailed discussion, but my conclusion was that in order to properly fix the instrument issue we'd need to confront adding in proper availability support for instruments in processor specifications. I haven't had time to actually start such a fix, but I don't want to let that make it so that we throw the baby (the bug fix for the germ selection as well as some bugfixes and documentation updates in the model construction routines) out with the bathwater (indefinite delays arising from everyone being crushingly overcommitted). So in this PR I've spun things off so we can at least make partial progress.

Included are:

- @pcwysoc 's bugfix for the qubit line labels in germ selection.
- Improved documentation for ProcessorSpec construction. Also documentation updates for some undocumented POVM model member constructor methods.
- Updates to model construction routines to patch in support for specification options we previously claimed (per docstrings) to support, but either didn't actually or it was broken. Unit tests added for coverage on these options.


pygsti/models/modelconstruction.py
```diff
@@ -903,7 +903,15 @@ def _spec_to_densevec(spec, is_prep):
                     num_qudits = len(qudit_labels)
                     if isinstance(spec, str):
                         if spec.isdigit():  # all([l in ('0', '1') for l in spec]): for qubits
-                            bydigit_index = effect_spec
+                            bydigit_index = spec
+                            assert (len(bydigit_index) == num_qudits), \
+                                "Wrong number of qudits in '%s': expected %d" % (spec, num_qudits)
+                            v = _np.zeros(state_space.udim)
+                            inc = _np.flip(_np.cumprod(list(reversed(processor_spec.qudit_udims[1:] + (1,)))))
```
sserita, line 910: Probably not worth changing since Instruments + Pspec + model construction needs a bit of an overhaul, but it's not immediately clear what this is doing. My best guess is pulling out the effect index in the case where we have MCMs and our POVM labels look something like ('0', '1') or something like that? A comment here would be super appreciated, but not PR blocking.

sserita (commented):
Two minor questions.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -42,10 +41,40 @@
...
+
+    Returns
+    -------
+    POVM
+        The constructed POVM object.
+    """
     povm_type_preferences = (povm_type,) if isinstance(povm_type, str) else povm_type
     if not isinstance(pure_vectors, dict):  # then assume it's a list of (key, value) pairs
-        pure_vectors = _collections.OrderedDict(pure_vectors)
+        pure_vectors = dict(pure_vectors)
```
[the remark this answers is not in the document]
coreyostrove: I don't think the order matters in this case

pygsti/models/modelconstruction.py
coreyostrove: I'm not 100% confident on this (it looks like I copied and pasted this code from existing POVM parsing logic since the effect logic was the same), but I think what this is doing is converting a string of digits (possibly corresponding to a mixed-base) integer into a base-10 integer. This is probably fragile against anything larger than a qunon.

coreyostrove: Comment added

sserita: Ty!

sserita (approved):

PR #545, opened 2025-02-21
Bugfixes for PGG FPR and Germ Selection

coreyostrove:
Three quick bugfixes:

1. Per-germ global FPR requires the matrix forward simulator for some subroutines, but the default forward simulator for models is now map. We now automatically make a copy of the model with the correct forward simulator as needed. Also for pgg FPR, fix to account for change to previously changed return value format when not passing in a precomputed germ spanning vector set. The same fix was also made to the `germ_set_spanning_vectors` function.
2. The change for germ selection line-labels in PR #540 introduced a bug caught by some annoyingly nondeterministic unit test failures. Germ selection returning None is a valid outcome when the initial candidate germ set is not amplificationally complete and we weren't checking for that case when doing the line label conversion. The test that caught this involved making random selections of the candidate set and wasn't seeded which is why it only failed some of the time. It is now seeded.

### DIFF
--- a/pygsti/algorithms/fiducialpairreduction.py
+++ b/pygsti/algorithms/fiducialpairreduction.py
@@ -19,22 +19,20 @@ import scipy.special as _spspecial
 import scipy.linalg as _sla
 
 from math import ceil
-import time
 
 from pygsti import baseobjs as _baseobjs
 from pygsti import circuits as _circuits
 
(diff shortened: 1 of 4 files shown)


sserita (approved):
This all looks good. Will merge as soon as I get the tests passing (with a develop merge).

Issue #546, opened 2025-02-24
BinaryRBDesign Not Fully Propagating Seed


coreyostrove:
**Describe the bug**
The construction of experiment designs for BinaryRBDesign is currently non-deterministic even when specifying a value for the seed. This looks to be because a seed isn't being propagated to `_sample_random_pauli` which means that the initial stabilizer state prep is still being randomized in an uncontrolled fashion.

**Expected behavior**
Design construction with seeding should be deterministically random for reproducibility.

**Environment (please complete the following information):**
 - pyGSTi version 0.9.13

sserita:
Closed with #547 merge.

PR #547, opened 2025-02-24
Add seeding for BiRB designs

coreyostrove:
This addresses issue #546 and adds in the rest of the plumbing needed for deterministically random BiRB experiment design construction. Previously even when specifying a seed this wasn't being propagated through to all of the different random sampling steps (or in some cases these sampling steps weren't configured to take a seed) meaning that the results were unexpectedly still non-deterministic (which is a problem where reproducibility guarantees are required.


sserita (approved):
Thanks for taking care of this. I'll note that this is using the new RNG interface instead of the old one that most of the codebase is using; however, we want to probably shift to this one, so I'll approve here and we'll migrate the others as part of dealing with #548 or a separate but related issue.

Issue #427, opened 2024-04-24
Replace OrderedDict with standard dictionaries where possible


coreyostrove:
In many parts of the codebase we opt to use OrderedDict instead of standard dictionaries in places where having dictionaries 'sorted' by insertion order is important. This is now mostly a vestige from pre-python 3.6/7 days when the standard python dictionary implementation did not remember insertion order, and made no guarantees on the order of returned elements from an iterated dictionary (the change in CPython was made in 3.6, and the announcement that this behavior was now part of the official language spec for all implementations was made with 3.7). This is no longer true, and so for most intents and purposes these two classes are now equivalent. I personally think that switching over to using standard dictionaries would improve readability in most places, but am open different perspectives. Thoughts?

sserita:
We no longer support versions of Python where dicts do not remember insertion order. I'm fine with this change!

enielse:
Yeah, I go back and forth on this.  I agree the readability is improved in that it's less clunky and cleaner syntax overall, but using OrderedDict _explicitly_ indicates that the ordering within the dictionary is important and we're relying on that, which isn't always that case, and maybe is typically not the case.  I could be convinced either way :)

sserita:
That's a fair counterpoint, the signaling of OrderedDict is much clearer that insertion order is important.Although I'll note that I'm not sure that distinction is immediately obvious to users. I was pair programming something yesterday, we got to an OrderedDict, and I could literally see their thought process settling on just using a regular dict instead.

But I can definitely see the value in signaling to developers that insertion order matters for that object, and not to manipulate it in careless ways... Hmm, maybe I'm changing my stance from pro this change to indecision on it

rileyjmurray:
I've come to see value in using ``OrderedDict`` instead of ``dict`` to flag that order preservation is important. So I could be convinced either way.

sserita:
I think my stance is now that I am for keeping `OrderedDict` in cases where the order preservation is specifically important. In my mind, this does come with a task of doing an audit of the codebase where we use `OrderedDict` and migrating to a `dict` when we *don't* care about insertion order to ensure this distinction is actually maintained. Probably this is scriptable somehow - if we are using methods that check the order explicitly, it's needed, or something like that.

But assuming that most places an `OrderedDict` was used pre-3.7 was a place where it was important, presumably most of them are legit uses of the class and so this is probably a relatively low priority thing to check.

coreyostrove:
[...]

PR #561, opened 2025-03-26
Change LGST Circuit Storage with FPR (Color Boxplot Fix)

coreyostrove:
This changes the way the LGST circuits are stored when using FPR. Previously these were stripped from the plaquettes and any needed circuits were added back in in `_additional_circuits`. Now we just don't remove those for L=1 if we shouldn't. This addresses a bug reported by @pcwysoc that was present in the color box plots where the result of the old behavior was that these LGST circuits weren't rendered with the plaquettes as expected.

### DIFF
--- a/pygsti/circuits/gstcircuits.py
+++ b/pygsti/circuits/gstcircuits.py
@@ -325,7 +325,8 @@ def create_lsgst_circuit_lists(op_label_src, prep_fiducials, meas_fiducials, ger
         indexing a string within prep_strs and effect_strs, respectively, so
         that prepStr = prep_strs[iPrepStr] and effectStr =
         effect_strs[iEffectStr].  If a dictionary, keys are germs (elements
-        of germ_list) and values are lists of 2-tuples specifying the pairs
+        of germ_list) or tuples of germs and length values
+        and values are lists of 2-tuples specifying the pairs
         to use for that germ.
 
     trunc_scheme : str, optional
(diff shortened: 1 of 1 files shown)


pygsti/circuits/gstcircuits.py
```diff
@@ -528,7 +527,7 @@ def add_to_plaquettes(pkey_dict, plaquette_dict, base_circuit, maxlen, germ, pow
     if nest:
         #keep track of running quantities used to build circuit structures
         running_plaquette_keys = {}  # base-circuit => (maxlength, germ) key for final plaquette dict
-        running_plaquettes = _collections.OrderedDict()  # keep consistent ordering in produced circuit list.
+        running_plaquettes = dict()  # keep consistent ordering in produced circuit list.
```
sserita, line 530: Just a general note that I've noticed several of these OrderedDict -> dict changes in recent PRs. It's not a problem, but just wanted to link it back to our discussion in #427 where we wanted to annotate dicts where order is explicitly important (e.g. where keys/values are iterated over with an expected order later).
I only bring this up here since "keeping consistent ordering" is part of the comment for this dict instantiation - is this one such place where order matters, and we should finalize our decision from #427?

sserita (commented):
One point of discussion.

coreyostrove:
Yup, good catch. I have been adding the annotations where relevant (which
has actually been pretty uncommon, it turns out), but forgot it seems. I'll
add these back in when I get a chance.

sserita (approved):

Issue #559, opened 2025-03-19
Figure out why "no cython" tests did not catch fastcalc import issue


sserita:
We recently had a case (#541) where one of the Cython extensions was not guarded properly, resulting in an ImportError when trying to use pyGSTi if the extensions were not available. Nominally this should have been caught by the beta No Cython tests but wasn't. We should figure out why this was the case, to ensure we catch these things in the future.

coreyostrove:
Completed with the merging in of #567.

Issue #566, opened 2025-04-01
Explicit flags to control Cython extension compilation


sserita:
**Is your feature request related to a problem? Please describe.**
We had some discussion in #541 of some downsides of the current try/except structure for building the Cython extensions, and there was some support for causing the pyGSTi installation to fail if the extensions failed to build.
A quick rundown of pros/cons:

Pros:
- The current workflow will attempt to build extensions if Cython is in the environment _whether or not Cython is being installed as part of the pyGSTi install_, i.e. with the extensions, testing, or complete optional dependencies. An environment with Cython in it already will automatically get the compiled extensions, if possible.
- The current workflow will fallback to no extensions and install a fully functional (albeit slow) pyGSTi; debatably, the Cython extensions are not needed for a complete installation (although in practice we really do want them if we are doing any sort of production forward simulation).

Cons:
- It may be unclear to the user when Cython extensions failed (we have a small but not super noticeable error message during install, and a suppressable warning on first import, but both can be missed).
- The try/except can eat the compiler error messages, making it difficult for the user to sometimes debug _why_ the compilation failed.

**Proposed solution**
The real problem is that the user cannot control whether the extensions are built or not on the command line, and the optional dependencies are not really the right mechanism to enforce this.
One option is to use something like [pip --install-options](https://stackoverflow.com/a/53833930) to pass a flag that allows the user to modify the behavior of the try/except. Maybe by default installation fails if there is Cython and extensions did not compile, and the user can pass a skip-extensions flag to get the current fallback behavior.

As a side note, this would also give us better controllability and maybe fix #559 on the runners by explicitly skipping the extension build.

coreyostrove:
Completed with the merging in of #567.

PR #567, opened 2025-04-01
CLI options for controlling Cython extension compliation

sserita:
This should fix both #566 and #559, giving the user better control over when Cython extensions are installed (and therefore us better control in our GitHub runners).

First, a clarification that took me a while to understand but is critical to understanding the issue in #559: whether or not pyGSTi's Cython extensions get compiled have nothing to do with which optional dependencies get installed, and everything to do with what is present in pip's isolated build system at install time. This, in turn, uses the `build-system.requires` section of `pyproject.toml`. Since `Cython` and `numpy` are included here, this means that pyGSTi will attempt to build Cython extensions **even if the user is not specifying `pygsti[extensions]` (or a superset, like `[testing]` or `[complete]`).**

It is not possible to make the `build-system.requires` field dynamic, so we instead always install Cython and provide a way for the user to turn this off via environment variables.

## Solution

During installation, pyGSTi will now check for two environment variables: `PYGSTI_CYTHON_SKIP` and `PYGSTI_CYTHON_EXCLUDE_FAILURES`. If the first is defined, then extension compilation will be skipped entirely. If the second is defined, then extension compilation will be performed with `exclude_failures=True` defined - this used to be the default but can suppress errors in the case where one of the extensions fails but the others succeed. (This is uncommon and is most likely to be useful to developers who are writing Cython extensions, but I opted for the more verbose failure behavior by default).

### Aside: Attempted PEP-517 Solution

In theory, it should be possible to do a pip-flag-based option via `--config-settings`. See an example package listed [here](https://discuss.python.org/t/explicit-optional-cythonization-via-config-setting-post-pep517-and-install-option-deprecation/25379) using this for optional Cythonization.

However, I could not figure out how to get this working and gave up after several hours of banging my head against it. The CLI flag seems to be working fine.

## New Behavior Examples

There is still a try-except block around the compilation. However, this now re-raises the error and I hope will prevent suppression of any error messages. We do still need it so that we can provide the user with the error message telling them about the new flags.

I'll use `CC=broken` to emulate a missing compiler that would trigger Cython compilation errors for testing. We can see that now we have our error message at the bottom.

```
$ CC=broken pip install -e .
[... MUCH ERROR MESSAGES ...]
          self._compile(obj, src, ext, cc_args, extra_postargs, pp_opts)
        File "/private/var/folders/xb/r8b8yj9d1kjgw1fr5x8bkls0003q1t/T/pip-build-env-5di3gmpi/overlay/lib/python3.11/site-packages/setuptools/_distutils/compilers/C/unix.py", line 223, in _compile
          raise CompileError(msg)
      distutils.compilers.C.errors.CompileError: command 'broken' failed: No such file or directory
      /private/var/folders/xb/r8b8yj9d1kjgw1fr5x8bkls0003q1t/T/pip-build-env-5di3gmpi/overlay/lib/python3.11/site-packages/setuptools/_distutils/dist.py:1021: _DebuggingTips: Problem in editable installation.
      !!
      
              ********************************************************************************
              An error happened while installing `pygsti` in editable mode.
      
              The following steps are recommended to help debug this problem:
      
              - Try to install the project normally, without using the editable mode.
                Does the error still persist?
                (If it does, try fixing the problem before attempting the editable mode).
              - If you are using binary extensions, make sure you have all OS-level
                dependencies installed (e.g. compilers, toolchains, binary libraries, ...).
              - Try the latest version of setuptools (maybe the error was already fixed).
              - If you (or your project dependencies) are using any setuptools extension
                or customization, make sure they support the editable mode.
      
              After following the steps above, if the problem still persists and
              you think this is related to how setuptools handles editable installations,
              please submit a reproducible example
              (see https://stackoverflow.com/help/minimal-reproducible-example) to:
      
                  https://github.com/pypa/setuptools/issues
      
              See https://setuptools.pypa.io/en/latest/userguide/development_mode.html for details.
              ********************************************************************************
      
      !!
        cmd_obj.run()
      
      An error occurred while compiling Cython extensions.
      Either fix the compilation issue or use the PYGSTI_CYTHON_SKIP to skip compilation,
      e.g. PYGSTI_CYTHON_SKIP=1 pip install pygsti
      To enable partial Cython failures (i.e. the exclude_failures=True flag of cythonize), use PYGSTI_CYTHON_EXCLUDE_FAILURES instead.
      
      error: command 'broken' failed: No such file or directory
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building editable for pygsti
Failed to build pygsti
ERROR: Failed to build installable wheels for some pyproject.toml based projects (pygsti)
```

Note that there is the full stack trace, an error message telling the user what environment variables should be used to avoid this issue, and then the actual relevant line of the error message.

Now with our newly introduced flag:

```
$ CC=broken PYGSTI_SKIP_CYTHON=1 pip install -e .
[...]
Successfully built pygsti
Installing collected packages: pygsti
  Attempting uninstall: pygsti
    Found existing installation: pygsti 0.9.13.1.post1.dev1+g86142fd1a.feature.extension.option.d20250401
    Uninstalling pygsti-0.9.13.1.post1.dev1+g86142fd1a.feature.extension.option.d20250401:
      Successfully uninstalled pygsti-0.9.13.1.post1.dev1+g86142fd1a.feature.extension.option.d20250401
Successfully installed pygsti-0.9.13.1.post1.dev3+g7c8b1afb5.feature.extension.option.clean
```

No error! And we can check none of the stuff got compiled, and that it does if we don't include the variable.

```
$ PYGSTI_SKIP_CYTHON=1 pip install -e .
[...]
Successfully installed pygsti-0.9.13.1.post1.dev3+g7c8b1afb5.feature.extension.option.clean
$ find . -name '*.so'
$ pip install -e .
[... same output as above since extension compilation is silent if no errors ...]
$ find . -name '*.so'
./pygsti/tools/fastcalc.cpython-311-darwin.so
./pygsti/forwardsims/termforwardsim_calc_stabilizer.cpython-311-darwin.so
./pygsti/forwardsims/mapforwardsim_calc_densitymx.cpython-311-darwin.so
./pygsti/forwardsims/termforwardsim_calc_statevec.cpython-311-darwin.so
./pygsti/circuits/circuitparser/fastcircuitparser.cpython-311-darwin.so
./pygsti/baseobjs/opcalc/fastopcalc.cpython-311-darwin.so
./pygsti/evotypes/statevec/statereps.cpython-311-darwin.so
./pygsti/evotypes/statevec/termreps.cpython-311-darwin.so
./pygsti/evotypes/statevec/effectreps.cpython-311-darwin.so
./pygsti/evotypes/statevec/opreps.cpython-311-darwin.so
./pygsti/evotypes/stabilizer/statereps.cpython-311-darwin.so
./pygsti/evotypes/stabilizer/termreps.cpython-311-darwin.so
./pygsti/evotypes/stabilizer/effectreps.cpython-311-darwin.so
./pygsti/evotypes/stabilizer/opreps.cpython-311-darwin.so
./pygsti/evotypes/densitymx/statereps.cpython-311-darwin.so
./pygsti/evotypes/densitymx/effectreps.cpython-311-darwin.so
./pygsti/evotypes/densitymx/opreps.cpython-311-darwin.so
./pygsti/evotypes/basereps_cython.cpython-311-darwin.so
```


.github/workflows/reuseable-main.yml
```diff
@@ -69,7 +69,7 @@ jobs:
     - name: Install package (No Cython)
       if: ${{ inputs.use-cython != 'true' }}
       run: |
-        python -m pip install -e .[testing_no_cython]
+        PYGSTI_CYTHON_SKIP=1 python -m pip install -e .[testing_no_cython]
```
sserita, line 72: This should fix #559

pygsti/layouts/prefixtable.py
```diff
@@ -1208,6 +1207,8 @@ def _draw_graph(G, node_label_key='label', edge_label_key='promotion_cost', figu
         An optional size specifier passed into the matplotlib figure
         constructor to set the plot size.
     """
+    import matplotlib.pyplot as plt
```
sserita, line 1210: Moved import that broke minimal install. Matplotlib is an optional dependency currently.

pyproject.toml
```diff
@@ -41,13 +41,13 @@ keywords=[
...
     "Development Status :: 4 - Beta",
     "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Python",
     "Topic :: Scientific/Engineering :: Physics",
     "Operating System :: Microsoft :: Windows",
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: Unix"
 ]
+license = "Apache-2.0"
```
sserita, line 50: Previous classifier was deprecated in favor of this.

setup.cfg
sserita, line 1: This file was unneeded - the [metadata] section is redundant with pyproject.toml, [bdist_wheel] for universal is actually deprecated with Python 2.7 EOL and would cause us issues in Aug 2025, and [nosetests] is out-of-date since we migrated to pytest.

setup.py
```diff
@@ -55,12 +54,219 @@ def build_extensions(self):
                 ext.extra_compile_args = args
         build_ext.build_extensions(self)
 
+# Check if environment can try to build extensions
```
sserita, line 57: Most of this is just moving stuff around, I'll note the critical changes.

setup.py
```diff
@@ -55,12 +54,219 @@ def build_extensions(self):
...
+            ),
+            Extension(
+                "pygsti.circuits.circuitparser.fastcircuitparser",
+                sources=["pygsti/circuits/circuitparser/fastcircuitparser.pyx"],
+                include_dirs=['.', np.get_include()],
+                language="c++",
+                extra_link_args=["-std=c++11"]
+            )
+        ]
+        extensions = cythonize(ext_modules, compiler_directives={'language_level': "3"}, exclude_failures="PYGSTI_CYTHON_EXCLUDE_FAILURES" in os.environ)
```
sserita, line 260: Here we went from exclude_failures=True to a check against `PYGSTI_CYTHON_EXCLUDE_FAILURES`

setup.py
```diff
@@ -55,12 +54,219 @@ def build_extensions(self):
...
+except ImportError:
+    warn("Extensions build tools are not available. Installing without Cython extensions...")
+    extensions = None
```
sserita, line 263: We still don't build extensions if no Cython. However, as mentioned in my main message, **this should basically never trigger** because it has to do with the build-system requirements, not the user-specified optional dependencies.

setup.py
```diff
@@ -146,213 +352,9 @@ def setup_with_extensions(extensions=None):
...
-    setup_with_extensions(cythonize(ext_modules, compiler_directives={'language_level': "3"}, exclude_failures=True))
-except ImportError:
-    # Cython or numpy is not available
-    warn("Extensions build tools are not available. Installing without Cython extensions...")
-    setup_with_extensions()
-except SystemExit:
-    # Extension compilation failed
-    warn("Error in extension compilation. Installing without Cython extensions...")
-    setup_with_extensions()
+except SystemExit as e:
```
sserita, line 355: We still have this try-except block so that we can tell the user about the env variables. However, we now re-raise the exception.

coreyostrove (approved):
Awesome work figuring this all out. I don't have any questions or concerns about any of these changes. Thanks, @sserita!

Issue #569, opened 2025-04-08
Target Eigenvalues Overwritten with Estimated in Germs Detail


pcwysoc:
**Describe the bug**
In the HTML report, the target eigenvalues are overwritten with the estimated eigenvalues in 'Germs Detail'. 

<img width="537" alt="Image" src="https://github.com/user-attachments/assets/a60333d7-0953-43d7-96bb-6edb79395d5d" />

**Environment (please complete the following information):**
 - pyGSTi version 'master'

coreyostrove:
Fixed in #572.

Issue #426, opened 2024-04-23
HTML Report Requests


pcwysoc:
Responding to @coreyostrove's call for HTML report requests. 

- [ ] **Quantum instrument scale**: Quantum instruments usually have a maximum element value of 0.5, but this varies. Currently, all process matrices (including QIs) are scaled the same which impacts visibility. The plot of the quantum instrument elements should take the maximum value in any of the estimates and Target to set the colorbar. However, this should be symmetric and the scale should be the same for all instrument elements. 
- [ ] **Scale of confidence regions for raw estimates**: It would be helpful to a) be able to find out the errorbars on a particular process matrix element by hovering over it (like is done for error generators currently) and b) be able to tell the scale of the errorbars at a glance (perhaps adding a colorbar). 
- [ ] **Consistent sizing of plots**: the auto-scaling to the size of figure size with the window means that some plots are tiny and some overlap. I know this could be hard to fix. 
- [ ] **Juan's point**: about adding something like "click for more info" on the titles of plots 
- [ ] **Predicted RB number in Gauge Invariant Estimates**: this never calculates correctly 
- [ ] **Downloading a plot as png**: rarely works, usually gives an empty plot 
- [ ] **The colorscale on raw estimates**: often we get such good results now that it can be hard to visually see how the estimates differ from the target. Maybe a log scale or something? 

I know some of these are probably more fixable than others, but I wanted to report everything just in case!

coreyostrove:
> Juan's point: about adding something like "click for more info" on the titles of plots

This was added by @sserita in 0.9.12.2, so this can be checked off.

coreyostrove:
Question: 

> Predicted RB number in Gauge Invariant Estimates: this never calculates correctly

By this do you mean that it doesn't show up, or that the value returned is wrong?

pcwysoc:
The returned value is -1 always.

pcwysoc:
Added another point about report generation.

pcwysoc:
@kmrudin and I added another 10 points.

coreyostrove:
![oinqy_scream](https://github.com/sandialabs/pyGSTi/assets/78768318/371367b6-7a61-43d0-89bb-6deb6c6db30c)

(Joking aside, thanks for the detailed updates!)

coreyostrove:
>Predicted RB number in Gauge Invariant Estimates: this never calculates correctly
> The returned value is -1 always.

[...]

PR #572, opened 2025-04-18
HTML Report Updates

coreyostrove:
This PR introduces a number of changes to the display, functionality and performance of HTML report generation.

[standard_report_high_depth.zip](https://github.com/user-attachments/files/19813859/standard_report_high_depth.zip)
[instruments_and_CIs.zip](https://github.com/user-attachments/files/19813860/instruments_and_CIs.zip)

# Visualization Changes (Existing Figures)
I will describe the changes briefly below, but it is likely most informative to see these in-situ so I have attached two demo 1Q GST reports, one with a really high maximum depth and another containing both instruments and error bars.

1. Updated per-sequence detail plots: Also called the color box plots, these have been reworked for clarity and with some additional features. It was sometimes hard to delineate between boxes of similar colors so thin outlines have been added to each box. It was also often tricky in the context of FPR to visually place a square within its prep-fid/meas-fid row-column pair by eye for very sparse plaquettes, so there is now a toggleable overlay which adds a grid to the help guide the eye. Sizing on squares has been made more consistent and spacings have been adjusted in accordance with this. One big request we had was to address difficulties with the hover box information being challenging to use due to it often obscuring underlying boxes. The behavior of the hover boxes has been adjusted to more consistently stay off to the right or left of the object being introspected, so this should hopefully reduce the frequency of those issues. The formatting of text in the hover boxes has also been tab aligned for readability. Another big request was the addition of a new interface for accessing the per-sequence detail information where one could click on a box and have the contents of the hover information printed to the side of the figure, and this has been added as well. Both the hover boxes and the click interface can be toggled on/off individually (live) in case either is not wanted for whatever reason.
2. Delineation between C and A error generators: There are now triangles marking which triangle is which.
3. Less funky SPAM and Choi states: The formatting the raw estimates for the SPAM has always been confusing, since it looks like it should be a density matrix, and it was, but written in a strange way where the upper and lower triangular parts of the plotted array were used for the real and imaginary parts of the off diagonals (except this wasn't ever explained anywhere). The new version splits these off into two arrays and plots the real and imaginary parts separately (and labels them). The same issue applied to choi states that were being reported so this change was made there too.
4. Error generator sector percentage: The text that used to accompany the error generator projections has been made more understandable. The original information is still there, but now with the clarification that it is frobenius norm, and a new line has been added with the percentage contribution to the generator infidelity (which is closer to what people have always wanted this percentage to mean).
5. Error bars on raw estimates for gates now shows up both in the hover info and in a separate column with the CI half widths. This was a request from some users for better visibility when errors are very small.
6. Error bars on raw estimates for SPAM now also show up as hover info.

# New Stuff
Here are some new additions to the reports.

1. There is a new plot on the model violation overview tab which gives a histogram with the distribution of the per-circuit TVDs, per a recent request from @omaupin.
2. Raw estimates for the SPAM operations now include the target and estimated SPAM operations as Hilbert-Schmidt vectors.
3. Generator infidelity has been added as an error metric in a few of the tables currently containing quality metrics.
 
# Performance Changes
100% honesty, report generation will often be at least a bit slower on the whole due to the changes on this branch. A lot of that has to do with the fact that we're now drawing a *lot* more geometry for the per-sequence details. There are some changes to the report generation that speed certain subroutines up and ameliorate this increased time cost quite a bit, though. The biggest change is completing some plumbing for reusing MDC store objects that are generated in the course of GST's iterative optimization to accelerate the construction of certain plots which rely on them (and previously were rebuilding new ones from scratch).  If my commit messages from a year ago are to be trusted then for smaller maximum circuit depths the speedup is rather modest, but for a maximum depth of 2048 with one-qubit GST this change sped up report generation by ~40%.

# Bugfixes

1. Fix for #569, which reported that on the second table of the germs detail tab the target eigenvalues were being pulled in from the wrong model. The underlying bug was just that the wrong model was being passed in from the switchboard, and it turns out the other table on this page had the exact same issue.
2. The predicted RB number table that had been previous disabled (years ago) is now enabled again and the code it points to has been updated to the current home of that RB theory code. This issue was originally report in #426.


README.md
```diff
@@ -197,3 +197,29 @@ Questions?
 ----------
 For help and support with pyGSTi, please contact the authors at
 pygsti@sandia.gov.
+
```
coreyostrove, line 200: I forgot I added this, I think I did so in response to a request from @dhothem. We should probably also include the citation for Zenodo.

pygsti/report/fonts/NotoSansMono-Regular.ttf
coreyostrove, line 1: For future reference. https://fonts.google.com/noto/specimen/Noto+Sans/license

pygsti/report/fonts/OpenSans-Regular.ttf
coreyostrove, line 1: For future reference. https://fonts.google.com/specimen/Open+Sans/license

pygsti/report/plotly_plot_ex.py
```diff
@@ -118,33 +113,20 @@ def plot_ex(figure_or_data, show_link=True, link_text='Export to plot.ly',
...
-        if plotly_version >= (3, 7, 0):  # then auto_play arg exists
-            kwargs['auto_play'] = False
-
-        #Note: removing width and height from layout above causes default values to
-        # be used (the '100%'s hardcoded below) which subsequently trigger adding a resize script.
-        plot_html, plotdivid, _, _ = _plot_html(
-            fig, config, validate, '100%', '100%',
-            global_requirejs=False,  # no need for global_requirejs here
-            **kwargs)  # since we now extract js and remake full script.
-    else:
```
coreyostrove, line 136: This change effectively kills support for plotly <3.8.0, so the minimum version requirement should be updated.

pygsti/report/plotly_plot_ex.py
```diff
@@ -103,11 +103,6 @@ def plot_ex(figure_or_data, show_link=True, link_text='Export to plot.ly',
         aspect_ratio = orig_width / orig_height
     else: aspect_ratio = None
 
-    #Remove original dimensions of plot so default of 100% is used below
-    # (and triggers resize-script creation)
-    if orig_width: del fig['layout']['width']
-    if orig_height: del fig['layout']['height']
-
```
coreyostrove, line 110: This change is related to the fix I found for the figure sizing issues I discussed live in the new per-sequence detail plots.

pygsti/report/workspace.py
```diff
@@ -2461,6 +2460,9 @@ def __init__(self, ws, fn, *args):
             The arguments to `fn`.
         """
         super(WorkspacePlot, self).__init__(ws)
+        #add a plot specific option for aspect ratios
+        self.options['lock_aspect_ratio'] = True
```
coreyostrove, line 2464: Adding a handle for this option was the second half of the fix for the sizing issue I reported for per-sequence detail plots.

pygsti/report/reportables.py
```diff
@@ -2218,9 +2259,7 @@ def predicted_rb_number(model_a, model_b):
     -------
     float
     """
-    #TEMPORARILY disabled b/c RB analysis is broken
-    #from ..extras.rb import theory as _rbtheory
-    return -1.0  # _rbtheory.predicted_rb_number(model_a, model_b)
+    return _tools.rbtheory.predicted_rb_number(model_a, model_b)
```
coreyostrove, line 2262: Ran this change past @tjproct and also asked him to take a look at the rest of the code for the predicted RB number table's construction and no immediate concerns were raised.

pcwysoc:
Hi Corey,

Excited to see these features getting close to the finish line! Here's a provisional list of feedback:

- I think the new TVD histogram should be blue, in order to provide greater contrast and maintain a similar look as other TVD figures.
- In the raw estimates, the instrument error bars are asymmetric as they are calculated using deriv_wrt_params which due to the parameterization of TPInstrument gives overly conservative error bars. I believe the correct way to do this is either using finite difference or calculating them like a reportable quantity using the confidence regions intervals. 
- Relatedly, this pull requests does not include some features for QI that we had at one point worked to incorporate into a similar branch (feature-QI-HTML-integration). These features included rescaling the color bar for quantum instruments in Raw Estimates and adding QIs to the Choi representation table in Gate Decompositions. 
- Is the "Poisson bump" correction implemented? This correction accounts for Wilks' Theorem breaking down for small numbers of counts. The code I wrote also accounts for differing degrees of freedom for different circuits (which often crops up with experiments that include MCMs). If not, I think we should consider including those features in this pull request. This will also require thinking about how to handle the loglikelihood histogram as the expected chi^2 distribution will be different. 
- You may wish to wrap the text for the error generator metrics. They get cut off on small screens. 
- Table 6 in Gauge Invariant Error Metrics does not have error bars for the quantum instrument. 

Overall, it looks great and we shouldn't let the perfect be the enemy of the good with regards to implementation.

coreyostrove:
Thanks for the feedback, @pcwysoc!


> 
>     * I think the new TVD histogram should be blue, in order to provide greater contrast and maintain a similar look as other TVD figures.
> 

That should be simple enough to change.

>     * In the raw estimates, the instrument error bars are asymmetric as they are calculated using deriv_wrt_params which due to the parameterization of TPInstrument gives overly conservative error bars. I believe the correct way to do this is either using finite difference or calculating them like a reportable quantity using the confidence regions intervals.
> 

Is this an HTML report issue, or is it one that goes deeper?

>     * Relatedly, this pull requests does not include some features for QI that we had at one point worked to incorporate into a similar branch (feature-QI-HTML-integration). These features included rescaling the color bar for quantum instruments in Raw Estimates and adding QIs to the Choi representation table in Gate Decompositions.
> 
>     * Is the "Poisson bump" correction implemented? This correction accounts for Wilks' Theorem breaking down for small numbers of counts. The code I wrote also accounts for differing degrees of freedom for different circuits (which often crops up with experiments that include MCMs). If not, I think we should consider including those features in this pull request. This will also require thinking about how to handle the loglikelihood histogram as the expected chi^2 distribution will be different.
> 

You're correct that a number of the instrument-related updates are on the aforementioned feature-QI-HTML-integration branch and not represented here. I haven't forgotten about those, but there are some instrument related plots that are currently broken on that branch (some issues related to error bars) which will take some effort to get fixed. I don't have a timeline on that, so decided to try and get the changes that were ready integrated in the meantime.

>     * You may wish to wrap the text for the error generator metrics. They get cut off on small screens.

Could you share a screenshot?

> 
>     * Table 6 in Gauge Invariant Error Metrics does not have error bars for the quantum instrument.
> 

This is also on my to-do list for whenever the feature-QI-HTML-integration branch gets merged in. This was listed in one of the open issue threads so I'll make sure that line item doesn't get prematurely deleted so we can track this.

pcwysoc:
@coreyostrove here's some answers to your questions: 

The blocking problem on the feature-QI-HTML branch is one and the same as the Raw Estimates error bar issue, which is as follows. Error bars on the quantum instrument process matrices are being calculated by falling back on using 'deriv_wrt_params,' since we don't know how to reshape the interval matrix. TPInstruments are parameterized as Q0 (fully parameterized) and Q0+Q1 (full TP parameterization). The error bars on Q1 are then effectively calculated summing the error bars on the Q0 and Q0+Q1 parameters. Consequently, the error bars on the second process matrix will spuriously be roughly twice those of the first process matrix as the positive covariance between the Q1 and Q0+Q1 parameters is ignored.

One way to avoid this issue is to calculate the error bars directly from the confidence region interval as one might a reportable quantity. It may also be possible to work out how the interval matrix should reshaped, which would be the optimal fix. 

Here's a screenshot of the error generator metrics issue: 

Screenshot 2025-04-24 at 2 17 23 PM

coreyostrove:
Pinging @pcwysoc and @rileyjmurray for re-review. I just pushed what I hope to be the final round of updates for this particular PR. The QI-related fixes and requests from @pcwysoc that aren't currently reflected on this branch will need to be addressed when we get ` feature-QI-HTML-integration` merged in. Regarding the other two pieces of feedback/requests:

> * I think the new TVD histogram should be blue, in order to provide greater contrast and maintain a similar look as other TVD figures.

Done. You'll likely be happy to know that this innocent sounding request not only led to me discovering and patching a bug in the colormap code, but also ultimately to a ~50% speedup in report generation (for the 1Q L=2048 test case I was using, at least) as a result of stumbling on some unexpected behavior in the course of trying to track down that bug. See the most recent commits on this branch if you're curious.

> * You may wish to wrap the text for the error generator metrics. They get cut off on small screens.

Text wrapping was behaving badly, but I did at least abbreviate these headings to ameliorate this a bit.

README.md
rileyjmurray: I don't know about Zenodo, but here's a reference that's tracked by Google Scholar.
```
@misc{ doecode_28250,
title = {Python GST Implementation (PyGSTi) v. 0.9},
author = {Nielsen, Erik and Blume-Kohout, Robin J. and Rudinger, Kenneth M. and Proctor, Timothy J. and Saldyt, Lucas and Young, Christopher and Gamble IV, John and Kelly, Robert and Russo, Antonio and Hothem, Daniel and Seritan, Stefan and Ostrove, Corey and Murray, Riley and Hasting Hines, Jordan},
abstractNote = {PyGSTi is a Python package for modeling and characterizing noise (errors) in small quantum information processors. In addition to being a basic framework for describing quantum circuits and noise models, it implements mainstream quantum characterization, verification, and validation (QCVV) protocols such as Gate Set Tomography (GST), Randomized Benchmarking (RB), Robust Phase Estimation, and Idle Tomography. It also implements prototype protocols used for timeseries analysis and crosstalk detection, all of which have the goal of better understanding the noise found in existing as-built experimental devices. The central protocol of pyGSTi (from where it derives its name) is Gate Set Tomography. GST is a theory and protocol for simultaneously estimating the state preparation, gate operations, and measurement effects of a physical system of one or many quantum bits (qubits). These estimates are based entirely on the statistics of experimental measurements, and their interpretation and analysis can provide a detailed understanding of the types of errors/imperfections in the physical system. In this way, GST provides not only a means of certifying the “goodness” of qubits but also a means of debugging (i.e. improving) them. The other protocols follow this similar pattern in that they use statistical inference and analysis of experimental data to estimate one or more properties of the noise in a device.},
doi = {10.11578/dc.20190722.2},
url = {https://doi.org/10.11578/dc.20190722.2},
howpublished = {[Computer Software] \url{https://doi.org/10.11578/dc.20190722.2}},
year = {2019},
month = {jul}
}
```

pygsti/algorithms/core.py
```diff
@@ -954,11 +954,13 @@ def _max_array_types(artypes_list):  # get the maximum number of each array type
...
-
-                yield (mdc_store.model, opt_result, final_objfn)
+                yield (mdc_store.model, opt_result, mdc_store, final_objfn)
             else:
-                #If not the final iteration then only send back a copy of the model and the optimizer results
-                yield (mdc_store.model.copy(), opt_result)
+                #If not the final iteration then send back a copy of the model and the optimizer results
+                #mdc_store gets re-initialized at the start of each circuit list iteration, and doesn't appear
+                #to get propagated beyond that point so sending it back directly without copying (which would
+                #probably require implementing a custom method for MDC store objects) should be fairly safe.
```
rileyjmurray, line 962: I don't follow what this is saying. Is it explaining why we send a _copy_ of mdc_store.model, rather than mdc_store.model itself?

pygsti/report/factory.py
```diff
@@ -367,6 +378,10 @@ def _create_master_switchboard(ws, results_dict, confidence_level,
 
             switchBd.mdl_target[d, i] = est.models['target']
             switchBd.mdl_gaugeinv[d, i] = est.models[GIRepLbl]
+            
+            switchBd.mdl_final_modvi[d, i]= est.models['final iteration estimate']
+            #print(est.models['final iteration estimate'])
```
rileyjmurray: Remove this commented-out code

pygsti/tools/optools.py
```diff
@@ -662,6 +662,50 @@ def entanglement_infidelity(a, b, mx_basis='pp', is_tp=None, is_unitary=None):
...
+    gen_infid = 0
+    for coeff_block in errgen.coefficient_blocks:
+        if coeff_block._block_type == 'ham': #H terms, get squared
+            gen_infid+= _np.sum(coeff_block.block_data**2)
+        if coeff_block._block_type == 'other_diagonal': #S terms, added directly
+            gen_infid+= _np.sum(coeff_block.block_data)
+        if coeff_block._block_type == 'other': #S terms on diagonal, added directly
+            gen_infid+= _np.sum(_np.diag(coeff_block.block_data))
+
+    return _np.real_if_close(gen_infid)
```
rileyjmurray: The error_generator function can raise exceptions under a range of scenarios. So this should be wrapped in a try-catch, with a catch that returns NaN.

pygsti/algorithms/core.py
coreyostrove: If I recall correctly this is explaining why, despite looking unsafe, it is safe in this instance to avoid doing a copy operation on mdc_store.model and mdc_store. These copy operations would likely be very expensive, so if we can avoid them it is a performance benefit.

README.md
coreyostrove: That might be better. I'd want to double-check how DOE Code is grabbing the author list, though. If it is through our internal CA process it might not be as flexible as we'd like.

pygsti/tools/optools.py
coreyostrove: Done

rileyjmurray (approved):

Issue #579, opened 2025-05-15
Non-Deterministic FOGI Models


juangmendoza19:
**Describe the bug**
FOGI models created from the same code can have discrepancies in the order of their parameters

**To Reproduce**
on the branch "develop" create your favorite FOGI model and print one of the labels. Re-run this multiple times (restart the kernel if you are running this on Jupyter notebooks), and within a couple of tries you should see the label change. Here is some sample code:

import pygsti
from pygsti.modelpacks import smq1Q_XY as std

fogi_model1 = std.target_model("GLND")

basis1q = pygsti.baseobjs.Basis.cast('pp', 4)

gauge_basis = pygsti.baseobjs.CompleteElementaryErrorgenBasis(
            basis1q, fogi_model1.state_space, elementary_errorgen_types='HSCA')

fogi_model1.setup_fogi(gauge_basis, None, op_abbrevs, include_spam=True, reparameterize=True)

print(fogi_model1.parameter_labels[1])

**Expected behavior**
It would be much more convenient to have some canonical FOGI model come out of pyGSTi.

**Environment (please complete the following information):**
 - pyGSTi version 0.9.13.1 (develop)
 - python version 3.10.16
 - OS macOS Sequoia 15.4.1

**Additional context**
This problem emerged after commit 3eda8c0 where some set algebra functions inside errgenbasis.py were modified

coreyostrove:
Fixed with the merging in of #580. Thanks for the fixes, @juangmendoza19.

PR #580, opened 2025-05-16
Bugfix deterministic fogi

juangmendoza19:
Resolves issue #579

Makes sure that some sets utilized within setup_fogi get sorted before they re utilized, resulting canonical/deterministic FOGI models

### DIFF
--- a/pygsti/baseobjs/errorgenbasis.py
+++ b/pygsti/baseobjs/errorgenbasis.py
@@ -262,7 +262,7 @@ class ExplicitElementaryErrorgenBasis(ElementaryErrorgenBasis):
         #Get the union of the two bases labels.
         union_labels = set(self._labels) | set(other_basis.labels)
         union_state_space = self.state_space.union(other_basis.state_space)
-        return ExplicitElementaryErrorgenBasis(union_state_space, union_labels, self._basis_1q)
+        return ExplicitElementaryErrorgenBasis(union_state_space, sorted(union_labels, key=lambda label: label.__str__()), self._basis_1q)
 
     def intersection(self, other_basis):
         """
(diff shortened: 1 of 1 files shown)


pygsti/baseobjs/errorgenbasis.py
```diff
@@ -262,7 +263,7 @@ def union(self, other_basis):
         #Get the union of the two bases labels.
         union_labels = set(self._labels) | set(other_basis.labels)
         union_state_space = self.state_space.union(other_basis.state_space)
-        return ExplicitElementaryErrorgenBasis(union_state_space, union_labels, self._basis_1q)
+        return ExplicitElementaryErrorgenBasis(union_state_space, sorted(union_labels, key=lambda label: label.__str__()), self._basis_1q)
```
juangmendoza19, line 266: Added lexicographic sorting to lists created from python sets

pygsti/baseobjs/errorgenbasis.py
```diff
@@ -232,6 +232,7 @@ def overlaps(sslbls):
             sub_labels, sub_indices = zip(*[(lbl, i) for i, lbl in enumerate(self._labels)
                                             if overlaps(lbl[0])])
             sub_state_space = self.state_space.create_subspace(sub_sslbls)
+            sub_sslbls = sorted(sub_sslbls)
```
coreyostrove, line 235: This sorting should happen before the preceding line in order to have an effect.

coreyostrove (changes requested):
One single line change requested.

pygsti/baseobjs/errorgenbasis.py
juangmendoza19: Good catch

coreyostrove (approved):
Looks great. Thanks for taking care of these changes!

PR #584, opened 2025-06-03
Allow state vectors of any size

### LINKED ISSUE #583: Creating a n-dimensional state vector space for nonperfect square n
I am opening this issue now, I will fix it next week.

On version '0.9.13.1', I am getting an error when I try to make a basis for a two dimensional state vector space.

```python
from pygsti.baseobjs.basis import BuiltinBasis
b = BuiltinBasis("sv", 2)
```

Suggests that the dimension needs to be a perfect square, but one should be able to make a state vector of any dimension.

Here is the error message.
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/name/pg-misc/pygsti-repo/pygsti/baseobjs/basis.py", line 1071, in __init__
    else _statespace.default_space_for_dim(dim_or_statespace)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/name/pg-misc/pygsti-repo/pygsti/baseobjs/statespace.py", line 1296, in default_space_for_dim
    assert(udim**2 == dim), "`dim` must be a perfect square: %d is not" % dim
           ^^^^^^^^^^^^^^
AssertionError: `dim` must be a perfect square: 2 is not
```

For a four dimensional state vector space we see that,
```python
print(BuiltinBasis("sv", 4))
```
returns
```
State-vector basis (dim=4), 4 elements of shape (4,) :
|0>, |1>, |2>, |3>
```

nkoskelo:
Fixes #583

### DIFF
--- a/pygsti/baseobjs/basis.py
+++ b/pygsti/baseobjs/basis.py
@@ -14,7 +14,7 @@ import copy as _copy
 import itertools as _itertools
 import warnings as _warnings
 from functools import lru_cache
-from typing import Union, Tuple, List
+from typing import Union
 
 import numpy as _np
 import scipy.sparse as _sps
@@ -1069,8 +1069,7 @@ class BuiltinBasis(LazyBasis):
             self.state_space = _statespace.ExplicitStateSpace([('L%d' % i,) for i in range(dim_or_statespace)])
         elif name == "sv":
             # A state vector can have any shape. It does not need to be a perfect square root.
-            udim = _np.sqrt(dim_or_statespace)
-            self.state_space = _statespace.ExplicitStateSpace(('all',), udims=(udim,), types=("quantum",))
+            self.state_space = _statespace.default_space_for_udim(dim_or_statespace)
         else:
             self.state_space = _statespace.default_space_for_dim(dim_or_statespace)
 
@@ -1092,6 +1091,14 @@ class BuiltinBasis(LazyBasis):
     def _from_nice_serialization(cls, state):
         statespace = _StateSpace.from_nice_serialization(state['state_space'])
         return cls(state['name'], statespace, state['sparse'])
+    
+    def _get_dimension_to_pass_to_constructor(self):
+        """
+        A basis in the state-vector (name= 'sv') case will correspond to a basis of vectors of length d.
+        This means that it will be operated on by matrices of shape (d \times d).
+        """
+        return self.state_space.udim if self.name == "sv" else self.state_space.dim
+
 
     @property
     def dim(self):
@@ -1100,7 +1107,7 @@ class BuiltinBasis(LazyBasis):
         spans.  Equivalently, the length of the `vector_elements` of the
         basis.
         """
-        size, dim, elshape = _basis_constructor_dict[self.name].sizes(dim=self.state_space.dim, sparse=self.sparse)
+        size, dim, elshape = _basis_constructor_dict[self.name].sizes(dim=self._get_dimension_to_pass_to_constructor(), sparse=self.sparse)
         return dim
 
     @property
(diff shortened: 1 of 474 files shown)


pygsti/baseobjs/basis.py
```diff
@@ -1063,12 +1063,16 @@ def __init__(self, name, dim_or_statespace, sparse=False):
...
-            self.state_space = dim_or_statespace if isinstance(dim_or_statespace, _statespace.StateSpace) \
-                else _statespace.ExplicitStateSpace([('L%d' % i,) for i in range(dim_or_statespace)])
+        if isinstance(dim_or_statespace, _statespace.StateSpace):
+            self.state_space = dim_or_statespace
+        elif name == 'cl':  # HACK for now, until we figure out better classical state spaces
+            self.state_space = _statespace.ExplicitStateSpace([('L%d' % i,) for i in range(dim_or_statespace)])
+        elif name == "sv":
+            # A state vector can have any shape. It does not need to be a perfect square root.
+            udim = _np.sqrt(dim_or_statespace)
+            self.state_space = _statespace.ExplicitStateSpace(('all',), udims=(udim,), types=("quantum",))
```
rileyjmurray: udim refers to "unitary dimension." For density matrix spaces this is the order $d$ of the $d \times d$ density matrix. For state-vector spaces this is the same as the dimension of the space itself. So I think udim should just be dim_or_statespace (which, by this point in the if-elif checks, is just the dim).

I'll grant that ExplicitStateSpace doesn't have a way to account for this. Consider the following:
```python
>>> ssawk = pg.baseobjs.statespace.ExplicitStateSpace(("all",), udims=(5**0.5,), types=('quantum',))
>>> ssawk
ExplicitStateSpace[all(5)]
>>> ssawk.dim
5
>>> ssawk.udim
2
```
So your hack at creating a statespace with non-perfect-square dimension by passing in non-integer udim _does_ set the dimension properly, but it _does not_ set the unitary dimension properly.

Ping @coreyostrove. It looks like we don't have an implementation of StateSpace that's suitable for state vectors. ExplicitStateSpace is only capable of representing tensor products of mixed state spaces (per [these lines](https://github.com/sandialabs/pyGSTi/blob/c3f1f84c1e360936292fe102387bf1694560228f/pygsti/baseobjs/statespace.py#L1069-L1075) in its constructor). Should @nkoskelo or I create a StateSpace class suitable for state vectors?

coreyostrove: Sorry for taking a while to chime in here, @rileyjmurray  and @nkoskelo. First off, thank you very much for taking the initiative here, @nkoskelo!

I've read through the relevant state space and basis code to refresh my memory on how this all works and it looks like 1) you're correct that there is any issue with the state space construction in `BuiltInBasis` for state vectors and 2) I think the proposed fix is *almost* what you want.

I'd suggest the following:

1. Instead of constructing and `ExplicitStateSpace` object directly there is a companion function to `default_space_for_dim` in the `statespace` module called `default_space_for_udim` which you can use for the state vector case. 
2. remove the definition of `udim` as the square root of `dim_or_statespace` and simply use this (in the case where it is an integer) as the udim directly. Then add a line to the docstring to indicate that this is the behavior when using state vectors.


@rileyjmurray: From having read through the `StateSpace` code I don't think there is any problem with those classes themselves, they should in principle store state vector data just fine, just an issue with the way those classes were being instantiated in the basis code.

test/unit/objects/test_basis.py
coreyostrove, line 1: Thanks for including unit tests!

coreyostrove:
@nkoskelo, could I ask you a quick favor while you're making patches to the basis/state space code? While I was looking through the state space code I noticed this errant property definition: https://github.com/sandialabs/pyGSTi/blob/3944ca53dab6e9d99080109e6c4a89dfebdf4f28/pygsti/baseobjs/statespace.py#L707

This is just missing the return value: `return self._qudit_udims`. Any chance you could patch that and add it to this PR?

coreyostrove (approved):
Everything looks good to go, thanks for the great work, @nkoskelo!

Issue #437, opened 2024-05-09
Replace CVXOPT dependency with Clarabel


sserita:
CVXPY now officially supports Clarabel as an alternative SDP solver to CVXOPT. There have been many pain points with building CVXOPT in various environments (including GitHub Actions runners and on HPC clusters). There is a general consensus that we would like to migrate pyGSTi dependencies away from CVXOPT and to Clarabel, provided that there is feature parity and comparable performance between the solvers (which we currently believe to be the case).

We'll deprecate the cvxopt dependency in 0.9.13 and officially switch solvers in 0.9.14.

rileyjmurray:
I think the only TODO is to specify the solver=Clarabel rather than solver=CVXOPT in the diamond norm calculation. See https://github.com/sandialabs/pyGSTi/issues/520#issue-2745584323. (Closing that issue as a duplicate of this one.

coreyostrove:
Completed in #588.

PR #588, opened 2025-06-17
Expire CVXOPT Dependency

coreyostrove:
This finishes the expiration of the dependency on CVXOPT that we announced in the release of 0.9.13. The cvxpy dependency has had an associated minimum version requirement added.

See #437 for more.

### DIFF
--- a/pygsti/tools/optools.py
+++ b/pygsti/tools/optools.py
@@ -305,7 +305,7 @@ def diamonddist(a, b, mx_basis='pp', return_x=False):
     mx_basis = _bt.create_basis_for_matrix(a, mx_basis)
 
     # currently cvxpy is only needed for this function, so don't import until here
-    import cvxpy as _cvxpy
+    import cvxpy as _cp
 
     # _jam code below assumes *un-normalized* Jamiol-isomorphism.
     # It will convert a & b to a "single-block" basis representation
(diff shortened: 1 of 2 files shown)


pygsti/tools/optools.py
```diff
@@ -323,7 +323,7 @@ def diamonddist(a, b, mx_basis='pp', return_x=False):
     prob, vars = _diamond_norm_model(dim, smallDim, J)
 
     try:
-        prob.solve(solver='CVXOPT')
+        prob.solve(solver='Clarabel')
     except _cvxpy.error.SolverError as e:
         _warnings.warn("CVXPY failed: %s - diamonddist returning -2!" % str(e))
         return (-2, _np.zeros((dim, dim))) if return_x else -2
```
rileyjmurray: Let's do the following instead
```python
    objective_val = -2
    varvals = [_np.zeros_like(J), None, None]
    sdp_solvers = ['MOSEK', 'CLARABEL', 'CVXOPT']
    for i, solver in enumerate(sdp_solvers):
        try:
            prob.solve(solver=solver)
            objective_val = prob.value
            varvals = [v.value for v in vars]
            break
        except (AssertionError, _sdps.cp.SolverError) as e:
            msg = f"Received error {e} when trying to use solver={solver}."
            if i + 1 == len(sdp_solvers):
                failure_msg = "Out of solvers. Returning -2 for diamonddist."
            else:
                failure_msg = f"Trying {sdp_solvers[i+1]} next."
            msg += f'\n{failure_msg}'
            _warnings.warn(msg)
```

rileyjmurray: (I left almost the same comment verbatim in some other GitHub interface. I'm not sure where that comment went.)

coreyostrove: What is the import for `_sdps`?

rileyjmurray: Oh, whoops. It's on a branch where I made an sdptools.py file. All that matters is that ``_sdps.cp.SolverError`` is ``cvxpy.SolverError``. So it'll suffice to import cvxpy as cp and then replace with ``cp.SolverError``.

coreyostrove: Ok, I made this change with a small modification to skip producing warnings is MOSEK isn't installed, since we can assume this is pretty much always going to be the case for most users. (Read: I don't have MOSEK and didn't want to get constant warnings about that...)

While going through I also noticed that there was a flag disabling the diamond distance unit test on windows that was added about 3 years ago. The commit message on that indicated it was due to random CVXOPT failures on the runners. I'm going to blindly assume that the changes we're making plus the intervening three years probably fixed things on their own so reenabled those.

coreyostrove:
Note to selves, @rileyjmurray passed along approval after most recent changes, so going ahead with merge.

PR #590, opened 2025-06-20
PyPI Deployment Fix (v0.9.13.2)

coreyostrove:
This PR fixes an issue with the deployment process for release on pypi which had resulted in the source distribution and a handful of wheels being missing from the latest release.

Unrelated, a minor import/dependency issue with matplotlib that was caught in testing the deployment changes.

To avoid name-reuse issues with pypi (my understanding is that once a version number is used it can't be reassigned after the fact even if you delete a release) this will become a new bugfix version 0.9.13.2.

### DIFF
--- a/.github/workflows/autodeploy.yml
+++ b/.github/workflows/autodeploy.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.1.2
         env:
-          CIBW_BUILD: cp38-* cp39-* cp310-* cp311-*
+          CIBW_BUILD: cp39-* cp310-* cp311-* cp312-*
           CIBW_BUILD_VERBOSITY: 1
           CIBW_BEFORE_ALL_LINUX: ./.github/ci-scripts/before_install.sh
           CIBW_BEFORE_ALL_MACOS: ./.github/ci-scripts/before_install_macos.sh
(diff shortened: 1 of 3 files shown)


.github/workflows/autodeploy.yml
```diff
@@ -66,7 +67,7 @@ jobs:
   upload_pypi:
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' && github.event.action == 'published'
+    #if: github.event_name == 'release' && github.event.action == 'published'
```
rileyjmurray, line 70: Why this change?

coreyostrove: This check made it so that the only way to trigger the final upload to pypi step of the deployment process was to have the workflow triggered by a webhook fired by publishing a new release post. I needed to disable that to be able to manually trigger the full deployment for testing. We should probably turn this back on at some point, but I made the mistake of already tagging the v0.9.13.2 commit, and don't have the permissions to delete it so would suggest we just revert that back for 0.9.14 and plan on manually deploying (by triggering the workflow ourselves) this time.

rileyjmurray: Got it. Works for me!

rileyjmurray (approved):

Issue #484, opened 2024-09-13
Serialization of PlaquetteGridCircuitStructure is incorrect if fiducial pair reduction is True


eendebakpt:
**Describe the bug**

To perform the GST measurements and analysis independently I tried to serialize the relevant structures after measurment, and then load the same structures later for analysis. This fails for the `PlaquetteGridCircuitStructure`

**To Reproduce**

A minimal example:
```
import pygsti
from pygsti.modelpacks import smq1Q_XY, smq2Q_XYCPHASE
import tempfile
import os
from pygsti.circuits.circuitstructure import PlaquetteGridCircuitStructure
import tempfile

model_pack = smq2Q_XYCPHASE
exp_design = model_pack.create_gst_experiment_design( max_max_length=[1,2], fpr=True)
       
experiment_list = exp_design.all_circuits_needing_data
f = tempfile.mktemp(suffix='.json')
experiment_list.write(f)

with open(f,  'r') as fid:
        experiment_list2=PlaquetteGridCircuitStructure.load(fid)
        
print(experiment_list[0])
print(experiment_list2[0])
```
Has output
```
Qubit 0 -----
Qubit 1 -----

Qubit 0 ---|     |-|     |-|     |-|     |---
Qubit 1 ---|Gypi2|-|Gxpi2|-|Gxpi2|-|Gxpi2|---
```

The issue does not occur then `fpr` is set to `False`.

**Expected behavior**

The two printed circuits should be identical. In particular, the deserialized object should be identical to the original object (in all aspects that are relevant for running GST). The same problem occurs when using the `PlaquetteGridCircuitStructure.to_nice_serialization`.

**Environment (please complete the following information):**
 - pyGSTi version [e.g. v0.9.7.4]: 0.9.12.3
 - python version [e.g. 3.7, 2.7]: 3.10
 - OS [e.g. OSX 10.14, Ubuntu 16.04 LTS] Windows

**Additional context**

coreyostrove:
Some more updates. I narrowed things down further, and am copying the commit message for my hacky patch for the specific issue reported here for posterity. 

[...]

PR #591, opened 2025-06-20
Hack For Issue #484

coreyostrove:
This is a hacky fix for the issue reported in #484. It was reported there that when serializing and deserializing certain `PlaquetteGridCircuitStructure` objects one wouldn't get exactly identical. Drilling down it was determined that the underlying issue wasn't serialization, but in fact one could reproduce this by taking the `PlaquetteGridCircuitStructure` produced by the function `pygsti.circuits.create_lsgst_circuit_lists` (which is used in the GST experiment design construction workflow internally) and then either copy, or simply instantiate a new `PlaquetteGridCircuitStructure` pulling all of the inputs from the corresponding attributes of the originally produced one and one would get a new instance apparently identical in all ways except for the ordering of the underlying `_circuits` attribute. For the life of me I can't figure out why that is, and there is nothing obvious in the construction logic used in `pygsti.circuits.create_lsgst_circuit_lists` that I can see which would cause this. So, since based on my testing the contents are identical up to the ordering, I've introduced a hacky patch which makes a copy of the `PlaquetteGridCircuitStructure`s in `pygsti.circuits.create_lsgst_circuit_lists` before returning them.

### DIFF
--- a/pygsti/circuits/circuitstructure.py
+++ b/pygsti/circuits/circuitstructure.py
@@ -719,16 +719,6 @@ class PlaquetteGridCircuitStructure(_CircuitList):
                                           if (self.circuit_weights is not None) else None)
                       })
 
-        #state['plaquette_types'] = XXX
-        #state['datacols'] = {
-        #    'Circuit': [circuits]
-        #    'PlaquetteX':
-        #    'PlaquetteY':
-        #    'X':
-        #    'Y':
-        #    'weight':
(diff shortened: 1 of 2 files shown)


rileyjmurray (approved):
This is fine with me. If you wanted, you could rephrase the TODO as an actual task, rather than just explaining why the code looks weird.

PR #597, opened 2025-06-24
Another round of small updates

enielse:
Another round of small updates from Erik.  Main changes are:
- adds more support for creating wildcard error budgets with `LocalNoiseModel` models (instead of just `ExplicitOpModels`).  Also adds several more error and warning messages during wildcard error construction to make debugging easier.
- adds ability to add & check compilation rules directly with a `CliffordCompilationRules` object.  And adds an iSWAP->CNOT compilation rule to the standard construction when relevant gates are available.
- adds `merge_with` methods to `CircuitListsDesign`, `ByDepthDesign`, `BenchmarkingDesign`, and `MirrorRBDesign` so make it easy to combine multiple of these experiment designs (especially useful for mirror RB designs, where a user may run an initial set of MRB circuits then follow up with another set containing wider or deeper circuits and want to analyze the entire set of circuits together.)

Note: please don't delete the `egn-updates` branch if/when this PR is approved, as Erik will continue to work on it.


coreyostrove (approved):
No concerns on my end, thanks for great work, @enielse!

Issue #595, opened 2025-06-23
Saving FOGI models to disk


juangmendoza19:
**Is your feature request related to a problem? Please describe.**
The creation of FOGI models for two qubits currently takes > 6 hours. Currently they are not able to be serialized due to the fogi_store attribute not having serialization methods.

**Describe the solution you'd like**
Working _to_nice_serialization() and from_nice_serialization methods for FOGI models

PR #598, opened 2025-06-24
FOGI model serialization and some FOGI model optimization 

juangmendoza19:
This resolves issue #595 by allowing FOGI models to be serialized. Now, any fogi model can be saved to disk with the method <fogimodel>.write('<filename>'), and loaded back with pygsti.models.Model.read('<filename>').

The main problem was the the class FirstOrderGaugeInvariantStore was not nicely serializable, which is stored in FOGI models. All appropriate serializations were added to this class, and other classes used within.

Additionally, @coreyostrove and I modified model.py:set_parameter_values() to be executed faster when used with FOGI models


pygsti/baseobjs/errorgenbasis.py
```diff
@@ -88,6 +89,7 @@ def __init__(self, state_space, labels, basis_1q=None):
             comprise the basis element labels for the values of the
             `ElementaryErrorgenLabels` in `labels`.
         """
+        super().__init__()
```
juangmendoza19, line 93: Every object that is nicely serializable needs to have this init statement

pygsti/baseobjs/errorgenbasis.py
```diff
@@ -116,6 +118,16 @@ def __init__(self, state_space, labels, basis_1q=None):
...
+    def _to_nice_serialization(self):
+        state = super()._to_nice_serialization()
+        state.update({'state_space' : self.state_space._to_nice_serialization(),
+                      'labels' : [label.__str__() for label in self.labels],
+                      '_basis_1q' : self._basis_1q if isinstance(self._basis_1q, str) else self._basis_1q._to_nice_serialization()
+        })
+        return state
+    @classmethod
+    def from_nice_serialization(cls, state):
+        return cls(_StateSpace.from_nice_serialization(state['state_space']), [_GlobalElementaryErrorgenLabel.cast(label) for label in state['labels']], state['_basis_1q'] if isinstance(state['_basis_1q'], str) else _Basis.from_nice_serialization(state['_basis_1q']))
```
juangmendoza19: Added support for _basis_1q variables which can be strings or Basis objects according to the docstring

pygsti/baseobjs/statespace.py
```diff
@@ -418,7 +418,7 @@ def create_subspace(self, labels):
         StateSpace
         """
         # Default, generic, implementation constructs an explicit state space
-        labels = set(labels)
+        labels = sorted(set(labels))
```
juangmendoza19, line 421: This change was made with @coreyostrove to make FOGI model construction deterministic

pygsti/models/fogistore.py
```diff
@@ -21,23 +21,46 @@
...
 
     Currently, it is only compatible with :class:`ExplicitOpModel` objects.
     """
+    def __init__(self, primitive_op_labels, gauge_space, elem_errorgen_labels_by_op, op_errorgen_indices, fogi_directions, fogi_metadata, dependent_dir_indices, fogv_directions, allop_gauge_action, gauge_space_directions, norm_order=None, dependent_fogi_action='drop'):
+        super().__init__()
+        self.primitive_op_labels = primitive_op_labels
+        self.gauge_space = gauge_space
+        self.elem_errorgen_labels_by_op = elem_errorgen_labels_by_op
+        self.op_errorgen_indices = op_errorgen_indices
+        self.fogi_directions = fogi_directions
```
juangmendoza19, line 117: The __init__ method of this class used to not only create a FOGIstore object, but also compute all fogi directions. In order to be able to load them from disk, we needed a simple __init__ method which just takes attributes and spits back the object without doing any math.

pygsti/models/fogistore.py
```diff
@@ -21,23 +21,46 @@
...
+        self.norm_order = norm_order
+        self.dependent_fogi_action = dependent_fogi_action
 
-    def __init__(self, gauge_action_matrices_by_op, gauge_action_gauge_spaces_by_op, errorgen_coefficient_labels_by_op,
+        self.errorgen_space_op_elem_labels = tuple([(op_label, elem_lbl) for op_label in self.primitive_op_labels
+                                                    for elem_lbl in self.elem_errorgen_labels_by_op[op_label]])
+        self.fogv_labels = ["%d gauge action" % i for i in range(self.fogv_directions.shape[1])]
+    
+    @classmethod
+    def compute_fogis(cls, gauge_action_matrices_by_op, gauge_action_gauge_spaces_by_op, errorgen_coefficient_labels_by_op,
```
juangmendoza19: The old __init__method became compute_fogis(). It is decorated as a classmethod such that it can be used without the existence of a fogi_store object. As it intended to be used, at the end it creates an object which is returned back to the user.

pygsti/models/fogistore.py
```diff
@@ -146,35 +171,86 @@ def __init__(self, gauge_action_matrices_by_op, gauge_action_gauge_spaces_by_op,
...
 
         #Store auxiliary info for later use
-        self.norm_order = norm_order
-        self._dependent_fogi_action = dependent_fogi_action
-
+        fogi_store = cls(primitive_op_labels, gauge_space, elem_errorgen_labels_by_op, op_errorgen_indices, fogi_directions, fogi_metadata, dependent_dir_indices, fogv_directions, allop_gauge_action, gauge_space_directions, norm_order, dependent_fogi_action)
+        fogi_store._check_fogi_store()
+        return fogi_store
         #Assertions to check that everything looks good
-        if True:
```
juangmendoza19, line 153: Erik coded this block to run some sanity checks on the resulting FOGIStore object. I moved it into its own method below, _check_fogi_store().

pygsti/models/model.py
```diff
@@ -657,56 +657,6 @@ def num_modeltest_params(self):
         self._clean_paramvec()
         return Model.num_modeltest_params.fget(self)
 
-    @property
```
juangmendoza19, line 660: It seems like a mistake that develop had all these methods duplicated

pygsti/models/model.py
```diff
@@ -1187,21 +1137,20 @@ def _build_index_mm_map(self):
...
+        index_mm_label_map = [[] for _ in range(len(ops_param_vec))]
+        
+        for lbl, obj in self._iter_parameterized_objs():
+            #if the gpindices are a slice then convert to a list of indices.
+            gpindices = _slct.indices(obj.gpindices) if isinstance(obj.gpindices, slice) else obj.gpindices
+            for gpidx in gpindices:
+                index_mm_map[gpidx].append(obj)
+                index_mm_label_map[gpidx].append(lbl)
+        self._index_mm_map = index_mm_map
+        self._index_mm_label_map = index_mm_label_map
```
juangmendoza19, line 1153: the resulting index_mm_map was incorrect for FOGI models. This is necessary for our set_parameter_values optimization

pygsti/models/model.py
```diff
@@ -1304,52 +1253,71 @@ def set_parameter_values(self, indices, values, close=False):
...
+            
+            original_errgen_vec = self._param_interposer.transform_matrix @ orig_param_vec
+            new_errgen_vec = self._param_interposer.transform_matrix @ self._paramvec
+            diff_vec =  original_errgen_vec -  new_errgen_vec
+            diff_vec[_np.abs(diff_vec) < 1e-14] = 0
+            non_zero_errgens = _np.nonzero(diff_vec)
+
+            indices = non_zero_errgens[0]
+            values = new_errgen_vec[indices]
+            vec_to_access = new_errgen_vec
```
juangmendoza19, line 1281: This if statement captures which error generators are affected by the FOGI vector changing.

pygsti/models/model.py
```diff
@@ -1304,52 +1253,71 @@ def set_parameter_values(self, indices, values, close=False):
...
+            obj.from_vector(vec_to_access[obj.gpindices].copy(), close, dirty_value=False)
+        
+        #go through the model members which have been updated and identify whether any of them have children
+        #which may be present in the _opcaches which have already been updated by the parents. I think the
+        #conditions under which this should be safe are: a) the layer rules are ExplicitLayerRules,
+        #b) The parent is a POVM (it should be safe to assume that POVMs update their children, 
+        #and c) the effect is a child of that POVM.
+        
+        if isinstance(self._layer_rules, _ExplicitLayerRules):
+            updated_children = []
```
juangmendoza19, line 1300: unique_mms identifies which operations correspond to the error generators that are changing, which then allows us to only update those in our model inside the loop. This way we only update the objects in our model which are affected by the values that were modified by the user.

pygsti/models/model.py
juangmendoza19: vec_to_access is a new variable which will always hold the vector of object parameters. In the case of FOGI models, it will always correspond to the *updated* error generator vector.

pygsti/models/model.py
```diff
@@ -1304,52 +1253,71 @@ def set_parameter_values(self, indices, values, close=False):
         -------
         None
         """
+        orig_param_vec = self._paramvec.copy()
```
juangmendoza19, line 1256: All of the code below assumes that parameter interposers are only used for FOGI models, and that FOGI models are built from GLND parameterizations (i.e. every FOGI quantity is defined in terms of error generators)

pygsti/models/model.py
```diff
@@ -2553,8 +2521,6 @@ def setup_fogi(self, initial_gauge_basis, create_complete_basis_fn=None,
                    dependent_fogi_action='drop', include_spam=True, primitive_op_labels=None):
 
         from pygsti.baseobjs.errorgenbasis import CompleteElementaryErrorgenBasis as _CompleteElementaryErrorgenBasis
-        from pygsti.baseobjs.errorgenbasis import ExplicitElementaryErrorgenBasis as _ExplicitElementaryErrorgenBasis
-        from pygsti.baseobjs.errorgenspace import ErrorgenSpace as _ErrorgenSpace
```
juangmendoza19, line 2557: These two imports were never used

pygsti/baseobjs/errorgenbasis.py
coreyostrove: The labels could either be local or global labels, at this point.

pygsti/baseobjs/errorgenbasis.py
coreyostrove, line 1: The classes CompleteElementaryErrorgenBasis and ExplicitElementaryErrorgenBasis have additional data and different init signatures, so these wouldn't be properly serialized using the parent classes `_to_nice_serialization` and `_from_nice_serialization` methods. These should be overridden in the child classes for full compatibility.

pygsti/baseobjs/errorgenspace.py
```diff
@@ -24,13 +24,22 @@ class ErrorgenSpace(object):
     """
 
     def __init__(self, vectors, basis):
+        super().__init__()
```
coreyostrove, line 36: Add a docstring, por favor.

pygsti/baseobjs/errorgenspace.py
```diff
@@ -24,13 +24,22 @@ class ErrorgenSpace(object):
...
-
+    def _to_nice_serialization(self):
+        state = super()._to_nice_serialization()
+        state.update({'vectors' : self._encodemx(self.vectors),
+                      'basis': self.elemgen_basis._to_nice_serialization()
+        })
+        return state
+    @classmethod
+    def from_nice_serialization(cls, state):
+        return cls(state['vectors'], ExplicitElementaryErrorgenBasis.from_nice_serialization(state['basis']))
```
coreyostrove: Vectors need decoding back into numpy array.

pygsti/models/fogistore.py
```diff
@@ -21,23 +21,46 @@
...
+                                                    for elem_lbl in self.elem_errorgen_labels_by_op[op_label]])
+        self.fogv_labels = ["%d gauge action" % i for i in range(self.fogv_directions.shape[1])]
+    
+    @classmethod
+    def compute_fogis(cls, gauge_action_matrices_by_op, gauge_action_gauge_spaces_by_op, errorgen_coefficient_labels_by_op,
                  op_label_abbrevs=None, reduce_to_model_space=True,
                  dependent_fogi_action='drop', norm_order=None):
         """
         TODO: docstring
         """
```
coreyostrove, line 170: Docstring, por favor.

coreyostrove (changes requested):
Mostly looks good. A few requests inline. Also a general request to identify whether or not we have unit test coverage for the new functionality that is being added for both the serialization cases and the updated code path for set_parameter_values when using a FOGI model/model with interposer.

test/unit/objects/test_fogi.py
```diff
@@ -228,3 +229,61 @@ def test_cloud_crosstalk_fogi(self):
...
+        self.assertTrue(equal_fogi_models(model3, model3), msg=msg)
+
+        msg = 'FOGI models that are different are not recognized as different by __eq__ methods'
+        self.assertFalse(equal_fogi_models(model, model2), msg=msg)
+        self.assertFalse(equal_fogi_models(model, model3), msg=msg)
+        self.assertFalse(equal_fogi_models(model2, model3), msg=msg)
+
+    #TODO: should this be in test_nice_serialization instead?
+    @with_temp_path
+    def test_fogi_serialization(self, temp_pth):
```
juangmendoza19, line 266: Do we think this test is fine here or is it more appropriate to have in test_nice_serialization.py? @coreyostrove

test/unit/io/test_nice_serialization.py
```diff
@@ -69,6 +69,8 @@ def test_cloudnoise_model(self, pth):
         self.assertTrue(mdl_cloud.is_equivalent(mdl_cloud2))
 
 
+#Somewhat random question for other pyGSTi developers: 
+#Am I missing something or does this class not test anything related to serialization?
 class ModelEquivalenceTester(BaseCase):
```
juangmendoza19: @coreyostrove while looking around these files I found this test which seems to not be related to the file name. Should these tests below be moved somewhere else or am I missing something? Do they have anything to do with serialization?

pygsti/models/modelparaminterposer.py
```diff
@@ -99,4 +99,11 @@ def _to_nice_serialization(self):
     @classmethod
     def _from_nice_serialization(cls, state):  # memo holds already de-serialized objects
         return cls(cls._decodemx(state['transform_matrix']))
+    
+    def __eq__(self, other):
```
juangmendoza19, line 103: After implementing all of these __eq__ methods, I realized that some classes have an is_equivalent() method instead. Let me know if I should change anything about the names I used for these, or their implementation to abide by a convention. @coreyostrove

test/unit/objects/test_fogi.py
coreyostrove: Having this in the fogi tests is fine

test/unit/io/test_nice_serialization.py
coreyostrove: Nope, not sure how they got there originally, but moving them to test_model.py would work.

pygsti/baseobjs/errorgenbasis.py
```diff
@@ -116,6 +119,24 @@ def __init__(self, state_space, labels, basis_1q=None):
         self._cached_dual_matrices = None
         self._cached_supports = None
 
+    def __eq__(self, other):
```
coreyostrove, line 122: Can you add a one-sentence docstring for this? The only possibly non-obvious behavior here is that equality will depend on the order of the labels as implemented. This instead could be a set equality check, for example. I think either way is fine.

pygsti/baseobjs/errorgenbasis.py
```diff
@@ -762,8 +802,10 @@ def label_index(self, label, ok_if_missing=False, identity_label='I'):
         elif eetype in ('C', 'A'):
             assert(len(trivial_bel) == 1)  # assumes this is a single character
             nontrivial_inds = [i for i, letter in enumerate(bels[0]) if letter != trivial_bel]
-            left_support = tuple([self.sslbls[i] for i in nontrivial_inds])
-
+            left_support = tuple([label.sslbls[i] for i in nontrivial_inds])
```
coreyostrove, line 815: Pinging @enielse regarding this change. The intention of this was to address an bug related to an edge case. I'll let @juangmendoza19 explain more.

pygsti/baseobjs/errorgenspace.py
```diff
@@ -75,6 +93,9 @@ def intersection(self, other_space, free_on_unspecified_space=False, use_nice_nu
 
         return ErrorgenSpace(intersection_vecs, common_basis)
 
+    def __eq__(self, other):
```
coreyostrove, line 96: Similar comment to the one on the __eq__ above. Can you add a one-sentence docstring indicating that the order of the vectors matters.

pygsti/models/fogistore.py
```diff
@@ -210,7 +393,82 @@ def find_nice_fogiv_directions(self):
         #gauge_space_directions = _np.dot(pinv_allop_gauge_action, self.fogv_directions)  # in gauge-generator space
         #assert(_np.linalg.matrix_rank(gauge_space_directions) <= self._gauge_space_dim)  # should be nearly full rank
         pass
+    
+    def __eq__(self, other):
```
coreyostrove, line 397: See comment above about order dependence.

pygsti/models/modelparaminterposer.py
coreyostrove: You can make a wrapper around __eq__ is you want that method as an option, but this is optional.

test/unit/objects/test_model.py
```diff
@@ -209,7 +213,7 @@ def test_jtracedist(self):
 
     @needs_cvxpy
     def test_diamonddist(self):
-        if SKIP_DIAMONDIST_ON_WIN and sys.platform.startswith('win'): return
+        if  SKIP_DIAMONDIST_ON_WIN and sys.platform.startswith('win'): return
```
coreyostrove: Can you do a merge of develop into your branch. This line was removed in PR #588 which was merged in a short while ago, so as is this would unintentionally add it back in.

pygsti/baseobjs/errorgenbasis.py
juangmendoza19: It seems like label_index() is supposed to identify different representations of the same label. For example, this can be seen in test_errorgenbasis.py:test_label_index(), where ('C', ['XI', 'YI']) is supposed to be identified as ('C', ('X', 'Y'), (0,)).

Conversely, the label ('C', ('X', 'Y'), (1,)) should be identified as equivalent to ('C', ['IX', 'IY']). This is not the case. The current label_index() code in develop returns None for this example. This problem prevented 2 qubit FOGI models from being constructed.

I was uncertain of what exactly the code within the function is doing, but by following how the variables transformed in each line, I concluded that maybe the line changed above was being indexed into the wrong array. The fix I provided passes test_label_index() and also seems to correctly identify when the trivial subspace is not in the first register.

We would like your input on this problem and solution @enielse

test/unit/objects/test_model.py
```diff
@@ -730,3 +746,55 @@ def tearDown(self):
 #    def test_randomize_with_unitary_raises(self):
 #        with self.assertRaises(AssertionError):
 #            self.model.randomize_with_unitary(1, rand_state=np.random.RandomState())  # scale shouldn't matter
+
```
juangmendoza19, line 749: All of the code below is not original to this PR, it was moved from test_nice_serialization.py

pygsti/baseobjs/errorgenbasis.py
enielse: Thanks for the explanation @juangmendoza19 , and sorry it took so long to get to this.  I agree the fix is good, and your explanation is correct.  

In the case when `label` is a `LocalElementaryErrorgenLabel` then `self.sslbls == label.sslbls` and the issue doesn't arise, which is my guess at why this didn't cause an issue sooner.  These lines should be computing the support of the requested `label` and therefore should use `label.sslbls` when there's a difference, i.e. for a `GlobalElementaryErrorgenLabel`, and so I think this was just a bug indexing the wrong array as you suggest.

pygsti/tools/fogitools.py
```diff
@@ -560,9 +560,39 @@ def resolve_norm_order(vecs_to_normalize, label_lists, given_norm_order):
                         # (A,B are faithful reps of gauge on intersection space, so pinv(ga_A) * ga_A
                         # restricted to intersection space is I:   int_spc.T * pinv(ga_A) * ga_A * int_spc == I
                         # (when int_spc vecs are orthonormal) and so the above redues to I - I = 0
+                        def scipy_pinv(A, rcond):
+                            # Step 1: Compute SVD
+                            U, S, VT = _scp.linalg.svd(A, lapack_driver='gesvd')
+
+                            # Step 2: Construct Sigma
+                            Sigma = _np.zeros((U.shape[0], VT.shape[0]))
+                            _np.fill_diagonal(Sigma, S)
```
coreyostrove: Pro-tip, use broadcasting here. It is much more efficient. This Sigma matrix is only being constructed here for the sake of the matrix multiplication on line 578, but what does matrix multiplication by a diagonal matrix, D, and some generic matrix A do? It rescales each row of A by the corresponding diagonal entry of D. If you have a vector V of the correct size (len=num_rows(A)) and you multiply it into A (using the standard multiplication operator, '*', not matrix multiplication's '@') then numpy's default behavior is to broadcast this multiplication across A, which is hard to [describe concisely](https://numpy.org/devdocs//user/basics.broadcasting.html), but has exactly the desired effect of rescaling the rows of A by the corresponding values of V. But you'll achieve this with better computational complexity (only a quadratic number of multiplications instead of the full arithmetic cost of matrix multiplication).

This isn't a bottleneck as far as I am aware, so I'm not going to hold things up for the sake of this change, but there's never a bad time to learn about nifty numpy performance tricks.

now-you-know-the-more-you-know

coreyostrove (approved):
Tests are now passing, which was the last thing I was waiting on before approval. Great work, @juangmendoza19!

PR #603, opened 2025-07-07
Plotly 6 Per-Sequence Detail Plot Rendering Bugfix

coreyostrove:
This PR addresses a bug that was reported by users internally and by email from some external users which resulted in figures which were instances of pyGSTi's `ColorBoxPlot` class to fail to render in HTML reports when using plotly 6.0.0 and above. I've traced this back to an incompatibility between version 6+ of the plotly python package and the version of plotly's corresponding javascript package (which is the actual code that runs in the browser for plot generation in reports) we were pinning to. This update bumps up the javascript package to version 3.0.1, which is the newest version as of the time of writing this. We had been pinning to a version called 'plotly-latest', which despite the name is in fact version 1.58.5 released in 2021, as it turns out plotly discontinued publishing a tracking version with the release of 2.0.0. See this github issue for more: https://github.com/plotly/plotly.js/issues/7315

Also included is a semi-related patch for some `titlefont` refactor stragglers needed for plotly 6. 

I was a bit concerned about creating incompatibilities in the opposite direction. I.e. incompatibilities induced by the use of the newest version of the javascript package with older versions of the plotly python package. I didn't do anything systematic here to find exactly where something breaks, but I did test report generation using v3.0.1 of the javascript package and v5.0.0 of the python package and didn't spot anything in the report that either was broken or behaving badly. v5.0.0 of the python package was released in June 2021, so that is good enough backwards compatibility as far as I am concerned.

We incidentally have other javascript packages associated with report generation with pinned version that should be evaluated, but I will open up a separate issue to track that since they don't currently seem to be causing problems.


coreyostrove:
Quick note for reviewers, the failing unit tests here are unrelated to this PR and are a separate which has already been patched on develop with #602, and can be safely ignored at the moment.

tjproct (approved):
Inasmuch as I understand it, it all looks good to me. The plotly code isn't something I'm very familiar with.

PR #605, opened 2025-07-07
Add type annotation for on_space parameter

nkoskelo:
When one asks for a dense representation of a `modelmember` via `to_dense(on_space=?)`, there are 3 options that the `on_space` parameter accepts, `Hilbert`, `HilbertSchmidt` and `minimal`. However, when one is coding, editors will not populate these options unless there is a type annotation on the parameter `on_space`. This could lead to unintended behavior, if there is a misspelling for instance. The method `to_sparse(on_space=?)` also has this problem.


rileyjmurray (changes requested):
The current changes have ``__init__.py`` do a relative import, ``from .typing import SpaceT``, from the new file ``pygsti/typing.py`` The fact that this is a relative import is easy to miss and led me to think that SpaceT was a built-in type of some kind. I'd prefer that ``__init__.py`` just define SpaceT directly, and that we remove ``pygsti/typing.py``.

@nkoskelo, if you're strongly opposed to that approach, then let's change the name of ``pygsti/typing.py`` to ``pgsti/pgtypes.py`` or ``pygsti/pygtypes``.

rileyjmurray (approved):

PR #611, opened 2025-07-21
Fix syntax error in `convert_to_openqasm`

judealnas:


coreyostrove:
Update: It turns out we actually have this line fixed on develop already (just using f-strings instead, not sure when it got fixed), so I'm going to close PR without merging. Thank you regardless for submitting this patch!

PR #612, opened 2025-07-22
Ensure Consistency of Global Parameter Indices.

nkoskelo:
The global parameter indices recorded by a layer with only one non-idle gate should be the same as those recorded by the gate itself.

Whenever a new parameterized object is added to a model, there is a possibility that we need to shift the global parameter indices stored by each object within the model.

A call to `shift_gpindices` will now track which changes have already been made.


rileyjmurray (changes requested):
Thanks for this fix, and for writing the tests!

I have two requests.
1. Please make the tests instance methods of a new class, TestRebuildParamvec, that subclasses unittest.TestCase.
2. Please update the PR opening comment to explain the purpose of the PR. You, me, and the other pyGSTi devs will know the purpose for the near future, but we'll inevitably forget.

I'll note that a search through the pyGSTi codebase shows that ModelMember.shift_gpindices is called only in two places: once recursively in shift_gpindices itself, and once in OpModel._rebuild_paramvec. The latter wasn't passing in any value for the "memo" argument in ModelMember.shift_gpindices.

coreyostrove:
We'll also need to track down what is going on with the failing unit tests. These look substantive, so there may be some unexpected consequences being caught. Do these fail locally on your system?

coreyostrove (approved):
These changes look good, and as far as I can tell it looks like the requested changes from @rileyjmurray have been integrated, so I think this is good to merge in now. Thanks for the hard work, @nkoskelo!

Issue #299, opened 2023-04-18
Less Fragile Report Generation


coreyostrove:
There are a number of places in the report generation that are more fragile than they should be. Examples include:

Issue #278 
Errors related to clifford compilations for custom gate sets
Etc. (other examples are escaping me, others should chime in here).

Ideally we should be more robust to these failures, sticking the generation of most plots in try except blocks and raising warnings when specific parts of the report generation fail rather than allowing the entire report generation to fail. We could then simply exclude the plots/tables that couldn't be generated with a message about why.

sserita:
Generally maybe have figures/tables always generate errors but we catch and keep going, i.e. HTML philosophy.

coreyostrove:
Note to selves, some of the changes in PR #615 and PR #572 constitute partial progress toward this (hazily defined) goal.

PR #615, opened 2025-07-25
à la carte standard GST reports, and exception handling for computing metrics

rileyjmurray:
Generating a standard report can take an excessively long time. It's also possible for uncaught exceptions to be raised in the report generation process. This PR addresses both of those issues, by
1. Adding a skip_sections argument to construct_standard_report (see docs for explanation of its behavior), and 
2. Adding exception handling to many (most?) modelmember metrics that might raise linear algebra errors or dimensional errors.

This PR also replaces instances of numpy.NaN with numpy.nan, for compatibility with numpy 2.0.

This splits out work from PR #410. Once this is merged into develop, I'll rebase PR #410 from that point of develop.

### DIFF
--- a/pygsti/baseobjs/basisconstructors.py
+++ b/pygsti/baseobjs/basisconstructors.py
@@ -59,7 +59,6 @@ MAX_BASIS_MATRIX_DIM = 2**6
 
 
 def _check_dim(dim):
-    global MAX_BASIS_MATRIX_DIM
     if not isinstance(dim, _numbers.Integral):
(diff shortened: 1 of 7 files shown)


coreyostrove (approved):
All of these changes look good to me, thank you for your work, @rileyjmurray! We're asymptotically approaching the more resilient report generation we highlighted as a goal in #299. Please feel free to merge this in upon the final set of unit tests completing (assuming they have passed).

PR #616, opened 2025-07-25
New gauge group, plus minor changes to error generator classes

rileyjmurray:
This is pulled out of PR #410. The main change is the new gauge group which handles subgroups of unitaries that have a block-diagonal structure. Along the way I added type annotations throughout error generator code. I also cleaned up code in LindbladCoefficientBlock._block_data_to_params.

### DIFF
--- a/pygsti/modelmembers/operations/fulltpop.py
+++ b/pygsti/modelmembers/operations/fulltpop.py
@@ -35,7 +35,8 @@ class FullTPOp(_DenseOperator, _Torchable):
     An operation matrix that is fully parameterized except for
     the first row, which is frozen to be [1 0 ... 0] so that the action
     of the operation, when interpreted in the Pauli or Gell-Mann basis, is
-    trace preserving (TP).
+    trace preserving (TP). Bases other than Pauli or Gell-Mann are supported
+    only if their first element is the identity matrix.
 
     Parameters
     ----------
@@ -43,7 +44,7 @@ class FullTPOp(_DenseOperator, _Torchable):
         a square 2D array-like or LinearOperator object representing the operation action.
         The shape of m sets the dimension of the operation.
 
-    basis : Basis or {'pp','gm','std'} or None
+    basis : Basis or {'pp','gm'} or None
         The basis used to construct the Hilbert-Schmidt space representation
         of this state as a super-operator.  If None, certain functionality,
         such as access to Kraus operators, will be unavailable.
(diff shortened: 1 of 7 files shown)


pygsti/modelmembers/operations/fulltpop.py
rileyjmurray, line 1: Changes to this file are incidental. They seemed in the spirit of type annotations, although the argument verification  at the end of ``__init__`` is new.

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -725,25 +725,24 @@ def _block_data_to_params(self, truncate=False):
...
+                     % (str(truncate), min(evals)))
 
                 if ttol < 0:  # if we're truncating and assert above allows *negative* eigenvalues
                     #push any slightly negative evals of other_projs positive so that
                     # the Cholesky decomp will work.
-                    Ui = _np.linalg.inv(U)
+                    Ui = U.T.conj()
                     pos_evals = evals.clip(1e-16, None)
-                    nonzero_block_data = _np.dot(U, _np.dot(_np.diag(pos_evals), Ui))
+                    nonzero_block_data = U @ (pos_evals[:, None] * Ui)
```
coreyostrove, line 739: Gotta love broadcasting!

pygsti/models/gaugegroup.py
```diff
@@ -1094,3 +1096,156 @@ def _to_nice_serialization(self):
...
+            'basis': self.basis if isinstance(self.basis, str) else self.basis.to_nice_serialization(),
+            'name': self.name
+        })
+        return state
+
+    @classmethod
+    def _from_nice_serialization(cls, state):
+        subgroups = []
+        for sg_dict in state['subgroups']:
+            if sg_dict['class'] == 'UnitaryGaugeGroup':
```
coreyostrove, line 1160: There is a classmethod for NicelySerializable called `_state_class` that will return the class (i.e. the class object) corresponding to a state dictionary which would let you turn this branching into a one-liner.

pygsti/models/gaugegroup.py
```diff
@@ -1094,3 +1096,156 @@ def _to_nice_serialization(self):
...
+    def num_params(self):
+        return self._num_params
+
+    def inverse(self):
+        return InverseGaugeGroupElement(self)
+
+    def _to_nice_serialization(self):
+        state = super()._to_nice_serialization()
+        state.update({
+            'class': self.__class__.__name__,
```
coreyostrove: This attribute is added by default by the NicelySerializable base class (which if plumbed correctly should be called , is this changing the value set there in some way?

pygsti/models/gaugegroup.py
rileyjmurray: Interesting point; I didn't know about NicelySerializable._state_class.

In this context, where  DirectSumUnitaryGroup only supports UnitaryGaugeGroup and TrivialGaugeGroup, I think the current implementation is easier to read. The current implementation is also cheaper. So I'd like to keep this as-is.

pygsti/models/gaugegroup.py
rileyjmurray: Nope! I just didn't see that NicelySerializable's implementation handled that.

I've pushed a change.

PR #617, opened 2025-07-25
Add type annotations

nkoskelo:


pygsti/baseobjs/basis.py
```diff
@@ -159,7 +168,7 @@ class Basis(_NicelySerializable):
     # could be static methods.
 
     @classmethod
-    def cast_from_name_and_statespace(cls, name: str, state_space: _StateSpace, sparse=None):
+    def cast_from_name_and_statespace(cls, name: str, state_space: _StateSpace, sparse: Optional[bool] = None) -> Basis:
```
rileyjmurray, line 171: Wait ... I thought signature type annotations within a class couldn't use that class' name.

For now I'll write a review assuming this sort of type annotation is allowed by all Python versions we support. If it isn't, then we'll need to revisit.

rileyjmurray (commented):
I was surprised by how you were able to use a class' name within function signature type annotations for that class. 

If tests pass with these sorts of annotations on all versions of Python that we support, then I'm fine with them.

One small note: when it comes to wrapping a variable in a tuple of length 1, I prefer ``(x,)`` rather than ``(x, )``.

pygsti/baseobjs/basis.py
rileyjmurray: Per live conversation: this is allowed as long as we do ``from __future__ import annotations``.

Issue #585, opened 2025-06-04
XZ modelpacks don't have correct fiducials


pcwysoc:
**Describe the bug**
The modelpacks using X(pi/2) and Z(pi/2) gates don't have the correct six antipodal fiducials. The incorrect fiducial is: Gxpi2:0.Gzpi2:0.Gxpi2:0.Gxpi2:0. 

These fiducials are informationally complete.

PR #618, opened 2025-07-28
Fix XZ Modelpack

coreyostrove:
It was recently pointed out by some external collaborators (and subsequently reported in #585) that there was an error on the preparation and measurement fiducials which made it such that, whereas the structure of the XZ modelpack's fiducials should have had a 6x6 structure, mapping the |0> state to each of the cardinal axes of the Bloch sphere, one of the fiducials was accidentally duplicitous of another (meaning we effectively only had 5 unique fiducials). Fortunately they were still IC, which minimized the amount of egg on face incurred.

This PR fixes the prep and measurement fiducials, updates the robust and standard germ sets, and recomputes the pre-computed FPR sets to account for the changes. For backwards compatibility there is a flag that can be set manually in the code to reproduce the older modelpack contents for legacy compatibility purposes.

### DIFF
--- a/pygsti/modelpacks/smq1Q_XZ.py
+++ b/pygsti/modelpacks/smq1Q_XZ.py
@@ -71,10 +71,10 @@ class _Module(GSTModelPack):
                         (('Gxpi2', 0), ('Gzpi2', 0), ('Gzpi2', 0), ('Gzpi2', 0))] #|->
         #Constructed by hand
         _measfiducials = [(), #<0|, <1|
-                        (('Gxpi2', 0),), # <-i|, <i|
+                        (('Gxpi2', 0),), # <i|, <-i|
                         (('Gzpi2', 0), ('Gxpi2', 0)), #<+|, <-| 
                         (('Gxpi2', 0), ('Gxpi2', 0)), # <1|, <0|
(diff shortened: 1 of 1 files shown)


coreyostrove:
Thanks, @rileyjmurray! I'm going to wait on a once-over from @kmrudin, and once he gives the thumbs-up will merge this in.

kmrudin:
Have confirmed the new prep and measure fiducials each generate all 6 antipodal points on the Bloch sphere so it looks good to me too!  One pretty minor thing:  Two of the outcome label pairs in the comments should be swapped (as they're in the comments it has no actual impact, but nice to be consistent in our labeling in the comments).  Lines 73-78 should read as follows (I've only changed the comments in 74 and 77):

```
        _measfiducials = [(), #<0|, <1|
                         (('Gxpi2', 0),), # <i|, <-i|
                         (('Gzpi2', 0), ('Gxpi2', 0)), #<+|, <-| 
                         (('Gxpi2', 0), ('Gxpi2', 0)), # <1|, <0|
                         (('Gzpi2', 0), ('Gzpi2', 0), ('Gxpi2', 0)), #<-i|, <i|
                         (('Gzpi2', 0), ('Gzpi2', 0), ('Gzpi2', 0), ('Gxpi2', 0))] # <-|, <+|
```

PR #621, opened 2025-07-30
Type annotations upstream

rileyjmurray:
This branch is just a vehicle for incorporating changes from PR #617 from a branch within the main pyGSTi repo. It seems to me that we need to do this to get CI to run.


rileyjmurray (commented):
[TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) was only introduced in Python 3.10. Please remove the couple of places where you use it in this PR, @nkoskelo.

nkoskelo (approved):

Issue #539, opened 2025-02-17
Error Generator Propagation Module v2.0 Tracking


coreyostrove:
This issue is intended to serve as a tracker for ideas and updates related to a future version 2.0 of the error generator propagation module.

- [ ]  Bulk methods for probability and pauli expectation value corrections: There is currently a lot of redundant computation when computing multiple output probability or pauli expecation value corrections. E.g. instantiating a fresh `stim.TableauSimulator` and computing the tableau inverse to set its state is a non-trivial fraction of the overall runtime. By reworking the implementation we should be able to amortize such costs over multiple corrections reducing overhead.
- [ ] Bugfix for random error generators: Figure out a corrected construction scheme for building CP-constrained error generators w/o needing both C and A errors to appear in conjunction.
- [ ] Refactor composition code: The error generator composition code is currently ~6000 lines. There is a whole bunch of structure here which we should be able to leverage to cut this down by at least a half.
- [ ] Refactor label propagation out of the `LocalStimErrorgenLabel` class into a separate function in `errorgenproptools`.
- [ ] Refactor `LocalStimErrorgenLabel` into two classes. One which contains only the metadata needed for Markovian models and a child class with additional metadata relevant for non-Markovian applications. This should reduce the overhead of object instantiation (a significant portion of the overall runtime).

PR #625, opened 2025-08-04
Magnus Expansions, Better Tableau Inference, and Errorgens on Paulis

coreyostrove:
This PR adds three new features/updates to the error generator propagation module:

1. Support for the analytic computation of Magnus expansions has been added and replaces the default behavior when computing the BCH approximation for an end-of-circuit error generator. The new implementation supports up to third-order in the Magnus expansion. The previous method, iterative pairwise application of the BCH approximation is still available, but is no longer the default method. A corresponding numerical implementation has been added to support new unit tests.
2. The logic for automatically mapping pyGSTi circuit line labels into the form required for conversion into a `stim.Tableau` has been extended to support automatic inference for the vast majority of use cases. For tricky or edge cases there are now options for manually controlling the label mapping behavior.
3. A new function has been added for analytically applying elementary error generators to paulis. A corresponding numerical implementation has been added to support corresponding unit tests.

### DIFF
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -9,7 +9,7 @@ Defines the Circuit class
 # in compliance with the License.  You may obtain a copy of the License at
 # http://www.apache.org/licenses/LICENSE-2.0 or in the LICENSE file in the root pyGSTi directory.
 #***************************************************************************************************
-
+from __future__ import annotations
 import collections as _collections
 import itertools as _itertools
 import warnings as _warnings
@@ -21,6 +21,8 @@ from pygsti.baseobjs import outcomelabeldict as _ld, _compatibility as _compat
 from pygsti.tools import internalgates as _itgs
 from pygsti.tools import slicetools as _slct
 from pygsti.tools.legacytools import deprecate as _deprecate_fn
+
+from typing import Union, Optional
 
 #Externally, we'd like to do thinks like:
 # c = Circuit( LabelList )
@@ -3762,7 +3764,9 @@ class Circuit(object):
         f.close()
 
 
-    def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=None, qubit_label_conversions=None):
+    def convert_to_stim_tableau_layers(self, gate_name_conversions: Optional[dict[str, stim.Tableau]] = None, 
+                                       num_qubits: Optional[int] = None, 
+                                       qubit_label_conversions: Optional[dict[Union[str, int], int]] = None) -> list[stim.Tableau]:
         """
         Converts this circuit to a list of stim tableau layers
 
@@ -3873,7 +3877,9 @@ class Circuit(object):
             stim_layers.append(stim_layer)
         return stim_layers
     
-    def convert_to_stim_tableau(self, gate_name_conversions=None, num_qubits=None, qubit_label_conversions=None):
+    def convert_to_stim_tableau(self, gate_name_conversions: Optional[dict[str, stim.Tableau]] = None, 
+                                       num_qubits: Optional[int] = None, 
+                                       qubit_label_conversions: Optional[dict[Union[str, int], int]] = None) -> stim.Tableau:
         """
         Converts this circuit to a stim tableau
 
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -3762,7 +3762,7 @@ class Circuit(object):
         f.close()
 
 
-    def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=None):
+    def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=None, qubit_label_conversions=None):
         """
         Converts this circuit to a list of stim tableau layers
 
@@ -3772,6 +3772,41 @@ class Circuit(object):
             A map from pygsti gatenames to standard stim tableaus. 
             If None a standard set of gate names is used from 
             `pygsti.tools.internalgates`
+
+        num_qubits : int, optional (default None)
+            Number of qubits which should be included in the each Tableau.
+            If None this value will be attempted to be inferred from the
+            Circuit's line_labels. 
+
+        qubit_label_conversions : dict, optional (default None)
+            A map from the circuit's qubit labels into integers in the range 0 to N-1,
+            where N is the number of qubits, where the integer indices indicate
+            which qubit in the stim Tableau to map a given circuit operation into.
+            If not specified an attempt will be made to infer this mapping based on the
+            circuit's qubit labels and an exception will be made if this inference is not possible.
+
+            Note: in the following line_labels = self.line_labels.
+
+            Inference is supported for line labels with the following format.
+
+            - Qubit labels that are either integers, or string representations of integers, prefixed by
+              the character 'Q' or 'q', that are monotonically increasing.
+
+            If num_qubits is not specified, then these qubit labels will be mapped into the range [0,N-1] based
+            on their index.
+
+            If num_qubits is specified then there are different subcases:
+                If num_qubits == len(line_labels) then these qubit labels will be mapped into the range [0,N-1] based
+                on their index.
+                If num_qubit > len(line_labels) we have two possibilities:
+                    1. If max(line_labels) <= num_qubits then the line_labels are mapped directly into their value
+                       in the range [0, N-1].
+                    2. If max(line_labels) > num_qubits then the line_labels are mapped into [0,N-1] using their index.
+                Note if num_qubits<len(line_labels) then an exception is raised.
+               
+            If the default conversion behavior doesn't suit your needs, or doesn't support your label format then
+            a manual dictionary should be specified.
+
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


pygsti/errorgenpropagation/errorpropagator.py
```diff
@@ -215,7 +215,7 @@ def propagate_errorgens(self, circuit, include_spam=True):
         return propagated_errorgen_layers
         
 
-    def propagate_errorgens_bch(self, circuit, bch_order=1, include_spam=True, truncation_threshold=1e-14):
+    def propagate_errorgens_bch(self, circuit, bch_order=1, include_spam=True, truncation_threshold=1e-14, mode='magnus'):
```
nkoskelo: It looks like ```mode``` is a variable that should only have 2 options at this time. Can add a type annotation for it?

At the top of the file one would need, ```from typing import Literal```
Then, in this function definition, ```mode='magnus'``` goes to ```mode: Literal['magnus', 'pairwise'] = 'magnus'```.

This way when someone tries to call the function using tab to propose options for mode, one will see ```'magnus'``` and ```'pairwise'``` as the valid strings.

test/unit/objects/test_errorgenpropagation.py
nkoskelo, line 1: Could you please update the following to make search terms easier in the future? We have spelled a variable ```max_stength``` which is actually the maximum strength.

pygsti/circuits/circuit.py
```diff
@@ -3773,6 +3773,41 @@ def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=
...
+            If num_qubits is specified then there are different subcases:
+                If num_qubits == len(line_labels) then these qubit labels will be mapped into the range [0,N-1] based
+                on their index.
+                If num_qubit > len(line_labels) we have two possibilities:
+                    1. If max(line_labels) <= num_qubits then the line_labels are mapped directly into their value
+                       in the range [0, N-1].
+                    2. If max(line_labels) > num_qubits then the line_labels are mapped into [0,N-1] using their index.
+                Note if num_qubits<len(line_labels) then an exception is raised.
+               
+            If the default conversion behavior doesn't suit your needs, or doesn't support your label format then
```
nkoskelo, line 3813: Should we also allow them to specify a callable function which can get this conversion?

pygsti/circuits/circuit.py
```diff
@@ -3814,12 +3883,46 @@ def convert_to_stim_tableau(self, gate_name_conversions=None):
...
```
nkoskelo, line 3926: Should we also allow them to specify a callable function which can get this conversion?

pygsti/errorgenpropagation/errorpropagator.py
nkoskelo, line 1: ```self.construct_stim_layers(circuit, drop_first_layer=True)``` will call into 

```python
stim_dict=standard_gatenames_stim_conversions()
stim_layers=circuit.convert_to_stim_tableau_layers(gate_name_conversions=stim_dict)
```
Shouldn't we propagate the additional parameter ```qubit_label_conversions``` from the ```convert_to_stim_tableau_layers``` to this function?

test/unit/objects/test_circuit.py
```diff
@@ -603,6 +603,46 @@ def test_from_cirq(self):
...
+        #TODO: Add correctness checks for generated Tableaus.
+        try:
+            import stim
+        except ImportError:
+            self.skipTest("Stim is required for this operation, and it does not appear to be installed.")
+
+        test_circuit = circuit.Circuit([Label('Gxpi2',0), Label('Gxpi2', 1)], line_labels=(0,1))
+        tableau0 = test_circuit.convert_to_stim_tableau()
+        #more qubits than line labels (with max line label less than number of qubits)
+        test_circuit.convert_to_stim_tableau(num_qubits=3)
```
nkoskelo, line 619: Please add a comment that this is a smoke test and if it fails will raise the error within the function.

nkoskelo (commented):
A few ease of development items which may be nice to have.

pygsti/errorgenpropagation/errorpropagator.py
coreyostrove: Will do

pygsti/circuits/circuit.py
coreyostrove: I’m not opposed to that, but probably won’t do this at the moment unless there is a clear use case for that additional flexibility. Historically we have had a maximalist tendency toward flexibility, which has added to maintenance overheads for rarely used functionality. (Just a peek behind the curtain on some of my current thinking more broadly).

pygsti/errorgenpropagation/errorpropagator.py
coreyostrove: *face palms*

Yup, that is a big oversight on my part. Thank you for catching this!

test/unit/objects/test_circuit.py
coreyostrove: I didn’t follow this, sorry. Which part of this test do you suggest I add a comment regarding? (Or maybe you meant at the top to cover the entire thing).

test/unit/objects/test_errorgenpropagation.py
coreyostrove: Will do

test/unit/objects/test_circuit.py
rileyjmurray: I think Nick is asking for a comment to the effect of "This test is just making sure we can make these calls without an exception getting raised. An uncaught exception is the only failure mode of this test." If that comment applies to an entire test then I'd say stick it at the top of said test.

pygsti/circuits/circuit.py
```diff
@@ -3783,12 +3818,46 @@ def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=
...
+                msg = 'Unable to infer a mapping from line labels of this circuit into integers in the range [0,N-1] '\
+                      +'as required for conversion to a stim.Tableau object. Please see the docstring for this method for the '\
+                      +'supported automatic inferencing. Please manually specify a qubit_label_conversions dictionary.'
+                raise RuntimeError(msg)
+        else: #validate qubit_label_conversion dictionary has the keys we need.
+            msg = 'qubit_label_conversions does not contain keys for all of the line_labels in this circuit.'
+            assert all([lbl in qubit_label_conversions for lbl in line_labels]), msg
+            msg1 = 'All qubit_label_conversions values should be ints in the range [0, num_qubits-1]'
+            assert all([isinstance(val,int) and (0 <= val <=num_qubits-1) for val in qubit_label_conversions.values()]), msg1
+
```
rileyjmurray, line 3866: Can we break out this logic into a helper function?

pygsti/tools/errgenproptools.py
```diff
@@ -459,6 +459,258 @@ def bch_approximation(errgen_layer_1, errgen_layer_2, bch_order=1, truncation_th
 
     return new_errorgen_layer_dict
 
+def magnus_expansion(errorgen_layers, magnus_order=1, truncation_threshold=1e-14):
+
```
rileyjmurray: Rouge blank line before function docstring.

pygsti/circuits/circuit.py
```diff
@@ -3762,7 +3762,7 @@ def _write_q_circuit_tex(self, filename):  # TODO
         f.close()
 
 
-    def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=None):
+    def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=None, qubit_label_conversions=None):
```
rileyjmurray: Can we add type annotations to this signature? Something like
```python
def convert_to_stim_tableau_layers(self, gate_name_conversions: Optional[dict[str,XXX]]=None, num_qubits: Optional[int]=None, qubit_label_conversions: Optional[dict[Union[int,str],int]=None) -> list[YYY]:
```
for suitable types XXX and YYY.

pygsti/circuits/circuit.py
```diff
@@ -3800,11 +3869,11 @@ def convert_to_stim_tableau_layers(self, gate_name_conversions=None, num_qubits=
             stim_layer = empty_tableau.copy()
             for sub_lbl in layer:
                 temp = gate_name_conversions[sub_lbl.name]    
-                stim_layer.append(temp, sub_lbl.qubits)
+                stim_layer.append(temp, [qubit_label_conversions[qubit_lbl] for qubit_lbl in sub_lbl.qubits])
             stim_layers.append(stim_layer)
         return stim_layers
     
-    def convert_to_stim_tableau(self, gate_name_conversions=None):
+    def convert_to_stim_tableau(self, gate_name_conversions=None, num_qubits=None, qubit_label_conversions=None):
```
rileyjmurray: I request type annotations

pygsti/tools/errgenproptools.py
```diff
@@ -459,6 +459,258 @@ def bch_approximation(errgen_layer_1, errgen_layer_2, bch_order=1, truncation_th
...
+    """
+    Function for computing the nth-order magnus expansion for a set of error generator layers.
```
rileyjmurray, line 467: Please add a reference for the math associated with this function (e.g., arXiv paper).

Please add type annotations, particularly for ``magnus_order: Literal[1,2,3]``.

pygsti/tools/errgenproptools.py
```diff
@@ -459,6 +459,258 @@ def bch_approximation(errgen_layer_1, errgen_layer_2, bch_order=1, truncation_th
...
+    Returns
+    -------
+    magnus_expansion_dict : dict
+        A dictionary with the same general structure as those in `errorgen_layers`, but with the
+        rates combined according to the selected order of the magnus expansion.
+    """
+
+    new_errorgen_layer = []
+
+    for curr_order in range(magnus_order):
```
rileyjmurray, line 502: Instead of a for-loop with if-elif-elif-else, how about 
```python
if magnus_order == 0 or magnus_order > 3:
    raise ValueError()

# do 1st order expansion
# ...
if magnus_order == 1:
    return new_errorgen_layer

# do 2nd order expansion
# ...
if magnus_order == 2:
    return new_errorgen_layer

# do 3rd order expansion
# ...
return new_errorgen_layer
```

pygsti/tools/errgenproptools.py
rileyjmurray, line 1: I have a request that in principle could apply everywhere, but I specifically ask for this file: can you use a space for the first after "``#``" when using in-line comments?

pygsti/tools/errgenproptools.py
```diff
@@ -6160,6 +6412,82 @@ def stim_pauli_string_less_than(pauli1, pauli2):
     
     return unsigned_pauli1_str < unsigned_pauli2_str
 
+def errorgen_pauli_action(errorgen, pauli):
```
rileyjmurray: Type annotations plz

pygsti/tools/errgenproptools.py
```diff
@@ -6527,6 +6855,134 @@ def pairwise_bch_numerical(mat1, mat2, order=1):
         bch_result += (1/120)*(commutator21212 - commutator12112)
     return bch_result
 
+def magnus_numerical(propagated_errorgen_layers, error_propagator, magnus_order=1):
```
rileyjmurray: Typzz

pygsti/tools/errgenproptools.py
```diff
@@ -6527,6 +6855,134 @@ def pairwise_bch_numerical(mat1, mat2, order=1):
...
+                        if j==k:
+                            magnus += (1/12)*_matrix_commutator(errorgen_layer_mats[k], _matrix_commutator(errorgen_layer_mats[j], errorgen_layer_mats[i]))
+                        else:
+                            magnus += (1/6)*_matrix_commutator(errorgen_layer_mats[k], _matrix_commutator(errorgen_layer_mats[j], errorgen_layer_mats[i]))
+        else:
+            raise NotImplementedError('Magnus beyond third order is not currently implemented.')
+        
+    return magnus  
+
+def errorgen_pauli_action_numerical(errorgen, pauli):
```
rileyjmurray: Types

test/unit/objects/test_circuit.py
nkoskelo: My understanding is that calling ```test_circuit.convert_to_stim_tableau()``` will not change the data structures within test_circuit but with create a new object, tableau. Line 616 is checking that the tableau can be made even if the number of qubits in the tableau is larger than the number of qubits used in the circuit. If that is no longer the case, then the call to ```test_circuit.convert_to_stim_tableau()``` is expected to raise the error.

In the event that one believes that the unit tests need to be executed quicker, I could see one removing this line from the test believing it does not perform any tests.

test/unit/tools/test_errgenproptools.py
rileyjmurray, line 1: Please import ``scipy.linalg as la``, then find-and-replace``np.linalg.norm`` with ``la.norm``.

rileyjmurray (changes requested):

test/unit/tools/test_errgenproptools.py
coreyostrove: Is there a particular reason to switch to scipy's implementation?

test/unit/objects/test_circuit.py
coreyostrove: Done

pygsti/circuits/circuit.py
coreyostrove: Done. Ditto for rest of the type annotation requests so will mark those as resolved.

pygsti/circuits/circuit.py
coreyostrove: I'm going to keep this as is for now. We don't have a present need to reuse this logic elsewhere, and the locality of behavior design principle would thus favor keeping it next to where it is immediately used.

pygsti/errorgenpropagation/errorpropagator.py
coreyostrove: Fixed

pygsti/tools/errgenproptools.py
coreyostrove: To be clear, you're simply asking that every comment of the form `#<some text>` -> `# <some text>`, correct?

If so, then sure I can do that, but it is going to utterly annihilate the diff (there are ~650 such comments in this file). So I will make that the very last thing I do on the branch just before merging after this is approved.

pygsti/tools/errgenproptools.py
coreyostrove: Thanks

pygsti/tools/errgenproptools.py
coreyostrove: Done

pygsti/tools/errgenproptools.py
coreyostrove: You're correct that the current version is clunky, but I don't have the time to properly refactor this now. There are other similarly clunky bits of logic in this module, and I have a tracking issue, #539, with todos for the next iteration of the module. I've added this refactor to that list.

coreyostrove:
@nkoskelo or @rileyjmurray: How do I avoid this linting error for these type annotations: https://github.com/sandialabs/pyGSTi/actions/runs/16983348064/job/48147611727#step:11:28

pygsti/tools/errgenproptools.py
rileyjmurray: Yup! That's what I'm hoping for.

pygsti/circuits/circuit.py
rileyjmurray: Reasonable! Marking as resolved.

test/unit/tools/test_errgenproptools.py
rileyjmurray: Hmm. Well, scipy.linalg has more features than numpy.linalg, but that's a mediocre case for making the replacement.

I vaguely remember some numpy documentation that said they encouraged people to move from numpy.linalg to scipy.linalg. When I look at docs now, there's no such statement. 

Given that I was most likely misremembering I think it's fine to go either way on this. I'll mark as resolved.

PR #626, opened 2025-08-04
Per seq threshold and miscellaneous QI serialization fixes

pcwysoc:
Hi Corey,

This should address the remainder of your email. There are three one line fixes:

1) Correcting the threshold calculation for the per sequence detail plot to use the number of circuits not the number of boxes.
2) Addressing a typo that means quantum instruments fail to deserialize correctly. 
3) Fixing a bug where integer qubit labels used with quantum instruments fail to serialize correctly.


pygsti/processors/processorspec.py
coreyostrove, line 1: Not to future selves, this will be getting a more permanent patch when we get `feature-instrument-availability` merged in. (Current serialization is related to hack for instrument availability.)

coreyostrove (approved):
Looks good, I left a note for our future selves with respect to one of these changes, but I think this is fine to merge in for now. Thank you for these patches, @pcwysoc!

PR #628, opened 2025-08-05
Feature mcfe integration

ndsieki:
Something


pygsti/protocols/protocol.py
```diff
@@ -1643,6 +1764,19 @@ class FreeformDesign(ExperimentDesign):
...
+
+        This interface is needed to ensure that all_circuits_needing_data
+        can be regenerated consistently during construction and deserialization.
+        
+        Returns
+        -------
+        list of Circuits
+            Keys of self.aux_info
+        """
+        return list(self.aux_info.keys())
```
nkoskelo: If aux_info is an ordinary dictionary, ```dict```, this can return a different ordering if the keys are added in a different order. This may not matter for your application though.

pygsti/tools/internalgates.py
```diff
@@ -535,6 +535,50 @@ def cirq_gatenames_standard_conversions():
     return cirq_to_standard_mapping
 
 
+def qiskit_gatenames_standard_conversions():
```
nkoskelo: Please add a type annotation.

nkoskelo, line 538: Also maybe update the function name to something like ```qiskit_gatenames_to_pygsti_stdnames```.

pygsti/tools/internalgates.py
```diff
@@ -828,12 +874,52 @@ def standard_gatenames_openqasm_conversions(version='u3'):
...
                                                           'rz(' + str(float(gatearg[0]) + _np.pi) + ')', 'sx',
                                                           'rz(' + str(float(gatearg[1]) + _np.pi) + ')']
+        std_gatenames_to_argmap['Gdelay'] = lambda gatearg: ['delay(' + str(gatearg[0]) + ')']
     else:
         raise ValueError("Unknown version!")
 
     return std_gatenames_to_qasm, std_gatenames_to_argmap
 
 
+def standard_gatenames_qiskit_conversions():
```
nkoskelo: Please add a type annotation.

pygsti/protocols/vbdataframe.py
```diff
@@ -11,9 +11,22 @@
 #***************************************************************************************************
 import pandas as _pandas
 import numpy as _np
+import tqdm as _tqdm
```
nkoskelo, line 23: Pygsti tools has version of tqdm which you could use.

pygsti/protocols/vbdataframe.py
```diff
@@ -480,3 +702,111 @@ def capability_regions(self, metric='polarization', threshold=1 / _np.e, signifi
             capreg = trimmed_capreg
 
         return capreg
+    
+
+
+    def create_vb_plot(self, title, accumulator=_np.mean, cp_or_rc='rc',
```
nkoskelo: Please add type annotations and a docstring.

pygsti/protocols/protocol.py
ndsieki: I believe this change was made in #379, which has yet to be merged in. You may get a better response to this comment if you post it there. For the applications that I am familiar with, the order does not matter.

pygsti/protocols/vbdataframe.py
ndsieki: I don't follow. Is there an issue with importing tdqm in this way?

coreyostrove: I believe @nkoskelo is alluding to the VerbosityPrinter class, which is built into pygsti and offers (among other things) a progress bar interface analogous to tdqm (with some additional niceties to account for being run with MPI).  We could refactor this to use VerbosityPrinter, but the initial use of tqdm as a dependency was/is introduced in #379 and I don't feel like having to refactor that too, so I think this is fine.

pygsti/baseobjs/label.py
```diff
@@ -1893,7 +1893,7 @@ def to_native(self):
         """
         return tuple(self)
 
-    def replacename(self, oldname, newname):
+    def replace_name(self, oldname, newname):
```
ndsieki, line 1896: Every other class that inherits from `Label` calls its corresponding function `replace_name()`. Calling `Circuit.replace_gatename()` would also return an error if any of the labels were of type `LabelTupWithArgs` due to this inconsistency.

pygsti/baseobjs/qubitgraph.py
```diff
@@ -878,7 +878,7 @@ def subgraph(self, nodes_to_keep, reset_nodes=False):
             qubit_labels = nodes_to_keep
 
         edges = []
-        for edge in self.edges():
+        for edge in self.edges(include_directions=True):
```
ndsieki: Having subgraphs retain the directions of their parent graph makes it easier to use stencils on subgraphs. I can't think of a downside to this behavior. If there is a good reason to discard the direction information in a subgraph, I think it would make sense to add a `include_directions` flag to `subgraph()`.

pygsti/tools/internalgates.py
```diff
@@ -23,7 +26,7 @@ class Gzr(_UnitaryGateFunction):
     shape = (2, 2)
 
     def __call__(self, theta):
-        return _np.array([[1., 0.], [0., _np.exp(-1j * float(theta[0]))]])
+        return _np.array([[1., 0.], [0., _np.exp(1j * float(theta[0]))]])
```
ndsieki, line 35: Defining the Gzr gate with a minus rotation is inconsistent with other resources. It is also internally inconsistent with the definition of the Gzpi2 / C14 gate: `_np.array([[1, 0], [0, 1j]], complex)`.

coreyostrove:
Quick note: I expect to be able to merge in #379 tomorrow morning. Once that is on develop I'll ping you and then let's merge develop into your branch to use as the baseline for reviewing. This will knock out a decent number of changes and reduce any redundancy.

coreyostrove:
Update: #379 is now merged into develop. It looks like there might be a couple conflicts to resolve when merging develop into your branch, so hopefully those won't be tricky. I'll start my review once that merge is done.

test/performance/mpi_2D_scaling/mpi_timings.py
ndsieki, line 1: This file was identical to `run_me_with_mpirun.py`.

ndsieki:
Merged in develop and fixed a couple of minor things. @coreyostrove, should be ready for your review.

jupyter_notebooks/Examples/MCFE-Example.ipynb
coreyostrove, line 1: It is hard to add this comment inline, but you have a cell in this example notebook under the heading "Compute process fidelity for each circuit" which contains a bunch of postprocessing functions which ultimately return a pandas dataframe with the estimate process fidelities. Is the functionality of these helper function implemented in pyGSTi proper? You mentioned in the PR notes a new method for VBDataFrame which sounded like it would superficially produce a comparable output to your helper functions.

pygsti/baseobjs/label.py
coreyostrove: Good catch!

pygsti/baseobjs/qubitgraph.py
coreyostrove: I agree, can you add this kwarg to subgraph (defaulted to True, for the reasons you've stated) and plumb this through?

pygsti/circuits/circuit.py
```diff
@@ -22,6 +25,16 @@
 from pygsti.tools import slicetools as _slct
 from pygsti.tools.legacytools import deprecate as _deprecate_fn
 
+try:
+    import qiskit
+    if qiskit.__version__ != '2.1.1':
+        _warnings.warn("Circuit class method `from_qiskit()` and method `convert_to_qiskit`"
+                       "is designed for qiskit version 2.1.1 and may not \
+                        function properly for your qiskit version, which is " + qiskit.__version__)
```
coreyostrove: We should revisit down the line whether to either revise this warning to include a broader range of acceptable versions as this gets tested with newer versions of qiskit, or if it is really important we're on a specific qiskit version lets pin that in the pyGSTi requirements.

pygsti/circuits/circuit.py
```diff
@@ -22,6 +25,16 @@
 from pygsti.tools import slicetools as _slct
 from pygsti.tools.legacytools import deprecate as _deprecate_fn
 
+try:
```
coreyostrove: Let's move the try except block for the qiskit import into the definitions of `from_qiskit` and `convert_to_qiskit`.  And with that being done have the except block actually raise that ImportError (with your additional context information), as at that point the failure of the import is critical.

pygsti/circuits/circuit.py
```diff
@@ -4072,6 +4085,197 @@ def from_cirq(cls, circuit, qubit_conversion=None, cirq_gate_conversion= None,
...
+
+            # for i in range(circuit.num_qubits):
+
+            # qubit_conversion = {i: f'Q{i}' for i in qiskit_qubits}
+
+            
+        line_labels = tuple(sorted([qubit_conversion[qubit] for qubit in qubits]))
+
+        if verbose:
+            print(qubit_conversion)
```
coreyostrove: Could you make these verbose print statements slightly more verbose? At a minimum including the variable name of the quantity being printed. (With f-strings you can do this in a one-liner, print(qubit_conversion) -> print(f'{qubit_conversion=}').

pygsti/circuits/circuit.py
```diff
@@ -4072,6 +4085,197 @@ def from_cirq(cls, circuit, qubit_conversion=None, cirq_gate_conversion= None,
...
+        else:
+            #default mapping is the identity mapping: qubit i in the Qiskit circuit maps to qubit i in the pyGSTi circuit
+            qubit_conversion = {circuit._qbit_argument_conversion(i)[0]: f'Q{i}' for i in range(circuit.num_qubits)} # in Qiskit 1.1.1, the method is called qbit_argument_conversion. In Qiskit >=1.2 (as far as Noah can tell), the method is called _qbit_argument_conversion. 
+
+            qubit_idx_conversion = {i: f'Q{i}' for i in range(circuit.num_qubits)}
+
+            # for i in range(circuit.num_qubits):
+
+            # qubit_conversion = {i: f'Q{i}' for i in qiskit_qubits}
+
```
coreyostrove: Stray comments.

pygsti/circuits/circuit.py
```diff
@@ -4072,6 +4085,197 @@ def from_cirq(cls, circuit, qubit_conversion=None, cirq_gate_conversion= None,
...
+        if verbose:
+            print(qubit_conversion)
+            print(line_labels)
+
+
+        # print(qubit_conversion)
+
+        # print(line_labels)
+
```
coreyostrove: Stray comments.

pygsti/circuits/circuit.py
```diff
@@ -4072,6 +4085,197 @@ def from_cirq(cls, circuit, qubit_conversion=None, cirq_gate_conversion= None,
...
+        for instruction in instructions:
+            # print(layer_indices)
+            if allow_different_gates_in_same_layer == False:
+                assert len(pygsti_circ_layers) == len(layer_names), "there must be one layer name for each layer!"
+
+            # print(instruction)
+            name = instruction.operation.name
+            num_qubits = instruction.operation.num_qubits
+            # instruction_qubit_indices = [qubit._index for qubit in instruction.qubits]
+            params = instruction.operation.params
```
coreyostrove, line 4311: Stray comments. I won't comment on the rest of these throughout all of the files, but if you could clean up any left over comment out print statements and whatnot from when you were debugging/testing that'd be appreciated. (Expository comments are of course welcome).

pygsti/circuits/circuit.py
```diff
@@ -4239,6 +4443,125 @@ def convert_to_quil(self,
...
+        if num_qubits is None:
+            num_qubits = self.width
+
+        # print(num_qubits)
+
+        if qubit_conversion is None:
+            qubit_conversion = {label: label for label in self.line_labels}
+        elif qubit_conversion == 'remove-Q':
+            qubit_conversion = {label: int(label[1:]) for label in self.line_labels}
+
```
coreyostrove, line 4614: The 'remove-Q' case should validate that the first character is 'Q' and that the remaining characters are numeric (and this can be cast to an int), otherwise just default to the same behavior as the case above. The following should do the trick:

`qubit_conversion = {label: (int(label[1:]) if (isinstance(label, str) and label[0]=='Q' and label[1:].isnumeric()) else label) for label in self.line_labels}`

pygsti/circuits/circuit.py
```diff
@@ -4239,6 +4443,125 @@ def convert_to_quil(self,
...
+        qiskit_gate_conversion = _itgs.standard_gatenames_qiskit_conversions()
+
+        for i in range(depth):
+            layer = self.layer_label(i).components
+            for gate in layer:
+                qiskit_gate, qiskit_gate_name, is_standard_gate = qiskit_gate_conversion[gate.name]
+                # pseudo-code
+                qiskit_qubits = [qubit_conversion[qubit] for qubit in gate.qubits]
+
+                args = [float(arg) for arg in gate.args] #mitigates an issue with float args being cast to strings somewhere in the mirror benchmarking pipeline, likely in the ibmqexperiment checkpoint and from_dir.
```
coreyostrove: This issue should be patched now (fixed it on #379).

pygsti/circuits/subcircuit_selection.py
```diff
@@ -0,0 +1,782 @@
+"""
+Utility functions for subcircuit selection
+"""
+
+#TODO: add copyright statement
```
coreyostrove: Please add this.

pygsti/circuits/subcircuit_selection.py
```diff
@@ -0,0 +1,782 @@
...
+
+try:
+    import qiskit
+    if qiskit.__version__ != '2.1.1':
+        _warnings.warn("The qiskit functionality of subcircuit selection is designed for qiskit 2.1.1." \
+        "Your version is " + qiskit.__version__)
+except:
+    _warnings.warn("Qiskit does not appear to be installed." \
+    "If providing a CouplingMap for `coupling_map`, qiskit must be installed.")
+
```
coreyostrove, line 38: Similar comment to the last qiskit import try-except block. Let's move this into the functions which accept a CouplingMap as an allowed input, or otherwise do not have a code path that works without qiskit installed. Then if the import fails (though that'd be hard to do in practice since you'd need qiskit imported to make a CouplingMap object to pass in) then you can check if you would actually hit the qiskit dependent code path and selectively raise and error only when that'd fail.

pygsti/circuits/subcircuit_selection.py
```diff
@@ -0,0 +1,782 @@
...
+            start = rand_state.choice(full_width - width)
+            end = start + width
+            qubit_subset = set(full_circ.line_labels[start:end])
+
+        elif isinstance(coupling_map, list): # this list needs to already include the 'Q' prefix on the qubits
+            G = _nx.Graph()
+            G.add_edges_from(coupling_map)
+            qubit_subset = random_connected_subgraph(G, width, rand_state)
+
+        else: # likely the coupling_map is a CouplingMap instance
```
coreyostrove: Should that be asserted?

pygsti/circuits/subcircuit_selection.py
```diff
@@ -0,0 +1,782 @@
...
+
+        for i in range(len(subcirc_layers)):
+            scl = subcirc_layers[i]
+            if len(scl) == 0: #layer is empty. add delay of appropriate duration
+                # handle the addition of a delay instruction based on the maximum duration of an instruction in this circuit layer
+                delay_layer = [_Label('Gdelay', qubit, args=[layer_durations[i]]) for qubit in qubit_subset]
+                subcirc_layers[i] = delay_layer
+
+        # Build subcircuit
+        subcirc = _Circuit(subcirc_layers, line_labels=sorted(list(qubit_subset)))
```
coreyostrove: Uber pedantic, but sorted's return type is already a list.

pygsti/extras/ibmq/ibmqexperiment.py
```diff
@@ -297,22 +306,29 @@ def partial_trace(ordered_target_indices, input_dict):
...
-                ds.add_count_dict(circ, counts_data)
-
         self.data = _ProtocolData(self.edesign, ds)
 
         if not self.disable_checkpointing:
             self.data.write(self.checkpoint_path, edesign_already_written=True)
 
-    def submit(self, ibmq_backend, start=None, stop=None, ignore_job_limit=True, wait_time=5, max_attempts=10, ibmq_session=None):
+    def submit(self, ibmq_backend, is_simulator=False, start=None, stop=None, ignore_job_limit=True, wait_time=5, max_attempts=10, ibmq_session=None):
         """
```
coreyostrove: It looks like the IBMBackend class (which I assume ibm_backend is an instance of?) has an attribute called `simulator` which returns whether a backend is a simulator or real device, see [here](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/ibm-backend). If correct then you could drop the kwarg and trigger your if statements off of that attribute.

pygsti/processors/random_compilation.py
```diff
@@ -0,0 +1,705 @@
+"""
+Randomized circuit compilation via random compiling and central Pauli propagation.
+"""
+
+#TODO: add copyright statement
```
coreyostrove: Please add this.

pygsti/processors/random_compilation.py
```diff
@@ -0,0 +1,705 @@
...
+
+
+def pauli_randomize_circuit(circ: _Circuit,
+                            rand_state: Optional[_np.random.RandomState] = None,
+                            return_bs: bool = False,
+                            return_target_pauli: bool = False,
+                            insert_test_layers: bool = False,
+                            test_layers: Optional[List[_np.ndarray]] = None
+                            ) -> _Circuit:
+    """
```
coreyostrove, line 163: I have a feature request. Not for now, but for a future revision of this code (I will open up a new github issue for tracking). At present this functionality relies on the use of the Gu3 gate for single-qubit gates, and does RC by updating those parameters. I would like to be able to RC other circuits, and a natural way to do that with minimal effort would be to add a pair of methods to the circuit class (probably together with a new mapping function in internalgates.py). One method for returning a version of a pyGSTi circuit where all of the standard gate names have been mapped to their Gu3 counterparts, and another for converting back to standard pyGSTi gate names (with proper handling for cases where that is not possible). With that it'd be easy to modify your code to work more generally with little to no change to the actual RC logic.

pygsti/protocols/mirror_edesign.py
```diff
@@ -0,0 +1,1018 @@
+"""
+Functions for creating mirror experiment designs for mirror circuit fidelity estimation (MCFE)
+"""
+
+#TODO: add copyright statement
```
coreyostrove: Please add this.

pygsti/protocols/mirror_edesign.py
```diff
@@ -0,0 +1,1018 @@
...
+
+
+try:
+    import qiskit
+    if qiskit.__version__ != '2.1.1':
+        _warnings.warn("The mirroring functions 'qiskit_circuits_to_mirror_edesign'," \
+        "'qiskit_circuits_to_fullstack_mirror_edesign', and" \
+        "'qiskit_circuits_to_subcircuit_mirror_edesign are designed for qiskit 2.1.1. Your version is " + qiskit.__version__)
+
+    from qiskit import transpile
```
coreyostrove: Similar comment to the previous ones related to this try-except block. Let's move these into the qiskit_* functions, as the core `make_mirror_edesign` doesn't rely on this, and we don't want to raise warnings for everyone if they're not using the qiskit functionality.

pygsti/protocols/vbdataframe.py
```diff
@@ -9,11 +9,42 @@
...
+
 import pandas as _pandas
 import numpy as _np
+import tqdm as _tqdm
+
+try:
+    import matplotlib.pyplot as _plt
+    import matplotlib as _mp
+except:
+    _warnings.warn('matplotlib is required for dataframe plotting, and does not appear to be installed.')
```
coreyostrove: Please move this into create_vb_plot

pygsti/tools/bsqale.py
```diff
@@ -0,0 +1,318 @@
+"""
+B-Sqale is a software tool for creating scalable, robust benchmarks from any quantum circuit.
+Please see <paper link forthcoming> for more information.
+"""
+
+# TODO: add copyright assertion
```
coreyostrove: Please add this.

pygsti/tools/bsqale.py
```diff
@@ -0,0 +1,318 @@
...
+import numpy as _np
+
+try:
+    import qiskit
+    if qiskit.__version__ != '2.1.1':
+        _warnings.warn("The B-Sqale functions 'qiskit_circuits_to_mirror_edesign'," \
+        "'qiskit_circuits_to_fullstack_mirror_edesign', and" \
+        "'qiskit_circuits_to_subcircuit_mirror_edesign are designed for qiskit 2.1.1. Your version is " + qiskit.__version__)
+    from qiskit import transpile
+
```
coreyostrove: See previous comments on moving these try and excepts to the relevant functions reliant on qiskit.

pygsti/tools/internalgates.py
```diff
@@ -18,12 +22,24 @@
...
 
+try:
+    import qiskit
+    from qiskit.circuit import Delay
+    from qiskit.circuit.library import standard_gates
+
+    if qiskit.__version__ != '2.1.1':
+        _warnings.warn("function 'standard_gatenames_qiskit_conversions()' is designed for qiskit version 2.1.1 \
+                and may not function properly for your qiskit version, which is " + qiskit.__version__)
+
```
coreyostrove: See previous comments regarding the qiskit try-except blocks.

coreyostrove (changes requested):
This is a true tour-de-force, thanks for the massive amount of effort that went into the development of these new capabilities, @ndsieki!

I have made a few suggestions and requests inline. Nothing that constituted a major concern as I recall (your set of unit tests looked rather comprehensive, which goes a long way to assuaging any concerns), and hopefully these should be fairly quick to enact. Please see in line comments, but a couple high-level recurring themes:

1. Please take a pass through the new code and give it a spring cleaning. I noticed a large number of commented out print statements (presumably from debugging and development), older/obsolete implementations, etc. If you could identify the stragglers which are no longer needed that'd be great.
2. Move try-except blocks into relevant functions, specifically for the qiskit and matplotlib imports. The rationale here is that not everyone will have qiskit installed, and most of the time that is fine, we should only raise warnings when touching functionality that requires it.

Also, we'll want to make sure all of your unit tests are passing. As of last night they weren't but it looks like you recently pushed some changes which may have addressed those.

Overall this all looks good though. Sorry it took a while to review (there was a lot of code here, in my defense). Ping me when you've had a chance to implement these changes and I strongly suspect we'll be good to go at that point to fully approve and merge this in.

pygsti/circuits/subcircuit_selection.py
```diff
@@ -0,0 +1,782 @@
...
+
+        # print(subcirc)
+        subcircs.append((subcirc, total_dropped_gates, compiled_depth, (start, end), total_dangling_gates, added_layer_indices))
+    
+        if verbosity > 0:
+            print(f'Found subcircuit with {total_dropped_gates} dropped gates, {compiled_depth} depth, and {total_dangling_gates} dangling gates')
+
+    if (failures == 1000):
```
nkoskelo: Should this be a variable global to the file which can be modified if the user is willing to wait longer?

pygsti/extras/ibmq/ibmqexperiment.py
ndsieki: One kind of simulator that is commonly used is `AerSimulator`, which can use this `submit` method but does not have the `simulator` attribute. I'm inclined to leave it as a kwarg for now. We could do something like this, I suppose:
```
is_simulator = True
if isinstance(ibmq_backend, IBMBackend):
   if not ibmq_backend.simulator:
      is_simulator = False
```
This code covers the three cases that seem relevant for what `ibmq_backend` could be: an `IBMBackend` where `simulator` is True, an `IBMBackend` where `simulator` is False, or an `AerSimulator`. Do you think this is a positive change?

pygsti/circuits/subcircuit_selection.py
coreyostrove: This is a good suggestion. A standard practice in that case is to name it using all caps, and to give it an evocative name unlikely to result in namespace conflicts. something like `MAX_STARTING_LAYER_ATTEMPTS`.

pygsti/extras/ibmq/ibmqexperiment.py
coreyostrove: Yes, but I'd go a step further and do something like this:

```
if isinstance(ibmq_backend, IBMBackend):
    is_simulator = ibmq_backend.simulator
elif isinstance(ibmq_backend, qiskit.AerSimulator):
    is_simulator = True
else:
   raise ValueError(f'ibmq_backend has an unsupported type, {type(ibmq_backend)}. Support is currently only available for arguments which are instances of `IBMBackend` or `AerSimulator`.')
```
Then just enumerate the allowed types for ibmq_backend in the docstring, and you'll be fully justified to just "nope out" on anything that isn't what you said was allowed.

tjproct (approved):
I have only given the code a brief review due to time constraints right now. It looks good to me. I suspect that we may want to, in the future, rename some functions and do bits and pieces of minor restructuring / removing redundancy with other benchmarking code (e.g., I think there might be redundancy with data analysis code used for mirror RB). However, I think it makes sense to accept this now, and to update the code when we have time in the future (and prior to final publication of Noah's paper that's based around this code). In particular, I (personally) anticipate spending some time in the fall revamping the broader pyGSTi benchmarking code as part of developing some new benchmarks + merging in some hacked-together new benchmarking capabilities that currently exist on a feature branch. During that process it would likely make sense to make small tweaks to this code and remove redundancies at that point.

PR #635, opened 2025-08-12
Add implicit zero to fractions where applicable

nkoskelo:
I noticed that we have some fractions of the form ``` .{0-9}{0-9}*```. This could result in someone missing the decimal point when reading the fraction quickly and depending on the specific number that could have varying degrees of consequence. This PR adds in the implicit ```0``` to the fractions.

We also want to be careful if we display users fractions with an implied zero.


pygsti/report/workspaceplots.py
```diff
@@ -3833,10 +3833,10 @@ def _create(self, projections,
             new_lmargin = fig.plotlyfig.layout.margin.l - 20
 
             #add annotations for the upper and lower triangles.
-            fig.plotlyfig.add_annotation(x= .92, y= -.16, xref= 'paper', yref= 'paper',
+            fig.plotlyfig.add_annotation(x= 0.92, y= -.16, xref= 'paper', yref= 'paper',
```
rileyjmurray: Missing the ``y=-0.16``; re-run the grep (or whatever else) to include strings of the form ``-.{0-9}{0-9}*``

test/unit/tools/test_lindbladtools.py
```diff
@@ -169,7 +169,7 @@ def test_error_metric_restrictions(self):
         random_errorgen_rates = lt.random_CPTP_error_generator_rates(num_qubits=2, errorgen_types=('H','S'), 
                                                                 error_metric= 'generator_infidelity', 
                                                                 error_metric_value=.99, 
-                                                                relative_HS_contribution=(.5, .5), seed=1234)
+                                                                relative_HS_contribution=(.5, 0.5), seed=1234)
```
rileyjmurray: missing the first ``.5``

test/unit/tools/test_lindbladtools.py
```diff
@@ -184,7 +184,7 @@ def test_error_metric_restrictions(self):
         random_errorgen_rates = lt.random_CPTP_error_generator_rates(num_qubits=2, errorgen_types=('H','S'), 
                                                                 error_metric= 'total_generator_error', 
                                                                 error_metric_value=.99, 
-                                                                relative_HS_contribution=(.5, .5), seed=1234)
+                                                                relative_HS_contribution=(.5, 0.5), seed=1234)
```
rileyjmurray: missing the first ``.5``

test/unit/tools/test_lindbladtools.py
```diff
@@ -184,7 +184,7 @@ def test_error_metric_restrictions(self):
         random_errorgen_rates = lt.random_CPTP_error_generator_rates(num_qubits=2, errorgen_types=('H','S'), 
                                                                 error_metric= 'total_generator_error', 
                                                                 error_metric_value=.99,
```
rileyjmurray: missing the leading zero, ``0.99``

Add a grep for ``=.{0-9}{0-9}*``

rileyjmurray (changes requested):

rileyjmurray (approved):

PR #639, opened 2025-08-15
Use truthy checks. Do not check explicitly against True and False.

### LINKED ISSUE #638: Direct comparisons to True and False can result in interesting behavior
```python
var = 2
if var:
     print(True)
```

will print True but

```python
var = 2
if var == True:
    print(True)
```

will not print True in python 3.11.

We have a few locations in ```germselection.py``` and ```fiducial_pair_reduction.py``` where we directly compare against True and False.

If ```var``` is restricted to the ```bool``` type then this could be okay. However, if the type changes, we could be back into the error regime.

nkoskelo:
fixes #638


rileyjmurray (approved):
LGTM

Issue #620, opened 2025-07-30
Parameterization-preserving gauge optimization


rileyjmurray:
Consider the following fact.

> Let $M_1$ be a model (e.g., an ExplicitOpModel) with parameterization $P_1$ (e.g., "CPTPLND" or "full TP"). Next, let $M_2$ be the representation of $M_1$ under a parameterization $P_2$, where $P_2$ allows for a strictly larger hypothesis class than does $P_1$.
> 
> We can't, in general, get confidence intervals for $M_1$ by looking at the likelihood function for $M_2$.

This fact is troublesome for computing confidence intervals of gauge-dependent quantities, since pyGSTi's gauge optimization tends to be performed in an expanded hypothesis class. This is the case even if we do gauge optimization on a CPTP model (CPTPLND parameterization) over the unitary group.

It's nominally straightforward to avoid this issue by gauge optimizing in whatever parameterization we want (e.g., Full), getting the gauge transformation, and then substituting modelmembers with suitably-transformed versions of themselves. The transformation + substitution can be facilitated with ComposedState, ComposedOp, and ComposedPOVM, as shown in the following code.
```python
def transform_composedop_model(mdl : ExplicitOpModel, s : GaugeGroupElement):
    oldmdl = mdl
    mdl = oldmdl.copy()

    U    = StaticArbitraryOp(s.transform_matrix,         basis=oldmdl.basis)
    invU = StaticArbitraryOp(s.transform_matrix_inverse, basis=oldmdl.basis) 

    for key, rho in oldmdl.preps.items():
        assert isinstance(rho, ComposedState)
        static_rho = rho.state_vec
        errmap  = ComposedOp([rho.error_map, invU])
        mdl.preps[key] = ComposedState(static_rho, errmap)

    for key, povm in oldmdl.povms.items():
        assert isinstance(povm, ComposedPOVM)
        static_povm = povm.base_povm
        errmap = ComposedOp([U, povm.error_map])
        mdl.povms[key] = ComposedPOVM(errmap, static_povm, mx_basis=oldmdl.basis)

    for key, op in oldmdl.operations.items():
        op_s = ComposedOp([U, op, invU])
        mdl.operations[key] = op_s

    assert len(oldmdl.factories) == 0
    assert len(oldmdl.instruments) == 0

    mdl._clean_paramvec()  # transform may leave dirty members
    return mdl
```
A whole gauge optimization pipeline could look something like this.
```python
[...]

PR #640, opened 2025-08-18
Miscellaneous Bugfixes for 0.9.14

coreyostrove:
This PR consists of miscellaneous bug fixes and patches in preparation of the 0.9.14 release which were being caught by tests within the `test_packages` submodule (which are run against pushes to beta and master). Here is a brief summary:

- Fixed bug in new logic for conversion of state preparation and POVMs to an error generator parameterization when using the 'statevec' evotype.
- Added documentation to `convert_members_inplace` and `set_default_gauge_group_for_member_type` methods of the `ExplicitModel` class. 
- Updated default behavior of  `set_default_gauge_group_for_member_type` for error generator parameterized models. Previously this was configured to set the default gauge group for such models to the `TrivialGaugeGroup`. Now we set this to be the `UnitaryGaugeGroup`, as one would more likely expect.
- Fixed deserialization of older models from before the addition of serialized FOGI stores.
- Fixed bug in the half diamond norm in `pygsti.report.reportables` related to a recent change in the return signature for the `diamonddist` function in `optools`.
- Minor bugfixes for edge cases introduced in recent report generation changes (unbound variable names for a code branch, and whatnot).
- Expired deprecations for legacy json and msgpack serialization interfaces. Some of the recent changes on develop broke some of the legacy serialization codepaths using json (note I am not referring to the newer json serialization support natively provided through the `NicelySerializable` interface) and msgpack. We formally deprecated this functionality just over two years and had these marked for removal, so any users who might've been using this have been getting warnings for some time. Rather than invest any time fixing this, it makes more sense to finally expire these deprecations. 
- New kwargs for controlling gauge optimization settings have been added to the GST driver functions `run_long_sequence_gst` and `run_long_sequence_gst_base`. At the same time the default gauge optimization suite used for these functions has been switched to 'stdgaugeopt'. This brings these functions in line with their OOP-based counterpart `GateSetTomography` which has used this gauge optimization suite as the default for a while.
- Fixed a bug in the `set_errorgen_coefficients` method of `ComposedOp` to add handling for the cases where the update dictionary is empty, or has keys which do not match the `ElementaryErrorgenLabel` type of the modelmember.
- Remove license field from pyproject.toml. See related commit message for more details, but TLDR not all versions of setuptools support properly this at the moment, which caused builds to fail on some versions of python we test against.
- Test file artifacts for `test_packages` have been regenerated to account for recent default setting changes.
- A bunch of unit tests have been updated to account for default setting changes.

With these updates we have the tests on beta now passing (at least when run locally on my system, TBD on runners).


pygsti/modelmembers/operations/composedop.py
```diff
@@ -637,8 +637,8 @@ def transform_inplace(self, s):
             #SPECIAL CASE / HACK: for 1 & 2Q, when holding e^L * T, where T is a static gate
             # then try to gauge transform by setting e^L directly and leaving T alone:
             Smx = s.transform_matrix; Si = s.transform_matrix_inverse
-            Tinv = _np.linalg.inv(self.factorops[0].to_dense("minimal"))
-            trans_eLT = _np.dot(Si, _np.dot(self.to_dense("minimal"), Smx))
+            Tinv = _np.linalg.inv(self.factorops[0].to_dense("HilbertSchmidt"))
```
nkoskelo, line 640: Why are we limiting this hack for ```state_space.dim <= 16```?

pygsti/modelmembers/povms/__init__.py
```diff
@@ -632,6 +632,7 @@ def _objfn(v):
             else:
                 raise ValueError("Invalid to_type argument: %s" % to_type)
         except Exception as e:
+            raise e
             error_msgs[to_type] = str(e)  # try next to_type
```
nkoskelo: This line should not run if we re-raise the error beforehand.

pygsti/modelmembers/operations/composedop.py
coreyostrove: Presumably just because it is expensive to perform the dense operations required with 3 or more qubits, but is faster than whatever other code path is used for 2Q and 3Q. This is incidentally not the biggest problem with this function at the moment, though. See #620 for more on that.

pygsti/modelmembers/povms/__init__.py
coreyostrove: Correct, that is left over from some debugging. I can nix it.

pygsti/models/explicitmodel.py
```diff
@@ -440,14 +493,14 @@ def set_all_parameterizations(self, gate_type, prep_type="auto", povm_type="auto
...
-            self.convert_members_inplace(povmtyp, 'povms', 'all', flatten_structure=True, ideal_model=static_model, cptp_truncation_tol = cptp_truncation_tol, spam_cp_penalty = spam_cp_penalty)
-        except ValueError as e:
-            raise ValueError("Failed to convert members. If converting to CPTP-based models, " +
-                "try providing an ideal_model to avoid possible branch cuts.") from e
+        #try:
+        self.convert_members_inplace(typ, 'operations', 'all', flatten_structure=True, ideal_model=static_model, cptp_truncation_tol = cptp_truncation_tol)
+        self.convert_members_inplace(ityp, 'instruments', 'all', flatten_structure=True, ideal_model=static_model, cptp_truncation_tol = cptp_truncation_tol)
+        self.convert_members_inplace(rtyp, 'preps', 'all', flatten_structure=True, ideal_model=static_model, cptp_truncation_tol = cptp_truncation_tol, spam_cp_penalty = spam_cp_penalty)
+        self.convert_members_inplace(povmtyp, 'povms', 'all', flatten_structure=True, ideal_model=static_model, cptp_truncation_tol = cptp_truncation_tol, spam_cp_penalty = spam_cp_penalty)
+        #except ValueError as e:
```
nkoskelo: Can we remove the commented out try-except?

coreyostrove: This is the second half of the left over debugging stuff from the other comment. I can revert this.

test/test_packages/drivers/test_nqubit.py
```diff
@@ -162,7 +162,8 @@ def test_2Q(self):
         results = pygsti.run_long_sequence_gst_base(ds, mdl_to_optimize,
                                                     lsgstLists, gauge_opt_params=False,
                                                     advanced_options={'tolerance': 1e-1, 'max_iterations': 5}, verbosity=0,
-                                                    disable_checkpointing= True) #probably don't care about convergence for same reason we
+                                                    disable_checkpointing= True,
+                                                    gauge_opt_suite_name='none') #probably don't care about convergence for same reason we
```
nkoskelo, line 166: string ```'none'``` -> ```None```.

coreyostrove: String 'none' is correct here. (None is also an allowed input, technically, though).

pygsti/report/workspacetables.py
```diff
@@ -460,6 +463,9 @@ def _create(self, models, titles, display_as, confidence_region_infos):
         
         if confidence_region_infos is None:
             confidence_region_infos = [None]*len(models)
+        elif not isinstance(confidence_region_infos, list):
+            confidence_region_infos = [confidence_region_infos]
```
nkoskelo, line 467: ```list(confidence_region_infos)```

pygsti/modelmembers/operations/lindbladerrorgen.py
```diff
@@ -1161,6 +1161,8 @@ def set_coefficients(self, elementary_errorgens, action="update", logscale_nonha
         -------
         None
         """
+        if not elementary_errorgens:
```
nkoskelo, line 1167: The parameter ```action``` appears to be a literal with the options ["update", "add", "reset"] could you please add a type annotation indicating as such?

pygsti/report/workspacetables.py
coreyostrove: This actually doesn't achieve what we want, as if confidence_region_infos is a non-list iterable it will unpack that into a list (I just want a list containing confidence_region_infos).

pygsti/modelmembers/operations/lindbladerrorgen.py
coreyostrove: done

pygsti/models/explicitmodel.py
coreyostrove: Fixed

pygsti/modelmembers/povms/__init__.py
coreyostrove: Fixed

PR #641, opened 2025-08-19
WIP: crosstalk-free GST

nkoskelo:
TODO: Add description.

PR #643, opened 2025-08-19
Upstream xfgst incidentals

rileyjmurray:
The changes in this PR were originally made for crosstalk-free GST work (see #641). We've factored out those changes into this PR to reduce the size of #641 (and get these incidental changes in sooner rather than later).


nkoskelo:
@coreyostrove, @rileyjmurray this is ready for your review when you get the chance.

pygsti/circuits/circuit.py
```diff
@@ -276,9 +294,11 @@ def from_tuple(cls, tup):
...
             return cls(tup)
 
-    def __init__(self, layer_labels=(), line_labels='auto', num_lines=None, editable=False,
-                 stringrep=None, name='', check=True, expand_subcircuits="default",
-                 occurrence=None, compilable_layer_indices=None):
+    def __init__(self,
+            layer_labels: Optional[Union[str, Iterable[Union[Iterable[_Label], _Label]]]]=(), line_labels: Union[str,Tuple[Any,...]] = 'auto', num_lines: Optional[int] = None,
+            editable: Optional[bool]=False, stringrep=None, name: Optional[str]='', check:Optional[bool]=True, expand_subcircuits: Union[bool, Literal['default']]="default", occurrence=None,
+            compilable_layer_indices=None
+        ):
```
rileyjmurray, line 308: todo for me: one line per argument in this bad boy

pygsti/circuits/circuit.py
```diff
@@ -380,22 +403,22 @@ def __init__(self, layer_labels=(), line_labels='auto', num_lines=None, editable
...
+                    raise ValueError(("Error initializing Circuit: "
                                       " `occurrence` and occurrence ID in `layer_labels` do not match: %s != %s")
                                      % (occurrence, chk_occurrence))
 
             if chk_compilable_inds is not None:
                 if compilable_layer_indices is None:  # Also acts as "auto"
                     compilable_layer_indices = chk_compilable_inds
                 elif compilable_layer_indices != chk_compilable_inds:
-                    raise ValueError(("Error intializing Circuit: `compilable_layer_indices` and markers"
+                    raise ValueError(("Error initializing Circuit: `compilable_layer_indices` and markers"
```
rileyjmurray, line 428: todo for me: try to define a helper function to handle the entire ``if isinstance(layer_labels, str)`` codepath.

pygsti/circuits/circuit.py
```diff
@@ -435,15 +458,15 @@ def __init__(self, layer_labels=(), line_labels='auto', num_lines=None, editable
...
+                    raise ValueError(("Error initializing Circuit: "
                                       " `occurrence` and occurrence ID in `layer_labels` do not match: %s != %s")
                                      % (occurrence, chk_occurrence))
 
             if chk_compilable_inds is not None:
                 if compilable_layer_indices is None:  # Also acts as "auto"
                     compilable_layer_indices = chk_compilable_inds
                 elif compilable_layer_indices != chk_compilable_inds:
-                    raise ValueError(("Error intializing Circuit:  `compilable_layer_indices` and markers"
+                    raise ValueError(("Error initializing Circuit:  `compilable_layer_indices` and markers"
```
rileyjmurray, line 476: todo for me: try and define a helper function to handle the ``if stringrep is not None and (layer_labels is None or check)`` codepath.

pygsti/circuits/circuit.py
```diff
@@ -1166,7 +1189,9 @@ def _replace_layer_component(self, ilayer, indx, val):
             assert(indx == 0), "Only index 0 exists for a single-simple-Label level"
             self._labels[ilayer] = val
 
-    def extract_labels(self, layers=None, lines=None, strict=True):
+    def extract_labels(self, layers:Optional[Union[int, slice, list[int], tuple[int, ...]]]=None,
+                       lines:Optional[Union[int, slice, list[int], tuple[int, ...]]]=None,
+                       strict: Optional[bool]=True) -> Union[Circuit, _Label]:
```
rileyjmurray: todo for me: search for the ``Optional[Union[...`` pattern and just use ``Union[None, ...`` instead. (I find the latter more readable.)

pygsti/circuits/circuit.py
```diff
@@ -52,6 +56,19 @@
...
+
+##############################################################################################
+# Some types
+_NestedLabelSeq = Sequence[Union[_Label, Sequence[_Label]]]
+#   ^ An alias to make it easier to see how subsequent types relate.
+#     Don't use this in function signatures.
+LayerTupLike = Union[Tuple[_LabelTupTup,    ...], _NestedLabelSeq, Tuple[_Label, ...]]
+LabelsLike   = Union[Tuple[_NestedLabelSeq, ...], _NestedLabelSeq]
+IndexStringOrInt = Union[Union[str, int], slice, List[Union[str, int]], Tuple[Union[str, int], ...]]
+IndexIntOnly = Union[int, slice, List[int], Tuple[int, ...]]
```
rileyjmurray: There are a ton of function signatures where an argument called ``lines`` is typed as ``IndexStringOrInt``. It that typing rule is universal, maybe we define a ``LinesLike`` type alias that's equal to ``IndexStringOrInt``. (I think it would be okay to use ``IndexStringOrInt`` on its own when not applying to ``lines`` variables.)

pygsti/circuits/circuit.py
```diff
@@ -4271,11 +4308,11 @@ def from_qiskit(cls,
             assert len(unmapped_qubits) == 0, f'Missing Qiskit to pygsti conversions for some qubits: {unmapped_qubits}'
 
             qubit_idx_conversion = {i: qubit_conversion[circuit._qbit_argument_conversion(i)[0]] for i in range(circuit.num_qubits)}
-                    
+
         #if it is None, build a default mapping.
         else:
             # default mapping is the identity mapping: qubit i in the Qiskit circuit maps to qubit i in the pyGSTi circuit
-            qubit_conversion = {circuit._qbit_argument_conversion(i)[0]: f'Q{i}' for i in range(circuit.num_qubits)} # in Qiskit 1.1.1, the method is called qbit_argument_conversion. In Qiskit >=1.2 (as far as Noah can tell), the method is called _qbit_argument_conversion. 
+            qubit_conversion = {circuit._qbit_argument_conversion(i)[0]: f'Q{i}' for i in range(circuit.num_qubits)} # in Qiskit 1.1.1, the method is called qbit_argument_conversion. In Qiskit >=1.2 (as far as Noah can tell), the method is called _qbit_argument_conversion.
```
rileyjmurray: todo for me: shorten line

pygsti/circuits/circuit.py
```diff
@@ -4970,7 +5010,7 @@ class CompressedCircuit(object):
         takes more time but could result in better compressing.
     """
 
-    def __init__(self, circuit, min_len_to_compress=20, max_period_to_look_for=20):
+    def __init__(self, circuit: Circuit, min_len_to_compress:Optional[int]=20, max_period_to_look_for:Optional[int]=20):
```
rileyjmurray: ``Optional[int]`` is equivalent to ``Union[int, None]``. Is that really intended?

pygsti/evotypes/evotype.py
```diff
@@ -23,6 +25,9 @@ class Evotype(object):
         in order to get a performance gain.
     """
     default_evotype = None
+    Castable = Union['Evotype',
+                     Literal['default'],
+                     str] # Type alias to be used for type checking.
```
rileyjmurray: How practical would it be for us to enumerate all evotypes? I dislike the use of ``str`` as an available type.

pygsti/circuits/circuit.py
rileyjmurray: Impractical. The issue is that the 140 lines of input handling involves overwriting the values of the input arguments. Very few variables are introduced for the input parsing that aren't used later on.

What we could do is pull the input handling out into a static method. This would have the advantage of showing non-dependence on ``self``. This is what it would look like. (Note that ``_normalize_init_inputs`` includes some references in the comments to ``self``; those references are spurious.)
```python
    @staticmethod
    def _normalize_init_inputs(
            layer_labels: Optional[Union[str, Iterable[Union[Iterable[_Label], _Label]]]],
            line_labels: Union[str,Tuple[Any,...]],
            num_lines: Optional[int],
            editable: Optional[bool],
            stringrep,
            name: str,
            check: bool,
            expand_subcircuits: Union[bool, Literal['default']],
            occurrence,
            compilable_layer_indices
        ):
        from pygsti.circuits.circuitparser import CircuitParser as _CircuitParser
        layer_labels_objs = None  # layer_labels elements as Label objects (only if needed)
        if isinstance(layer_labels, str):
            if stringrep is None:  # then take the given string as the initial string rep
                stringrep = layer_labels
                check = False  # no need to check whether this matches since we're parsing it now (below)
            cparser = _CircuitParser(); cparser.lookup = None
            layer_labels, chk_labels, chk_occurrence, chk_compilable_inds = cparser.parse(layer_labels)
            if chk_labels is not None:
                if line_labels == 'auto':
                    line_labels = chk_labels
                elif tuple(line_labels) != chk_labels:
                    raise ValueError(("Error initializing Circuit: "
                                      " `line_labels` and line labels in `layer_labels` do not match: %s != %s")
                                     % (line_labels, chk_labels))
            if chk_occurrence is not None:
                if occurrence is None:  # Also acts as "auto"
                    occurrence = chk_occurrence
                elif occurrence != chk_occurrence:
                    raise ValueError(("Error initializing Circuit: "
                                      " `occurrence` and occurrence ID in `layer_labels` do not match: %s != %s")
                                     % (occurrence, chk_occurrence))

            if chk_compilable_inds is not None:
                if compilable_layer_indices is None:  # Also acts as "auto"
                    compilable_layer_indices = chk_compilable_inds
                elif compilable_layer_indices != chk_compilable_inds:
                    raise ValueError(("Error initializing Circuit: `compilable_layer_indices` and markers"
                                      " in `layer_labels` do not match: %s != %s")
                                     % (compilable_layer_indices, chk_compilable_inds))

        if expand_subcircuits == "default":
            expand_subcircuits = Circuit.default_expand_subcircuits
        if expand_subcircuits and layer_labels is not None:
            layer_labels_objs = tuple(_itertools.chain(*[x.expand_subcircuits() for x in map(to_label, layer_labels)]))

        #Parse stringrep if needed
        if stringrep is not None and (layer_labels is None or check):
            cparser = _CircuitParser()
            cparser.lookup = None  # lookup - functionality removed as it wasn't used
            chk, chk_labels, chk_occurrence, chk_compilable_inds = cparser.parse(stringrep)  # tuple of Labels
            if expand_subcircuits and chk is not None:
                chk = tuple(_itertools.chain(*[x.expand_subcircuits() for x in map(to_label, chk)]))

            if layer_labels is None:
                layer_labels = chk
            else:  # check == True
                if layer_labels_objs is None:
                    layer_labels_objs = tuple(map(to_label, layer_labels))
                if layer_labels_objs != tuple(chk):
                    raise ValueError(("Error initializing Circuit: "
                                      " `layer_labels` and `stringrep` do not match: %s != %s\n"
                                      "(set `layer_labels` to None to infer it from `stringrep`)")
                                     % (layer_labels, stringrep))
            if chk_labels is not None:
                if line_labels == 'auto':
                    line_labels = chk_labels
                elif tuple(line_labels) != chk_labels:
                    raise ValueError(("Error initializing Circuit: "
                                      " `line_labels` and `stringrep` do not match: %s != %s (from %s)\n"
                                      "(set `line_labels` to None to infer it from `stringrep`)")
                                     % (line_labels, chk_labels, stringrep))

            if chk_occurrence is not None:
                if occurrence is None:  # Also acts as "auto"
                    occurrence = chk_occurrence
                elif occurrence != chk_occurrence:
                    raise ValueError(("Error initializing Circuit: "
                                      " `occurrence` and occurrence ID in `layer_labels` do not match: %s != %s")
                                     % (occurrence, chk_occurrence))

            if chk_compilable_inds is not None:
                if compilable_layer_indices is None:  # Also acts as "auto"
                    compilable_layer_indices = chk_compilable_inds
                elif compilable_layer_indices != chk_compilable_inds:
                    raise ValueError(("Error initializing Circuit:  `compilable_layer_indices` and markers"
                                      " in `layer_labels` do not match: %s != %s")
                                     % (compilable_layer_indices, chk_compilable_inds))

        if layer_labels is None:
            raise ValueError("Must specify `stringrep` when `layer_labels` is None")

        # Set self._line_labels
        if line_labels == 'auto':
            if layer_labels_objs is None:
                layer_labels_objs = tuple(map(to_label, layer_labels))
            explicit_lbls = _accumulate_explicit_sslbls(layer_labels_objs)
            if len(explicit_lbls) == 0:
                if num_lines is not None:
                    assert(num_lines >= 0), "`num_lines` must be >= 0!"
                    if len(layer_labels) > 0:
                        assert(num_lines > 0), "`num_lines` must be > 0!"
                    my_line_labels = tuple(range(num_lines))
                elif len(layer_labels) > 0 or not editable:
                    my_line_labels = ('*',)  # special single line-label when no line labels are given
                else:
                    my_line_labels = ()  # empty *editable* circuits begin with zero line labels (this is ok)
            else:
                my_line_labels = tuple(sorted(explicit_lbls))
        else:
            explicit_lbls = None
            my_line_labels = tuple(line_labels)

        if (num_lines is not None) and (num_lines != len(my_line_labels)):
            if num_lines > len(my_line_labels) and \
               set(my_line_labels).issubset(set(range(num_lines))):
                # special case where we just add missing integer-labeled line(s)
                my_line_labels = tuple(range(num_lines))
            else:
                raise ValueError("`num_lines` was expected to be %d but equals %d!" %
                                 (len(my_line_labels), num_lines))

        if check:
            if explicit_lbls is None:
                if layer_labels_objs is None:
                    layer_labels_objs = tuple(map(to_label, layer_labels))
                explicit_lbls = _accumulate_explicit_sslbls(layer_labels_objs)
            if not set(explicit_lbls).issubset(my_line_labels):
                raise ValueError("line labels must contain at least %s" % str(explicit_lbls))

        #Set compute self._labels, which is either a nested list of simple labels (non-static case)
        #  or a tuple of Label objects (static case)
        if not editable:
            if layer_labels_objs is None:
                layer_labels_objs = tuple(map(to_label, layer_labels))
            labels = layer_labels_objs
        else:
            labels = [_label_to_nested_lists_of_simple_labels(layer_lbl)
                      for layer_lbl in layer_labels]

        # check that all the compilable layer indices are valid
        if compilable_layer_indices is not None:
            max_layer_index = len(labels) - 1
            if any([(i < 0 or i > max_layer_index) for i in compilable_layer_indices]):
                raise ValueError("Entry out of range in `compilable_layer_indices`!")
            compilable_layer_indices_tup = tuple(compilable_layer_indices)
        else:
            compilable_layer_indices_tup = ()

        return labels, my_line_labels, editable, name, stringrep, occurrence, compilable_layer_indices_tup

    def __init__(self,
            layer_labels: Optional[Union[str, Iterable[Union[Iterable[_Label], _Label]]]]=(),
            line_labels: Union[str,Tuple[Any,...]] = 'auto',
            num_lines: Optional[int]=None,
            editable: Optional[bool]=False,
            stringrep=None,
            name: Optional[str]='',
            check: Optional[bool]=True,
            expand_subcircuits: Union[bool, Literal['default']]="default",
            occurrence=None,
            compilable_layer_indices=None
        ):
        """
        Creates a new Circuit object, encapsulating a quantum circuit.

        You only need to supply the first `layer_labels` argument, though
        usually (except for just 1 or 2 qubits) you'll want to also supply
        `line_labels` or `num_lines`.  If you'll be adding to or altering
        the circuit before using it, you should set `editable=True`.

        Parameters
        ----------
        layer_labels : iterable of Labels or str
            This argument provides a list of the layer labels specifying the
            state preparations, gates, and measurements for the circuit.  This
            argument can also be a :class:`Circuit` or a string, in which case
            it is parsed as a text-formatted circuit.  Internally this will
            eventually be converted to a list of `Label` objects, one per layer,
            but it may be specified using anything that can be readily converted
            to a Label objects.  For example, any of the following are allowed:

            - `['Gx','Gx']` : X gate on each of 2 layers
            - `[Label('Gx'),Label('Gx')]` : same as above
            - `[('Gx',0),('Gy',0)]` : X then Y on qubit 0 (2 layers)
            - `[[('Gx',0),('Gx',1)],[('Gy',0),('Gy',1)]]` : parallel X then Y on qubits 0 & 1

            `Note` Labels stored in an iterable type with a non-deterministic
            iteration order like dictionaries and sets, may result in unexpected layer orders.

        line_labels : iterable, optional
            The (string valued) label for each circuit line.  If `'auto'`, then
            `line_labels` is taken to be the list of all state-space labels
            present within `layer_labels`.  If there are no such labels (e.g.
            if `layer_labels` contains just gate names like `('Gx','Gy')`), then
            the special value `'*'` is used as a single line label.

        num_lines : int, optional
            Specify this instead of `line_labels` to set the latter to the
            integers between 0 and `num_lines-1`.

        editable : bool, optional
            Whether the created `Circuit` is created in able to be modified.  If
            `True`, then you should call `done_editing()` once the circuit is
            completely assembled, as this makes the circuit read-only and
            allows it to be hashed.

        stringrep : string, optional
            A string representation for the circuit.  If `None` (the default),
            then this will be generated automatically when needed.  One
            reason you'd want to specify this is if you know of a nice compact
            string representation that you'd rather use, e.g. `"Gx^4"` instead
            of the automatically generated `"GxGxGxGx"`.  If you want to
            initialize a `Circuit` entirely from a string representation you
            can either specify the string in as `layer_labels` or set
            `layer_labels` to `None` and `stringrep` to any valid (one-line)
            circuit string.

        name : str, optional
            A name for this circuit (useful if/when used as a block within
            larger circuits).

        check : bool, optional
            Whether `stringrep` should be checked against `layer_labels` to
            ensure they are consistent, and whether the labels in `layer_labels`
            are a subset of `line_labels`.  The only reason you'd want to set
            this to `False` is if you're absolutely sure `stringrep` and
            `line_labels` are consistent and want to save computation time.

        expand_subcircuits : bool or "default"
            If `"default"`, then the value of `Circuit.default_expand_subcircuits`
            is used.  If True, then any sub-circuits (e.g. anything exponentiated
            like "(GxGy)^4") will be expanded when it is stored within the created
            Circuit.  If False, then such sub-circuits will be left as-is.  It's
            typically more robust to expand sub-circuits as this facilitates
            comparison (e.g. so "GxGx" == "Gx^2"), but in cases when you have
            massive exponents (e.g. "Gx^8192") it may improve performance to
            set `expand_subcircuits=False`.

        occurrence : hashable, optional
            A value to set as the "occurrence id" for this circuit.  This
            value doesn't affect the circuit an any way except by affecting
            it's hashing and equivalence testing.  Circuits with different
            occurrence ids are *not* equivalent.  Occurrence values effectively
            allow multiple copies of the same circuit to be stored in a
            dictionary or :class:`DataSet`.

        compilable_layer_indices : tuple, optional
            The circuit-layer indices that may be internally altered (but retaining the
            same target operation) and/or combined with the following circuit layer
            by a hardware compiler when executing this circuit.  Layers that are
            not "compilable" are effectively followed by a *barrier* which prevents
            the hardware compiler from restructuring the circuit across the layer
            boundary.
        """
        labels, my_line_labels, editable, name, stringrep, occurrence, compilable_layer_indices_tup = \
            Circuit._normalize_init_inputs(
                layer_labels, line_labels, num_lines, editable, stringrep,
                name, check, expand_subcircuits, occurrence, compilable_layer_indices
            )

        #Set *all* class attributes (separated so can call bare_init separately for fast internal creation)
        self._bare_init(labels, my_line_labels, editable, name, stringrep,
                        occurrence, compilable_layer_indices_tup)
```

pygsti/circuits/circuit.py
rileyjmurray: Impractical. See https://github.com/sandialabs/pyGSTi/pull/643#discussion_r2718316838.

pygsti/circuits/circuit.py
rileyjmurray: Turns out that 9 out of the 12 instances of this pattern do care about None as a value. Most of those use None to signal an index that would usually be represented as -1 assuming standard Python wrap-around indexing. Alas, None is officially supported in those functions.

I'm fixing the other three instances.

rileyjmurray: Also fixing similar issues with ``Optional[bool]``.

pygsti/circuits/circuit.py
rileyjmurray: Turns out that IndexStringOrInt was _only_ used for these ``lines`` variables. So I've renamed to LinesLike.

@nkoskelo, are you cool with the rename? Seems ``IndexIntOnly`` is only used for variables named ``layers``, but it's certainly a bad idea to rename ``IndexIntOnly`` to ``LayersLike``.

pygsti/evotypes/evotype.py
rileyjmurray: Perfectly practical! Turns out the the string is passed as the ``name`` argument to the Evotype constructor, which looks like this
```python
    def __init__(self, name, prefer_dense_reps=False):
        self.name = name
        self.module = _importlib.import_module("pygsti.evotypes." + self.name)
        self.prefer_dense_reps = prefer_dense_reps
```
The import will fail if ``name`` is not one of the sub-packages inside pygsti.evotypes (of which there are 7).

pygsti/circuits/circuit.py
nkoskelo: How about something like `IndexingLayers` since an `int` is not similar to a `layer` but can be used to index into a `circuit` to get a `layer`?

rileyjmurray: Good idea! Will make the change.

rileyjmurray: Done!

rileyjmurray:
@coreyostrove can you merge once tests pass?

pygsti/circuits/circuit.py
```diff
@@ -916,9 +946,9 @@ def __add__(self, x):
 
         return Circuit._fastinit(self.layertup + x.layertup, new_line_labels, editable=False, name='',
                                  stringrep=s, occurrence=None)
-    
 
-    def sandwich(self, x, y):
+
```
rileyjmurray: Add whitespace?

pygsti/circuits/circuit.py
```diff
@@ -1166,7 +1196,9 @@ def _replace_layer_component(self, ilayer, indx, val):
             assert(indx == 0), "Only index 0 exists for a single-simple-Label level"
             self._labels[ilayer] = val
 
-    def extract_labels(self, layers=None, lines=None, strict=True):
+    def extract_labels(self, layers: Union[None, int, slice, list[int], tuple[int, ...]]=None,
+                       lines: Union[None, int, slice, list[int], tuple[int, ...]]=None,
+                       strict: bool=True) -> Union[Circuit, _Label]:
```
rileyjmurray: Do we have better type aliases for this?

pygsti/circuits/circuit.py
```diff
@@ -1453,15 +1485,20 @@ def insert_idling_layers_inplace(self, insert_before, num_to_insert, lines=None)
...
 
         if lines is None:  # insert complete layers
             for i in range(num_to_insert):
                 self._labels.insert(insert_before, [])
 
             #Shift compilable layer indices as needed
             if self._compilable_layer_indices_tup:
+                if num_to_insert <= 0 and insert_before < len(self._labels):
+                    raise ValueError('Undefined behavior (at least until the ' \
+                    'documentation is updated).')
```
rileyjmurray: raise an error if num_insert is negative, quick return if num_insert is zero.

pygsti/circuits/circuit.py
```diff
@@ -2526,14 +2563,23 @@ def replace(obj):  # obj is either a simple label or a list
             if isinstance(obj, _Label):
                 if obj.name == old_gatename:
                     newobj = _Label(new_gatename, obj.sslbls)
-                else: newobj = obj
+                else:
+                    newobj = obj
+            elif obj == old_gatename:
+                if len(obj) == 0:
+                    sslbls = self.line_labels
```
rileyjmurray: TODO: check if this is where we handle the empty label. (Is it sometimes an empty string?)

pygsti/circuits/circuit.py
rileyjmurray: TODO: see if "simple label" is being used in a colloquial sense or the sense of a Label. Seems like it's the former, and that ``obj`` can be a plain string (not even a LabelStr).

pygsti/circuits/circuit.py
```diff
@@ -2969,7 +3015,7 @@ def idling_lines(self, idle_layer_labels=None):
             return tuple([x for x in self._line_labels
                           if x not in all_sslbls])  # preserve order
 
-    def delete_idling_lines_inplace(self, idle_layer_labels=None):
+    def delete_idling_lines_inplace(self, idle_layer_labels: Optional[Sequence[_Label]]=None):
```
rileyjmurray, line 3029: We have a LayersIndex type alias. Should that be used here? TODO for me to check.

pygsti/circuits/circuit.py
```diff
@@ -4001,10 +4045,10 @@ def convert_to_cirq(self,
...
 
         return cirq.Circuit(moments)
-    
-    @classmethod
-    def from_cirq(cls, circuit, qubit_conversion=None, cirq_gate_conversion= None,
-                  remove_implied_idles = True, global_idle_replacement_label = 'auto'):
+
+    @staticmethod
+    def from_cirq(circuit: CirqCircuit, qubit_conversion=None, cirq_gate_conversion=None,
+                  remove_implied_idles=True, global_idle_replacement_label='auto') -> Circuit:
```
rileyjmurray, line 4062: TODO: check for other uses of ``@classmethod``

pygsti/circuits/circuit.py
```diff
@@ -4935,10 +4981,13 @@ def done_editing(self):
...
         if not self._static:
             self._static = True
-            self._labels = tuple([layer_lbl if isinstance(layer_lbl, _Label) 
-                                  else _Label(layer_lbl) for layer_lbl in self._labels])
+            t = [LL if isinstance(LL, _Label) else _Label(LL) for LL in self._labels]
+            self._labels = tuple(t) # type: ignore
         self._hashable_tup = self.tup
         self._hash = hash(self._hashable_tup)
+        self._str = None
+        self._str = self.str # this accessor recomputes the value of self._str
```
rileyjmurray: Clarify: accessor needs self._str to be None in order for it to recompute self._str.

pygsti/modelmembers/modelmember.py
```diff
@@ -903,7 +915,7 @@ def is_similar(self, other, rtol=1e-5, atol=1e-8):
         bool
             True if structure (all but parameter *values*) matches
         """
-        if type(self) != type(other): return False
+        if not isinstance(self, type(other)): return False
```
rileyjmurray: linebreak

pygsti/circuits/circuit.py
rileyjmurray: Yup! LayersIndex and LinesIndex.

pygsti/circuits/circuit.py
rileyjmurray: I've decided to handle this by falling back to Label.replace_name.

For static analysis purposes I added a replace_name function to the Label base class. Note that Label objects are never instantiated explicitly, and all concrete Label classes have a replace_name implementation.

pygsti/circuits/circuit.py
rileyjmurray: Nope! This function would error _hard_ on integer arguments.

pygsti/circuits/circuit.py
rileyjmurray: There were a few other places. Changed those.

PR #646, opened 2025-08-28
Adds fix for pandas "KeyError: slice(None, None, None)" failures.

enielse:
Single line fix to convert pandas CategoricalIndex to a normal Index, which seems necessary for `pandas.merge` to function as intended after a recent update (of pandas).  This issue was causing "`KeyError: slice(None, None, None)` failures when trying to convert results objects to pandas DataFrames, failing within `_to_dataframe(...)` calls.


rileyjmurray (approved):
Looks good to me!

PR #649, opened 2025-09-10
TVD and other objectives in GST

rileyjmurray:
This PR adds support for new objective types (TVD, normalized TVD, general Lp^p). These are demonstrated in jupyter_notebooks/Examples/Robust-GST.ipynb.

In order to document the available objective types I created a new ``ObjectiveType`` union type alias. It supports literals 'logl', 'chi2', 'tvd', 'normalized-tvd', and tuples form ('Lp^p', p). 

This PR makes it possible to specify a black-box penalty function that maps a Model to a nonnegative scalar. It's supported by any TimeIndependentMDCObjectiveFunction. The implementation follows the existing approach to adding penalties: we just add another component to the terms vector and another row to the Jacobian. The current method of computing the Jacobian row is inefficient.

An integration test has been added under the new ``test/workflows`` folder. I've moved an integration test for leakage GST into that folder as well.

Incidental changes to logging that we can discuss:
 * ``printer.warning`` now uses indentation in the same way as ``printer.log``. This makes it much easier to understand the context of the warning. It also now adds another blank line underneath it. The latter change avoids giving the impression that the warning applies to the line of text that immediately follows.
 * I added ``message_level=2`` to several calls to ``printer.log``.
 * I guarded calls to ``printer.log`` with ``comm is not None``, since lots of log messages are only really of interest when using MPI.

#### Some details

Changed the builder API:

   - code now uses ObjectiveFunctionBuilder(Class, ...) rather than Class.builder(), and  
   - ObjectiveFunctionBuilder.create_from behavior has been extended to accommodate the new objective types (e.g., TVD) and the possibility of a callable penalty function.
  
Changes to MDCObjectiveFunction
 - ``__init__`` and ``create_from`` methods now have explicit ``**kwargs`` for passthrough from child class implementations.
 - Stateless evaluation of ``MDCObjectiveFunction.fn`` and ``MDCObjectiveFunction.fn_local`` by passing kwarg ``stateless=True``.
 - New ``MDCObjective.fn_from_model`` function, similar to ``MDCObjectiveFunction.fn`` but accepts a Model instead of an (optional) paramvec.
 
Use docstring injection for TimeIndependentMDCObjectiveFunction and its __many__ subclasses.

New classes:
 - ``TermWeighted(TimeIndependentMDCObjective)``
 - ``RawAbsPower(RawObjectiveFunction)``.
 - The RawTVDObjective class was already in place; I just had to implement a few functions for it.


pygsti/data/datasetconstruction.py
```diff
@@ -209,6 +209,39 @@ def simulate_data(model_or_dataset, circuit_list, num_samples,
...
+    if choose_machine:
+        if seed is None:
+            _warnings.warn('Set the random seed! Using 42.')
+            seed = 42
+        rngstate = _np.random.default_rng(seed)
+        interp_weights = rngstate.uniform(low=0, high=1, size=num_circuits)
+        interp_weights[interp_weights < p] = 0.0
+        interp_weights[interp_weights > 0] = 1.0 
+    else:
+        interp_weights = p * _np.ones(num_circuits)
```
nkoskelo: Are we allowing ```p``` to be a arraylike here, so that one can fully specify the distribution of weights?

pygsti/protocols/gst.py
```diff
@@ -751,20 +752,22 @@ def cast(cls, obj):
...
+    @staticmethod
+    def create_from(objective='logl', freq_weighted_chi2=False, always_perform_mle=False, only_perform_mle=False):
         """
         Creates a common :class:`GSTObjFnBuilders` object from several arguments.
 
         Parameters
         ----------
-        objective : {'logl', 'chi2'}, optional
+        objective : {'logl', 'chi2', 'tvd'}, optional
             Whether to create builders for maximum-likelihood or minimum-chi-squared GST.
```
nkoskelo: Comment needs to be updated to include total variational distance GST.

pygsti/protocols/gst.py
```diff
@@ -786,10 +789,15 @@ def create_from(cls, objective='logl', freq_weighted_chi2=False, always_perform_
         chi2_builder = _objfns.ObjectiveFunctionBuilder.create_from('chi2', freq_weighted_chi2)
         mle_builder = _objfns.ObjectiveFunctionBuilder.create_from('logl')
 
-        if objective == "chi2":
+        if not isinstance(objective, str):
+            import warnings
+            warnings.warn(f'Trying to create an objective function from non-string specification, "{objective}". \
+                           \nSupport for this kind of specification is experimental!')
+            iteration_builders = [chi2_builder]
```
nkoskelo: Do not need line 796.

pygsti/report/section/summary.py
```diff
@@ -28,8 +28,9 @@ def final_model_fit_progress_bar_plot_sum(workspace, switchboard=None, max_lengt
     def final_model_fit_histogram(workspace, switchboard=None, linlog_percentile=5, comm=None, bgcolor='white',
                                   **kwargs):
         return workspace.ColorBoxPlot(
-            switchboard.objfn_builder, switchboard.circuits_final,
-            switchboard.modvi_ds, switchboard.mdl_final_modvi,
+            switchboard.objfn_builder_modvi,
+            switchboard.circuits_final,
+            switchboard.modvi_ds, switchboard.mdl_current_modvi,
```
nkoskelo: ```switchboard.mdl_current_modvi``` seems inaccurate. This function is for the final_model_fit. Thus, shouldn't the appropriate model be ```switchboard.mdl_final_modvi```?

pygsti/tools/optools.py
```diff
@@ -99,6 +110,18 @@ def assert_hermitian(mat):
     assert_hermitian(a)
     assert_hermitian(b)
 
+    def check_unit_trace(mat):
```
nkoskelo: Do we want ```check_unit_trace``` to be only callable within ```_flat_mut_blks```?

pygsti/drivers/longsequence.py
```diff
@@ -386,7 +386,14 @@ def run_long_sequence_gst(data_filename_or_set, target_model_filename_or_object,
...
         the objective function or expert-level functionality.  The allowed keys
         and values include:
-        - objective = {'chi2', 'logl'}
+
+        HISTORICAL NOTE: "XX" indicates that we've at least _intended_ for the
+        keyword argument to be removed. I've removed documentation for options
+        that we never reference in the code (truncScheme, estimate_label, and
+        missingDataAction).
+
+        - objective = typically, a string in {'chi2', 'logl', 'tvd'}. But this can
```
nkoskelo, line 393: Add the ```Literal``` typing to the parameter objective.

pygsti/objectivefns/objectivefns.py
```diff
@@ -160,22 +169,24 @@ def cast(cls, obj):
...
         elif isinstance(obj, str): return cls.create_from(objective=obj)
         elif isinstance(obj, dict): return cls.create_from(**obj)
         elif isinstance(obj, (list, tuple)): return cls(*obj)
         else: raise ValueError("Cannot create an %s object from '%s'" % (cls.__name__, str(type(obj))))
 
-    @classmethod
-    def create_from(cls, objective='logl', freq_weighted_chi2=False):
+    # This was a classmethod, but I made it static because this class has no derived classes.
+    @staticmethod
+    def create_from(objective='logl', freq_weighted_chi2=False, callable_penalty=None):
```
nkoskelo: ```objective``` should be typed as a ```Literal```.

nkoskelo (commented):
Some items for consideration.

pygsti/drivers/longsequence.py
rileyjmurray: These arguments are only passed in the advanced_options dict. I can add the type annotation that advanced_options is ``dict[str,Any]``. I'll leave other type annotations of this function to later work.

pygsti/layouts/copalayout.py
rileyjmurray, line 1: I needed to make these changes to resolve exceptions that were getting raised. I need to find an example showing how to trigger the error without these changes.

pygsti/data/datasetconstruction.py
rileyjmurray: Good catch. That functionality seems possible. Although, I've decided to remove this function altogether. It's needed for some experiment notebooks that I have, so I've moved the definition into those notebooks as needed. (Its current form would take nontrivial work to make robust enough to integrate into our public API.)

pygsti/report/section/summary.py
rileyjmurray: Good catch. Fixed!

.gitignore
rileyjmurray, line 1: revert

pygsti/tools/locking.py
rileyjmurray, line 1: Move to pygsti.tools.locking

wip_notebook_sharing/README.md
rileyjmurray, line 1: delete

wip_notebook_sharing/case0-gst-with-outliers.ipynb
rileyjmurray, line 1: delete

rileyjmurray:
Failures will be resolved once https://github.com/sandialabs/pyGSTi/pull/703 is merged in.

pygsti/models/explicitmodel.py
```diff
@@ -985,6 +994,14 @@ def all_objects(self):
                                          self.factories.items()):
             yield (lbl, obj)
 
+    def members(self, include=('preps', 'operations', 'povms')):
+        out = dict()
+        for member_type in include:
+            member_dict = self.__getattribute__(member_type)
+            for k,v in member_dict.items():
+                out[k] = v
+        return out
```
rileyjmurray: Update to include instruments

pygsti/models/explicitmodel.py
```diff
@@ -241,6 +241,15 @@ def default_gauge_group(self, value):
...
         """
+        if isinstance(value, str):
+            value = value.lower()
+            if value == 'unitary':
+                value = _gg.UnitaryGaugeGroup(self.state_space, self.basis, self.evotype)
+            elif value == 'tp':
+                value = _gg.TPGaugeGroup(self.state_space, self.basis, self.evotypes)
+            else:
+                msg = f'string "{value}" cannot be used to set this model\'s gauge group.'
+                raise ValueError(msg)
```
rileyjmurray, line 255: Add type annotation.

pygsti/protocols/gst.py
```diff
@@ -819,8 +826,13 @@ def create_from(cls, objective='logl', freq_weighted_chi2=False, always_perform_
                 iteration_builders = [chi2_builder]
                 final_builders = [mle_builder]
         else:
-            raise ValueError("Invalid objective: %s" % objective)
-        return cls(iteration_builders, final_builders)
+            iteration_builders = [chi2_builder]
+            try:
+                final_builders = [_objfns.ObjectiveFunctionBuilder.create_from(objective)]
+            except Exception as e:
+                raise ValueError("Invalid objective: %s" % objective)
```
rileyjmurray: pointless exception handling

pygsti/protocols/gst.py
```diff
@@ -3116,6 +3128,9 @@ def add_estimate(self, estimate, estimate_key='default'):
             _warnings.warn("Re-initializing the %s estimate" % estimate_key
                            + " of this Results object!  Usually you don't"
                            + " want to do this.")
+            
+        if estimate.parent is None:
+            estimate.set_parent(self)
```
rileyjmurray: Why is this needed?

pygsti/tools/optools.py
```diff
@@ -40,7 +40,18 @@
...
+"""^
+__SCALAR_TOL_EXPONENT__ is used when checking properties of matrices and vectors.
+It's intended only when we can check the property without incurring rounding
+errors from some reduction (like a sum or a matrix-vector product). If we're
+working with a matrix whose dtype is `d`, then we set
+
+    __SCALAR_TOL__ = d.eps ** __SCALAR_TOL_EXPONENT__,
+
+or a modest multiple thereof.
+"""
```
rileyjmurray, line 56: Seems unlikely that I could add this definition without the PR diff showing ``__SCALAR_TOL_EXPONENT__`` being used in a new place.

pygsti/models/explicitmodel.py
rileyjmurray: Turns out this was unused in this PR. Removing.

pygsti/protocols/gst.py
rileyjmurray: Certain instance methods of Estimate objects need to refer to a parent ModelEstimateResults object to which they belong. This makes sure those instance methods never fail.

pcwysoc (approved):
The one file I own doesn't appear to have changed, so good on my front. 👍

rileyjmurray:
@coreyostrove this has the changes we discussed in the dev meeting. Please review when you're able!

pygsti/forwardsims/mapforwardsim.py
```diff
@@ -278,12 +278,12 @@ def create_layout(self, circuits, dataset=None, resource_alloc=None, array_types
...
         
-        printer.log(f'Num Param Processors {npp}')
-        
-        printer.log("MapLayout: %d processors divided into %s (= %d) grid along circuit and parameter directions." %
-                    (nprocs, ' x '.join(map(str, (na,) + npp)), _np.prod((na,) + npp)))
-        printer.log("   %d atoms, parameter block size limits %s" % (natoms, str(param_blk_sizes)))
-        assert(_np.prod((na,) + npp) <= nprocs), "Processor grid size exceeds available processors!"
+        if comm is not None:
+            printer.log(f'Num Param Processors {npp}', message_level=2)
+            printer.log("MapLayout: %d processors divided into %s (= %d) grid along circuit and parameter directions." %
```
coreyostrove, line 283: A hopefully low-effort request. It is suboptimal, but iirc currently the only obvious clue to which forward simulator was being used in the GST logging is the identity of the corresponding layout. So without this change that would only be getting printed with MPI. Could you update the layout creation log message (line 270 in this diff log) to include the type of the layout? This request applies to the changes in matrixforwardsim.py and termforwardsim.py.

pygsti/layouts/copalayout.py
coreyostrove: That’s strange, maybe a numpy change? This looks like an innocuous enough change though.

pygsti/modelmembers/povms/composedpovm.py
```diff
@@ -50,59 +50,34 @@ class ComposedPOVM(_POVM):
     """
 
     def __init__(self, errormap, povm=None, mx_basis=None):
-        """
```
coreyostrove, line 53: Is the complete removal of this docstring intentional? It looks like some of the class names are out of date, but it should just be a couple of minor changes needed.

pygsti/modelmembers/states/composedstate.py
```diff
@@ -43,21 +47,23 @@ class ComposedState(_State):  # , _ErrorMapContainer
         parameters with other gates and spam vectors.)
     """
 
+    STATIC_STATE_ERROR_MSG = "`static_state` 'reference' must have *zero* parameters!"
```
coreyostrove, line 50: This looks like it is used in just one place, so why class variable?

pygsti/modelpacks/_modelpack.py
```diff
@@ -448,6 +448,38 @@ def create_gst_circuits(self, max_max_length, qubit_labels=None, fpr=False, lite
                                   list(_gen_max_length(max_max_length)),
                                   fidpairs,
                                   **kwargs)
+        return lists
+
+    # TODO: deprecate this function?
```
coreyostrove: That’s fine, just scan the codebase for instances and refactor those first if needed.

pygsti/objectivefns/objectivefns.py
```diff
@@ -246,8 +302,9 @@ def _to_nice_serialization(self):
     @classmethod
     def _from_nice_serialization(cls, state):
         from pygsti.io.metadir import _class_for_name
+        kwargs = state['additional_arguments'].pop('kwargs', dict())
```
coreyostrove, line 299: why pop instead of get?

pygsti/objectivefns/objectivefns.py
```diff
@@ -1400,13 +1469,20 @@ def fn(self, paramvec=None):
...
         result, result_shm = _smt.create_shared_ndarray(self.resource_alloc, (1,), 'd')
-        local = _np.array([self.fn_local(paramvec)], 'd')
+        local = _np.array([self.fn_local(paramvec, stateless)], 'd')
         unit_ralloc = self.layout.resource_alloc('atom-processing')  # proc group that computes same els
         self.resource_alloc.allreduce_sum(result, local, unit_ralloc)
         global_fnval = result[0]
         _smt.cleanup_shared_ndarray(result_shm)
         return global_fnval
 
+    def fn_from_model(self, model):
```
coreyostrove: docstring

pygsti/objectivefns/objectivefns.py
```diff
@@ -28,8 +28,14 @@
 from pygsti.baseobjs.resourceallocation import ResourceAllocation as _ResourceAllocation
 from pygsti.baseobjs.nicelyserializable import NicelySerializable as _NicelySerializable
 from pygsti.baseobjs.verbosityprinter import VerbosityPrinter as _VerbosityPrinter
-from pygsti.models.model import OpModel as _OpModel
-from pygsti import SpaceT
+from pygsti.models.model import OpModel as _OpModel, Model as _Model
+
+
+def set_docstring(docstr):
```
coreyostrove: This is not the first time you've defined this sort of decorator, right? Should this be made into a reusable utility somewhere?

pygsti/objectivefns/objectivefns.py
```diff
@@ -563,6 +620,8 @@ def lsvec(self, probs, counts, total_counts, freqs, intermediates=None):
         """
         return _np.sqrt(self.terms(probs, counts, total_counts, freqs, intermediates))
 
+    # Infinite loop in evaluation of "dterms" and "dlsvec".
```
coreyostrove: That sounds bad... elaborate?

pygsti/objectivefns/objectivefns.py
```diff
@@ -4060,7 +4143,7 @@ def hterms(self, probs, counts, total_counts, freqs, intermediates=None):
         numpy.ndarray
             A 1D array of length equal to that of each array argument.
         """
-        raise NotImplementedError("Derivatives not implemented for TVD yet!")
+        raise NotImplementedError("TVD is not twice-differentiable.")
```
coreyostrove: Would bad stuff happen if this were defined to just be all zeros?

pygsti/objectivefns/objectivefns.py
```diff
@@ -4133,8 +4218,39 @@ def zero_freq_hterms(self, total_counts, probs):
...
+
+
+class RawAbsPower(RawObjectiveFunction):
+    """
+    The function `(1/power) * |p-f|^power`.
+    """
+
+    def __init__(self, power: float,  regularization=None, resource_alloc=None,
+                 name='Lp^p', description="Elementwise absolute value and raising to a power.", verbosity=0):
```
coreyostrove: Pedantry: The L_p norm doesn't typically include the normalization by p. This non-standard definition should be clarified in the doctsring.

pygsti/objectivefns/objectivefns.py
```diff
@@ -5016,230 +5012,144 @@ def __init__(self, mdc_store, regularization=None, penalties=None, name=None, de
...
+        super().__init__(raw_objfn, mdc_store, penalties, verbosity, **kwargs)
+        self.normalized = name == 'normalized tvd'
+        self._terms_weights = None
+        self._update_terms_weights()
+
+    def _update_terms_weights(self):
+        if self.normalized:
+            num_elements = self.layout.num_elements
+            circuit_sizes = _np.zeros((num_elements,))
+            for elind, circuit, _ in self.layout:
```
coreyostrove, line 5168: The term weights should be cacheable, we shouldn't need to recompute these weights every time

pygsti/objectivefns/objectivefns.py
```diff
@@ -5016,230 +5012,144 @@ def __init__(self, mdc_store, regularization=None, penalties=None, name=None, de
...
-    """
 
-    @classmethod
-    def create_from(cls, model, dataset, circuits, regularization=None, penalties=None,
-                    resource_alloc=None, name=None, description=None, verbosity=0,
-                    method_names=('fn',), array_types=()):
-        mdc_store = cls._create_mdc_store(model, dataset, circuits, resource_alloc, method_names,
-                                          array_types, verbosity)
-        return cls(mdc_store, regularization, penalties, name, description, verbosity)
+class LpNormToPowerP(TermWeighted):
```
coreyostrove, line 5184: See comment about about non-standard definition.

jupyter_notebooks/Examples/Robust-GST.ipynb
coreyostrove, line 1: Please annotate this example notebook with an explaining what it is that is demonstrating at a high-level, as well as the key analysis steps/results being shown. I have no idea what I am looking at here (not entirely true since I've sat in the relevant meetings, but this will be very challenging to parse for the uninitiated).

pygsti/protocols/gst.py
```diff
@@ -3019,7 +3028,7 @@ def to_nameddict(self):
             ret[k] = v
         return ret
 
-    def add_estimates(self, results, estimates_to_add=None):
+    def add_estimates(self, results, estimates_to_add=None, silent_steal=False):
```
coreyostrove: Document new kwarg

coreyostrove (changes requested):
A few minor requests, thanks for the hard work as always, @rileyjmurray !

pygsti/layouts/copalayout.py
rileyjmurray: Maybe. Lemme ask the robot about that conjecture.

jupyter_notebooks/Examples/Robust-GST.ipynb
rileyjmurray: That's super valid. Thanks for catching me on this!

pygsti/forwardsims/mapforwardsim.py
rileyjmurray: Will do!

pygsti/forwardsims/matrixforwardsim.py
```diff
@@ -1103,7 +1103,7 @@ def create_layout(self, circuits, dataset=None, resource_alloc=None, array_types
         if mem_limit is not None:
             if mem_limit <= 0:
                 raise MemoryError("Attempted layout creation w/memory limit = %g <= 0!" % mem_limit)
-            printer.log("Layout creation w/mem limit = %.2fGB" % (mem_limit * C))
+            printer.log("Layout creation w/mem limit = %.2fGB" % (mem_limit * C), message_level=2)
```
rileyjmurray, line 1106: Need to update the message to indicate forward simulator type, and to show the type even if mem_limit is None.

pygsti/forwardsims/termforwardsim.py
```diff
@@ -316,15 +316,16 @@ def create_layout(self, circuits, dataset=None, resource_alloc=None, array_types
         if mem_limit is not None:
             if mem_limit <= 0:
                 raise MemoryError("Attempted layout creation w/memory limit = %g <= 0!" % mem_limit)
-            printer.log("Layout creation w/mem limit = %.2fGB" % (mem_limit * C))
+            printer.log("Layout creation w/mem limit = %.2fGB" % (mem_limit * C), message_level=2)
```
rileyjmurray, line 319: Need to update so a level-2 message is printed about forward simulator type, regardless of whether mem_limit is None.

pygsti/layouts/copalayout.py
rileyjmurray: Ah, so it turns out I've already resolved that comment. The current code only differs from develop by adding an explicit type annotation and de-nesting some expressions.

Will mark this thread as resolved.

pygsti/modelmembers/povms/composedpovm.py
rileyjmurray: Yup! I believe the standard practice is for class documentation to appear either under the class name or under the ``__init__`` signature. So I viewed this as removing duplicate code. I'll add type annotations to the constructor to avoid losing any useful information.

pygsti/modelmembers/states/composedstate.py
rileyjmurray: I don't remember my precise reasoning. I think it's that I like to have error messages defined as class constants so we can reference those messages in testing. (Granted, I didn't add any test to trigger this erroring codepath.)

pygsti/objectivefns/objectivefns.py
rileyjmurray: Ya caught me! Will define one in a utility file.

pygsti/objectivefns/objectivefns.py
rileyjmurray: If I only ``get`` then the additional_arguments kwarg will also appear in the unpacked ``**kwargs``, which would raise an error.

pygsti/objectivefns/objectivefns.py
rileyjmurray: No need to worry. The comment is just getting at how a concrete ObjectiveFunction class needs to override one and only one of the ``dterms`` and ``dlsvec`` functions. (The ObjectiveFunction implementation of dterms reduces the problem to an equivalent call to dlsvec, while its implementation of dlsvec reduces the problem to an equivalent call to dterms.)

I'll update the comment to be clearer. The point of the comment is to get the code-reader's attention and say "hey, you aren't losing your mind if you can't figure out which of these two functions is more fundamental."

pygsti/objectivefns/objectivefns.py
rileyjmurray: I think that'd actually be fine. It's consistent with how we handle TVD's lack of a first derivative at the origin. Will make that change.

pygsti/objectivefns/objectivefns.py
rileyjmurray: This is a good point. It might actually be better to remove the scaling. Will think more before merging.

pygsti/objectivefns/objectivefns.py
rileyjmurray: My concern was/is that if self.layout was updated without our knowledge (say, to have more circuits) then things would break if we used the same weights as before the update.

Having the TermWeighted._reweight_terms and TermWeighted._reweight_jac functions call _update_terms_weights, and recomputing those weights here, prevents the situation above.

That said, modifications to self.layout are probably very rare, even if not forbidden. So I'll modify the implementations of TermWeighted._reweight_terms and TermWeighted._reweight_jac to check for possible changes before recomputing.

pygsti/objectivefns/objectivefns.py
rileyjmurray: Will do (or will revert to standard definition).

pygsti/protocols/gst.py
rileyjmurray: Will do

pygsti/objectivefns/objectivefns.py
rileyjmurray: reverted to standard definition

pygsti/objectivefns/objectivefns.py
rileyjmurray: reverted to standard definition

pygsti/modelmembers/povms/composedpovm.py
coreyostrove: Cool. I agree about the standard practice (I personally prefer it with the __init__, but that is neither here nor there). In this case the class docstring was missing 2 out of 3 of the parameters compared to the __init__ docstring which is what flagged this for me. Thanks for adding the type annotations.

pygsti/modelmembers/states/composedstate.py
coreyostrove: Good enough for me.

coreyostrove (approved):
Thank for addressing my requests, @rileyjmurray! Approving this now so that once you're done with the last couple changes (all of which are minor or documentation related) you're making you can merge freely without waiting on me for re-review.

rileyjmurray:
Still need to make the type annotation changes to composedpovm.py. Need to take a break for dinner. Will wait until tests pass, then make + push the composedpovm.py changes and immediately merge.

Issue #655, opened 2025-09-18
Purge the pickles


rileyjmurray:
The (vast) majority of datastructures in PyGSTi have a very nice JSON serialization capabilities. Despite this, there are places where we save pickles to disk. This was brought to my attention by non-QPL Sandians.

The place we need to make the change for our colleagues is in [``Report.write_notebook``](https://github.com/sandialabs/pyGSTi/blob/5bd509fce3c1bd0b22bc5393a758c8a7631dc95c/pygsti/report/report.py#L248-L252). I'll take care of fixing that. While I'm at it I'll try to get rid of all pickling in pyGSTi.

rileyjmurray:
In release 0.10.

PR #658, opened 2025-09-23
Enhancements and bugfixes in Report.write_notebook

rileyjmurray:
This resolves issue #655.

There's a change to the public API for Report.write_notebook and changes to the code in the notebook it generates. Report.write_notebook only saves data as a pickle if it's called with kwarg ``use_pickle=True``. There's a new Report.results_dict_from_dir function that can accept _either_ the a filename with a ``.pkl`` extension or a directory name. If the pkl extension is present then we try to load from a pickle (to ensure backward-compatibility). Otherwise, we try to load from a directory generated by the new Report.write_results_dict function.

I introduced a little testing infrastructure so our write_notebook test actually runs the notebooks that it generates. I also added a test that that calls Report.write_notebook with kwarg ``use_pickle=True``.


pygsti/report/templates/report_notebook/gauge_variant.txt
pcwysoc, line 1: Glad to see the ErrgenTable fixed!

pygsti/report/workspacetables.py
pcwysoc, line 1: I appreciate this warning.

pygsti/baseobjs/label.py
```diff
@@ -201,7 +201,6 @@ def is_simple(self):
         return self.IS_SIMPLE
 
 
-
```
rileyjmurray, line 204: There should only be two blank lines between class definitions.

pygsti/circuits/circuit.py
```diff
@@ -512,7 +511,6 @@ def __init__(self, layer_labels=(), line_labels='auto', num_lines=None, editable
         self._bare_init(labels, my_line_labels, editable, name, stringrep, 
                         occurrence, compilable_layer_indices_tup)
 
-
```
rileyjmurray, line 515: There should only be one blank line between method definitions.

pygsti/circuits/circuit.py
```diff
@@ -4942,6 +4940,7 @@ def done_editing(self):
         self._hashable_tup = self.tup
         self._hash = hash(self._hashable_tup)
 
+
```
rileyjmurray, line 4943: Two blank lines between class definitions.

test/test_packages/report/test_report.py
rileyjmurray, line 1: Deleted several spurious blank lines

pygsti/protocols/estimate.py
rileyjmurray, line 1: Changes to this file fix a longstanding bug about calling ``.write()`` on Estimate objects whose circuit_weights member is a non-empty dict. The fix is to transform circuit_weights from a ``dict[Circuit,float]`` to a ``dict[str,float]`` before serializing.

pygsti/protocols/estimate.py
```diff
@@ -236,6 +240,8 @@ def write(self, dirname):
         -------
         None
         """
+        if self.circuit_weights is not None:
+            self.circuit_weights = {getattr(c,'str',c): v for c,v in self.circuit_weights.items()}
```
rileyjmurray: The dict comprehension uses getattr in case self.circuit_weights is already keyed by strings.

rtd-requirements.txt
rileyjmurray, line 1: Added newline at end of file

requirements.txt
rileyjmurray, line 1: Added newline at end of file

pyproject.toml
rileyjmurray, line 1: nbformat is needed for tests in test_packages to run successfully. I can't figure out if nbformat is an explicit dependency of ``notebook`` (it is, after all, the reference implementation of the Jupyter notebook format), but there's no harm in making its dependency explicit for us.

pygsti/protocols/estimate.py
```diff
@@ -83,6 +84,9 @@ def from_dir(cls, dirname, quick_load=False):
         ret.__dict__.update(_io.load_meta_based_dir(_pathlib.Path(dirname), 'auxfile_types', quick_load=quick_load))
         for crf in ret.confidence_region_factories.values():
             crf.set_parent(ret)  # re-link confidence_region_factories
+        if ret.circuit_weights is not None:
+            from pygsti.circuits.circuitparser import parse_circuit
+            ret.circuit_weights = {parse_circuit(c, True, True): v for c,v in ret.circuit_weights.items()}
```
rileyjmurray: Turns out parse_circuit doesn't return a Circuit object. It turns a 4-tuple whose first component is a list of Labels that can be cast to a Circuit. I'm not sure how I missed that. Will fix.

rileyjmurray: Fixed in most recent commit.

Issue #652, opened 2025-09-17
Diamond-distance wildcard incorrect for leakage-aware reporting


coreyostrove:
**Describe the bug**
In 0.9.14 new subspace restricted metrics were introduced as part of new reporting/modeling capabilities for measuring leakage and tailoring gauge optimization to leaky systems. These new metrics were not integrated into the diamond distance wildcard model workflow, and as a result the wildcard values obtained when constructing a leakage-aware report are pretty crazy. We should automatically detect when we're using leakage-aware analysis and switch over to subspace restricted diamond distance for that case.

PR #664, opened 2025-10-08
WIP: Improvements to leakage modeling and gauge optimization

rileyjmurray:
EDIT: closing this PR since its scope got too large. Moving preliminary changes into a new PR.

I made this branch to resolve issue #652. While working on that I made various improvements to gauge optimization.

1. Fixed bugs in definitions of subspace-restricted Frobenius norm objective. There was no meaningful impact to the bug as far as gates were concerned. There was an impact in how POVMs contributed to the objective. That hadn't showed up before since leakage-aware gauge optimization placed zero weight on SPAM. 
2. Changed the definition of fidelity as a gauge optimization objective. It's now defined in such a way that if two models exhibit no relational errors with respect to one another, then it's possible to drive the fidelity-based loss function to zero even if the models are not gauge-equivalent.
3. Added the ability to separately weight state prep and POVMs differently by introducing a parameter called prepWeight. The spamWeight option is used as the default for prepWeight. I'm open to rolling this change back since I'm not actually using it.

I made incidental changes to objectivefns.py. The default values that various objective classes use for "min_prob_clip" (or similarly named variables) are now controlled by module-wide or class-wide constants that can be changed at runtime, rather than being hard-coded to defaults of 1e-4.

Changes to gst.py:
1. Removed unused functions ``_load_dataset`` and ``_load_model`` that were not referenced anywhere in the pyGSTi repo.
2. Refactored ``_compute_wildcard_budget_1d_model`` and ``_compute_1d_reference_values_and_name`` to eliminate nesting and code duplication.
3. Modified ``_compute_1d_reference_values_and_name`` to use subspace-restricted diamond distance if the provided gauge optimization suite includes the flag n_leak=1.
4. Modified ``_add_param_preserving_gauge_opt`` to also call ``_add_bad_fit_estimates``.

Other leakage-related changes:
1. Moved several function definitions from report/reportables.py to tools/leakage.py.
2. Changed leakage gauge optimization suite to have two stages. I'll edit this comment with an explanation of the second stage later.
3. Changed add_lago_models so that it works even if the provided estimate has no model called "stdgaugeopt".

### DIFF
--- a/jupyter_notebooks/Examples/Leakage-automagic.ipynb
+++ b/jupyter_notebooks/Examples/Leakage-automagic.ipynb
@@ -2,14 +2,21 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pygsti.modelpacks import smq1Q_XY, smq1Q_ZN\n",
-    "from pygsti.tools.leakage import leaky_qubit_model_from_pspec, construct_leakage_report\n",
+    "%load_ext autoreload\n",
+    "%autoreload 2\n",
+    "import pygsti\n",
+    "from pygsti.modelpacks import smq1Q_XYI as mp\n",
+    "from pygsti.tools.leakage import leaky_qubit_model_from_pspec, construct_leakage_report, add_lago_models, gate_leakage_profile\n",
+    "from pygsti.report import construct_standard_report\n",
+    "from pygsti.tools.optools import unitary_to_superop\n",
     "from pygsti.data import simulate_data\n",
-    "from pygsti.protocols import StandardGST, ProtocolData"
+    "from pygsti.protocols import StandardGST, ProtocolData, GateSetTomography, GSTGaugeOptSuite, GSTBadFitOptions\n",
+    "import numpy as np\n",
+    "import scipy.linalg as la"
    ]
   },
   {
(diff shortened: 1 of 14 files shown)


pygsti/baseobjs/basisconstructors.py
```diff
@@ -10,6 +10,7 @@
 # http://www.apache.org/licenses/LICENSE-2.0 or in the LICENSE file in the root pyGSTi directory.
 #***************************************************************************************************
 import itertools as _itertools
+import re as _re
```
rileyjmurray: Need to remove this.

rileyjmurray: Done (though the commit not yet pushed).

rileyjmurray:
Splitting into ...
 * a PR targeting develop https://github.com/sandialabs/pyGSTi/pull/669. The scope includes a heavy refactor of gaugeopt.py, the introduction of eigenvalue_fidelity, a change in the mathematical meaning of fidelity as a gauge optimization objective, and new ways of setting default floating point values in objectivefns.py. These changes are backward-compatible, so PR 669 could target bugfix instead of develop.
 * a PR targeting bugfix: wildcard analysis with leakage models; fixes to LAGO (two-stage, where a second stage uses a direct sum gauge group).
 * a PR targeting develop: general leakage modeling by making inferences from ``basis.ellookup['I']``. This includes some backward compatible enhancements to the Basis class. The changes for leakage will NOT be backward compatible. This branch (``resolve-issue-652``) which I've abandoned, shows how messy it would be to make these changes backward compatible.

PR #668, opened 2025-10-13
Update germ selection and FPR to better work with implicit models

enielse:
Adds logic to allow implicit models to be used for germ selection and fiducial pair reduction.  This basically just adds case logic to interpret an `ImplicitOpModel`'s `prep_blks['layers']`, `povm_blks['layers']`, and `operation_blks['layers']` elements as the state prep, POVM, and gate operations of that model (parallel to the `.preps`, `.povms`, and `.operations` of an `ExplicitOpModel`).  I don't think this necessarily handles all the edge cases, but I've found it to work fine when using `LocalNoiseModel` objects when you also specify the `num_nongauge_params` when performing germ selection.

### DIFF
--- a/pygsti/algorithms/fiducialpairreduction.py
+++ b/pygsti/algorithms/fiducialpairreduction.py
@@ -2074,6 +2074,32 @@ def _make_spam_static(model):
 
 
 def _copy_to_static_explicitop_model(mdl):
+    """Create and return an "effective" a copy of `mdl` that is an ExplicitOpModel with static elements.
+    
+    If `mdl` is already an ExplicitOpModel it is copied and returned
+     after setting all paramterizations to 'static'.
+    Otherwise an ExplicitOpModel is constructed from quantities within
+     `mdl` (for instance, if `mdl` is an ImplicitOpModel), namely:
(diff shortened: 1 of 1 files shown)


pygsti/algorithms/fiducialpairreduction.py
```diff
@@ -2071,6 +2071,29 @@ def _make_spam_static(model):
...
+                           simulator=mdl.sim.copy(), evotype=mdl.evotype)
+        for k, v in mdl.prep_blks['layers'].items():
+            ret.preps[k] = v.to_dense()
+        for k, v in mdl.povm_blks['layers'].items():
+            ret.povms[k] = _convert_povm(v.copy(), 'static', mdl.basis)
+        for k, v in mdl.operation_blks['layers'].items():
+            ret.operations[k] = v.to_dense()
+
+    assert ret.num_params == 0
+    return ret
```
rileyjmurray, line 2121: Please add a docstring explaining the limitations of this function. Presumably there are situations where I could call ``mdl.probabilities(c)`` for some Circuit c, yet ``ret.probabilities(c)`` would raise an error. What are those cases?

PR #669, opened 2025-10-17
Refactoring gaugeopt.py: maintainability and fidelity as an objective

rileyjmurray:
This PR is split off of #664, which I've since closed for being too large in scope.

This PR has major backward-compatible changes to the ``gaugeopt_to_target`` and ``_create_objective_fn`` functions in gaugeopt.py. 
 * Changes to gaugeopt_to_target are designed to clarify how it serves as a wrapper to gaugeopt_custom (which is also a wrapper, of wrappers, ...).
 * Changes to _create_objective_fn are to split out its two codepaths into separate functions: _legacy_create_scalar_objective_fn and _legacy_create_least_squares_objective_fn.

Both of these changes are stepping stones so we can introduce new types of gauge optimization in the future. My vision is that one day we'll be able to include user-defined black-box objective functions in gauge optimization suites. This will enable future prototyping without reaching into gauge optimization plumbing.

This PR also changes the definition of fidelity as a gauge optimization objective. It's now defined in such a way that if two models exhibit no relational errors with respect to one another, then it's possible to drive the fidelity-based loss function to zero even if the models are not gauge-equivalent. As part of this effort I introduced a new function in optools.py called eigenvalue_fidelity. This function computes fidelity of two states in a way that pretends they share an eigenbasis. There's an option of ordering eigenvalues in a gauge-invariant way or in a way that's technically gauge-dependent yet still robust.

Small changes:
  * model.py and explicitopmodel.py: type annotations for copy methods.
  * gaugeopt.py no longer calls ExplicitOpModel._excalc(). Calls to _excalc were only introduced in PR 672, but at the time they were identified as a temporary fix. This change resolves https://github.com/sandialabs/pyGSTi/pull/672#pullrequestreview-3373424254.

### DIFF
--- a/pygsti/algorithms/gaugeopt.py
+++ b/pygsti/algorithms/gaugeopt.py
@@ -20,34 +20,79 @@ from pygsti import optimize as _opt
 from pygsti import tools as _tools
 from pygsti.tools import mpitools as _mpit
 from pygsti.tools import slicetools as _slct
+from pygsti.models import (
+    ExplicitOpModel as _ExplicitOpModel
+)
+from pygsti.models.model import OpModel as _OpModel
 from pygsti.models.gaugegroup import (
     TrivialGaugeGroupElement as _TrivialGaugeGroupElement,
     GaugeGroupElement as _GaugeGroupElement
 )
 
-from typing import Callable, Union, Optional
+from typing import Callable, Union, Optional, Any
 
 
-def gaugeopt_to_target(model, target_model, item_weights=None,
-                       cptp_penalty_factor=0, spam_penalty_factor=0,
-                       gates_metric="frobenius", spam_metric="frobenius",
-                       gauge_group=None, method='auto', maxiter=100000,
-                       maxfev=None, tol=1e-8, oob_check_interval=0,
-                       convert_model_to=None, return_all=False, comm=None,
-                       verbosity=0, check_jac=False, n_leak=0):
+class GaugeoptToTargetArgs:
     """
-    Optimize the gauge degrees of freedom of a model to that of a target.
+    This class is basically a namespace. It was introduced to strip out tons of complexity
+    in gaugeopt_to_target(...) without breaking old code that might call gaugeopt_to_target.
+    """
 
-    Parameters
-    ----------
-    model : Model
-        The model to gauge-optimize
+    old_trailing_positional_args = (
+        'item_weights', 'cptp_penalty_factor', 'spam_penalty_factor',
+        'gates_metric', 'spam_metric', 'gauge_group', 'method',
+        'maxiter', 'maxfev', 'tol', 'oob_check_interval', 
+        'convert_model_to', 'return_all', 'comm', 'verbosity',
+        'check_jac'
+    )
 
-    target_model : Model
-        The model to optimize to.  The metric used for comparing models
-        is given by `gates_metric` and `spam_metric`.
+    @staticmethod
+    def parsed_model(model: Union[_OpModel, _ExplicitOpModel], convert_model_to: Optional[Any]) -> _OpModel:
+        if convert_model_to is None:
+            return model
+        
+        if not isinstance(model, _ExplicitOpModel):
+            raise ValueError('Gauge optimization only supports model conversion for ExplicitOpModels.')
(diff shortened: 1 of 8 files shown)


rileyjmurray:
Could also make changes to eigenvalue_entanglement_fidelity. Here's a new implementation from #664 ([click here for the diff](https://github.com/sandialabs/pyGSTi/pull/664/files#diff-83f2b5353882febad629524a6ba19ae25609d387737b8eaf579a31f12c2c7e20R1607-R1674). It requires a new [is_trace_preserving](https://github.com/sandialabs/pyGSTi/pull/664/files#diff-2d4af4c9e9d9960f49ce364c3d695cd583f1042752214f83c0301ae335b075dfR397-R414) function that's also on 664's branch.
```python

def eigenvalue_entanglement_infidelity(a: _np.ndarray, b: _np.ndarray, mx_basis: BasisLike, is_tp=None, is_unitary=None, tol=1e-8) -> _np.floating:
    """
    Eigenvalue entanglement infidelity is the infidelity between certain diagonal matrices that
    contain [see Note 1] the eigenvalues of J(a) and J(b), where J(.) is the Jamiolkowski
    isomorphism map. This is equivalent to computing infidelity of (J(a), J(b)) while pretending
    that they share an eigenbasis.

    Parameters
    ----------
    is_tp : bool, optional (default None)
        Flag indicating that a and b are TP in their provided basis. If None (the default),
        an explicit check is performed up to numerical tolerance `tol`. If True/False, then
        the check is skipped and this is used as the result of the check as though it were 
        performed.

    is_unitary : bool, optional (default None)
        Flag indicating that b is unitary. If None (the default) an explicit check is performed
        up to tolerance `tol`. If True/False, then the check is skipped and this is used as the
        result of the check as though it were performed.

    Notes
    -----
    [1] The eigenvalues are ordered using a min-weight matching algorithm.

    [2] If a and b are trace-preserving (TP) and b is unitary, then we compute eigenvalue
        entanglement fidelity by leveraging three facts:

        1. If (x,y) share an eigenbasis and have (consistently ordered) eigenvalues (v(x),v(y)),
           then Tr(x y^{\\dagger}) is the inner product of v(x) and v(y).

        2. J(a) and J(b) share an eigenbasis of if `a` and `b` share an eigenbasis.

        3. If `a` and `b` are TP and `b` is unitary, then their entanglement fidelity can be
           expressed as |Tr( a.b^{\\dagger} )| / d^2.
     
       Then together, we see that if (a,b) are TP and `b` is unitary, then their eigenvalue 
       entanglement fidelity is expressible as the inner product of v(a) and v(b), where 
       v(a) and v(b) vectors holding are suitably-ordered eigenvalues of a and b.
    
    """
    d2 = a.shape[0]

    if is_unitary is None:
        is_unitary = _np.allclose(_np.eye(d2), b @ b.T.conj(), atol=tol, rtol=tol)
    if is_tp is None:
        is_tp = _tools.is_trace_preserving(a, mx_basis, tol) and _tools.is_trace_preserving(b, mx_basis, tol) 

    if is_unitary and is_tp:
        evA = _np.linalg.eigvals(a)
        evB = _np.linalg.eigvals(b)
        _, pairs = _tools.minweight_match(evA, evB, lambda x, y: abs(x - y),
                                        return_pairs=True)  # just to get pairing
        fid = abs(_np.sum([_np.conjugate(evB[j]) * evA[i] for i, j in pairs])) / d2
    else:
        Ja = _tools.fast_jamiolkowski_iso_std(a, mx_basis)
        Jb = _tools.fast_jamiolkowski_iso_std(b, mx_basis)
        from pygsti.tools.optools import eigenvalue_fidelity
        fid = eigenvalue_fidelity(Ja, Jb, gauge_invariant=True)
    return 1.0 - fid

```

rileyjmurray:
Note: the fact that eigenvalue_entanglement_fidelity in reportables.py rather than optools.py is a weirdness related to issue #526.

rileyjmurray:
TODO: make changes throughout gaugeopt.py to avoid accessing ExplicitOpModel._excalc(), which we currently need to do if we want gauge optimization to consider instruments.

coreyostrove (approved):
Looks good, thanks for the hard work. You were right, the diff for gaugeopt.py was a nightmare and I was holding on by my fingernails, but the context from the walkthrough of the changes you gave us a week or two ago went a long way to making it comprehensible.

PR #670, opened 2025-10-17
abandoned; see 671

### LINKED ISSUE #652: Diamond-distance wildcard incorrect for leakage-aware reporting

rileyjmurray:
This PR is split off from #664. It fixes how wildcard error is computed in leakage modeling. When merged it will resolve #652. It also includes changes to the default leakage-aware gauge optimization suite. The latter changes are needed to prevent gauge optimization from trying to spurriously "spread out" relational errors across available gates.

Ping @coreyostrove. I made this PR by starting from develop. It seems that targeting bugfix after branching from develop is carrying those extra commits (I mean, that _is_ the expected behavior, but I hadn't considered that when branching from develop). How would you like to proceed? Option 1 is I reply these commits on top of a new branch that starts from bugfix (which I think requires closing this PR and opening a new one). Option 2 is I change the target branch to develop. If we go with option 2 then we'd be asking leakage modeling collaborators to use develop until the release of 0.9.15.


coreyostrove:
I like option 1. Having this in with 0.9.14.2 would be preferable.

rileyjmurray:
Closing to start from a different branch.

PR #672, opened 2025-10-21
Ensure that gauge optimization incorporates instruments

rileyjmurray:
This resolves #644. The fix was easier than I expected for two reasons. First, gauge optimization w.r.t. the Frobenius norm already incorporated instruments by way of ExplicitOpModelCalc.frobeniusdist and ExplicitOpModelCalc.residuals. Second, gauge optimization w.r.t. trace distance or infidelity can incorporate instruments just by working with operation dicts from ``target_model._excalc()`` and ``mdl._excalc()``, rather than the operation dicts in ``target_model`` and ``mdl``.

While I was making these changes, I (1) replaced instances of np.dot with the matmul infix operator, and (2) simplified branching in explicitcalc.py.

PR #675, opened 2025-10-24
Merge bugfix into develop (resolve failed auto-merge)

rileyjmurray:
The automatic ``develop <- bugfix`` merge from Corey's PR yesterday (#674) [failed](https://github.com/sandialabs/pyGSTi/actions/runs/18768908655/job/53549598808#step:3:20).

The current state of bugfix includes changes from #674 and the PR I just merged, #672. There's no way that the automatic ``develop <- bugfix`` merge will work since it includes the changes from yesterday that failed the auto-merge. So this PR catches develop up with collective changes in #672 and #674.

Tests for the candidate ``develop <- bugfix`` merge workflow passed (all that failed was the auto-merge), so I'm going to merge this PR without waiting for tests to pass.

Notes:

It's not clear to me why the auto-merge for #674 failed. It might have something to do with the weird commit history I call out in the screenshot below.

image

### DIFF
--- a/pygsti/algorithms/gaugeopt.py
+++ b/pygsti/algorithms/gaugeopt.py
@@ -24,6 +24,8 @@ from pygsti.models.gaugegroup import (
     TrivialGaugeGroupElement as _TrivialGaugeGroupElement,
     GaugeGroupElement as _GaugeGroupElement
 )
+from pygsti.models import ExplicitOpModel
+from pygsti.modelmembers.operations import LinearOperator as _LinearOperator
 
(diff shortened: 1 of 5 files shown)


coreyostrove:
Yeah, I'm not sure why the auto-merging has been failing. Presumably git detects a merge conflict of some sort which would need to be resolved? If this continues being a recurring problem then we should probably just remove the auto-merge from the workflow and plan on doing this manually going forward (we already are atm anyhow).

PR #677, opened 2025-11-05
Make better use of context when calling eig vs eigh

### LINKED ISSUE #582: Audit use of eig versus eigh (and eigvals versus eigvalsh)
While reviewing a PR by Juan I came across a strange comment explaining deliberate use of eig instead of eigh, even though the matrix was Hermitian:
https://github.com/sandialabs/pyGSTi/pull/480#discussion_r2104767418

We should audit the codebase for these instances and review if those uses are sensible. I'm opening as a separate issue from Juan's PR in order to not slow him down.

rileyjmurray:
This PR resolves #582.


pygsti/report/reportables.py
```diff
@@ -1666,6 +1671,9 @@ def eigenvalue_avg_gate_infidelity(a, b, mx_basis):
     -------
     float
     """
+    # TODO: discuss correctness of this function. I think it might assume that both are
+    # tp and that b is unitary. We should probably just transform the result of 
+    # eigenvalue_entanglement_infidelity, rather than compute from scratch.
```
rileyjmurray: Potential correctness issue in reportables.py::eigenvalue_avg_gate_infidelity.

pygsti/report/reportables.py
```diff
@@ -1698,6 +1706,8 @@ def eigenvalue_diamondnorm(a, b, mx_basis):
     -------
     float
     """
+    # TODO: discuss correctness of this function. I think it might assume that both are
+    # tp and that b is unitary.
```
rileyjmurray: Potential correctness issue in reportables.py::eigenvalue_diamondnorm.

pygsti/report/reportables.py
```diff
@@ -1728,6 +1738,7 @@ def eigenvalue_nonunitary_diamondnorm(a, b, mx_basis):
     -------
     float
     """
+    # TODO: discuss meaning of this function.
```
rileyjmurray: It's not obvious to me what eigenvalue_nonunitary_diamondnorm is supposed to do. It has a similar structure to other functions whose correctness I'm unsure of.

pygsti/tools/matrixtools.py
```diff
@@ -807,6 +807,40 @@ def _objective(flat_logm):
...
+        # Make sure it's Hermtian in exact arithmetic. This helps with
+        # reproducibility across different implementations of LAPACK.
+        m += m.T.conj()
+        m /= 2
+        evals, evecs = _np.linalg.eigh(m)
+        inv_evecs = evecs.T.conj()
+    else:
+        evals, evecs = _np.linalg.eigh(m)
+        inv_evecs = _np.linalg.inv(evecs)
+    return evecs, evals, inv_evecs
```
rileyjmurray, line 840: The star in the function signature means that only one positional argument is allowed. All remaining arguments have to be specified by kwargs. This is to give us more flexibility to change these signatures in the future.

pygsti/tools/optools.py
```diff
@@ -57,6 +57,65 @@ def _flat_mut_blks(i, j, block_dims):
...
+        return 0, _np.zeros(n)
+    
+    if abs(alpha - 1) > scalar_tol:
+        message = f"The input matrix is not trace-1 up to tolerance {scalar_tol}. Beware result!"
+        _warnings.warn(message)
+        candidate_v *= _np.sqrt(alpha)
+
+    return 1, candidate_v
+
+
```
rileyjmurray: This function used to be defined locally inside the ``fidelity`` function.

pygsti/tools/optools.py
```diff
@@ -1482,6 +1479,8 @@ def std_process_mx_to_unitary(superop_mx):
             Uj = U[:, j]
             Ui = _np.dot(UijU, Uj)
         else:
+            # TODO: discuss the significance of this i==0 codepath.
+
```
rileyjmurray, line 1501: Please review optools.py::std_process_mx_to_unitary. I'd like to discuss the significance of this i==0 codepath.

pygsti/tools/optools.py
```diff
@@ -2495,204 +2494,136 @@ def compute_best_case_gauge_transform(gate_mx, target_gate_mx, return_all=False)
...
-        start = 0
-        for i, k in enumerate(kite):
-            slc = slice(start, start + k)
-            dstart = start + k
-            for kk in kite[i + 1:]:
-                if k == kk and _np.isclose(evals_tgt[start], evals_tgt[dstart].conj()):  # conjugate block!
-                    dslc = slice(dstart, dstart + kk)
-                    # enforce block conjugacy needed to retain Uproj conjugacy structure
-                    D[dslc, dslc] = D[slc, slc].conj()
+    def _get_eigenspace_pairs(mx, tol=1e-6):
```
rileyjmurray, line 2515: All changes past this point are from simplifying an ``if True: ...`` block to remove the unreachable code.

rileyjmurray:
@coreyostrove and @sserita, this is ready for review.

@enielse I have correctness concerns for three functions in reportables.py. These functions were written over 6 years ago, so that means they were likely written by you. Can you take a look?

pygsti/report/reportables.py
rileyjmurray: Plan: have this wrap eigenvalue_entanglement_infidelity.

pygsti/tools/optools.py
rileyjmurray: Marking this comment resolved to reduce review burden.

pygsti/tools/matrixtools.py
coreyostrove: Despite having seen this many times, I'm embarrassed to say that I didn't actually know that this is what the star convention signified.

pygsti/algorithms/contract.py
```diff
@@ -243,7 +243,7 @@ def _contract_to_cp_direct(model, verbosity, tp_also=False, maxiter=100000, tol=
             #  print "DEBUG: orig evals == ",evals                #DEBUG
             assert(abs(sum(new_evals) - 1.0) < 1e-8)
 
-            new_Jmx = _np.dot(evecs, _np.dot(_np.diag(new_evals), _np.linalg.inv(evecs)))
+            new_Jmx = evecs @ _np.diag(new_evals) @ inv_evecs
```
coreyostrove: Minor request. When doing matrix multiplication with a diagonal matrix it is typically much faster to do element-wise multiplication with the vector of diagonal matrix entries. This will get broadcast into an O(n^2) rather than O(n^3) operation.

pygsti/tools/matrixtools.py
rileyjmurray: I only learned about it recently!

pygsti/algorithms/contract.py
```diff
@@ -408,7 +411,7 @@ def _contract_to_valid_spam(model, verbosity=0):
                 for (k, ev) in enumerate(evals):
                     if ev < 0.0: evals[k] = 0.0
                     if ev > 1.0: evals[k] = 1.0
-                mx = _np.dot(evecs, _np.dot(_np.diag(evals), _np.linalg.inv(evecs)))
+                mx = evecs @ _np.diag(evals) @ inv_evecs
```
coreyostrove: See comment above (I won't comment on all of the instances).

pygsti/models/qutrit.py
```diff
@@ -22,11 +22,12 @@
 from pygsti.tools import unitary_to_superop, change_basis
 
 #Define 2 qubit to symmetric (+) antisymmetric space transformation A:
-A = _np.matrix([[1, 0, 0, 0],
+A = _np.array([[1, 0, 0, 0],
```
coreyostrove, line 25: Thanks, this rouge use of matrix has been a minor annoyance for a while (one less deprecation warning).

pygsti/tools/jamiolkowski.py
```diff
@@ -341,9 +342,8 @@ def sum_of_negative_choi_eigenvalues_gate(op_mx, op_mx_basis):
     """
     sumOfNeg = 0
     J = fast_jamiolkowski_iso_std(op_mx, op_mx_basis)  # Choi mx basis doesn't matter
-    evals = _np.linalg.eigvals(J)  # could use eigvalsh, but wary of this since eigh can be wrong...
-    for ev in evals:
-            if ev.real < 0: sumOfNeg -= ev.real
+    evals = _mt.eigenvalues(J, assume_hermitian=True)
+    sumOfNeg = _np.sum(evals[evals < 0])
```
coreyostrove: Is this change of sign convention intentional? If not can we revert this? (i.e. add in a minus). I'm not sure anything relies on this convention, but would prefer to be safe if there isn't a reason to change it.

pygsti/tools/matrixtools.py
```diff
@@ -66,7 +66,7 @@ def gram_matrix(m, adjoint=False):
     return out
 
 
-def is_hermitian(mx, tol=1e-9):
+def is_hermitian(mx, tol=1e-9) -> bool:
```
coreyostrove: incomplete typing

pygsti/tools/optools.py
```diff
@@ -57,6 +57,65 @@ def _flat_mut_blks(i, j, block_dims):
     return ret
 
 
+def check_rank_one_density(mat, scalar_tol: Optional[Union[float,_np.floating]]=None) -> tuple[int, _np.ndarray]:
```
coreyostrove: incomplete typing

pygsti/tools/optools.py
```diff
@@ -57,6 +57,65 @@ def _flat_mut_blks(i, j, block_dims):
...
+    # ^ use for checks that do have dimensional dependence (will naturally increase for larger matrices)
+
+    n = mat.shape[0]
+
+    if _np.linalg.norm(mat) < vec_tol:
+        # We prefer to return the zero vector instead of None to simplify how we handle
+        # this function's output.
+        return 0, _np.zeros(n, dtype=complex)
+
+    rng = _np.random.default_rng(0)
```
coreyostrove: Naive question, is it fine that this has a fixed seed? I.e. that all matrices of a given dimension N use the same test vector for the power iteration.

coreyostrove (changes requested):
Thanks for all of this hard work, @rileyjmurray! I've left a couple minor suggestions and questions, but this is pretty close to being good to go.

pygsti/algorithms/contract.py
rileyjmurray: Good catch; will fix!

pygsti/tools/jamiolkowski.py
rileyjmurray: It was not! Good catch.

pygsti/tools/optools.py
```diff
@@ -1382,18 +1382,15 @@ def dmvec_to_state(dmvec, tol=1e-6):
...
-    psi = evecs[:, k] * _np.sqrt(evals[k])
-    psi.shape = (d, 1)
+    dim = int(dmvec.size**0.5)
+    dm = dmvec.reshape((dim, dim))
+    _mt.assert_hermitian(dm, tol)
+    r, psi = check_rank_one_density(dm, tol)
+    if r == 0:
+        raise ValueError('Cannot convert zero dmvec to pure state!')
+    if r == 2:
+        raise ValueError('Cannot convert mixed dmvec to pure state!')
```
rileyjmurray: ~~The case r == 2 doesn't certify that the density is rank >= 2. We should do an eigendecomposiiton in this case.~~

^ That wasn't true. There is a case where an eigendecomposition is needed, but that case is only introduced in this PR's changes to check_rank_one_density (renamed to fast_density_rank).

pygsti/tools/optools.py
rileyjmurray: One could argue both sides. The worst-case scenario is that 
```python
    candidate_v = mat @ test_vec
    candidate_v /= _np.linalg.norm(candidate_v)
```
divides by zero. This would require an adversarial input.

I'll add a check to guard against this and return with a different status code if we encounter this situation. I'll also think about renaming the function to be clearer about it's role: if the returned value of ``r`` is something other than zero or one then the onus is on the calling function to do a more expensive check, like an eigendecomposition.

rileyjmurray: renamed to fast_density_rank.

rileyjmurray:
All comments addressed. If all tests pass when I'll merge!

coreyostrove (approved):
Looks good, thanks for addressing these requests. Merge at will (conditional on passing tests).

PR #679, opened 2025-11-07
Bugfix IBMQ Retrieve Results

pcwysoc:
Hopefully fixes an issue where the code could not handle more bits returned from IBM than the number of qubits in the experiment. This is the case in my default workflow. I ended up reverting to older code that I had locally. This older code also handles MCMs, which I do not believe current code does. @ndsieki I would be very interested in your thoughts on this, as you've been maintaining the IBMQ code recently and may be using a different workflow. 

We should also investigate why retrieving results is so slow. It takes several hours to load larger batches of jobs from IBM. My current guess is that it is due to a checkpoint being written when each job is retrieved, which is probably not necessary and slowing things down wildly.


coreyostrove:
> I think we'd also like this code to be able to handle the case where only some of the qubits on the device are measured. Given that we also need to handle mid-circuit measurements (which also affect the length of the bitstring results), perhaps one approach would be to provide two arguments to `retrieve_results()`: one that specifies a list of ordered target indices for the data bits in the bitstring results, and another that specifies a list of ordered target indices for the mid-circuit measurement bits in the bitstring results. These arguments would supercede the `ordered_target_indices` variable that is currently used internally in `retrieve_results()`.
> 
> Re: the speed of `retrieve_results()`, I have found the checkpointing to be a bottleneck. Given the speed of the result retrieval for each batch, and the fact that the job IDs are checkpointed during `submit()`, I think it makes sense to only checkpoint once, after all results have been retrieved.

I have no immediate concerns about the proposed approach. @ndsieki, @pcwysoc: Are you two cool with taking the lead on making and validating these changes?

pcwysoc:
@coreyostrove sounds good. I can work on implementing a fix, but it will difficult for me to test right now as I don't have IBM access right now. @ndsieki happy to jump on a call and help you work through this if you're open to taking the lead.

ndsieki:
I've made the following changes to address the bitstring and checkpointing issues in `IBMQExperiment`:

Bitstrings: `retrieve_results()` now uses metadata attached to each Qiskit circuit (those in `qiskit_isa_circuit_batches`) to determine which bits in the Qiskit bitstring are mid-circuit measurements (MCMs), and which bits are data bits. All other bits are marginalized over. The MCM metadata is called `im_dict`, and the data bit metadata is called `ordered_data_indices`. The MCM metadata is a dictionary because I am anticipating that it may be useful to attach additional information to each MCM, such as which qubit it occurred on and how many MCMs have been performed on that qubit. The metadata is attached to each Qiskit circuit during the `transpile()` call. Note that there are two pathways in the `transpile()` call: either a direct conversion (pyGSTi -> Qiskit), or an indirect conversion (pyGSTi -> OpenQASM -> Qiskit). Adding the metadata to the Qiskit circuit is more awkward with the indirect conversion. It may be worth discussing whether we even need the indirect conversion anymore, but that goes beyond the scope of a bugfix.

Checkpointing: The checkpointing that occurred after each batch retrieval has been removed, and has been replaced by one `write()` call after all batch results have been retrieved *and* the `ProtocolData` object has been created from these results. There would have been a separate `_write_checkpoint()` and `data.write()` call, but looking at the functionality of `write()` it seemed appropriate to combine them. I am less familiar with the checkpointing system, so I would certainly appreciate another set of eyes there.

I've added unit tests for the bitstring changes. They cover the cases I know to be relevant, which are: indirect and direct conversion to Qiskit; with and without MCMs; measuring all qubits on the device, and measuring only those qubits that are used in the circuit. @pcwysoc, please feel free to add more test cases, and let me know if there are outstanding concerns.

rileyjmurray (approved):
I resolved the merge conflicts between this branch and develop, then applied modest changes to get CI tests to pass. Approving now and merging.

@pcwysoc and @ndsieki: please open GitHub issues as follow-up work if you feel that's needed!

PR #687, opened 2025-11-19
Zassenhaus Formula and Error Generator Propagation Performance Improvements

coreyostrove:
This PR contains two mostly disjoint sets of updates for error generator propagation: an implementation of the Zassenhaus formula (up to second order), and performance improvements for linear sensitivities for probabilities and expectation values (and related functions).  

## Zassenhaus Formula:
The Zassenhaus formula is kind of like a dual to the well known BCH approximation. In a nutshell, it allows one to take an exponentiated sum of operators and write it as a product of single-exponentials with some multiplicative corrections:

$\exp(\sum_i A_i) = (\prod_i \exp(A_i)) (\prod_j \exp(C_j))$

Where the $C_j$ terms are Lie polynomials of order $j$ in the $A_i$s. At present I have only implemented the second-order correction term ($C_2$). I found a paper that had purported to derive the higher-order terms of the generalized Zassenhaus formula (99.9% of the time online only the special case for a sum of exactly two operators is given), but when I implemented and tested the third order term I concluded the published result wasn't correct and the mistake in it wasn't immediately obvious. So, if we ever want the third order or higher corrections we'll probably need to sit down and do the algebra ourselves. The basic idea is straightforward and I can describe how one would do so if they're in the mood for some tedious algebra. Conveniently, the second-order generalized Zassenhaus correction is exactly the same as the second-order Magnus expansion correction, so I've refactored that code to enable reuse.

The function for computing this is in `errgenproptools.py`, and some basic correctness tests have been added.

## Performance Improvements for Linear Sensitivities:
A few weeks ago during a meeting we were talking about a computation which was being done that relied on the linear sensitivity (alpha) code and @jordanh6 proclaimed, and I quote, "the alpha code is *obnoxiously* slow!" Taking this as the assault on my pride it was I've been plotting my revenge ever since. Since I had this code fresh in my mind for the Zassenhaus thing I decided to piggyback off of that and enact my scheme. Changes include:

- New functions, `bulk_alpha` and `bulk_alpha_pauli`, for computing linear sensitivities for a number of error generators and bitstrings/paulis all at once. This reduces overhead associated with instantiating `stim.TableauSimulator` instances and enables other economies of scale (see next bullet).
- New `bulk_phi` function for computing a number of phi values all at once. A big part of the performance gains here come from performing some additional preprocessing caching intermediate results and identifying unique pauli effective pauli pairs to reduce redundancy.
- Reduced copy overhead in `amplitude_of_state` by resetting the `TableauSimulator` after use rather than working on copies (much faster, it turns out).
- Cythonization: All of the core subroutines, as well as `bulk_alpha` and `bulk_alpha_pauli` now have cythonized counterparts for even more efficiency. These are aliased in automatically when cython extensions have been built, and we still have fallback pure python implementations. For some of the subroutines, e.g. the `pauli_phase_update` function, the mostly C-code cython implementations are a couple orders of magnitude faster. For others (the `bulk_alpha_pauli` for example) the difference is meaningful but less stark (anywhere from a 20% improvement to a factor of 2). 

Only semi-related note, I also ran into an edge case with underflowing values in certain `phi` computation. I've also included a patch for this issue too by adding a new option to `amplitude_of_state` which returns just the phase of a complex amplitude for a given state (which is all that is actually needed for `phi` computations).

A couple other things. I've reworked the stabilizer probability and expectation value correction functions to use these new bulk methods, so those should benefit from this change. The bulk methods also support computing multiple bitstring/paulis at once. There are a couple of functions which could benefit from this, but I haven't reworked those yet to take advantage of this (so there are a few mostly low effort performance wins to be had if anyone is bored).

### Benchmarks

Here is a table of benchmarks comparing this branch to the current tip of develop. Benchmarking notebook (and helper module) attached for those curious in the details. 
[2025-11-17_Profile_Alpha_Comps.ipynb](https://github.com/user-attachments/files/23619218/2025-11-17_Profile_Alpha_Comps.ipynb)
[model.py](https://github.com/user-attachments/files/24410421/model.py)

Note that for "bulk" methods, since these don't directly exist on develop, the comparison being made is to computing these value for a full list or error generators using a simple for loop (as one would have done in practice presently). For both the 4- and 100-qubit models the noise model is a mostly arbitrarily selected H+S error model. The 4 qubit circuits are depth 4 random circuits. The 100 qubit circuit is depth 100 and also randomly sampled. I.e. the 4 qubit results are the average over 10 random circuits, but the 100 qubit result is just for a single random instantiation (I don't expect this to make a big difference, but didn't feel like waiting around for statistics). The error generator list for the alpha computations in both cases come from the first-order magnus approximation of the propagated generators for that circuit. 

| Function                    | Num Qubits | Develop | Current | Improvement (X) |
|------------------------|---------------|-----------|----------|----------------|
| `bulk_alpha`               |  4                  |  48.2ms  |   821us   |   58.7             |
| `bulk_alpha`               |  100              |  42.4s     |    851ms   |   49.8              |
|`bulk_alpha_pauli`       |  4                  |  3.2ms   |    134us  |    23.9             |
|`bulk_alpha_pauli`       |  100              |  5.91s    |  6.14ms  |    962.5*          |
| `amplitude_of_state`  |  4                  |  120us   |   67us     |     1.8              |
| `amplitude_of_state`  |  100              |  1.66ms |   1.65ms |       1^              |
| `pauli_phase_update` |  4                  |   47us   |   164ns   |      286             |
| `pauli_phase_update` |  100              |  123us  |    426ns  |      289             |

*=When I first saw this value it was very unexpected (all of my detailed profiling was done in the 4-qubit case). But I double checked and the answer is same in both versions of the code. I'd have to do some more profiling to see what exactly made such a big difference between the two. After I did confirm it was correct the fact that I was so tantalizingly close to @rileyjmurray's namesake wasn't lost one me.
^= Most of the change to this function was the simulator copying overhead. I guess this isn't a big deal for the many-qubit case.

As for my revenge, that was secured when I got @jordanh6 to agree to review this PR last week before he knew these changes (and their many associated lines of code) were included 😈.


pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,991 @@
...
+    Returns the phase correction for an operator character when the input bit is '0'.
+    
+    For op:
+      '_' -> 1
+      'X' -> 1
+      'Y' -> 1j  if dual is False, or -1j if dual is True.
+      'Z' -> 1
+    For any unrecognized operator, defaults to 1.
+    """
+    if op == 95:  # '_' ASCII 95
```
nkoskelo: Should we use the short circuiting approach here? Are we expecting `_` or `X` to be more common than `Y` for the value of ```op```? If not we could refactor this to just check for `Y` first and then return 1 if it is not `Y`.

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,991 @@
...
+        # empty pauli string
+        raise ValueError("pauli_str must not be empty")
+
+    n = pauli_len - prefix_len  # Number of operator characters.
+    # Allocate a C char buffer for output with space for n characters plus a null terminator.
+    out_buffer = <char*> PyMem_Malloc((n + 1) * sizeof(char))
+    if not out_buffer:
+        raise MemoryError("Failed to allocate memory for output buffer")
+     # Initialize the buffer with the ASCII code for '0' (48).
+    for i in range(n):
```
nkoskelo: We do not need to initialize the output buffer. We can just set the value to 48 if it fails the ```pauli_flip_ascii(op)``` check. Yes the compiler may be able to vectorize the first loop easier but loop fusion may still result in a slight speed up.

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,991 @@
...
+    if not out_buffer:
+        raise MemoryError("Failed to allocate memory for output buffer")
+
+    # Process each qubit position.
+    for i in range(n):
+        # Get the operator character from the pointer.
+        op = p[prefix_len + i]
+        # Get the corresponding input bit from bit_str.
+        # (Assume bit_str characters are '0' or '1')
+        if b[i] == 49: # ASCII '1'
```
nkoskelo: If ```b[i] == 49 or 48```, then we could just use ```bit_val = 49 - b[i]``` since subtraction is typically cheaper than a jump.

coreyostrove:
@nkoskelo: Thanks for the suggestions! Question for you related to a couple of your suggestions: How much experience do you have with profiling tools/workflows for C/C++ code? 

E.g. in "If b[i] == 49 or 48, then we could just use bit_val = 49 - b[i] since subtraction is typically cheaper than a jump." your suggestion here is to make this branchless which is good for reducing potential overheads from branch mispredictions, but I don't currently have a handle on how often branch misses are happening. I know tools exist for profiling things like branch misses and other low-level execution information, but I've never used them. Do you have any experience with these? I suspect going further with optimization would benefit heavily from access to lower-level profiling.

There were a few places where I tried replacing some of the nested if-else table lookups with switch statements with this in mind, for example, but didn't find this made much difference in my testing (but I didn't have access to the lower-level profiling for those changes).

pygsti/tools/fasterrgencalc.pyx
coreyostrove: This will be workload dependent, but there are some workloads where I think we might periodically see Ys. This would benefit from some statistics gathered on user workloads. Regardless, this is a good suggestion and would bring the average number of checks to a bit over 1 (right now it would probably be a bit over 2), so we should do it.

pygsti/tools/fasterrgencalc.pyx
coreyostrove: Good suggestion, I'll give this a try and report back.

pygsti/tools/fasterrgencalc.pyx
coreyostrove: Very clever! Will report back!

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,991 @@
...
+    if len(desired_state) != n:
+        raise ValueError("desired_state length must equal number of qubits")
+
+    bs = <unsigned char*>PyMem_Malloc(n * sizeof(unsigned char))
+    if not bs:
+        raise MemoryError("allocating desired_state buffer failed")
+    dptr = PyUnicode_AsUTF8(desired_state)
+    for q in range(n):
+        # ASCII '1' = 49
+        bs[q] = 1 if dptr[q] == 49 else 0
```
nkoskelo: If we assume `dptr[q]` is only 48 or 49 we can use the same trick again as in `pauli_phase_update`.

nkoskelo:
Commit c10583edb9953f5c1facb57c8b3efef54ca2c4fa resulted in a small slowdown compared to 68828d335617d7c2f558be1ebd667acbc602ef37 for the `pauli_phase_update` with 100 qubits in my testing. However, it was still faster than those quoted in 68828d335617d7c2f558be1ebd667acbc602ef37 completing on the order of 60 ns instead of 426ns. The other test cases saw a 1.01x to 1.11x speedup compared to the prior commit. I converted the jupyter notebook into a python file for ease of testing.
[benchmark.py](https://github.com/user-attachments/files/24658799/benchmark.py)

rileyjmurray:
> Commit [c10583e](https://github.com/sandialabs/pyGSTi/commit/c10583edb9953f5c1facb57c8b3efef54ca2c4fa) resulted in a small slowdown compared to [68828d3](https://github.com/sandialabs/pyGSTi/commit/68828d335617d7c2f558be1ebd667acbc602ef37) for the `pauli_phase_update` with 100 qubits in my testing. However, it was still faster than those quoted in [68828d3](https://github.com/sandialabs/pyGSTi/commit/68828d335617d7c2f558be1ebd667acbc602ef37) completing on the order of 60 ns instead of 426ns. The other test cases saw a 1.01x to 1.11x speedup compared to the prior commit. I converted the jupyter notebook into a python file for ease of testing. [benchmark.py](https://github.com/user-attachments/files/24658799/benchmark.py)

Consider trying -O2 instead of -O3. There are situations where -O3 leads to worse performance than -O2.

coreyostrove:
What is the practical effect of including the -fopenmp flag in this context?

rileyjmurray:
@coreyostrove the failing tests are from numpy 2 incompatibility. You might need to merge develop changes into this branch.

Let me know when it's ready for me to pull down and review locally. (I'll wait for you to say "now" or to do the develop merge so tests pass.)

setup.py
```diff
@@ -72,7 +72,19 @@ def build_extensions(self):
...
+            Extension(
+                "pygsti.tools.fasterrgencalc",
+                sources=["pygsti/tools/fasterrgencalc.pyx"],  # , "fasterrgencalc.c
+                # # Cython docs on NumPy usage should mention this!
+                # define_macros = [('NPY_NO_DEPRECATED_API','NPY_1_7_API_VERSION')],
+                # # leave above commented
+                # # see http://docs.cython.org/en/latest/src/reference/compilation.html#configuring-the-c-build
+                # define_macros = [('CYTHON_TRACE','1')], #for profiling
+                include_dirs=['.', np.get_include()],
+                extra_compile_args=['-O3','-fopenmp']
```
rileyjmurray: Some compilers (particularly, defaults on macOS) don't have OpenMP, and will error with `-fopenmp`. Removing.

rileyjmurray:
I've pushed small changes:
 * Merged develop into this branch. (Catch up on numpy 2.0 compatibility, among other things.)
 * Fixed missing end-of-file in fastcalc.pyx.
 * Removed `-fopenmp` compiler flag, since it wasn't strictly necessary and it can cause build errors on macOS. (It caused some for me).

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,1039 @@
+# encoding: utf-8
+# cython: profile=False
+# cython: linetrace=False
+# filename: fasterrgencalc.pyx
+
+#***************************************************************************************************
+# Copyright 2015, 2019, 2025 National Technology & Engineering Solutions of Sandia, LLC (NTESS).
```
rileyjmurray: 2026

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,1039 @@
...
+                codes[i] = ET_A1
+            else:
+                codes[i] = ET_A
+
+    #
+    # 4) Pre‐allocate the output sensitivity arrays:
+    #
+    #cdef list sensitivities_by_bitstring = [None] * n_b
+    #for i in range(n_b):
+    #    sensitivities_by_bitstring[i] = [0.0] * n_e
```
rileyjmurray: Remove

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,1039 @@
...
+    # P1 and P2 either commute or anticommute.
+    if P1.commutes(P2):
+        return None
+    else:
+        P3 = P1*P2
+        return (P3.sign*2, P3 / P3.sign)
+
+cdef inline tuple _pauli_product(object P1, object P2):
+    P3 = P1*P2
+    return (P3.sign, P3 / P3.sign)
```
rileyjmurray, line 1108: Please add a comment that these `object` types will actually be `stim.PauliString`.

pygsti/tools/fasterrgencalc.pyx
```diff
@@ -0,0 +1,1039 @@
...
+                    else:
+                        com_BP = B.commutes(paulis[i])
+                        if com_BP:
+                            sensitivities_by_pauli_view[i, j] = 0.0
+                        else:
+                            ABP = _pauli_product(A * B, paulis[i])
+                            expectation = ABP[0] * sim.peek_observable_expectation(ABP[1])
+                            sensitivities_by_pauli_view[i, j] = _real_if_close(1j * 4 * expectation)
+                            
+    return sensitivities_by_pauli
```
rileyjmurray: Missing blank line at EOF.

pygsti/tools/fasterrgencalc.pyx
rileyjmurray, line 1: The complex numbers used in this file seem to take on a tiny range of values (only the 4th roots of unity?). If that's the case, we should try to use complex64 rather than complex128. If there are external API reasons why we prefer complex128 then it's fine to leave precision as-is.

I've tried a find-and-replace of np.complex128 with np.complex64 and all tests passed (test/unit, test/test_packages, and test/integration). I won't commit that change though.

pygsti/tools/fastcalc.pyx
rileyjmurray, line 1: Leaving a comment to make note of the fact that we seem to use this functionality in a handful of places. 

When we import this module we usually do so as `_fastcalc`. A search for `_fastcalc.` in the pyGSTi source and then excluding commented-out lines yields the following hits.

Image

We also have hits for `_fc.`

Image

test/unit/tools/test_errgenproptools.py
```diff
@@ -671,3 +784,9 @@ def _compare_analytic_numeric_iterative_composition(num_qubits):
             print('analytic_composition_mat=')
             print_mx(analytic_composition_mat)
             raise ValueError('Numeric and analytic error generator compositions were not found to be identical!')
+
+#helper function for zassenhaus formula testing
+def _dense_zassenhaus_generators_analytic(errorgen_groups, errogen_propagator, zassenhaus_order=1):
+    zassenhaus_gens= _eprop.zassenhaus_formula(errorgen_groups, zassenhaus_order)
+    dense_zassenhaus_generators = [errogen_propagator.errorgen_layer_dict_to_errorgen(gen) for gen in zassenhaus_gens]    
+    return dense_zassenhaus_generators
```
rileyjmurray: Missing blank line at EOF

test/unit/tools/test_errgenproptools.py
rileyjmurray, line 1: Given that the PR has two disjoint sets of changes, I'd expect there to be two test files that were changed. I see that's not the case because the user-facing changes are all in `pygsti/tools/errorgenproptools.py`.

I think this is a good time to have a `test/unit/tools/errorgenproptools/` directory that contains files for testing distinct error generator propagation subsystems.

On that note, error generator propagation is a gigantic feature. I think we should consider restructuring the code as part of #715. (Out of scope for this PR.)

rileyjmurray (changes requested):

pygsti/tools/fasterrgencalc.pyx
coreyostrove: I'm going to leave this as is for now, mostly because the default numpy data type for complex arrays is complex128, and there isn't a practical need to deviate from that default at the moment. It would primarily be memory, and memory is a non-issue for this part of the computation presently. 

You're right about the phases only being 4th roots of unity. If we wanted to be extra slick we could even rework this to get rid of the complex numbers entirely and just use integers mod 4. GPT incidentally suggested this change at one point while I was using it. I decided not too for readability sake, the complex arithmetic is very far from being the bottleneck right now (that should still be the postselection), so I decided the extra work was not worth it.

test/unit/tools/test_errgenproptools.py
coreyostrove: I've left a note in #715 about this change.

coreyostrove:
Update 4/13: I've just finished pushing up the last round of substantive technical changes on this branch. Doing some profiling using py-spy I identified the main remaining bottleneck in the alpha (for bitstring) computation was the computational cost of doing postselections with stim. The latest round of changes both reduces the total number of postselections required, and implements a new pair of bulk_amplitude_of_state functions which amortize the cost of some of the remaining postselections across multiple bit strings. Most of the numbers in the table above didn't meaningfully change. `bulk_alpha` did and for the 4Q test case was slightly slower with these changes, but for the 100Q case much faster (12.8X -> 49.8X compared to current develop).

@rileyjmurray: I've implemented the smaller changes you requested. Let me know if you wanted to do any additional review, if not I'll merge this after tests pass.

rileyjmurray:
No other comments from me

coreyostrove:
Oh, a math thing worth documenting here before I forget. While conversing with our computational overlords it was suggested that we can get rid of postselection for computing the amplitudes altogether by constructing something called the "quadratic form triple" representation for the target stabilizer state. It even wrote me some complicated code (which I have not tested at all, because I'd need to understand this myself first before being comfortable reviewing it for correctness) for taking a tableau representation and constructing this "quadratic form triple" representation. This is indeed a thing (with the first work on this going back to 2002), and there are some recent papers about fast algorithms for doing stabilizer math in this representation and converting between representations (bookmarking this [paper ](https://quantum-journal.org/papers/q-2025-01-08-1586/pdf/)). If this becomes important enough we can give these algorithms a try to see if we can't make these computations even faster.

PR #690, opened 2025-11-25
New Feature:  GST iteration-specific optimizers

juangmendoza19:
Corey and I pair-programmed a new feature (in hopes it will be useful for AMS) that allows for the specification of iteration-specific optimizers.

This new feature allows the user to pass in many optimizers into GST, such that each GST iteration can use a different optimizer.

For example, if a user wants to have a different number of iterations only for the final logl GST iteration, it can specify a different optimizer just for the final iteration.

To use this feature, simply create a list of the optimizers to be used:
```
optimizers = [optimizer_1, optimizer_2, ..., optimizer_k]
your_favorite_gst_protocol.run(data, optimizers=optimizers)
```

Where k is the total number of iterations for the GST protocol (k == len(circuit_lists)). 

Iteration i within GST will use optimizer_i

On a separate note, I found two GST unit test classes that were not being executed: TestStandardGST and TestGateSetTomography. Corey suggested their names must end with "Tester", so I renamed them "GateSetTomographyTester" and "StandardGSTTester".

This revealed two failing tests:

FAILED protocols/test_gst.py::GateSetTomographyTester::test_run_custom_sim - assert False
FAILED protocols/test_gst.py::StandardGSTTester::test_run_custom_sim - assert False

These tests are failing due to code that is not touched in this branch, and also fail in develop if named appropriately such that they are not skipped.

Also Corey promised I would get "street cred" for doing type annotations of any arguments touched in this PR. I hope this promise is fulfilled.


test/unit/algorithms/test_core.py
```diff
@@ -451,3 +451,46 @@ def test_iterative_gst_generator_starting_index(self):
...
+        
+        #Test that passing a different optimizer per iteration works as intended
+        optimizers = [] 
+        tols = [1e1, 1e-8]
+        maxiters = [10, 150]
+
+        #First create substantially different optimizers
+        for i in range(len(self.lsgstStrings)):
+            optimizers.append({'tol': tols[i], 'maxiter':maxiters[i]})
+
```
nkoskelo, line 464: We may want to add some more test cases. I think we are missing the case where the optimizer is a list of one dictionary.

```
optimizer:

Optimizer, or dict, or list of Optimizer, or list of dict (default None)
        The optimizer to use, or a dictionary of optimizer parameters
        from which a default optimizer can be built. If a list, the length
        of the list should either be 1 or equal to the number of iterations. 
        If 1, then this optimizer is used for every iteration, otherwise
        each optimizer is used for its corresponding iteration.```

File: test_packages/temp_test_files/test_GateSetTomography_serialization/badfit_options.json
Comment (coreyostrove) [inline test_packages/temp_test_files/test_GateSetTomography_serialization/badfit_options.json:1]: You accidentally committed some testing artifacts. Please remove these.

File: test/unit/algorithms/test_core.py (hunk 1 above)
Comment (juangmendoza19) [reply test/unit/algorithms/test_core.py:464]: Thank you! Done in the latest commit

File: test_packages/temp_test_files/test_GateSetTomography_serialization/badfit_options.json
Comment (juangmendoza19) [reply test_packages/temp_test_files/test_GateSetTomography_serialization/badfit_options.json:1]: Removed in the latest commit

File: test/unit/algorithms/test_core.py [hunk 2]
```diff
@@ -451,3 +451,70 @@ def test_iterative_gst_generator_starting_index(self):
...
+            )
+            
+            models1.append(next(generator_step)[0])
+            
+            self.assertArraysAlmostEqual(models0[-1].to_vector(), models1[-1].to_vector())
+
+        # we also test use case of optimzer=[optimizer] being equivalent to optimzer=optimizer
+        generator_single_item_optimizers0 = core.iterative_gst_generator(
+            self.ds, self.mdl_clgst, self.lsgstStrings,
+            optimizer=[optimizers[0]],
```
nkoskelo: I would prefer to test that the generators get created have the same parameters if passed as `[optimizer]` or as `optimizer`. The test case may have been slightly faster to run. However, since the `core.iterative_gst_generator` generator does not return the `optimizer` object it used for that iteration, this test works. This test function has really two test cases. One for the test that you can have different optimizers for each iteration and one to test `[optimizer]` is functionally equivalent to `optimizer`. It may be easier to debug in the future if these test cases are split.

pygsti/protocols/gst.py
```diff
@@ -1388,6 +1397,38 @@ def run(self, data, memlimit=None, comm=None, checkpoint=None, checkpoint_path=N
                                                       data.dataset, comm)
         if simulator is not None:
             mdl_start.sim = simulator
+        
+        if optimizers is None:
```
nkoskelo, line 1412: This logic is repeated from the `algorithms/core.iterative_gst_generator` function. I think we should refactor to call the same utility helper function in both cases. That way if we update the allowed types to pass in we will only need to update the logic in one location.

test/unit/protocols/test_gst.py
```diff
@@ -357,6 +382,30 @@ def _test_run_custom_sim(self, mode, parent_capfd, check_output):
             for model in estimate.models.values():
                 assert isinstance(model, MapForwardSimulatorWrapper)
```
nkoskelo: `model` should not be an instance of `MapForwardSimulatorWrapper`; it is usually either a subclass of `ImplicitOpModel` or `ExplicitOpModel`. If you want to ensure that the `model` used the `MapForwardSimulatorWrapper` you should check if `isinstance(model.sim, MapForwardSimulatorWrapper)`.

test/unit/protocols/test_gst.py
```diff
@@ -357,6 +382,30 @@ def _test_run_custom_sim(self, mode, parent_capfd, check_output):
...
+        for mode in ["full TP","CPTPLND","Target"]:
+            mdl_result1 = results1.estimates[mode].models['stdgaugeopt']
+            mdl_result2 = results2.estimates[mode].models['stdgaugeopt']
+            mdl_result3 = results3.estimates[mode].models['stdgaugeopt']
+            mdl_result4 = results4.estimates[mode].models['stdgaugeopt']
+
+            assert _np.allclose(mdl_result2.to_vector() , mdl_result1.to_vector())
+            assert _np.allclose(mdl_result3.to_vector() , mdl_result1.to_vector())
+            assert _np.allclose(mdl_result4.to_vector() , mdl_result1.to_vector())
+
```
nkoskelo, line 421: Did you also want to test if we can have different optimizer settings for each iteration of the `GST` experiment?

nkoskelo (commented):
Just a few comments.

coreyostrove:
Quick comment: Github auto selected me as a reviewer, but since I helped write this code it wouldn't make sense for me to do so, in the off chance anyone was waiting for one from me.

pygsti/protocols/gst.py
coreyostrove: I'll take this as an opportunity to soap box against DRY in this particular instance, and instead advocate for LoB (https://htmx.org/essays/locality-of-behaviour/).

test/unit/protocols/test_gst.py
juangmendoza19: I can add a test to check if it runs, but I don't see an easy way to test its valid. At that point it might not have any more value than the 

results3 = proto1.run(self.gst_data, optimizers=optimizers)

test.

test/unit/algorithms/test_core.py
juangmendoza19: changed in the most recent commit

test/unit/protocols/test_gst.py
juangmendoza19: Changed in the most recent commit

nkoskelo:
There is currently a failure in the checks for python version 3.12. It is trying to set the `i`th entry of a probability vector using `np.linalg.multi_dot`. This can work however, it will return a numpy array and in python 3.12 a 1x1 array is not automatically convertible to a scalar of the appropriate type. If we instead use `np.einsum` this will return an appropriate scalar.

Also when we call ```assert isinstance(model.sim, MapForwardSimulator)``` we have a built in check to see if the simulator points back to the model. In this setup it appears to be seven different models being tested. Index 1 appears to be pointing to an incorrect model.

nkoskelo:
> There is currently a failure in the checks for python version 3.12. It is trying to set the `i`th entry of a probability vector using `np.linalg.multi_dot`. This can work however, it will return a numpy array and in python 3.12 a 1x1 array is not automatically convertible to a scalar of the appropriate type. If we instead use `np.einsum` this will return an appropriate scalar.

PR #702 will be doing this conversion.

rileyjmurray:
> > There is currently a failure in the checks for python version 3.12. It is trying to set the `i`th entry of a probability vector using `np.linalg.multi_dot`. This can work however, it will return a numpy array and in python 3.12 a 1x1 array is not automatically convertible to a scalar of the appropriate type. If we instead use `np.einsum` this will return an appropriate scalar.
> 
> PR #702 will be doing this conversion.

Uh oh. I just closed that because I've made another branch with slightly larger scope.

juangmendoza19:
Is there any other work needed in this PR? @rileyjmurray @nkoskelo

coreyostrove:
> Is there any other work needed in this PR? @rileyjmurray @nkoskelo

Bump

pygsti/protocols/gst.py
rileyjmurray: @coreyostrove, I have to disagree about LoB in this case.

There's enough nested branching here that I'm unable to understand this code by a quick skim. If I were to change this code in the future then I'd worry that I'd break something. If we move this code into a helper function then it becomes much easier to reason about its behavior. I'd advocate that it be made into a free function, rather than an instance method or even a static method.

^ I say those things even _without_ considering how much of this logic is duplicated across ``algorithms/core.iterative_gst_generator``.

rileyjmurray:
> > Is there any other work needed in this PR? @rileyjmurray @nkoskelo
> 
> Bump

@coreyostrove and @juangmendoza19 I left a comment here: https://github.com/sandialabs/pyGSTi/pull/690#discussion_r2713334949.

coreyostrove:
> > > Is there any other work needed in this PR? @rileyjmurray @nkoskelo
> > 
> > 
> > Bump
> 
> @coreyostrove and @juangmendoza19 I left a comment here: [#690 (comment)](https://github.com/sandialabs/pyGSTi/pull/690#discussion_r2713334949).

Cool. @juangmendoza19: could you make the requested refactor? Once that is done I think you're good to go to merge.

rileyjmurray:
> > > > Is there any other work needed in this PR? @rileyjmurray @nkoskelo
> > > 
> > > 
> > > Bump
> > 
> > 
> > @coreyostrove and @juangmendoza19 I left a comment here: [#690 (comment)](https://github.com/sandialabs/pyGSTi/pull/690#discussion_r2713334949).
> 
> Cool. @juangmendoza19: could you make the requested refactor? Once that is done I think you're good to go to merge.

Agreed.

rileyjmurray:
ping @juangmendoza19. Making sure this is still on your radar.

juangmendoza19:
> ping @juangmendoza19. Making sure this is still on your radar.

I created free floating functions for the optimizer validation code. I just threw them in the same file, but they seem out of place there. Any suggestions for where these should live?

Also, the two instances of code under scrutiny have slightly different logic, so I could not refactor them into a single function.

rileyjmurray (approved):
Thanks for making the changes, Juan!

Note: I see that you have two validation free-functions for the two different contexts (pygsti.algorithms.core.py and pygsti.protocols/gst.py). While I consider that undesirable, I see why you did it in this case. Since these functions are documented and marked as private (via the leading underscore) I see no reason to delay merging.

Will merge once the conflicts are resolved and tests pass.

rileyjmurray:
I think I fixed the lone failing test.

PR #692, opened 2025-11-26
Ensure that logically equivalent circuits have the same hash value

### LINKED ISSUE #684: Add logical equality check for circuits
**Is your feature request related to a problem? Please describe.**
When running subcircuit MCFE, it is possible that a circuit with the same operations, on the same qubits, is sampled multiple times. In these cases, we would like to avoid running more circuits than we need to by deduplicating any circuit that is sampled multiple times. This logic is currently implemented for MCFE experiment designs by using the pyGSTi circuit as a key in a dictionary, and therefore the deduplication is based on checking whether two circuits are equal based on their hashes. However, consider the following two circuits `circ_1` and `circ_2` given by:

```
circ_1 = pygsti.circuits.Circuit([[("Gxpi2", "Q0"), ("Gypi2", "Q1")]])
circ_2 = pygsti.circuits.Circuit([[("Gypi2", "Q1"), ("Gxpi2", "Q0")]])
```

These two circuits are logically equivalent. Each has a `Gxpi2` gate on `Q0` and a `Gypi2` gate on `Q1`. Thus, for the deduplication use case, we want to consider these circuits as equivalent. However, their hashes are different, and so we fail to deduplicate. In this particular case, `circ_1.__hash__() = 2869224465175841073` and `circ_2.__hash__() = -9005778824968725551`. The issue is that since the order of the gates within the layer is different, the hash is different, even though this has no effect on the logical behavior of the quantum circuit.

**Describe the solution you'd like**
I would like a second kind of circuit equality check that checks for logical equivalence between circuits, and is thereby independent of the insertion order of gates within a layer.

**Describe alternatives you've considered**
One could instead sort the layer labels upon construction of the circuit (or once the circuit is made static), such that the logical equivalence check that I am describing and the current equality check would coincide.

nkoskelo:
Fixes #684.

### DIFF
--- a/pygsti/baseobjs/label.py
+++ b/pygsti/baseobjs/label.py
@@ -239,11 +239,12 @@ class Label(object):
         In all other situations, we return a LabelTupTup whose inner labels are sorted
         according to their sslbls. This may raise an error if any of the sslbls are not
         comparable to one another.
+        
+        Note that we are sorting in increasing order of the first qubit that the label
+        acts on and respect the order of the sslbls for each label.
         """
 
-        if self.is_simple:
-            # ^ We check is_simple rather than isinstance(self, LabelTupTup), since
-            #   LabelTupTupWithTime and LabelTupTupWithArgs don't subclass LabelTupTup.
+        if not isinstance(self, LabelTupTup):
             return self  # type: ignore
         
         assert isinstance(self, tuple)
@@ -254,8 +255,8 @@ class Label(object):
             # The inner labels are trivially sorted.
             return self  # type: ignore
         
-        # if self.is_sorted:
-        #     return self
+        if self.is_sorted:
+            return self
 
         tmp1 = dict()
         for inner in self.components:
--- a/pygsti/baseobjs/label.py
+++ b/pygsti/baseobjs/label.py
@@ -218,11 +218,7 @@ class Label(object):
         """
         return self.IS_SIMPLE  # type: ignore
 
-    def with_sorted_inner_labels(self) -> Union[
-        LabelTup,    LabelTupWithArgs,    LabelTupWithTime,
-        LabelTupTup, LabelTupTupWithArgs, LabelTupTupWithTime,
-        LabelStr, CircuitLabel
-        ]:
+    def with_sorted_inner_labels(self) -> ConcreteLabel:
         """
         Returns `self` if either (1) we aren't a LabelTupTup or (2) we are a LabelTupTup,
         but one or more inner Label objects have sslbls == None.
--- a/pygsti/baseobjs/label.py
+++ b/pygsti/baseobjs/label.py
@@ -347,6 +347,7 @@ class LabelTup(Label, tuple):
         sslbls = tuple(integerized_sslbls)
         tup = (_sys.intern(name),) + sslbls
         obj = tuple.__new__(cls, tup)
+        obj._is_sorted = True
         return obj
 
     __new__ = tuple.__new__
@@ -675,24 +676,6 @@ class LabelTupWithTime(LabelTup, tuple):
         else:
             raise NotImplementedError("Cannot add %s to a Label" % str(type(s)))
 
-    def __eq__(self, other) -> bool:
-        """
-        Defines equality between gates, so that they are equal if their values
-        are equal.
-        """
-        # NOTE: this does not depend on self.time!
-        return tuple.__eq__(self, other)
-
-    def __lt__(self, x) -> bool:
-        # NOTE: this does not depend on self.time!
-        return tuple.__lt__(self, tuple(x))
-
-    def __gt__(self, x) -> bool:
-        return tuple.__gt__(self, tuple(x))
-
-    def __pygsti_reduce__(self) -> tuple[type, tuple[tuple, float], None]:
-        return self.__reduce__()
-
     def __reduce__(self) -> tuple[type, tuple[tuple, float], None]:
         # Need to tell serialization logic how to create a new Label since it's derived
         # from the immutable tuple type (so cannot have its state set after creation)
@@ -781,6 +764,7 @@ class LabelStr(Label, str):
     def __new__(cls, name, time=0.0) -> LabelStr:
         ret = str.__new__(cls, name)
         ret.time = time
+        ret._is_sorted = True # there is only one value so it must be sorted.
         return ret
 
     @property
@@ -989,6 +973,7 @@ class LabelTupTup(Label, tuple):
     def __new__(cls, tup_of_labels: tuple[ConcreteLabel, ...]) -> LabelTupTup:
         ret = tuple.__new__(cls, tup_of_labels)
         ret._sslbls : Optional[tuple] = None  # type: ignore
+        ret._is_sorted = False
         return ret
 
     @property
@@ -1175,7 +1160,7 @@ class LabelTupTup(Label, tuple):
         if len(self.components) == 0: return 1  # still depth 1 even if empty
         return max([x.depth for x in self.components])
 
-    def expand_subcircuits(self) -> tuple[LabelTupTup]:
+    def expand_subcircuits(self) -> tuple[LabelTupTup, ...]:
         """
         Expand any sub-circuits within this label.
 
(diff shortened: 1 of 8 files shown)


test/unit/objects/test_circuit.py
```diff
@@ -469,6 +469,21 @@ def test_compress_depth(self):
...
+
+        self.assertTrue(circ1 != circ2)
+
+        tmp = circ1.canonicalize_circuit()
+        self.assertTrue(tmp, circ1)
+        self.assertTrue(tmp, circ2)
+
+        tmp2 = circ2.canonicalize_circuit()
+        self.assertTrue(tmp, tmp2)
+
```
rileyjmurray, line 482: Bad test case. The two circuits are not equivalent. Successive strings in the list passed to the Circuit constructor specify successive __layers__. A valid test case would be
```python
circ1 = circuit.Circuit(["Gxpi2:0Gypi2:1"])  # happens to already be canonicalized
circ2 = circuit.Circuit(["Gypi2:1Gxpi2:0"])
self.assertTrue( circ1 != circ2 )
canon2 = circ2.canonicalize_circuit()
self.assertTrue( circ1 == canon2 )
```
(Your tests pass as-written because of malformed calls to assertTrue.)

rileyjmurray (changes requested):

rileyjmurray:
This PR has become FAR more involved.

#### new context
The original ``Circuit.__hash__`` function was ever-so-slightly incompatible with ``Circuit.__eq__``. This persisted with Nick's changes. 

The issue stemmed from how ``Circuit.__hash__`` would call ``self.done_editing()`` if ``self._static`` was False, while ``Circuit.__eq__`` did no such thing. So it was possible for Circuit objects to compare-equal at one point in execution, while they could have different hash values at a later point in execution.

#### original motivation
The issue that motivated this PR was a desire to recognize that if we look within a single "layer" of a LabelTupTup<like> (formally, a single Label object living in ``LabelTupTup<like>.components``) __and__ all constituent Label objects of that layer have sslbls, then we prefer a canonical ordering that sorts lexicographically by those sslbls.

#### current solution
If L is a ``Label<TupTup-like>`` that doesn't respect the preferred intra-layer Label ordering, we can get an equivalent ``Label<TupTup-like>`` __with__ that ordering by evaluating ``L_sorted = L.with_sorted_inner_labels()``. 

Circuit.done_editing() now iterates over the layers and calls this ``.with_sorted_inner_labels()`` method. This ensures that all static circuits have canonically formatted layers. We call Circuit.done_editing() as needed within ``Circuit.__eq__`` and ``Circuit.__hash__``.

#### alternative solution
An alternative to this sorting mechanism would be to have a Circuit represent a sequence of sets of LabelTup-like objects. This seemed like a far more invasive change than what I'm proposing with my latest commits.

#### other discussion
Making the type annotations throughout label.py made clear how much code duplication there is in that file. The duplication is partly because of the lack of inheritance. E.g., LabelTupTupWithArgs does not inherit from LabelTupTup. I imagine there are undesirable side-effects of using inheritance in the many classes within this file (although they don't jump out at me). A middle of the road approach could be to define free functions that handle common operations across related classes. I actually introduced one such free-function, ``_tuptup_sslbls``, to handle accessing the ``sslbls`` property of LabelTupTup, LabelTupTupWithTime, and LabelTupTupWithArgs.

pygsti/baseobjs/label.py
```diff
@@ -193,12 +212,62 @@ def expand_subcircuits(self):
...
         """
         Whether this is a "simple" (opaque w/a true name, from a circuit perspective) label or not.    
         """
+        return self.IS_SIMPLE  # type: ignore
+
+    def with_sorted_inner_labels(self) -> Union[
+        LabelTup,    LabelTupWithArgs,    LabelTupWithTime,
+        LabelTupTup, LabelTupTupWithArgs, LabelTupTupWithTime,
+        LabelStr, CircuitLabel
+        ]:
```
rileyjmurray: Use ConcreteLabel type alias.

pygsti/circuits/circuit.py
```diff
@@ -4935,8 +4973,13 @@ def done_editing(self):
         """
         if not self._static:
             self._static = True
-            self._labels = tuple([layer_lbl if isinstance(layer_lbl, _Label) 
-                                  else _Label(layer_lbl) for layer_lbl in self._labels])
+            labels_list = []
+            for layer_lbl in self._labels: # type: ignore
+                if not isinstance(layer_lbl, _Label):
+                    layer_lbl = _Label(layer_lbl)
+                layer_lbl = layer_lbl.with_sorted_inner_labels()
```
nkoskelo: We should have an early exit condition if we know that the layer is already sorted.

pygsti/baseobjs/label.py
```diff
@@ -193,12 +212,62 @@ def expand_subcircuits(self):
...
+        but one or more inner Label objects have sslbls == None.
+
+        In all other situations, we return a LabelTupTup whose inner labels are sorted
+        according to their sslbls. This may raise an error if any of the sslbls are not
+        comparable to one another.
+        """
 
-        return self.IS_SIMPLE
+        if self.is_simple:
+            # ^ We check is_simple rather than isinstance(self, LabelTupTup), since
```
nkoskelo: In `label.py` the early exits are only for sorting `len <= 1` or if the `layer` is simple. `LabelTup`, `LabelTupWithTime`, `LabelTupWithArgs`, `LabelStr`, and `CircuitLabel` are the classes which are currently simple. `CircuitLabel` can be a full circuit which is not ensured to have internally sorted layers.

pygsti/baseobjs/label.py
```diff
@@ -193,12 +212,62 @@ def expand_subcircuits(self):
...
+        # ^ We override an empty tuple with None to make sure we hit the right
+        #   codepath in Label.__new__.
+        out = Label(tmp2,
+            time=getattr(self, 'time', None),
+            args=args
+        )
+        # ^ We don't pass state_space_labels=self.sslbls in order to make sure
+        #   we hit the right codepath in Label.__new__; all codepaths that
+        #   lead to LabelTupTup-like objects require state_space_labels=None.
+        return out
```
nkoskelo, line 296: We know that this `layer` is sorted. It would be nice to not have to resort it again when the circuit is `done_editing` and this `layer` did not change.

pygsti/tools/exceptions.py
```diff
@@ -125,3 +125,12 @@ class UnnamedReportWarning(UserWarning):
     and randomly, since a user didn't provide a name.
     """
     pass
+
+
+class PoorPerformanceWarning(RuntimeWarning):
```
nkoskelo: This warning is not being used at the moment. Do we want to keep it in this PR?

pygsti/circuits/circuit.py
rileyjmurray: Agreed! Please pull down the latest code and implement. I think it would be nice for the implementation to be part of layer_lbl.with_sorted_labels(), although I'm open to having it here instead if you prefer.

pygsti/baseobjs/label.py
rileyjmurray: Good point about CircuitLabel. I think I punted on that thinking "well, existing tests pass, so we'll leave it at that." 

I'd welcome changes from you (again, just pull down the branch and commit to it) including tests for CircuitLabel handling. I think the trickiness in CircuitLabel is distinguishing between a single layer with ``k`` components versus having ``k`` distinct layers.

pygsti/baseobjs/label.py
rileyjmurray: Good point. Feel free to propose changes.

pygsti/tools/exceptions.py
rileyjmurray: Good catch. Please remove it.

rileyjmurray: Vaguely on this topic, can you please rename the HashingEditableCircuitWarning class to ImplicitDoneEditingCall (or something along those lines), and have that warning be raised when we call ``done_editing`` in ``__eq__`` and ``__hash__``?

pygsti/circuits/circuit.py
```diff
@@ -985,7 +1019,11 @@ def __eq__(self, x):
         if isinstance(x, Circuit):
             if len(self) != len(x):
                 return False
-            elif self._static and x._static and self._hash != x._hash:
+            if not self._static:
+                self.done_editing()
+            if not x._static:
+                x.done_editing()
+            if self._hash != x._hash:
```
rileyjmurray: Let's create copies of self and x as needed, then call done_editing on those copies. Raise no warning there. Keep behavior where done_editing canonicalizes.

rileyjmurray: Leave a PR comment with test suite runtime before and after the changes int he comment above. If the times are similar then we don't worry about cost overhead of the copies. If the runtimes are materially longer then we investigate ways of mitigating performance impact to users. (E.g., maintain a counter in Circuit that we increment in ``__eq__`` and start to raise a warning once it's been incremented past some threshold.)

rileyjmurray: Timing results below for two different approaches to ``__eq__``. The first approach calls done_editing() as needed, while the latter approach calls copy(editable=False) as needed. All tests were invoked with pytest -n 4, twice, on both test/unit and test/test_packages.

We see that the change is a wash. We'll move forward with the copy approach.

#### the copy(editable=False) approach
unit
```
2017 passed, 78 skipped, 281 warnings in 151.94s (0:02:31) =====================================================
2017 passed, 78 skipped, 281 warnings in 152.11s (0:02:32) =====================================================
```
test_packages
```
226 passed, 12 skipped, 24619 warnings in 127.01s (0:02:07) =====================================================
226 passed, 12 skipped, 24618 warnings in 121.72s (0:02:01) =====================================================
```

#### the done_editing() approach
unit
```
2017 passed, 78 skipped, 280 warnings in 152.86s (0:02:32) =====================================================
2017 passed, 78 skipped, 280 warnings in 152.21s (0:02:32) =====================================================
```
test_packages
```
226 passed, 12 skipped, 24618 warnings in 123.14s (0:02:03) =====================================================
226 passed, 12 skipped, 24618 warnings in 122.17s (0:02:02) =====================================================
```

rileyjmurray:
Corey requests a change. The codeblock
```
labels_list = []
for layer_lbl in self._labels: # type: ignore
    if not isinstance(layer_lbl, _Label):
        layer_lbl = _Label(layer_lbl)
        layer_lbl = layer_lbl.with_sorted_inner_labels()
        labels_list.append(layer_lbl)
```
is used a few times. It should be factored out into a new function.

rileyjmurray (approved):

Issue #398, opened 2024-02-06
Use ``@`` for matrix-matrix and matrix-vector multiplications


rileyjmurray:
Numpy's ``dot`` function executes matrix-matrix multiplication. There are many places where we use this function in pyGSTi. Python 3.5 introduced the ``@`` operator for matrix multiplication. Since then the preferred way to express ``C = np.dot(A, B)`` is ``C = A @ B``. Using ``@`` has a _big_ advantage of working with matrix/array datatypes from libraries other than numpy (like PyTorch Tensors or Dask Arrays).  In view of this, I suggest we replace calls to ``np.dot`` with an appropriate use of ``@``.

How do @sandialabs/pygsti-maintainers feel about this?

sserita:
I'm generally for replacing `np.dot` with `@` in places where we mean matrix multiplication. That sort of modernization has been a low priority but would certainly be worthwhile IMO.

eendebakpt:
There is a slight performance penalty for `@`:
```
In [167]: %timeit _np.dot(L, R)
1.49 µs ± 65 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [168]: %timeit L @ R
1.84 µs ± 137 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
(`L` and `R` dimension (16,16) from one of the inner loops in 2-qubit GST)

(maybe the overhead is due to `@` being able to dispatch to libraries like dask and pytorch, I did not check)

coreyostrove:
I suspect that any performance differences are likely going to be system configuration dependent and hard to predict. I did a quick check on my machine and found that the two were within error bars of each other.

```
import numpy as np
from numpy.random import default_rng
rng = default_rng()
L = rng.random((16,16))
R = rng.random((16,16))
```

```
%%timeit
L@R
```

1.67 µs ± 81.7 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

```
%%timeit
np.dot(L,R)
```

1.61 µs ± 16.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

That said, even if it had been the case that I'd found a performance hit on my machine, I think it would still be generally worth it for the sake of readability (if we spot any particularly hot bits of the codebase where this difference could add up we can always selectively keep dot there).

rileyjmurray:
[...]

PR #693, opened 2025-12-03
Effect-then-gate parameterization for Instruments

rileyjmurray:
This PR adds support for efficient Lindblad parameterizations of quantum instruments. Recall that our model of a quantum instrument is a collection of CPTR (completely positive trace reducing) maps that sum to a CPTP map.

The approach uses an effect-then-gate decomposition. Every CPTR member `I_k` factors as a measurement effect `E_k` followed by a post-measurement CPTP gate `G_k`,

    I_k(ρ) = G_k(E_k^½ ρ E_k^½),   E_k = I_k^†(I),

i.e. a "soft" measurement of the POVM effect `E_k` -- which sets the outcome probability `tr I_k(ρ)` -- followed by the trace-preserving back-action `G_k` on the surviving state. The two physical constraints then decouple cleanly: joint trace-preservation is exactly the statement that the effects `{E_k}` form a single POVM (`Σ_k E_k = I`), and complete positivity of a member is exactly the statement that its gate `G_k` is CP (the root-conj part `ρ ↦ E_k^½ ρ E_k^½` is always CP). So the effects are gathered into one shared `ComposedPOVM` -- which can carry any of pyGSTi's positivity-ensuring POVM parameterizations -- and each `G_k` is given its own gate parameterization, CP-constrained or not. An `n`-outcome instrument therefore needs only `n` effects and `n` gates. The decomposition is read straight off the dense superoperator (the effect is the Heisenberg dual `I_k^†(I)`; the gate is an on-support inverse plus a kernel completion).

An earlier draft of this PR --- which was very nearly merged --- used a complicated Kraus-polar approach to representing instruments. I used that approach because it it avoided an ambiguity in how to define the (ideal) conditional CPTP maps in the typical situation where the (ideal) CPTR maps were rank-deficient. I figured out how to resolve that ambiguity with the help of a robot.

### Core changes

Defines two new static methods for the Instrument class, in `instruments/instrument.py`. **These are the main entry-points into the new functionality.**
- `Instrument.from_effects` builds a parameterized instrument from a dict of POVM effects, or of (POVM effect, CPTP gate) pairs. The former case is equivalent to the post-measurement gates being the identity matrix.
- `Instrument.from_cptr_superops` builds a parameterized instrument from arbitrary dense CPTR member superoperators, recovering the `(E_k, G_k)` decomposition for you.

Construction logic for the new instruments is handled in `instruments/_construction.py`. It never imports `Instrument` to avoid an import cycle. The `_decompose_cptr(superop, basis, ...)` recovers `(E_k, G_k)` from a dense CPTR superop. The effect is `E_k = I_k^†(I)`; the gate is the CPTP completion `G_k = Q_k + P_k`, with `Q_k = I_k · pinv(rootconj(E_k))` the on-support part and `P_k` a conjugation by the projector onto `ker(E_k)`.

Changes to `instruments/__init__.py`
- `convert` delegates to `Instrument.from_cptr_superops` for Lindblad target types, promoting the shared POVM error map to the least-expressive CP parameterization that subsumes the request via `LindbladParameterization.minimal_cp_paramtype` (so `"GLND"` -> `"CPTPLND"`, `"H+s"` -> `"H+S"`).
- `instrument_type_from_op_type` is now a trivial wrapper.

Added a file, `modelmembers/operations/cptrop.py`, that defines the following pair of linear operator classes used by the new instrument members.
- `RootConjOperator`: implements the map `ρ ↦ E^½ ρ E^½`, where `E` is represented by a `POVMEffect`. Its `_rep` is recomputed via `rootconj_superop()` whenever parameters change. Parameters are shared with (and owned by) the underlying effect.
- `SummedOperator`: wraps a list of `LinearOperator`s and presents their superoperator sum as a single operator. It is **no longer used in instrument construction** (every member is now a single `ComposedOp`), but is retained and unit-tested as a general-purpose building block.

Changes in `tools/optools.py`.
- `minimal_kraus_decomposition(op_x, op_basis, ...)`: returns the shortest list of Kraus operators for a CP superoperator by eigendecomposing its Choi matrix. Now also used to validate complete positivity in `_decompose_cptr` and as the default for `LinearOperator._kraus_operators()`.
- `rootconj_superop(effect_superket, basis, ...)`: computes the superoperator for `ρ ↦ E^½ ρ E^½` given a superket representation of `E`. Validates that eigenvalues of `E` lie in `[0, 1]`; the warn/raise tolerances are the named module constants `EFFECT_EIGVAL_ABSTOL_WARN` / `EFFECT_EIGVAL_ABSTOL_ERROR`. These constants are also referenced by `RootConjOperator`.
- `superket_trace(superket, basis)`: returns `tr(ρ)` given a superket representation of `ρ`; used in `Instrument.acton` to compute outcome probabilities correctly for bases other than pauli-product or gell-mann.

Changes in `operations/lindbladerrorgen.py`.
- Adds `LindbladParameterization.minimal_cp_paramtype(abbrev)`: the least-expressive CP parameterization subsuming a requested type. Used by `convert` to keep the shared POVM CP regardless of the requested gate type.

Demo notebook
- `docs/markdown/objects/Instruments.md` (the jupytext-paired Instruments tutorial) has been substantially reframed around the effect-then-CPTP-gate representation. It builds a physically-motivated noisy `Z` measurement out of three mechanisms (amplitude damping, imperfect collapse, assignment error), recovers it with GST under both `"full TP"` and `"CPTPLND"`, and shows why a per-member completely-positive parameterization matters.
- `modelpacks/_modelpack.py`: extend `create_gst_experiment_design` to allow custom germs. This is useful for the demo notebook about GST with instruments.

Updated `models/explicitmodel.py::transform_composed_model` so that it works with `Instrument` and `TPInstrument` objects. The two are handled differently: an `Instrument`'s component operations are gauge-transformed and replaced individually, whereas a `TPInstrument` is updated via its own `transform_inplace`.

### Tests

- New file `test/unit/modelmembers/test_instrument.py`: covers `Instrument`, `TPInstrument`, `RootConjOperator`, `SummedOperator`, and the two new classmethods (`FromCptrSuperopsTester`, `FromEffectsTester`). Includes round-trip checks (construct -> `to_dense` -> compare), `from_vector`/`to_vector` consistency, and basic `acton` tests. New coverage targeting the effect-then-gate design.
- `TPInstrumentOp` tests moved from `test/unit/modelmembers/test_operation.py` to the new `test_instrument.py`; `test/unit/objects/test_instrument.py` removed.
- Add tests for `transform_composed_model` to `test/unit/objects/test_explicitmodel.py`.
- `test/test_packages/objects/test_instruments.py`: add a save/load + pickle round-trip over an instrument built with `from_cptr_superops`, exercising serialization of the nested member structure.
- `test/unit/tools/test_optools.py`: add regression tests for `minimal_kraus_decomposition` (Kraus action matches the channel; completeness relation).

### Minor changes

instrument.py
- `acton` now uses `superket_trace` to compute outcome probabilities correctly for non-PP bases.
- `__init__` was refactored to push its validation/inference logic into small module helpers (`_as_ordered_member_list`, `_infer_space_and_evotype`, `_validate_member_consistency`); the instrument modules also gained type annotations throughout.

SDPs and diamond distance
- `tools/sdptools.py`: adds a new helper, `solve_sdp(prob, ...)`, that wraps the solver-priority loop (MOSEK -> CLARABEL -> CVXOPT) for easy use in different contexts. Returns `(objective_val, varvals_dict)`.
- Updated `diamonddist` in `optools.py` simplified to call `solve_sdp`.

Kraus operators in LinearOperator classes.
- `modelmembers/operations/linearop.py`: added `LinearOperator._kraus_operators()` with a default implementation via `minimal_kraus_decomposition`.
- `modelmembers/operations/denseop.py`: replaced the in-line implementation with delegation to `LinearOperator._kraus_operators()`.

Almost entirely incidental
- `tools/exceptions.py`: added `DubiousTargetWarning` for alerting users to unusual target specifications.
- `modelmembers/states/cptpstate.py`: minor formatting cleanup in `deriv_wrt_params`; removed a `hessian_wrt_params` override that was raising `NotImplementedError` and is now handled by the base class.
- `baseobjs`: export the `BasisLike` type alias (used by the new annotations).
- `modelmembers/modelmember.py`: allow array-valued submember `rpindices` (`list_to_slice(..., array_ok=True)`), needed for the shared-effect parameter indices.
- A handful of one-line touch-ups (type-annotation imports, `lstsq(..., rcond=None)`, return-type annotations).


pcwysoc:
The changes to the analysis pipeline do not play nice with reduced models (see error message). I forgot to switch branches and tried to fit a reduced model on a dataset not including instruments. It might make sense to check if there are instruments in the model and only then throw up an error.

Screenshot 2025-12-09 at 1 48 38 PM

pcwysoc:
Here's the most recent version of my CPTPInstrument which includes Corey's improvements to make it write to and load from disk: 
[ExampleSavingComposedInst_rev1.ipynb](https://github.com/user-attachments/files/24062199/ExampleSavingComposedInst_rev1.ipynb)

rileyjmurray:
Idea for GLND parameterization: use GLND trace-reducing operators. Create a new ProjectedPOVM modelmember that wraps a "not actually positive POVM" (e.g., TPPOVM or a POVM in GLND parameterization) and its from_vector works by materializing POVM effects (indefinite Hermitian matrices in general) and projecting them (in some metric) onto the set of psd Hermitian matrices that sum to the identity.

pygsti/tools/matrixtools.py
```diff
@@ -150,6 +150,54 @@ def is_valid_density_mx(mx, tol=1e-9):
...
+
+
+def project_onto_densities(herm_mx: _np.ndarray, ord:Literal['fro','nuc']='fro') -> _np.ndarray:
+    if ord != 'fro':
+        raise NotImplementedError()
+    assert_hermitian(herm_mx, 1e-8)
+    u, v, udag = eigendecomposition(herm_mx, assume_hermitian=True)
+    proj_v = project_onto_simplex(v, 2)
+    density_mx = u @ (proj_v[:, None] * udag)
+    return density_mx
```
rileyjmurray: Update type annotations to indicate what we currently support.

pygsti/modelmembers/povms/__init__.py
```diff
@@ -26,7 +27,7 @@
 from .fulleffect import FullPOVMEffect
 from .fullpureeffect import FullPOVMPureEffect
 from .marginalizedpovm import MarginalizedPOVM
-from .povm import POVM
+from .povm import POVM  # TODO: replace implementation with that of _BasePOVM ?
```
rileyjmurray: Raise this point in a GitHub issue and remove from this PR.

pcwysoc:
@rileyjmurray The notebook was shared because at one point you thought it would be useful to have the alternate CPTP Instrument approach documented here. I can't remember why right now. I would just ignore it, as none of the functionality is distinct from functionality already present in this PR.

rileyjmurray:
> The notebook was shared because at one point you thought it would be useful to have the alternate CPTP Instrument approach documented here. I can't remember why right now. I would just ignore it, as none of the functionality is distinct from functionality already present in this PR.

@pcwysoc, we had three reasons to be interested in your alternate implementation.
 1. It had clean support for reduced-order Lindblad error generator models and non-CP models. The implementation on this branch fixes most of that, but the way this branch supports non-CP models is different than (and maybe not preferable to) the way done in your notebook.
 2. The parameterization in your notebook seemed to play nicer with pyGSTi's optimizer. IIRC, you needed to tune the optimizer step size using the parameterization on this branch, but you didn't need to do such a thing with your notebook's parameterization.
 3. The notebook parameterization is what's described in the paper. 

Given that info, do you still recommend we ignore that branch? In some just just the fact that we've had this conversion on the GitHub PRs serves as documentation in case we want to come back to this, but I figure I'd ask now.

pcwysoc:
@rileyjmurray answers to your points: 

1. This may be worth considering. My implementation gives access to at least one useful reduced model. Setting the constituent error generator to an 'S' parameterization effectively produces a uniform stochastic instrument. Other models are less useful, as the second order effects of the elementary error generators can become significant and as such not completely remove unwanted FOMGIs. 
2. Both parameterizations experience suboptimal optimizer behavior when ran on the paper data. With my CPTP Instrument implementation, I handled this by running an initial analysis with a TP Instrument, truncating the model to a nearby model with a CPTP Instrument, and using that to seed the CPTP Instrument analysis. I expect both models would benefit from similar adjustments to the optimizer settings. 
3. This is not actually the case. I switched at last minute to your implementation of a CPTP Instrument for forward-compatibility.

rileyjmurray:
Thanks for the quick reply, @pcwysoc! Your response raises some more questions for me. I've send you an email so we can discuss without making a mess of this GitHub PR's comment history.

EDIT: Short update. While I knew the paper was updated to use the then-latest feature on this branch, I wasn't sure if it was important that the paper's mathematical formalism had a 1-to-1 implementation in pyGSTi. My new impression is that it's _not_ terribly important.

rileyjmurray:
The failing tests are unrelated to this PR. They'll be fixed once https://github.com/sandialabs/pyGSTi/pull/724 is merged in.

EDIT: I don't think PR 724 is actually relevant. The failing test is due to a plotly issue. The diff on PR 724 claims changes to fix the plotly issue, but it seems the fix landed on develop in a different PR. See below for a screenshot of PR 724's claimed diff, and then a screenshot of the current lines on develop.
image
image

I'm going to trigger tests for this PR again.

rileyjmurray:
> I'm going to trigger tests for this PR again.

Tests failed in the exact same way they did before. I'll try merging develop into cptp-instruments. If that merge doesn't bring the plotly fix in with it then I'll just duplicate that change here.

docs/markdown/objects/Instruments.md
rileyjmurray, line 1: @pcwysoc can you please review this file and either approve as-is or request changes? I had a robot write it. I iterated with the robot a little but I want your eyes on it.

rileyjmurray:
@coreyostrove I added an `instrument_superops_from_effects` function based on our conversation yesterday. This should be good to go. Will wait for approval from you and @pcwysoc before merging.

docs/markdown/objects/Instruments.md
```diff
@@ -12,164 +12,229 @@ kernelspec:
...
 
-In pyGSTi, instruments are represented as collections of gates, one for each classical "outcome" of the instrument.  This tutorial will demonstrate how to add instruments to `Model` objects, compute probabilities using such `Model`s, and ultimately perform tomography on them.  We'll start with a few familiar imports:
+This tutorial demonstrates how to model, simulate, and perform tomography on quantum *instruments*: maps that act on a qubit state (density matrix) and produce a qubit state *together with* a classical outcome.  Formally, an instrument is a map from $\mathcal{B}(\mathcal{H})$, the space of density matrices, to $\mathcal{B}(\mathcal{H}) \otimes K(n)$, where $K(n)$ is a classical space of $n$ elements.  Instruments are the natural model for **mid-circuit measurements** (MCMs): operations that read out a qubit partway through a circuit and leave behind a (possibly disturbed) post-measurement state.
```
coreyostrove: qubit -> qudit. It sounds like instruments are definitionally restricted otherwise.

docs/markdown/objects/Instruments.md
```diff
@@ -12,164 +12,229 @@ kernelspec:
...
-target_model = std.target_model()
+mdl_ideal = std.target_model()
+
+# Build the ideal projective Z-measurement instrument from the POVM effects.
+E0 = mdl_ideal.effects['0']
+E1 = mdl_ideal.effects['1']
+# Alternate indexing that names the POVM explicitly:
+#   E0 = mdl_ideal['Mdefault']['0']   # 'Mdefault' = POVM label, '0' = effect label
+Gmz_plus  = E0 @ E0.T   # note: effect vectors are stored as column vectors
+Gmz_minus = E1 @ E1.T
```
coreyostrove: This syntax will be going away when we get rid of the dense operator interface, so we should include `to_dense` calls.

docs/markdown/objects/Instruments.md
```diff
@@ -12,164 +12,229 @@ kernelspec:
...
+- The post-conjugation channels take whatever parameterization you requested.  If that parameterization is itself CP-constrained (`"CPTPLND"`, `"H+S"`), each member is completely positive.  If it is only trace-preserving (`"GLND"`, `"H+s"`), the members remain jointly trace-preserving but may individually fail to be CP.
+
+So complete positivity of the members is something you *opt into* by choosing a CP-constrained post-conjugation parameterization -- it is not an automatic consequence of the Kraus-polar structure, which by itself guarantees only trace preservation.
+
+You can build such an instrument directly with `kraus_polar_instrument`, passing the members as dense superoperators:
 
 ```{code-cell} ipython3
-dict(target_model.probabilities( pygsti.circuits.Circuit(( ('Gxpi2',0) , ('Iz',0) )) ))
+kp = kraus_polar_instrument({'p0': Gmz_plus, 'p1': Gmz_minus}, basis)
```
coreyostrove: This should demonstrate the kwargs that control the CPness of the kraus-polar instrument.

docs/markdown/objects/Instruments.md
```diff
@@ -12,164 +12,229 @@ kernelspec:
...
+
+    # Each "true" outcome is weak-measurement-after-damping (Kraus rank 2).
+    S0 = FullArbitraryOp.from_kraus_operators([M0 @ A0, M0 @ A1], 'pp').to_dense()
+    S1 = FullArbitraryOp.from_kraus_operators([M1 @ A0, M1 @ A1], 'pp').to_dense()
+
+    # Classical assignment error mixes the recorded labels.
+    R0 = (1 - eps) * S0 + eps * S1
+    R1 = eps * S0 + (1 - eps) * S1
+    return Instrument({'p0': R0, 'p1': R1})
+```
```
coreyostrove, line 141: This is actually not the same error process as is considered in the paper that is being referred to. The subtlety here is that the noisy instruments in that setting are specified in terms of generators. I.e. all three errors are happening at the same time, not as the result of sequential compositions. Since these don't commute you won't get the same thing in both cases (though the differences should be at second-order). Not necessarily a problem, per se (except insofar as end-users are wont to pattern match off of these things)

docs/markdown/objects/Instruments.md
coreyostrove, line 1: Great tutorial!

pygsti/modelmembers/operations/cptrop.py
```diff
@@ -0,0 +1,208 @@
+"""
+The classes in this file help us represent CPTR (completely positive trace-reducing)
+channels using the error generator formalism.
```
coreyostrove, line 13: In the tutorial you defined CPTR to be trace non increasing.

pygsti/modelmembers/operations/cptrop.py
```diff
@@ -0,0 +1,208 @@
...
+Kraus-rank-1 terms together when needed.
+"""
+#***************************************************************************************************
+# Copyright 2015, 2019, 2025 National Technology & Engineering Solutions of Sandia, LLC (NTESS).
+# Under the terms of Contract DE-NA0003525 with NTESS, the U.S. Government retains certain rights
+# in this software.
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License.  You may obtain a copy of the License at
+# http://www.apache.org/licenses/LICENSE-2.0 or in the LICENSE file in the root pyGSTi directory.
+#***************************************************************************************************
```
coreyostrove: Copyright marking should be at the top of the file.

pygsti/modelmembers/states/cptpstate.py
```diff
@@ -363,28 +363,3 @@ def has_nonzero_hessian(self):
         bool
         """
         return True
-
-    def hessian_wrt_params(self, wrt_filter1=None, wrt_filter2=None):
```
coreyostrove, line 367: Why delete this?

test/unit/modelmembers/test_instrument.py
```diff
@@ -0,0 +1,285 @@
...
+    __test__ = False
+    constructor: type
+
+    def setUp(self):
+        self.n_elements = 32
+        self.model = std.target_model()
+        E = self.model.povms['Mdefault']['0']
+        Erem = self.model.povms['Mdefault']['1']
+        self.Gmz_plus = np.dot(E, E.T)
+        self.Gmz_minus = np.dot(Erem, Erem.T)
```
coreyostrove: See comment above about dense operator interfaces.

test/unit/modelmembers/test_instrument.py
```diff
@@ -0,0 +1,285 @@
...
+        self.assertIsInstance(converted, Instrument)
+        self.assertSetEqual(set(converted.keys()), {'plus', 'minus'})
+
+    def test_member_dense_shapes(self):
+        dim = self.basis.dim
+        for op in self.instrument.values():
+            mx = op.to_dense('HilbertSchmidt')
+            self.assertEqual(mx.shape, (dim, dim))
+
+    def test_sum_of_members_is_cptp_like(self):
```
coreyostrove: This is only testing the TP constraint, so rename function to reflect that.  Could you add a test that coves the CP constraint too, though? E.g. testing that an exception is raised when trying to build a CP constrained instrument from non CPTR maps?

coreyostrove (changes requested):
A few requests for changes.

docs/markdown/objects/Instruments.md
rileyjmurray: Fair enough! I'll add a disclaimer that this isn't _quite_ the model from that paper.

pygsti/modelmembers/operations/cptrop.py
rileyjmurray: I'll add a qualifier after the sentence. Something like
> By "reducing" we mean "non-increasing."

pygsti/modelmembers/states/cptpstate.py
rileyjmurray: The base class already raises a NotImplementedError. Redundant not-implemented-errors bad.

test/unit/modelmembers/test_instrument.py
coreyostrove: Oh, and also #398 (favoring @ instead of dot).

rileyjmurray:
Ping @coreyostrove while making edits to Instrument.md I finally appreciated your push-back on the Kraus-polar approach. At the end of the day that was an implementation detail. Making it the main entry point for the new functionality led to very confusing documentation. Also, turns out there's a way to bypass the parameter-count blow up when the CPTR operators had Kraus rank > 1. This realization led me to do a pretty comprehensive refactor of the PR. Please review again at your earliest convenience.

test/unit/modelmembers/test_operation.py
rileyjmurray, line 1: Content removed from this file was ported to test/unit/modelmembers/test_instrument.py.

docs/markdown/objects/Instruments.md
```diff
@@ -4,172 +4,244 @@ jupytext:
...
-```
+### The effect-then-CPTP-gate representation
+
+Every completely-positive instrument member factors as a **measurement effect followed by a post-measurement CPTP gate**:
 
-## Generating probabilities 
-Instrument labels (e.g. `"Iz"`) may be included within `Circuit` objects, and `Model` objects are able to compute probabilities for them just like normal (non-instrument) operation sequences.  The difference is that probabilities are labeled by tuples of instrument and POVM outcomes - referred to as **"outcome tuples"** - one for each instrument and one for the final POVM:
+$$\mathcal{I}_k(\rho) = \mathcal{G}_k\!\left( E_k^{1/2}\,\rho\,E_k^{1/2} \right), \qquad E_k = \mathcal{I}_k^{\dagger}(I).$$
+
+This is a "soft" measurement of the POVM effect $E_k$ -- which sets the outcome probability $\mathrm{tr}\,\mathcal{I}_k(\rho)$ -- followed by the trace-preserving back-action $\mathcal{G}_k$ on the surviving state.  The effect $E_k$ is just the Heisenberg-dual of the member applied to the identity. This is the *measurement-then-disturbance* picture used to describe the physical mechanisms in [arXiv:2602.03938](https://arxiv.org/abs/2602.03938).
```
coreyostrove: I don’t think this reference is correct. I didn’t recall this representation appearing directly in that paper so I just skimmed through it to confirm. A similar looking, but somewhat different one called MPR does appear. I also can’t find (at least via ctrl+F) the phrase “measurement-then-disturbance” in this paper. @pcwysoc can correct me if I am missing something.

coreyostrove (approved):
Thanks for the hard work, @rileyjmurray! Sorry for the delay, I needed to stare at this for a while an convince myself that it was true that we could *always* write an instrument in this effect-then-gate form (which I eventually was convinced of).  

I’m approving this now to avoid adding bottlenecks (since the below is not a code concern), but could you please address the question/concern I flagged before merging? Flagging @pcwysoc for tag-team support.

docs/markdown/objects/Instruments.md
rileyjmurray: Good catch. Will revise before merge.

PR #699, opened 2025-12-17
Leakage modeling: confidence intervals and more robust gauge optimization

rileyjmurray:
Changes
 * Remove unnecessary model copies in confidenceregionfactory.py.
 * Log a warning if we try to add confidence regions to a model whose number of parameters differs from that of "final iteration estimate".
 * Add a helper to parse the confidence_level argument in construct_standard_report. We now handle fractional values in (0,1) in the natural way.
 * Leakage-aware gauge optimization no longer has a first stage that puts all weight on SPAM. Instead, it puts equal weight on SPAM and gates, and uses Frobenius loss instead of Frobenius-squared. The latter change reduces bias toward gauges that spread error out across gates.
 * Update construct_leakage_report to accept a confidence_level argument. When provided, we automatically add the confidence region factories needed for report generation.

### DIFF
--- a/pygsti/protocols/confidenceregionfactory.py
+++ b/pygsti/protocols/confidenceregionfactory.py
@@ -1033,9 +1033,8 @@ class ConfidenceRegionFactoryView(object):
         #elements of fn_dependencies are either 'all', 'spam', or
         # the "type:label" of a specific gate or spam vector.
         all_gpindices = []
+        mdl = self.model
         for dependency in fn_dependencies:
-            mdl = self.model.copy()  # copy that will contain the "+eps" model
-
             if dependency == 'all':
                 all_gpindices.extend(range(mdl.num_params))
             else:
@@ -1054,15 +1053,12 @@ class ConfidenceRegionFactoryView(object):
                     elif typ == "instrument": modelObj = mdl.instrument_blks['layers'][lbl]
                     else: raise ValueError("Invalid dependency type: %s" % typ)
                 all_gpindices.extend(modelObj.gpindices_as_array())
-
-        vec0 = mdl.to_vector()
         all_gpindices = sorted(list(set(all_gpindices)))  # remove duplicates
-
+        
+        vec0 = mdl.to_vector()
         for igp in all_gpindices:  # iterate over "global" Model-parameter indices
             vec = vec0.copy(); vec[igp] += eps
             mdl.from_vector(vec)
-            mdl.basis = self.model.basis  # we're still in the same basis (maybe needed by fn_obj)
-
             f = fn_obj.evaluate_nearby(mdl)
             if isinstance(f0, dict):  # special behavior for dict: process each item separately
                 for ky in gradF:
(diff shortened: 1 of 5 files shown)


pygsti/protocols/confidenceregionfactory.py
rileyjmurray, line 1: It was unnecessary to make copies to ``self.model``; it suffices to mutate``self.model`` directly and restore it to the state it had on entry before returning.

pygsti/tools/leakage.py
```diff
@@ -509,7 +517,14 @@ def add_lago_models(results: ModelEstimateResults, est_key: Optional[str] = None
             else:
                 gop_params = std_lago_gopsuite(results.estimates[est_key].models['target'])
             gos = GSTGaugeOptSuite(gaugeopt_argument_dicts=gop_params)
-        _add_param_preserving_gauge_opt(results, est_key, gos, verbosity)
+        try:
+            _add_param_preserving_gauge_opt(results, est_key, gos, verbosity)
```
rileyjmurray, line 521: This parameterization-preserving gauge optimization only happens when using ``add_lago_models``. If we don't call this function (say, because we're doing full qutrit modeling) then the "stdgaugeopt" model will most likely have full TP parameterization. This invalidates the assumptions required to build confidence intervals, per the discussion here: https://github.com/sandialabs/pyGSTi/issues/620#issue-3275460535.

We should update the confidence-region-generation pipeline to raise an error (or, at the very least, a warning) if the model parameterization of "stdgaugeopt" differs materially from that of "final iteration estimate".

rileyjmurray:
There were several test failures in the Ubuntu 3.12 build. It's unclear which, if any, are related to changes in this PR.  This comment contains my findings for the various errors.

This specific test failure:
https://github.com/sandialabs/pyGSTi/actions/runs/20318418382/job/59600801805#step:12:1122
Is triggered here when running numpy >= 2:
https://github.com/sandialabs/pyGSTi/actions/runs/20318418382/job/59600801805?pr=699#step:12:355.
This error can be fixed by replacing the lines 
```python
        for i, interpolator in enumerate(self.interpolators.flat):
            value.flat[i] = interpolator(*v)
```
in pygsti.extras.interpygate.core with
```python
        for i, interpolator in enumerate(self.interpolators.flat):
            u = interpolator(*v)
            w = u.item() if isinstance(u, _np.ndarray) else u
            value.flat[i] = w
```

pygsti/tools/leakage.py
rileyjmurray: Made changes to log a warning.

rileyjmurray:
Failures will be resolved once #703 is merged in.

pygsti/protocols/estimate.py
coreyostrove, line 1: Given our current standard practice workflow for error bars this is going to get hit pretty much every single time. What other changes would be needed to avoid hitting this all the time? Is it parameterization-preserving gauge transformations?

coreyostrove (approved):
Approving changes, but if you could answer the question I left prior to merging that would be appreciated.

pygsti/protocols/estimate.py
rileyjmurray: We'd need to make parameterization-preserving gauge optimization the default. I recognize that this warning will be triggered for the vast majority of times that people currently compute error bars. I'd like to merge as-is despite that. We should have another PR that fixes this issue properly by changing how we do gauge optimization.

rileyjmurray:
I re-ran tests after merging #703 and I still got the exact same error. I'm guessing I would have needed to merge develop back into this branch in order for the tests to have updated. Not worth doing that and waiting when the failures will clearly be resolved by #703. Merging now!

PR #701, opened 2026-01-06
MAINT: LindbladCoefficientBlock blocktype branching

rileyjmurray:
This PR has major changes to LindbladCoefficientBlock. I've taken care to produce code that is equivalent to the old code from an input-output perspective.

After this PR we'll be well-positioned to make a base class for LindbladCoefficientBlock (or just make LindbladCoefficientBlock a base class) from which new classes can inherit. I want to do this because I have an idea for a sparse Cholesky approach to reduced-order CPTP models.

The diff for this PR is kind of awful. Honestly, it's a little hopeless to track what happened. My ask is that we merge anyway once all tests pass. The approach I took to the refactor was to go function by function, preserving as much of the original implementation as possible. My first pass used an LLM for applying the refactoring pattern below. Once some tests failed after this, I did another pass that manually copy-pasted relevant codeblocks from the old ``perform_operation`` functions into the new ``_perform_operation_<blocktype>`` helper functions.

### general approach
I took functions like
```python
def perform_operation(self, *args, **kwargs):
    # some pre-branching work ...
    if self._blocktype == 'ham':
        # insert 3 to 5 lines here; call this codeblock "X"
        pass
    elif self._blocktype == 'other_diagonal':
        # insert 10 to 20 lines of code here; call this codeblock "Y"
        pass
    elif self._blocktype == 'other':
        # insert 20 to 80 lines of code here; call this codeblock "Z"
        pass
    else:
        raise ValueError('unsupported block type')
     # some post-branching work ...
     return
```
and refactored them to
```python
def perform_operation(self, *args, **kwargs):
    # some pre-branching work ...
    if self._blocktype == 'ham':
        self._perform_operation_ham(*args, **kwargs)
    elif self._blocktype == 'other_diagonal':
        self._perform_operation_otherdiag(*args, **kwargs)
    elif self._blocktype == 'other':
        self._perform_operation_other(*args, **kwargs)
    else:
        raise InvalidBlockTypeError()
     # some post-branching work ...
    return

def _perform_operation_ham(self, *args, **kwargs):
    # insert codeblock "X" here
    return

def _perform_operation_otherdiag(self, *args, **kwargs):
    # insert codeblock "Y" here
    return
    
def _perform_operation_other(self, *args, **kwargs):
    # insert codeblock "Z" here
    return

### DIFF
--- a/pygsti/modelmembers/operations/lindbladcoefficients.py
+++ b/pygsti/modelmembers/operations/lindbladcoefficients.py
@@ -704,32 +704,13 @@ class LindbladCoefficientBlock(_NicelySerializable):
             raise InvalidParamModeError(self._param_mode, self._block_type)
         return labels
 
-    def _block_data_to_params(self, truncate=False):
+    def _block_data_to_params(self, truncate: Union[bool, float]=False):
         """
-        Compute parameter values from coefficient values.
-
-        Constructs an array of paramter values from an arrays of
-        coefficient values for this block.
-
-        Parameters
-        ----------
-        truncate : bool or float, optional
-            Whether to truncate the coefficients given by `block_data` in
-            order to meet constraints (e.g. to preserve CPTP) when necessary.
-            If >= 0 or False, then an error is thrown when the given projections
-            cannot be parameterized as specified, using the value given as the
-            the maximum negative eigenvalue that is tolerated (`False` is equivalent
-            to 1e-12).  True tolerates *any* negative eigenvalues.
-
-        Returns
-        -------
-        param_vals : numpy.ndarray
-            A 1D array of real parameter values. Length is `len(self.basis_element_labels)`
-            in the case of `'ham'` or `'other_diagonal'` blocks, and `len(self.basis_element_labels)**2`
-            in the case of `'other'` blocks.
+        Compute parameter values implied by self.block_data.
+        (The public method now simply dispatches on block type.)
         """
         if truncate is False:
-            ttol = -1e-14  # (was -1e-12) # truncation tolerance
+            ttol = -1e-14
         elif truncate is True:
             ttol = -_np.inf
         else:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pygsti/modelmembers/operations/lindbladcoefficients.py [hunk 1]
```diff
@@ -23,6 +23,70 @@
...
+
+    if sparse:
+        # Note: complex OK here sometimes, as only linear combos of "other" gens
+        # (like (i,j) + (j,i) terms) need to be real.
+        superops = [leftTrans @ (mx @ rightTrans) for mx in superops]
+        for mx in superops:
+            mx.sort_indices()
+    else:
+        # superops = _np.einsum("ik,akl,lj->aij", leftTrans, superops, rightTrans)
+        superops = _np.transpose(_np.tensordot(
```
nkoskelo: Are we sure that this ordering is always going to be the best ordering performance wise?

rileyjmurray: Not at all! But the point of this PR is to be minimally-invasive. In an ideal world we'd formally certify that the current refactor does not affect input-output behavior of the existing functions.

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -23,6 +23,70 @@
...
+
+def _custom_superops_stdbasis_conversion(mx_basis, sparse, superops):
+    """ 
+    The input superops is either:
+        (1) a list of CSR matrices, or 
+        (2) an ndarray of shape (n, d, d) for some d,
+    where superops[i] is a superoperator in the mx_basis representation.
+
+    The output superops has the same datatype as the input, although
+    the superoperators are now in the standard (matrix-unit) basis.
```
coreyostrove: I am pretty sure the output superoperators should be in mx_basis, not the standard basis.

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -23,6 +23,70 @@
...
+    """
+    mxBasisToStd = mx_basis.create_transform_matrix(_BuiltinBasis("std", mx_basis.dim, sparse))
+    # use BuiltinBasis("std") instead of just "std" in case mx_basis is a TensorProdBasis
+
+    rightTrans = mxBasisToStd
+    leftTrans = _spsl.inv(mxBasisToStd.tocsc()).tocsr() if _sps.issparse(mxBasisToStd) \
+        else _np.linalg.inv(mxBasisToStd)
```
coreyostrove: Rather than computing the inverse we should try using the `reverse_transform_matrix` method of mx_basis. I believe this should be equivalent, but has the benefit of having an associated LRU cache for its output.

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -124,225 +188,214 @@ def _set_block_data(self, block_data, truncate):
...
-                for i, Lm in enumerate(mxs):
-                    superops[i] = _lt.create_lindbladian_term_errorgen('O', Lm, Lm, sparse)  # in std basis
-            elif self._block_type == 'other':
-                superops = [None] * nMxs**2 if sparse else _np.empty((nMxs**2, d2, d2), 'complex')
-                for i, Lm in enumerate(mxs):
-                    for j, Ln in enumerate(mxs):
-                        superops[i * nMxs + j] = _lt.create_lindbladian_term_errorgen('O', Lm, Ln, sparse)
+        if flat or self._block_type in ('ham', 'other_diagonal'):
+            if include_1norms:
+                return _copy.deepcopy(cached_superops), cached_superops_1norms.copy()
```
coreyostrove, line 254: What purpose is the deepcopy serving? In my experience deepcopy is a performance nightmare that should be avoided unless absolutely needed (and when needed it is often way more performant to write a tailored deepcopy routine).

coreyostrove (approved):
A couple non-blocking questions/requests. Generally this looks good and I appreciate the hard work and your patience waiting on my review. git diff really made a mess of this one...

A provocative question for you to seed future refactoring discussions. Should we (eventually) go even one step further and split the three block types off into separate classes, ditching the branching and dispatch structure entirely? There is a small shared core, but really the three blocks do quite different things with their internal parameters and it might be even cleaner longer term.

rileyjmurray:
> A provocative question for you to seed future refactoring discussions. Should we (eventually) go even one step further and split the three block types off into separate classes, ditching the branching and dispatch structure entirely? There is a small shared core, but really the three blocks do quite different things with their internal parameters and it might be even cleaner longer term.

Not provocative at all! That's my evil plan 😈. This is just a stepping stone.

pygsti/modelmembers/operations/lindbladcoefficients.py
rileyjmurray: Good catch. Fixed.

pygsti/modelmembers/operations/lindbladcoefficients.py
rileyjmurray: I've changed to use mx_basis.inverse_transform_matrix for most codepaths in this function. The left change-of-basis operator (now called std_to_mxbasis) can still be constructed as _spsl.inv(mxbasis_to_std.tocsc()).tocsr() to preserve the old behavior. IMO, it's not worth it right now to figure out if it's safe to change that codepath to also use mx_basis.inverse_transform_matrix.

pygsti/modelmembers/operations/lindbladcoefficients.py
rileyjmurray: I'm not sure. My changes don't affect whether we deepcopy though.

Current state of develop:
https://github.com/sandialabs/pyGSTi/blob/37e3f8ce1fce8987346f98ce501f5eddff52cb7e/pygsti/modelmembers/operations/lindbladcoefficients.py

Current state of proposed changes:
https://github.com/sandialabs/pyGSTi/blob/a419af1bdb83b85ee9a14dade842579ba1dbe1d8/pygsti/modelmembers/operations/lindbladcoefficients.py

Both reference copy.deepcopy exactly twice, and in the same way. I'll mark this thread as resolved.

rileyjmurray:
Original response to Corey's comments referenced a wrong function name. I've fixed in the latest commit. Will see if tests pass.

PR #702, opened 2026-01-06
Use np.einsum instead of np.linalg.multi_dot

nkoskelo:
In python 3.12 it appears as though `np.linalg.multi_dot(x.T, A, y)` will retain the output as a 1x1 array which is not convertible directly to a scalar value. If we instead use `np.einsum` ensuring that the output has no indices remaining then the output will be a scalar value as intended.


rileyjmurray:
I just created another branch to resolve these issues and related issues. Thanks though!

PR #704, opened 2026-01-14
Interpygate example: add import for csaps and replace log with logm

rileyjmurray:
This tiny PR makes it so the notebook runs without error.


coreyostrove (approved):
Looks good!

Issue #542, opened 2025-02-18
customsolve.py: incorrect number of arguments to _find_pivot.


rileyjmurray:
Here's a code snippet in pygsti/optimize/customsolve.py on the head of master.

https://github.com/sandialabs/pyGSTi/blob/c74fcb2d4e61cd5fae573d55a0baf6f8618be800/pygsti/optimize/customsolve.py#L157-L168

The smbuf1b variable is never used. It looks like it should be passed as an argument to _find_pivot, after smbuf1, per the signature below (variable names in the signature are just ``buf1`` and ``buf1b``):

https://github.com/sandialabs/pyGSTi/blob/c74fcb2d4e61cd5fae573d55a0baf6f8618be800/pygsti/optimize/customsolve.py#L213-L214

As a consequence of smbuf1b _not_ being passed to _find_pivot, we're calling _find_pivot with only 15 positional arguments when it requires 16. 

The appearance of this bug is uncommon because there's a dedicated code path for handling linear solves with matrices of order < 10,000 (below) that gets triggered in our testing and even in most large GST runs:
https://github.com/sandialabs/pyGSTi/blob/c74fcb2d4e61cd5fae573d55a0baf6f8618be800/pygsti/optimize/customsolve.py#L98-L126

Possible ways to move forward:
1. __Temporary__. Just change the dimension threshold to 100,000 instead of 10,000. Storing one double-precision matrix of this size requires ~80GB. The overhead of the Python call means this will require 160 GB of "free" RAM to run. That's a reasonable amount for modern compute servers.
2. Try and fix the bug in the custom Gaussian elimination code. Have the branch on matrix size check a user-modifiable constant instead of a hard coded value. Some of our tests can run this code where that matrix size threshold is quite small.
3. __Ambitious__. Create a new ArrayInterface class that relies on Dask arrays. Replace the custom Gaussian elimination code with calls to Dask's distributed linear algebra functions: https://docs.dask.org/en/stable/_modules/dask/array/linalg.html.

rileyjmurray:
It looks like the bug was introduced here:
https://github.com/sandialabs/pyGSTi/commit/057383207f043020399a9e7f01ac045691387909

That commit leaves the call to _find_pivot unchanged:
https://github.com/sandialabs/pyGSTi/commit/057383207f043020399a9e7f01ac045691387909#diff-b5f5b8b216bb22ad8b9236aee7c88e7f141562fb4a7584e15e5a885f26361cbdR166-R168
[...]

PR #707, opened 2026-01-22
General leakage modeling

rileyjmurray:
This PR promotes `pygsti.tools.leakage` from a monolithic 616-line module into the
top-level subpackage `pygsti.leakage`, and simultaneously generalizes leakage analysis
so that the computational subspace is inferred from the basis rather than supplied as
an explicit `n_leak: int` parameter.

***Future work:*** figure out the basis element labeling scheme for the `l2p1` basis.
I don't consider that labeling scheme part of the public API. However, I want to have
open discussions with collaborators about the options, and we don't have time to
do that before the PR merge.

### Subpackage layout

`pygsti/tools/leakage.py` is deleted. Its contents are reorganized into five modules:

| Module | Contents |
|---|---|
| `leakage/core.py` | `computational_effect`, `computational_superkets`, `computational_projector`, `augment_for_leakage_modeling` |
| `leakage/metrics.py` | Choi-induced metrics, projected metrics, transport profiles |
| `leakage/models.py` | `leaky_qubit_model_from_pspec`, `promote_bb_to_bt`, `random_unitary_excitation` |
| `leakage/gaugeopt.py` | `lagoified_gopparams_dicts`, `std_lago_gopsuite`, `add_lago_models` |
| `leakage/reports.py` | `construct_leakage_report` |

`pygsti/__init__.py` now imports `leakage`, making `pygsti.leakage` a first-class
subpackage alongside `pygsti.algorithms`, `pygsti.models`, etc.

`pygsti/tools/__init__.py` replaces `from .leakage import *` with explicit re-exports
from `pygsti.leakage`, so all `pygsti.tools.*` leakage names continue to resolve without
any changes to their callers.

### Basis-aware leakage detection of leakage modeling

The central design change is a new `Basis.implies_leakage_modeling` property. A basis
*implies leakage modeling* when it has a unique element whose label consists entirely of
the character `'I'` (e.g., `'I'`, `'II'`, `'III'`) and that element is proportional to
an orthogonal projector of rank strictly less than the full Hilbert space dimension.

This property drives three new functions in `leakage/core.py`:

- `computational_effect(basis)` — extracts the Hermitian operator that orthogonally
  projects onto the computational subspace.
- `computational_superkets(basis)` — returns a column-orthonormal matrix whose columns
  span the computational subspace of superoperator space.
- `computational_projector(basis)` — the Hermicity-preserving superoperator that
  projects M[H] onto M[C].

The vestigial `n_leak: int` flag — formerly used to signal "activate leakage modeling" —
is retired. Leakage behavior is now inferred from the basis: `_create_objective_fn`,
`_legacy_create_scalar_objective`, and the metric functions (`subspace_jtracedist`,
`diamond_distance_projection_model`, `leaky_maximum_trace_dist`,
`diamonddist_to_leakfree_cptp`) drop `n_leak` from their signatures and instead branch on
`mx_basis.implies_leakage_modeling`, looking up the projector via
`computational_projector(mxBasis)` when needed.

Where an explicit request for subspace-restricted behavior is still meaningful in the
gauge-optimization path, `n_leak` is replaced by a boolean `leakage_modeling` flag rather
than dropped. `gaugeopt_to_target` reads `leakage_modeling` from its keyword arguments —
set by `lagoified_gopparams_dicts` when building a leakage-aware gauge-opt suite — and
raises `ValueError` if it is `True` for a basis that does not imply leakage modeling.

`Basis.is_hermitian()` is a new method used internally by `computational_superkets` to
validate its input.

### Basis-aware report metrics

Reports carry no report-wide leakage display flag. In the gates-vs-target tables (both the
summary section and the gauge-variant gauge section), the choice between subspace-restricted
and standard full-space metric columns is made **per cell from that cell's own model basis**:
a model whose basis implies leakage modeling gets subspace metrics (`sub-inf`, `sub-trace`,
`sub-diamond`, `plf-sub-diamond`, leak/seep rates); every other model gets the standard
full-space columns. Because the columns and the underlying computation are both driven by
the same basis, the displayed header can never disagree with what was computed — a single
report can freely mix leakage and non-leakage models.

Subspace metrics are shown by default wherever a basis permits. Whenever any model in the
report carries a leakage basis, the top switchboard gains an interactive **"Metrics"**
switch (`Subspace` / `Full-space`, defaulting to `Subspace`) that flips all gates-vs-target
tables to standard full-space metrics — useful, e.g., for gates that intentionally drive
population between levels. When no model implies leakage modeling the switch is hidden and
adds no cost. This is implemented by passing the tables' `display` argument as a per-cell
`SwitchValue` over the `(Dataset, Estimate, Gauge-Opt, Metrics)` axes, which the existing
switchboard machinery already renders as one independent table per position.

As defense-in-depth, the leak/seep reductions return `NaN` on an empty transport profile
rather than reducing an empty sequence, and `GatesVsTargetTable` cell evaluation degrades a
failed metric to a `NaN` cell (with a warning) instead of aborting the whole report.

### Basis construction: `augment_for_leakage_modeling`

`augment_for_leakage_modeling(basis, E)` takes any Hermitian basis and a real orthogonal
projector `E`, and returns a new `ExplicitBasis` that implies leakage modeling with
computational subspace `C = range(E)`. Calling this function on the Gell-Mann basis
with ``E = np.diag([1, 1, 0])`` basically gives you the l2p1 built-in basis.

The returned basis has these properties (k = rank(E)):

- Element 0 is proportional to `E`, labeled with all `'I'`s (longest available).
- Elements 0 ... k²−1 span `M[C]`.
- Elements k² ... dim−1 span `M[H] \\ M[C]`. All are labeled `'L[orig_label]'`, except the
  final element, which is proportional to `I − E` and labeled `'L'`.
- Every element has unit Frobenius norm.

Element selection uses `pivot_indices_after_deflation` (pivoted QR after orthogonal
deflation of a fixed vector) to prefer the original basis elements that are "most
supported" on each subspace.

Input validation: `E` must be real (‖imag(E)‖ ≤ 1e-10), Hermitian, and (proportional
to) an orthogonal projector; otherwise `ValueError` is raised.

### Leakage gauge groups and new model functions

Both `leaky_qubit_model_from_pspec` and `promote_bb_to_bt` now reject non-Hermitian
Hilbert-Schmidt bases (via the shared `_assert_hermitian_basis` helper, which checks
`Basis.real`). These models store their members in a *real* parameter vector, which is
only consistent with a Hermitian basis; a non-Hermitian basis such as the matrix-unit
`'std'` basis yields complex gate superoperators whose imaginary parts would be silently
discarded when the parameter vector is rebuilt, quietly corrupting the model. The guard
raises `ValueError` and points to Hermitian alternatives (`'l2p1'`, `'gm'`, `'qt'`).

`leaky_qubit_model_from_pspec` now assigns a `DirectSumUnitaryGroup` as its default
gauge group. This group is a block-diagonal subgroup of U(3), decomposed as U(2) ⨁ U(1):
U(2) acts on the 2-dimensional computational subspace and U(1) acts on the 1-dimensional
leakage level. The two summands are specified to `_direct_sum_unitary_group` by their
bases in *Hilbert-Schmidt* space — `'pp'` of dimension 4 for the U(2) factor, `'pp'` of
dimension 1 for the U(1) factor — rather than by the dimension of the underlying Hilbert
space itself.

Supporting U(1) as a gauge-group factor requires two new classes in `gaugegroup.py`:

- `U1Group` — a one-parameter `GaugeGroup` whose elements are complex phases.
- `U1GroupElement` — implements `transform_matrix`, `transform_matrix_inverse`,
  `from_vector`, `to_vector`, and `inverse`.

`DirectSumUnitaryGroup` and `DirectSumUnitaryGroupElement` are extended to accept
`U1Group` / `U1GroupElement` components alongside the existing `UnitaryGaugeGroup` and
`TrivialGaugeGroup` types.

`promote_bb_to_bt` (qubit ⨂ qutrit) is a new function in `leakage/models.py`. Given a
2-qubit `ExplicitOpModel`, it produces a 6-dimensional model in a tensor-product basis
`pp ⨂ l2p1` where the second qubit factor has an accessible leakage level. The function
handles the lifting of 2-qubit unitaries to 2-qudit unitaries (via the helper
`_lift_unitary_bb_to_bt`) and POVM construction with four effects (`'00'`, `'01'`,
`'10'`, `'11'`). Its default gauge group is a plain `UnitaryGaugeGroup` over the full
6-dimensional space.

The gate lift `_lift_unitary_bb_to_bt` adds the leakage level to the second register,
acting as the input unitary on the computational subspace and as the identity on the
leakage level — i.e. the block-diagonal unitary `u ⊕ I2`, which is unitary for every
unitary input.

`random_unitary_excitation` is a helper in `leakage/models.py`. Given a state space, a
target subsystem, and a `ground_level`, it draws a random rank-one Hermitian generator
supported on levels `ground_level` and `ground_level + 1`, exponentiates it to a unitary,
and embeds that unitary (as an `EmbeddedOp`) into the full state space. It returns the
embedded operation together with the complex unit vector defining the generator, and is
useful for injecting controllable leakage into a model for testing.

### Other changes

- `matrixtools.is_projector` — new utility function (used in `computational_superkets`
  and `pop_transport_profile`).
- `jamiolkowski_iso` — adds a fallback when `basis.create_simple_equivalent()` raises,
  which happens for leakage tensor-product bases.
- `find_closest_unitary_opmx` — gains an `op_basis` parameter (defaults to `'pp'`).
- `EmbeddedOp` — early return in `_iter_matrix_elements` when the embedded op fills
  the full state space, avoiding unnecessary iteration.
- `UnitaryGaugeGroup.__init__` — raises `ValueError` when `state_space.udim == 1`;
  callers should use `U1Group` instead.
- `gaugeopt_to_target` — docstring consolidation: the Notes section that was split
  across two separate `"""..."""` blocks is merged into one.
- Type annotation typos in `gaugegroup.py` corrected (`_np.array` → `_np.ndarray`,
  `_np.ndarry` → `_np.ndarray`).
- `algorithms/core.py` import cleanup: `BasisLike` imported from `baseobjs.basis`
  directly rather than via the star-import chain.

## Tests

A new `test/unit/leakage/` package adds unit tests for the subpackage's public API:

- `test_core.py` — `Basis.implies_leakage_modeling`, `Basis.is_hermitian()`,
  `matrixtools.is_projector`, `computational_effect`, `computational_superkets`, and
  `computational_projector` (both the 1-arg basis-aware form and the 3-arg explicit form,
  plus consistency between them); `augment_for_leakage_modeling` (structural properties,
  spanning of `M[C]` and `M[C]⊥`, label format, non-diagonal projector round-trip
  through `computational_effect` / `computational_superkets`, and input validation).
- `test_metrics.py` — `choi_state` (Hermitian and positive-semidefinite output),
  self-distance = 0 and self-fidelity = 1 for `subspace_jtracedist`,
  `subspace_superop_fro_dist`, and `subspace_entanglement_fidelity`; transport profiles
  (`gate_leakage_profile`, `gate_seepage_profile`, `pop_transport_profile`) verified
  against identity and explicitly-constructed leaky gates; `subspace_diamonddist`
  guarded with `@needs_cvxpy`.
- `test_models.py` — four `TestCase` classes:
  - `LeakyQubitModelTester` — structural checks (state-space dimension,
    `DirectSumUnitaryGroup` gauge group, POVM completeness), exact `rho0` and `E0`,
    gate-action correctness via forward simulation, agreement of outcome probabilities
    across the Hermitian HS bases `l2p1` / `gm` / `qt`, rejection of the non-Hermitian
    `'std'` basis, and `levels_readout_zero` variation.
  - `LiftUnitaryBBtoBTTester` — `_lift_unitary_bb_to_bt` on separable and entangling
    (CNOT, Haar-random) inputs: the computational block equals `u`, the leakage block is
    `I2`, and the coupling blocks vanish; composition (`lift(u1 @ u2) == lift(u1) @
    lift(u2)`) and rejection of non-4×4 input.
  - `PromoteBBToBTTester` — structural checks, exact `rho0` and four-effect projectors,
    forward-simulation correctness (single-qubit marginal, CNOT entanglement), and
    rejection of a non-Hermitian register basis.
  - `RandomUnitaryExcitationTester` — return types, unit-norm generator supported on the
    two coupled levels, zero-strength ↦ identity, unitary-channel superoperator, seed
    determinism, and out-of-range `ground_level` validation, over qutrit, ququart, and
    qubit⨂qutrit state spaces.

`test_gaugeopt_correctness.py` adds `FindPerfectGauge_DirectSumGaugeGroupTester`, which:

1. Builds a leaky-qubit model via `leaky_qubit_model_from_pspec`.
2. Applies a random gauge transformation using a block-diagonal unitary (2×2 block on
   the computational subspace, a U(1) phase on the leakage level).
3. Runs `gaugeopt_to_target` with `method='L-BFGS-B'` over a `DirectSumUnitaryGroup`.
4. Verifies that all gate metrics return to near zero after optimization.

The test runs with both a `U1Group` and a `TrivialGaugeGroup` for the leakage factor,
confirming that both configurations recover the gauge.

`test_gaugegroup.py` gains `U1GroupTester`, covering construction, the identity transform
at angle zero, unitarity for arbitrary angles, `from_vector`/`to_vector` round-trip, 2π
angle wrapping, and the inverse element.

`test/unit/report/test_basis_aware_display.py` covers the report side: per-cell column
selection for a switchboard mixing a leakage (`l2p1`) and a plain (`pp`) model, including
the `Subspace`/`Full-space` positions and `NotApplicable` cells; the leak/seep NaN
hardening; a leakage display forced onto a plain model degrading to NaN cells instead of
raising; and an end-to-end mixed-basis report that builds and exposes the interactive
`Metrics` switch.

## Migration notes

Code that imported directly from `pygsti.tools.leakage` will need to update to
`pygsti.leakage`. Anything that accessed leakage names through `pygsti.tools.*` requires
no change.

The integer `n_leak` flag is gone. Leakage behavior is inferred from the basis, so using a
leakage basis (e.g., `'l2p1'`) lets `Basis.implies_leakage_modeling` drive the behavior
automatically. Two call sites are affected:

- `construct_standard_report` — the report-wide `n_leak` display option is removed.
  Passing in `advanced_options` emits a `DeprecationWarning` and is
  otherwise ignored; leakage metrics are now shown automatically per basis, with the
  interactive `Subspace`/`Full-space` `Metrics` switch to view full-space metrics instead.
- `gaugeopt_to_target` — `leakage_modeling=True` raises `ValueError` when the model's
  basis does not imply leakage modeling. This flag is normally set for you by
  `lagoified_gopparams_dicts`; it was never a documented `gaugeopt_to_target` argument, so
  there is no back-compat shim here.


pygsti/baseobjs/basis.py
```diff
@@ -317,6 +319,23 @@ def elsize(self):
...
+    def implies_leakage_modeling(self) -> bool:
+        if (not hasattr(self, '_implies_leakage')) or self._implies_leakage is None:
+            id_label = 'I'
+            if id_label not in self.labels:
+                id_label = 'II'
+            if id_label not in self.labels:
+                return False
+            I = self.ellookup[id_label]
+            self._implies_leakage = _np.linalg.matrix_rank(I) < I.shape[0]
+        return self._implies_leakage
```
rileyjmurray: We can characterize id_label more reliably (e.g., for 3Q systems) as
> ``argmax([ len(ell) for ell in self.labels if len(ell.strip('I')) == 0])``

^ That's math, not valid python code.

pygsti/optimize/customsolve.py
```diff
@@ -165,7 +165,7 @@ def custom_solve(a, b, x, ari, resource_alloc, proc_threshold=100):
         potential_pivot_indices = all_row_indices[potential_pivot_mask]
         ibest_global, ibest_local, h, k = _find_pivot(a, b, icol, potential_pivot_indices, my_row_slice,
                                                       shared_floats, shared_ints, resource_alloc, comm, host_comm,
-                                                      smbuf1, smbuf2, smbuf3, host_index_buf, host_val_buf)
+                                                      smbuf1, smbuf1b, smbuf2, smbuf3, host_index_buf, host_val_buf)
```
rileyjmurray: I could have sworn we fixed this at least twice. Why is it here ???

rileyjmurray: Oh no. https://github.com/sandialabs/pyGSTi/issues/542#issuecomment-2667030124.

It's languishing on the riley-xfgst branch.

pygsti/protocols/gst.py
```diff
@@ -1534,6 +1534,46 @@ def run(self, data, memlimit=None, comm=None, checkpoint=None, checkpoint_path=N
...
+                f.writelines(script_contents)
+            # step 4: launch script with mpiexec.
+            try:
+                command = f'which python; {mpicommand} -n {num_ranks} python {scriptloc_str}'
+                subprocess.run(command, shell=True)
+            except subprocess.CalledProcessError as e:
+                print(f"{mpicommand} failed with error : {e}!")
+            # step 4: read in the results
+            res = ModelEstimateResults.from_dir(tempdir_str, self.name)
+        return res
```
rileyjmurray: Move this to another PR.

rileyjmurray:
image

^ Need to fix.

pygsti/optimize/customsolve.py
rileyjmurray: Fixed by merging in develop.

rileyjmurray:
Todo: add the following function. Maybe not strictly necessary, but I wrote it in another PR where it's distinctly _less_ necessary.
```python
def trace_effect(op: _np.ndarray, op_basis: BasisLike, on_space: SpaceT = 'HilbertSchmidt'):
    """
    Let `op` be the array representation of a superoperator G in `op_basis`,
    where G maps from and to the space of order-d Hermitian operators.
    
    The trace effect of G is the Heritian operator E that satifies

        trace(G(ρ)) = trace(E ρ) for all order-d Hermitian matrices ρ.

    If on_space='HilbertSchmidt', then this function returns a superket representation
    of E in `op_basis`. If on_space='Hilbert', then we return E itself.
    """
    d = int(_np.round(op.size ** 0.25))
    assert op.shape == (d**2, d**2)
    basis = op_basis if isinstance(op_basis, _Basis) else _Basis.cast(op_basis, dim=d**2)
    vecI = _bt.stdmx_to_vec(_np.eye(d), basis)
    vecE = op.T.conj() @ vecI
    if on_space == 'HilbertSchmidt':
        return vecE
    else:
        E = _bt.vec_to_stdmx(vecE, op_basis)
        return E
```

rileyjmurray:
The test in `test/integration/test_leakage_gst.py` is failing due to numerical issues. The body of the test includes the following info in a long-form comment.
```markdown
        Original results are shown below. We don't rely on the exact numbers here. What matters is
        qualitative aspects of how Gxpi2 and Gypi2 deviate from their respective targets. Since our
        data generating model only applied leakage to Gxpi2, a "good" result reports much more error
        in Gxpi2 than Gypi2. (It's not clear to me why stdgaugeopt lacks wildcard error.)

        Leakage-aware guage optimization.

            | Gate    | ent. infidelity | 1/2 trace dist | 1/2 diamond dist | Max TOP  | Unmodeled error |
            |---------|-----------------|----------------|------------------|----------|-----------------|
            | []      | 0.000009        | 0.002639   	 | 0.003724	        | 0.000014 | 0.00007         |
            | Gxpi2:0 | 0.000965        | 0.030946       | 0.040751	        | 0.001629 | 0.000764        |
            | Gypi2:0 | 0.000375        | 0.019367       | 0.025308         | 0.000531 | 0.000475        |

        
        Standard gauge optimization

            | Gate    | ent. infidelity | 1/2 trace dist | 1/2 diamond dist | Max TOP  |
            |---------|-----------------|----------------|------------------|----------|
            | []      | 0.000016        | 0.003644       | 0.005146         | 0.000026 |
            | Gxpi2:0 | 0.000620        | 0.024758       | 0.032616         | 0.001000 |
            | Gypi2:0 | 0.000611        | 0.024714       | 0.033323         | 0.001001 |

        We'll run tests with subspace entanglement infidelity.

            * For LAGO, infidelity of Gxpi2 is 2.57x larger than that of Gypi2;
              we'll test for a 2x difference.
            
            * For standard gauge optimization, Gxpi2 and Gypi2 have almost identical infidelities;
              we'll test for a factor 1.1x there.
```
The issue is that the factor 1.1x is, apparently, not enough to cover the change after merging in develop, using the latest numpy. The tests pass locally when I run with older numpy. I don't know exactly _what_ changes in numpy / develop would cause the changes here, but that's the nature of integration tests. I'll probably regenerate the table and pick a new tolerance.

Tests pass as of March 24 with the attached conda environment. Unclear how this is different from what's on CI.
[leakage_integration_ok_env.txt](https://github.com/user-attachments/files/26232649/leakage_integration_ok_env.txt)

pygsti/leakage/core.py
```diff
@@ -0,0 +1,230 @@
...
+    assert len(candidates) > 0
+    label = candidates[0]
+    E = basis.ellookup[label].copy()  # type: ignore
+    k = np.linalg.matrix_rank(E)
+    E *= (k/np.trace(E))
+    return E
+
+
+@set_docstring(NOTATION)
+# TODO: document me
```
rileyjmurray, line 91: Must address before merge.

pygsti/baseobjs/basis.py
```diff
@@ -317,6 +319,23 @@ def elsize(self):
         if self.elshape is None: return 0
         return int(_np.prod(self.elshape))
 
+    # TODO: convert labels, ellookup, and elements to properties.
+    #       Type annotations for Basis objects are of limited use
+    #       without the Basis class declaring that these members exist.
```
rileyjmurray, line 324: Should open an issue and remove this comment from the PR changes.

pygsti/leakage/core.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+Basic definitions for leakage modeling
+--------------------------------------
+Let B denote a Hermitian basis for the Hilbert--Schmidt space M[H]. We say that
+B supports leakage modeling if
+
+    (1) it has a unique element whose label consists of (and only of) one
+        or more copies of the character 'I', and
+    (2) that element is proportional to a real orthogonal projector on H.
```
rileyjmurray, line 59: Not quite the case. More accurate to say that there is at least one element whose label is a string consisting of one or more copies of `I` (and no other characters), and that the element with the longest such label is proportional to a real orthogonal projector on H.

pygsti/leakage/core.py
```diff
@@ -0,0 +1,230 @@
...
+    dim = basis.dim
+    if basis.first_element_is_identity:
+        return np.eye(dim)
+    U = computational_superkets(basis)
+    P = U @ U.T
+    return P
+
+
+# TODO: figure out where this helper should go. We use its basic functionality
+# a couple of times in this file and we use it once in basis.py.
```
rileyjmurray, line 130: Resolve before merge.

pygsti/leakage/core.py
```diff
@@ -0,0 +1,230 @@
...
+        and is the longest such label in the basis.
+
+      * The first rank(E)^2 elements span M[C]. Labels of these elements (other
+        than the very first) match those of their corresponding elements in `basis`. 
+
+      * All subsequent elements span M[H] \\ M[C]. Their labels are of the form
+        'L[ell]', where 'ell' is the label of their corresponding element in `basis`.
+
+      * The final element is proportional to the projector onto C^⟂
+        and is labeled 'L'.
```
rileyjmurray, line 160: Potentially dubious change worth discussing: use the letter `'C'` for the first label, then strings of the form `'C[lbl]'` with existing labels lbl for the next k^2-1 labels, then strings of the form `'L[lbl]'` with exsiting labels lbl, and finally just the letter `'L'`.

pygsti/leakage/core.py
```diff
@@ -0,0 +1,230 @@
...
+
+    return new_basis
+
+
+
+if __name__ == "__main__":
+    l2p1 = Basis.cast('l2p1', 9)
+    gm   = Basis.cast('gm', 9)
+    b = augment_for_leakage_modeling(gm, computational_effect(l2p1))
+    print()
```
rileyjmurray, line 230: Remove before merge

pygsti/leakage/reports.py
```diff
@@ -0,0 +1,109 @@
...
+def _add_lago_estimates(mer: ModelEstimateResults, gaugeopt_verbosity: int = 0):
+    # This is broken out into its own function in case someone wants to use
+    # it in a report-generation function other than construct_leakage_report.
+    for ek in mer.estimates:
+        assert isinstance(ek, str)
+        if ek == 'Target':
+            # There's no need to gauge-optimize in this case.
+            continue
+        add_lago_models(mer, ek, verbosity=gaugeopt_verbosity)
+    return
```
rileyjmurray, line 53: This should probably be moved into pygsti.leakage.gaugeopt and made public, rather than kept as a private helper.

pygsti/leakage/reports.py
```diff
@@ -0,0 +1,109 @@
...
+from typing import Any, Optional, Union, TYPE_CHECKING
+
+from pygsti.leakage.gaugeopt import add_lago_models
+
+if TYPE_CHECKING:
+    from pygsti.protocols.gst import ModelEstimateResults
+
+
+def _add_all_hessians(mer: ModelEstimateResults, kwargs_for_projhess=None):
+    # NOTE: this function is not leakage-specific.
```
rileyjmurray, line 23: Maybe add a TODO that we should revisit the question of where this lives the next time we're making changes to pygsti.report.factory.py. Could open a GitHub issue about that and point to this line so we don't lose track of it.

pygsti/tools/optools.py
```diff
@@ -1111,7 +1111,7 @@ def povm_diamonddist(model, target_model, povmlbl, _premultiplier=None):
         return diamonddist(povm_mx, target_povm_mx, target_model.basis)
     except AssertionError as e:
         assert '`dim` must be a perfect square' in str(e)
-        return _np.NaN
+        return _np.nan
```
rileyjmurray, line 1114: TODO: open a GitHub issue about np.NaN vs np.nan.

test/unit/leakage/test_core.py
rileyjmurray, line 1: Need close review

test/unit/leakage/test_gaugeopt.py
rileyjmurray, line 1: Need close review

test/unit/leakage/test_metrics.py
rileyjmurray, line 1: Need close review

pygsti/tools/optools.py
rileyjmurray: No need. Lowercase is the standard in numpy 2.x.

pygsti/leakage/core.py
rileyjmurray: Corrected.

pygsti/leakage/core.py
rileyjmurray: Created helpers `basis.py::_eye_label` and `matrixtools.py::induced_projector`.

pygsti/leakage/reports.py
rileyjmurray: Deciding against that for now. We can make it public if people need it later on.

pygsti/leakage/reports.py
rileyjmurray: Moved logic to a new `ModelEstimateResults.add_hessians` instance method.

pygsti/baseobjs/basis.py
rileyjmurray: Done.

pygsti/algorithms/gaugeopt.py
```diff
@@ -674,6 +670,8 @@ def _objective_fn(gauge_group_el, oob_check):
 
         if "frobenius" in spam_metric:
             # SPAM and gates can have different choices for squared vs non-squared.
+            #
+            # TODO: remove support for this codepath. It's ridiculous.
```
sserita, line 670: Pinging an active TODO. Not something we should fix in this PR, but is there an active issue for it?

pygsti/algorithms/gaugeopt.py
```diff
@@ -568,54 +570,49 @@ def _legacy_create_scalar_objective(model, target_model,
     #  since from_vector will clear any basis info)
     if mxBasis.name == "unknown" and target_model is not None:
         mxBasis = target_model.basis
+    # ^ TODO: look into if this comment about from_vector is still valid.
+    #   my instinct tells me that we should be raising an error if model.basis
+    #   doesn't match target_model.basis.
```
sserita: Pinging TODO. Not blocking, up to you whether this stays as a comment or escalates to a new issue (seems minor). Seems to me like from_vector does not touch basis info, but perhaps I'm missing something.

pygsti/algorithms/gaugeopt.py
```diff
@@ -699,6 +696,9 @@ def _objective_fn(gauge_group_el, oob_check):
                     M_target = target_model.povms[povmlbl]
                     M_curest =        model.povms[povmlbl]
                     Es_target, ts = povm_fidelity_targets[povmlbl]
+                    # TODO: extend optools.povm_fidelity to a leaky version, and an eigenvalue
+                    # version that accounts for leakage. Once we have those in place we can
+                    # avoid iterating over individual POVM effects.
```
sserita, line 697: Pinging TODO. Issue-ify?

pygsti/baseobjs/basis.py
```diff
@@ -317,6 +355,22 @@ def elsize(self):
...
+                self._implies_leakage = False
+                return False
+            I_before = self.ellookup[label]
+            from pygsti.tools.matrixtools import induced_projector
+            I_after  = induced_projector(I_before, require_real=True)
+            if I_after.size == 0:
+                self._implies_leakage = False
+            else:
+                self._implies_leakage = round(_np.trace(I_after).real)**2 < I_after.size
+        return self._implies_leakage  # type: ignore
```
sserita: The only other option is default None - could we just annotate return type as Optional[bool] and remove this type ignore?

(I generally like to minimize `#type: ignore` when fix is not convoluted)

pygsti/leakage/core.py
sserita: This maybe makes things easier to match notation-wise, but loses some easy information about the size of the C subspace (i.e. the number of I's in the first label).
Maybe one potential compromise would be the first label is C["I"*k], making it more consistent with the rest of the C[lbl] while still imparting the computational subspace info I think you were hoping to get across with the new convention.

I don't feel particularly strongly about this though, and will let you decide if this is something you want to do now, or make an Issue and punt down the line.

pygsti/leakage/core.py
sserita, line 1: Awesome job on the docstrings here!

pygsti/leakage/metrics.py
```diff
@@ -0,0 +1,361 @@
...
+expressed in the standard (matrix-unit) basis of H⨂H.
+
+When `op_basis` implies leakage modeling, ρ is the maximally entangled state of the
+computational subspace C (against a reference copy of H), so the result is the Choi matrix
+of X seen only through inputs supported on C. Otherwise ρ is the maximally entangled state
+of all of H and the result is the ordinary Choi state of X.
+""" + NOTATION)
+def choi_state(op_x: np.ndarray, op_basis: BasisLike) -> np.ndarray:
+    udim = int(op_x.size ** 0.25)
+    dim = udim**2
```
sserita: Is this the only use of udim? Why not just do dim = int(op_x.size ** 0.5) or dim=int(np.sqrt(op_x.size))?

pygsti/leakage/metrics.py
```diff
@@ -0,0 +1,361 @@
...
+The pair (op_x, op_y) represent some superoperators (X, Y) in S[H], using op_basis.
+
+Let rho be the rank-1 density in M[H⨂H] induced by op_basis' computational effect,
+and let I denote the element of S[H] that acts as the identity on M[H].
+
+This function returns the %s between X⨂I(rho) and Y⨂I(rho).
+""" + NOTATION
+
+
+@set_docstring(CHOI_INDUCED_METRIC_TEMPLATE % 'entanglement fidelity')
```
sserita, line 145: Nice, this pattern is cool.

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+    if not mx_basis.real:
+        raise ValueError(
+            f"mx_basis {mx_basis.name!r} is not Hermitian: superoperators of physical "
+            f"operations would be complex, but these models use a real parameter vector "
+            f"(the imaginary parts would be silently discarded). Use a Hermitian basis "
+            f"such as 'l2p1', 'gm', or 'qt'."
+        )
+
+
+# TODO: write a version of this that's flexible in what it promotes from and to.
```
sserita, line 53: Promote to Issue?
At this point, maybe there is an "enhance leakage" master issue with some subtasks?

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+        tm_3level.operations[gatekey] = u2x2_to_9x9_superoperator(unitary)
+
+    subgroup_bases = [Basis.cast('pp', 4), Basis.cast('pp', 1)]
+    g_full = _direct_sum_unitary_group(subgroup_bases, mx_basis)
+    tm_3level.default_gauge_group = g_full
+    tm_3level.sim = 'map'  # can use 'matrix', if that's preferred for whatever reason.
+    return tm_3level
+
+
+def _lift_unitary_bb_to_bt(u: np.ndarray) -> np.ndarray:
```
sserita, line 158: In the future, would we want a general utility that can take some number of qubits and lift an arbitrary subset to qutrits? If so, do we toss that on the future Issue?

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+    u_6x6[np.ix_(comp, comp)] = u
+
+    expect_zero = np.eye(6, dtype='complex') - u_6x6 @ u_6x6.T.conj()
+    if (nrm := np.linalg.norm(expect_zero)) > 1e-12:
+        warnings.warn(f'Nominally-unitary operator {u_6x6} fails adjoint-inverse check with norm {nrm}.')
+
+    return u_6x6
+
+
+def promote_bb_to_bt(
```
sserita, line 195: same comment as above

pygsti/leakage/core.py
rileyjmurray: Something that occurred to me:

The "C[]" wrapper might help when understanding the labels in tensor product bases of the form (3-level)-tensor-(2-level).

I'm not going to make the change in this PR. If we find the current string rules are problematic then we'll just revisit.

rileyjmurray:
@coreyostrove this is ready for review

pygsti/models/explicitmodel.py
```diff
@@ -119,7 +119,13 @@ class ExplicitOpModel(_mdl.OpModel):
...
     """
 
-    #Whether access to gates & spam vecs via Model indexing is allowed
+    # Whether access to gates & spam vecs via Model indexing is *forbidden*.
+    # This is consulted per-instance (``self._strict``) in ``__getitem__``/
+    # ``__setitem__``; the value here is only the default for instances that
+    # don't set their own.  Set ``some_model._strict = True`` to enable strict
+    # mode on a single model without mutating shared class state -- important
+    # so that tests toggling strict mode stay isolated under parallel/xdist
+    # execution rather than racing on this class attribute.
```
sserita, line 128: docstring flipped from allowed to forbidden. Is this just a clarification or semantic change?

pygsti/report/factory.py
```diff
@@ -1252,9 +1252,18 @@ def construct_standard_report(results, title="auto",
     confidence_level = _validated_confidence_level(confidence_level)
 
     advanced_options = advanced_options or {}
-    n_leak = advanced_options.get('n_leak', 0)
-    # ^ It would be preferable to store n_leak in a Basis object, or something similar.
+    leakage_modeling = advanced_options.get('leakage_modeling', False)
+    # ^ It would be preferable to store this on a Basis object, or something similar.
```
sserita: Pinging as possible future leakage TODO

pygsti/tools/__init__.py
```diff
@@ -20,7 +20,28 @@
...
+    pop_transport_profile,
+    gate_leakage_profile,
+    gate_seepage_profile,
+    leaky_qubit_model_from_pspec,
+    promote_bb_to_bt,
+    lagoified_gopparams_dicts,
+    std_lago_gopsuite,
+    add_lago_models,
+    construct_leakage_report,
+)
```
sserita: Is there a way we can throw a deprecated warning if these are hit? I imagine we'd eventually want to not re-export here, so deprecate in 0.10, remove in 0.11?

pygsti/tools/jamiolkowski.py
```diff
@@ -149,7 +155,7 @@ def jamiolkowski_iso(operation_mx: Union[_np.ndarray, Expression], op_mx_basis:
     # This construction results in a Jmx with trace == dim(H) = sqrt(operation_mx.shape[0])
     #  (dimension of density matrix) but we'd like a Jmx with trace == 1, so normalize:
     if normalized:
-        choiMx /= dmDim
+        choiMx = choiMx / dmDim
```
sserita, line 162: For my own curiosity: Why is this preferable?

test/integration/test_leakage_gst.py
```diff
@@ -116,10 +120,10 @@ def test_pipeline_1Q_XYI(self):
 
         infids = dict()
         for lbl in ['LAGO', 'stdgaugeopt']:
-            infids[lbl] = {g: 1 - fidelity(gates[lbl][g], gates['target'][g], 'l2p1', n_leak=1) for g in ['x', 'y'] } 
+            infids[lbl] = {g: 1 - fidelity(gates[lbl][g], gates['target'][g], 'l2p1') for g in ['x', 'y'] } 
 
-        self.assertGreater( infids['LAGO']['x'],        2.0 * infids['LAGO']['y']        )
-        self.assertLess(    infids['stdgaugeopt']['x'], 1.1 * infids['stdgaugeopt']['y'] )
+        self.assertGreater( infids['LAGO']['x'],        2.0  * infids['LAGO']['y']        )
+        self.assertLess(    infids['stdgaugeopt']['x'], 1.25 * infids['stdgaugeopt']['y'] )
```
sserita, line 126: Why the tolerance bump? is the new behavior less tight?

test/unit/leakage/test_core.py
```diff
@@ -0,0 +1,240 @@
...
+    def test_oc_elements_span_mc_perp(self):
+        # The complement projector I - E lives entirely in M[C]⊥.
+        k = self.k
+        W = np.column_stack([e.ravel() for e in self.b_aug.elements[k**2:]])
+        E_comp = np.eye(self.E.shape[0]) - self.E
+        a_vec  = E_comp.ravel()
+        proj   = W @ np.linalg.lstsq(W, a_vec, rcond=None)[0]
+        self.assertArraysAlmostEqual(proj, a_vec)
+
+    def test_leakage_labels_format(self):
```
sserita, line 256: If we do end up going with C[lbl] format comment, this test can be enhanced.

test/unit/leakage/test_gaugeopt.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_default_flags_uses_trivial_for_dim1(self):
+        # triviality_flags defaults to [dim==1 for each basis]:
+        # pp(4): dim=4 → False → UnitaryGaugeGroup
+        # pp(1): dim=1 → True  → TrivialGaugeGroup (0 params)
+        gg = _direct_sum_unitary_group(self.subspace_bases, self.full_basis)
+        self.assertIsInstance(gg, DirectSumUnitaryGroup)
+
+    # Lines 29 (assert) and 38 (U1Group branch):
+    # Passing triviality_flags explicitly hits the assert on line 29.
+    # Setting tf=False for the dim-1 basis triggers line 38 (U1Group()).
```
sserita: Using line numbers seems fragile

test/unit/leakage/test_metrics.py
```diff
@@ -0,0 +1,228 @@
...
+        n = E.shape[0]
+        E_perp_vec = pgbt.stdmx_to_vec(np.eye(n) - E, self.basis)
+        transport_mat = vec_to_stdmx(self.op_leaky.T @ E_perp_vec, self.basis, keep_complex=True)
+        transport_mat = E @ transport_mat @ E
+        for rate, state in zip(rates, states):
+            self.assertArraysAlmostEqual(transport_mat @ state, rate * state)
+
+
+class NonLeakageBasisTester(BaseCase):
+    """
```
sserita, line 176: Again, appreciate the detailed line numbers. Still seems fragile.

test/unit/tools/test_jamiolkowski.py
```diff
@@ -12,8 +11,10 @@
 
 class JamiolkowskiBasisTester(BaseCase):
     def setUp(self):
-        #Set Model objects to "strict" mode for testing
-        ExplicitOpModel._strict = True
+        # (This test never indexes an ExplicitOpModel, so it does not need
+        # strict mode.  It used to set ``ExplicitOpModel._strict = True`` at the
+        # class level with no tearDown, which leaked into and broke unrelated
+        # tests -- non-deterministically under parallel/xdist execution.)
```
sserita, line 17: yeesh that sounds super annoying...

sserita (changes requested):
A few minor changes requested, many "let's not forget about this" pings, and a couple questions.

Overall, amazing job. Docstrings and tests are awesome on top of everything else! 👍

pygsti/baseobjs/basis.py
rileyjmurray: My first instinct is that the stronger return value (bool) would help with static analysis. Then I remembered that None is False-y, so `if None` and `if (None or True)` work as though None were False.

Changing now!

pygsti/leakage/metrics.py
rileyjmurray: Looks like it. I've changed to `dim = round(op_x.size ** 0.5)`. The round is important so that something like `2.999999999999` gets rounded to 3, rather than floored to 2.

pygsti/leakage/metrics.py
rileyjmurray: Looks like I've got someone on my side of the debate, @coreyostrove! ;)

(Our truce remains in effect, Corey. I'm not advocating for universal adoption.)

pygsti/leakage/models.py
rileyjmurray: I don't think this is promote-to-issue worthy. The comment is getting at how any particular set of conventions will end up clashing with a collaborator's conventions, and probably require changes to this function. It's hard to anticipate what the clashing conventions will end up being. I don't want to come out of the gate with an over complicated interface. Maybe once people try working with these features for a bit we can create such an issue.

I'd be open to creating a GitHub Discussion thread about it. The downside is that many of our collaborators prefer to communicate via email rather than out in the open.

pygsti/leakage/models.py
rileyjmurray: I think we cross that bridge when someone requests it. I don't anticipate that being a request in the near future, since modeling a six-level system with pyGSTi is already quite expensive.

rileyjmurray: I could modify things so the subsystems could be EITHER 2x3 or 3x2, though, if you think I'll regret _not_ doing that out of the gate.

pygsti/models/explicitmodel.py
rileyjmurray: This is clarification. When strict=True, these functions *forbid* certain behavior. If the original comment was to be correct then we should have used a variable name like `permissive` rather than `strict`.

pygsti/report/factory.py
rileyjmurray: Ah, glad you caught this. The comment is stale. The changes on this branch specifically have leakage modeling info stored in a basis. This is complicated though because the `construct_standard_report` has no canonical `Basis` object to inspect. I'll explore what happens in the many scenarios.

pygsti/tools/__init__.py
rileyjmurray: Good point. Will investigate!

pygsti/tools/jamiolkowski.py
rileyjmurray: The choiMx can be a CVXPY Expression or a numpy ndarray. The former will error on in-place modifications.

test/integration/test_leakage_gst.py
rileyjmurray: For some reason the behavior of stdgaugeopt changed slightly. At the end of the day this is a somewhat fragile test. For now I decided it was better to bump the tolerance for now than investigate in detail. If this test fails again then I'll think about how to qualitatively revise for robustness.

pygsti/leakage/models.py
sserita: I'm fine with things falling below the "promote to issue" threshold, just wanted to bring it up before we forget about it forever.

test/unit/leakage/test_gaugeopt.py
rileyjmurray: Good point. Will remove.

pygsti/leakage/models.py
sserita: No, I don't think so. This is just my ["pass arbitrary condiments"](https://xkcd.com/974/) brain kicking in.

test/unit/leakage/test_metrics.py
rileyjmurray: Definitely too fragile. Thanks for flagging.

test/unit/tools/test_jamiolkowski.py
rileyjmurray: It wasn't as painful a debug as you'd expect. I had run into this error before. In those contexts I just replaced the offending indexing with indexing into an appropriate member dict. Here I decided it was better to resolve the issue once and for all.

pygsti/report/factory.py
rileyjmurray: Update: I decided to do away with the `leakage_modeling` advanced-option for report generation. The impact on report generation is what metrics are displayed for gates. I've changed things so that if leakage modeling is possible then we add a switch (default to active) showing leakage-aware metrics for that basis. The current basis is defined by the basis used for the selected gauge optimized model.

pygsti/tools/__init__.py
rileyjmurray: Applied a shim to raise deprecation warnings.

test/unit/leakage/test_core.py
rileyjmurray: I did end up switching to the `C[lbl]` format for the helper function augment_for_leakage_modeling. I consider that part of the function's API.

I also switched the `l2p1` built-in basis to use the `C[lbl]` format, although I don't consider that part of the public API. I added a TODO at the top of this PR about revisiting the `l2p1` basis element labels.

I'll defer enhancements to these tests to later. For now I'm freezing work on this until Corey can review.

pygsti/algorithms/gaugeopt.py
coreyostrove: I am fairly certain that from_vector doesn't interact with basis information at any point. Perhaps that wasn't always true (too lazy to look at the git blame to date this comment). I think the comment this TODO refers to can be deleted safely.

pygsti/baseobjs/basis.py
```diff
@@ -56,6 +58,48 @@ def _sparse_equal(a, b, atol=1e-8):
...
+    Pick a default basis spec for a system whose per-qudit Hilbert-space dimensions
+    are given by `udims`.
+
+    Returns a basis-name string ('pp', 'gm') when every qudit has the same dimension,
+    because a name is dimension-agnostic and resolves correctly at every downstream
+    use site (e.g. when a single-qudit sub-op needs the per-qudit basis rather than
+    the full product basis). Only genuinely mixed-dimension systems return an
+    explicit `TensorProdBasis`.
+    """
+    udim_to_name = {2: 'pp', 3: 'gm'}
```
coreyostrove: 'gm' works for d>3, why not use this as the default?

pygsti/baseobjs/basis.py
```diff
@@ -317,6 +369,22 @@ def elsize(self):
         if self.elshape is None: return 0
         return int(_np.prod(self.elshape))
 
+    @property
+    def implies_leakage_modeling(self) -> Optional[bool]:
```
coreyostrove, line 374: Please add a note for our future selves on what this is checking for. I had a hard time following this.

pygsti/leakage/core.py
```diff
@@ -0,0 +1,256 @@
...
+is denoted S[U].
+
+The orthogonal complement of U in H is denoted U^⟂. There are multiple ways
+to extend a linear operator Φ ∈ M[U] to all of H = U ⨁ U^⟂.
+
+    We call an extension ...
+
+      * direct if U and U^⟂ are invariant subspaces for both Φ and Φ^†,
+      * annhilating if ker(Φ) and ker(Φ^†) contain U^⟂, and
+      * unitary if it is direct and A is unitary on U^⟂.
```
coreyostrove: "A" wasn't defined

pygsti/leakage/core.py
```diff
@@ -0,0 +1,256 @@
...
+
+P is assembled from the computational superkets U as P = U Uᵀ. Applied to a superket, P
+annihilates every component outside M[C] while leaving components within M[C] untouched.
+
+If `basis` does not imply leakage modeling then M[C] = M[H] and P is the identity of
+S[H] (an order-basis.dim identity matrix).
+""" + NOTATION)
+def computational_projector(basis: Basis) -> np.ndarray:
+    dim = basis.dim
+    if basis.first_element_is_identity:
```
coreyostrove: Wouldn't it be cleaner to check the implies_leakage_modeling property directly?

pygsti/leakage/core.py
```diff
@@ -0,0 +1,256 @@
...
+    E *= (k/np.trace(E))
+    if not pgmt.is_projector(E):
+        raise ValueError("E must be (proportional to) a projector")
+    num_I = max(_eye_label(b_in).count('I'), 1)
+    I_lbl = 'C[' + 'I' * num_I + ']'
+    L_lbl = 'L[' + 'I' * num_I + ']' 
+
+    # Step 1: build computational subspace (cs) basis elements and labels.
+    cs_elements = [ E @ B @ E            for B in b_in.elements ]      # type: ignore
+    cs_elements = [ (B + B.T.conj()) / 2 for B in cs_elements   ]
```
coreyostrove, line 245: Am I correct that this implies the resulting augmented basis will be hermitian by construction (even if the input basis is not)? If so, could you add a note to the documentation indicating this.

pygsti/leakage/core.py
```diff
@@ -0,0 +1,256 @@
...
+    k  = np.linalg.matrix_rank(E)
+    E *= (k/np.trace(E))
+    if not pgmt.is_projector(E):
+        raise ValueError("E must be (proportional to) a projector")
+    num_I = max(_eye_label(b_in).count('I'), 1)
+    I_lbl = 'C[' + 'I' * num_I + ']'
+    L_lbl = 'L[' + 'I' * num_I + ']' 
+
+    # Step 1: build computational subspace (cs) basis elements and labels.
+    cs_elements = [ E @ B @ E            for B in b_in.elements ]      # type: ignore
```
coreyostrove, line 244: You can reuse this intermediate result in computing oc_elements below to reduce the number of matrix multiplications you need.

pygsti/leakage/gaugeopt.py
```diff
@@ -0,0 +1,202 @@
...
+            # parameterization-preserving gauge optimization fails if the seed model
+            # uses members other than ComposedState, ComposedOp, ComposedPOVM, etc..
+            # So we add a non-preserving gauge optimization here.
+            tm = results.estimates[est_key].models['target']
+            _add_gauge_opt(results, est_key, gos, tm, tuple())
+    elif est_key is None:
+        for est_key in results.estimates.keys():
+            add_lago_models(results, est_key, gos, verbosity)
+    else:
+        raise ValueError()
```
coreyostrove: Add descriptive message.

pygsti/leakage/gaugeopt.py
```diff
@@ -0,0 +1,202 @@
...
+        # ^ The precise weights here are somewhat arbitrary. We place weight on
+        #   SPAM because SPAM plays a distinguished role in defining the 
+        #   computational subspace.
+    
+    # Below, we add a final gauge optimization step that preserves separation between 
+    # what we've identified as the computational subspace and leakage space. This
+    # step uses squared Frobenius norm as an objective, since that's better-behaved
+    # than plain Frobenius norm.
+    inner_dict = inner_dict.copy()
+    gg = _direct_sum_unitary_group([Basis.cast('pp', 4), Basis.cast('pp', 1)], tm.basis)
```
coreyostrove: This hardcodes for a qubit, but I don't see any reason why that is necessary. Given the code I've read thus far I would think everything would be happy with a higher-dimension computational space, including _direct_sum_unitary_group.

pygsti/leakage/metrics.py
```diff
@@ -0,0 +1,360 @@
...
+Return the rank-1 density in M[H⨂H] induced by op_basis' computational effect.
+""" + NOTATION)
+def tensorized_teststate_density(op_basis: Basis) -> np.ndarray:
+    if not op_basis.implies_leakage_modeling:
+        udim = int(np.sqrt(op_basis.dim))
+        E = np.eye(udim)
+    else:
+        E = computational_effect(op_basis)
+        if la.norm(E.imag) > 0:
+            raise ValueError()
```
coreyostrove: Add descriptive error message.

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+    The model is built as follows:
+
+        The Hermitian matrix representation of `rho0` is the 6-by-6 matrix with a 1 in the
+        upper-left corner and zeros elsewhere.
+
+        `Mdefault` has four effects, labeled "00", "01", "10", and "11". Effect "ab" is
+        ``kron(Ea_b, Eb_t)``, where `Ea_b` is the qubit projector onto computational level
+        `a` and `Eb_t` is the qutrit readout projector for outcome `b`. The qutrit's "0"
+        outcome projects onto the levels in `levels_readout_zero`; its "1" outcome projects
+        onto the complementary levels (including the leakage level).
```
coreyostrove: parenthetical confusing, change to "including the leakage level if not in `levels_readout_zero`".

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+    Promote a two-qubit model to a six-dimensional "qubit ⨂ qutrit" model in which the
+    second register carries an accessible leakage level.
+
+    The input describes two qubits ("bit ⨂ bit", hence *bb*). The output describes a
+    qubit tensored with a qutrit ("bit ⨂ trit", hence *bt*): the first register (`sys0`)
+    stays a 2-level system, while the second register (`sys1`) is extended to a 3-level
+    system whose third level models leakage. The returned model is represented in the
+    tensor-product basis `sys0_basis ⨂ sys1_basis` (by default `pp ⨂ l2p1`).
+
+    The model is built as follows:
```
coreyostrove, line 211: General comment on model construction. I think the newly constructed model should inherit the parameterization types of the lifted operations from the input model, where possible. Currently this always returns a fully unconstrained process-matrix model.

pygsti/leakage/models.py
coreyostrove, line 195: Should raise an exception if the model contains any instruments or factories, since these aren't currently being lifted. (In a perfect world we'd implement that, but this can always wait until there is a concrete need.)

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+        raise ValueError(
+            f"mx_basis {mx_basis.name!r} is not Hermitian: superoperators of physical "
+            f"operations would be complex, but these models use a real parameter vector "
+            f"(the imaginary parts would be silently discarded). Use a Hermitian basis "
+            f"such as 'l2p1', 'gm', or 'qt'."
+        )
+
+
+# TODO: write a version of this that's flexible in what it promotes from and to.
+def leaky_qubit_model_from_pspec(
```
coreyostrove, line 54: Should raise an exception if the processor spec contains any instruments or factories (I think these are called something like unitary functions in the pspec context), since these aren't currently being lifted. (In a perfect world we'd implement that, but this can always wait until there is a concrete need.)

pygsti/leakage/models.py
```diff
@@ -0,0 +1,376 @@
...
+
+    target_subsys : hashable
+        The label (within `state_space`) of the subsystem whose levels are coupled.
+
+    subsys_basis : str or Basis
+        Hilbert-Schmidt basis for the target subsystem, used to represent the excitation
+        as a superoperator. Its dimension must be the square of the subsystem's Hilbert
+        space dimension (e.g. `'gm'` or `'l2p1'` of dimension 9 for a qutrit).
+
+    ground_level : int
```
coreyostrove: Would recommend changing the name of this. Any physicist reading will instantly substitute in the zero state in their head.

pygsti/models/modelconstruction.py
```diff
@@ -1471,6 +1474,13 @@ def _setup_local_gates(processor_spec, evotype, modelnoise=None, custom_gates=No
     gatedict : dict
         A dictionary mapping gate names to local gate operations.
     """
+    if isinstance(basis, _TensorProdBasis):
+        b0 : _Basis = basis.component_bases[0]
+        if all(b0.name == bi.name for bi in basis.component_bases[1:]):
+            basis = b0.name
+        else:
+            raise NotImplementedError()
```
coreyostrove: Add descriptive error message.

pygsti/protocols/gst.py
```diff
@@ -2473,43 +2476,44 @@ def _memberdicts(_m) -> tuple[dict,dict,dict,dict]:
...
-            P = P.real
+        if not leakage_modeling:
+            P = _tools.matrixtools.IdentityOperator()
+            def diamonddist_fn(*args, **kwargs):
+                return 0.5 * _tools.diamonddist(*args, **kwargs)  # type: ignore
+            jtracedist_fn = _tools.jtracedist
+        else:
+            from pygsti.leakage.core import computational_projector
+            from pygsti.leakage.metrics import subspace_diamonddist, subspace_jtracedist
+            P = computational_projector(basis)
```
coreyostrove, line 2492: P doesn't look to be used, and I think the projection is handled implicitly as part of the application to the restricted tensorized test state.

pygsti/protocols/gst.py
```diff
@@ -3292,6 +3296,39 @@ def add_model_test(self, target_model, themodel,
         test_result = mdltest.run(self.data, simulator=simulator)
         self.add_estimates(test_result, silent_steal=True)
 
+    def add_hessians(
```
coreyostrove, line 3299: missing docstring

pygsti/report/section/__init__.py
coreyostrove, line 1: I think this would be better living in report.factory. Not a perfect fit, but it is where most of the report switchboard processing is happening.

pygsti/tools/jamiolkowski.py
```diff
@@ -128,7 +128,13 @@ def jamiolkowski_iso(operation_mx: Union[_np.ndarray, Expression], op_mx_basis:
     if not isinstance(choi_mx_basis, _Basis):
         choi_mx_basis = _Basis.cast(choi_mx_basis, N)  # we'd like a basis of dimension N
 
-    BVec = choi_mx_basis.create_simple_equivalent().elements
+    try:
+        temp = choi_mx_basis.create_simple_equivalent()
+        BVec : Union[_np.ndarray, list] = temp.elements  # type: ignore
+    except:
```
coreyostrove: Please catch a specific set of exceptions, or at least Exception. Bare except also catches keyboard interrupts, so it is dangerous pattern.

pygsti/tools/matrixtools.py
```diff
@@ -101,6 +101,50 @@ def assert_hermitian(mat : _np.ndarray, tol: Union[float,_np.floating]) -> None:
         raise ValueError(message)
 
 
+def is_projector(mx, tol=1e-12):
```
coreyostrove: docstring and type annotations

pygsti/tools/matrixtools.py
```diff
@@ -101,6 +101,50 @@ def assert_hermitian(mat : _np.ndarray, tol: Union[float,_np.floating]) -> None:
...
+    if tol == _np.inf:
+        return True
+    if not is_hermitian(mx, tol):
+        return False
+    mx2 = mx @ mx
+    tol = _np.sqrt(mx.size) * tol
+    return _np.allclose(mx, mx2, atol=tol, rtol=tol)
+
+
+def assert_projector(mx, tol=1e-12):
```
coreyostrove: docstring and type annotations

pygsti/tools/matrixtools.py
```diff
@@ -101,6 +101,50 @@ def assert_hermitian(mat : _np.ndarray, tol: Union[float,_np.floating]) -> None:
...
+    if not is_projector(mx, tol):
+        message = f"""
+            Input matrix 'mx' is not a projector, up to tolerance {tol}.
+            The absolute values of entries in (mx^2 - mx) are \n{_np.abs(mx @ mx - mx)}. 
+        """
+        raise ValueError(message)
+
+
+def induced_projector(mx: _np.ndarray, tol=1e-12, *, require_real=False) -> _np.ndarray:
```
coreyostrove: needs docstring, looks like a projector onto the range of mx?

pygsti/tools/matrixtools.py
```diff
@@ -101,6 +101,50 @@ def assert_hermitian(mat : _np.ndarray, tol: Union[float,_np.floating]) -> None:
...
+
+    if require_real and not _np.allclose(mx, mx.conj(), atol=tol, rtol=tol):
+        return _np.empty((0,))
```
coreyostrove: I would have expected this to raise an exception rather than returning a size zero array.

pygsti/tools/matrixtools.py
```diff
@@ -101,6 +101,50 @@ def assert_hermitian(mat : _np.ndarray, tol: Union[float,_np.floating]) -> None:
...
+
+    if not is_hermitian(mx, tol):
```
coreyostrove, line 198: Same comment as above.

pygsti/report/factory.py
coreyostrove, line 1226: Any chance you'd be willing to add the automatic hessian computation to construct_standard_report, as you added for the new leakage report constructor? With the updated results object returned as a second output in that case. Your idea to do so for the leakage report code is a really good one. This would save a lot of users from the hassle of needing to look up the syntax for adding confidence_region_factories to their estimates.

test/unit/protocols/test_gst.py
```diff
@@ -441,7 +441,28 @@ def test_write_and_read_to_dir(self):
...
+    Regression test: StandardGST must be able to infer a target model from
+    the 2-qubit ProcessorSpec attached to the experiment design when the
+    caller does not pass `target_model` explicitly.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.gst_design = smq2Q_XYICNOT.create_gst_experiment_design(max_max_length=1)
+        mdl_datagen = smq2Q_XYICNOT.target_model().depolarize(op_noise=0.01, spam_noise=0.005)
+        ds = simulate_data(mdl_datagen, cls.gst_design.all_circuits_needing_data,
```
coreyostrove: We don't actually need to simulate anything here. We should be able to just pass in a DataSet object with counts for each circuit initialized to some arbitrary values using the add_count_list method.

coreyostrove (changes requested):
Thanks for the hard work, Riley. I have left a number of requests/comments and suggestions below. Let me know if any of them aren't clear.

To pull out a more substantive concern, there are a few places in the code where you've hardcoded in single-qubit or qutrit behavior where it didn't seem necessary given the generality of the backend implementations. Examples include in  `baseobjs.basis`, `leakage.gaugeopt`.

The majority of these should hopefully be quick and easy to knock out. I will second Stefan's request to issueify TODOs where appropriate for tracking purposes. Now that we can point robots to a github issue and say 'Go!' (at least for simplish things) there is more incentive than before to do so to keep things surfaced.

pygsti/algorithms/gaugeopt.py
sserita: I can help with that one since I was looking into it. :P It's 8 months ago Riley, so maybe out of mental cache. Commit d2eafd5, a refactor of gaugeopt_to_target. Regardless, I agree with Corey.

test/integration/test_leakage_gst.py
sserita: Sounds good, maybe let's just note it down in an Issue somewhere. I feel like a change in stdgaugeopt behavior is something we should at least understand and know when/why it happened. Agree it's not a now thing though.

pygsti/baseobjs/basis.py
rileyjmurray: Done!

pygsti/baseobjs/basis.py
rileyjmurray: Done!

pygsti/leakage/core.py
rileyjmurray: Agreed; fixed

pygsti/leakage/core.py
rileyjmurray: fixed

pygsti/leakage/core.py
rileyjmurray: Yes, it'll be Hermitian. I've updated to set the `real=True` flag in the constructor.

pygsti/leakage/core.py
rileyjmurray: I (strongly) prefer to pay the price of the matmuls in order to have better readability here.

pygsti/leakage/gaugeopt.py
rileyjmurray: Message added

pygsti/leakage/gaugeopt.py
rileyjmurray: Good catch! Fixed.

pygsti/leakage/metrics.py
rileyjmurray: Updated to include message

pygsti/leakage/models.py
rileyjmurray: I feel that's out of scope here. I've updated to raise a warning if the input model isn't static, and I've filed an issue to follow up.

pygsti/leakage/models.py
rileyjmurray: Done

pygsti/leakage/models.py
rileyjmurray: Prefer to wait for a concrete need; issue filed.

pygsti/leakage/models.py
rileyjmurray: Fair point; should be fixed

pygsti/models/modelconstruction.py
rileyjmurray: Should be fixed

pygsti/protocols/gst.py
rileyjmurray: It is used, just about 25 lines down.

pygsti/protocols/gst.py
rileyjmurray: fixed

pygsti/tools/jamiolkowski.py
rileyjmurray: Done!

pygsti/tools/matrixtools.py
rileyjmurray: should be fixed

pygsti/tools/matrixtools.py
rileyjmurray: should be fixed

pygsti/tools/matrixtools.py
rileyjmurray: should be fixed. Yes, a projector onto the range of mx. But we only allow matrices that are already _proportional to_ a projector.

pygsti/tools/matrixtools.py
rileyjmurray: Fair enough; switched

pygsti/report/factory.py
coreyostrove: Documented in issue.

coreyostrove (approved):
Re-reviewed diffing changes since my first review. It looks all of my feedback has been addressed and then some. Tests are also passing, so as far as I am concerned this is good to go. Thanks for the impressive work as always, @rileyjmurray!

coreyostrove:
Interpreting the wave of emojis to my last comment as permission to merge.

PR #710, opened 2026-01-29
Updates wildcard budget invocation from GST to work with gaugeopt_suite=None.

enielse:
Some recent updates cause `protocols.gst._compute_wildcard_budget_1d_model` to now return a dictionary of budget objects rather than a single one, and this introduced two issues:

1. `_compute_wildcard_budget_1d_model` takes an *optional* `gaugeopt_suite`
 argument (that should be allowed to be `None`) but internally it assumes
 a valid budget object.  This PR adds a line creating an empty default
 `GSTGaugeOptSuite` object when `None` is passed in.

2. the processing code "bad fit" processing code in
 `protocols.gst._add_badfit_estimates` was only updated to account
 for _compute_wildcard_budget_1d_model returning a dict when
 `gaugeopt_suite is not None` was True.  This PR removes this
 conditional branch so that returned value is always processed as
 a dictionary of budgets, as it should be.

Also, this PR adds an unrelated update to wildcardbudget.py's `precompute_for_same_probs_freqs` function to raise a `ValueError` when the probabilities given are out of range (outside [0.0,1.0]). This results in a more informative error message to users, as out-of-range probabilities will cause downstream assertion failures that are difficult to make sense of.

### DIFF
--- a/pygsti/objectivefns/wildcardbudget.py
+++ b/pygsti/objectivefns/wildcardbudget.py
@@ -234,8 +234,10 @@ class WildcardBudget(_NicelySerializable):
             elInds = layout.indices_for_index(i)
             fvec = freqs[elInds]
             qvec = probs_in[elInds]
-            if any(el > 1.0 or el < 0.0 for el in qvec):
+            eps = 1e-8 # tolerance for checking valid probabilities, should be ~ machine precision
+            if any(el > 1.0 + eps or el < 0.0 - eps for el in qvec):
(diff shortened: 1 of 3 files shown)


rileyjmurray (changes requested):
Can I get a type annotation from ya on ``_compute_wildcard_budget_1d_model``?

enielse:
Added type annotations and a couple other minor related fixes :)

rileyjmurray (approved):
Looks great!

PR #724, opened 2026-03-17
Helpers for MPI-accelerated GST and SLURM runs

rileyjmurray:
This PR adds two new methods to `Protocol` that together make MPI-accelerated
protocols accessible without requiring users to write their own MPI launcher
scripts or SLURM batch files:

- **`Protocol.run_mpi()`** — launches MPI workers as a subprocess and returns
  results directly.  Can be called from a Jupyter notebook, an interactive
  session, or a plain script.
- **`Protocol.stage_slurm()`** — writes all working files to a permanent
  directory and auto-generates a ready-to-submit SLURM `.sbatch` script.
  The companion **`SlurmSettings`** dataclass carries the SLURM options
  (partition, time limit, node count, etc.).

### Key changes

**`run_mpi(data, num_ranks, *, …)`**
- Accepts `num_ranks`, an optional `persistent_dir`, forwarded `**run_kwargs`,
  and keyword arguments for launcher selection, thread counts, and dry-run
  behavior.
- When `num_ranks=1`, short-circuits to `self.run()` with no subprocess overhead.
- Auto-detects the MPI launcher (`mpiexec`, `mpirun`, or `mpiexec.hydra`) when
  `mpiexec='auto'` (raises `FileNotFoundError` with a clear message otherwise).
- Auto-computes BLAS thread counts per rank (`OMP_NUM_THREADS`, `MKL_NUM_THREADS`,
  etc.) from the physical core count, with a `blas_threads_per_rank` override.
- `persistent_dir` (required when `dry_run=True`): writes data and runner
  artifacts to a permanent directory.  Without it a temporary directory is used
  and cleaned up on return.
- `dry_run=True`: writes artifacts without executing.  Useful for embedding into
   PBS, LSF, or other scheduler scripts. Prints the launch command to console
   if self.verbosity > 0 (or is unset).

**`stage_slurm(data, num_ranks, slurm, work_dir, *, …)`**
- The SLURM-specific staging path: writes data, runner script, and a
  ready-to-submit `.sbatch` file.  Does **not** launch any processes.
- `SlurmSettings` dataclass holds `script_path`, `partition`, `time`, `nodes`,
  `job_name`, `output`, and `error`.  `None` fields emit `FILL_IN` placeholders.
- Generated script uses `srun` (SLURM-native) and includes a commented block of
  optional directives (`--account`, `--qos`, `--constraint`, etc.).
- `--cpus-per-task` and BLAS environment variables are derived from
  `blas_threads_per_rank`.

**Helpers** (in `pygsti/tools/mpitools.py`):
`resolve_mpiexec`, `compute_blas_threads`, `write_mpi_runner_artifacts`, and
`build_slurm_script`.

### Tests

- `test/unit/protocols/test_run_mpi.py` — 43 unit tests covering all helpers,
  error paths, `dry_run=True`, and `stage_slurm`; no MPI launcher required.
- `test/integration/test_gst_run_mpi.py` — integration tests for live MPI
  execution including a `persistent_dir` run and `stage_slurm` script
  validation; demonstrated in
  `jupyter_notebooks/Examples/MPI-RunningGST-Automagic.ipynb`.

### Incidental changes

- `pygsti/tools/optools.py`: extract real part of trace for Hermitian matrices
  (makes warning messages clearer).
- `numpy.core` namespace access removed (deprecated in NumPy 2.x).
- Removed a spurious warning about non-MPI-accelerated gauge optimization.


pygsti/protocols/protocol.py
```diff
@@ -124,6 +124,218 @@ def run(self, data, memlimit=None, comm=None):
...
+                    break
+            if _found is None:
+                raise FileNotFoundError(
+                    "run_mpi: could not find an MPI launcher on PATH. "
+                    "Tried: mpiexec, mpirun, mpiexec.hydra. "
+                    "Install an MPI distribution or pass mpiexec=<path> explicitly."
+                )
+
+        if num_ranks == 1:
+            return self.run(data, **run_kwargs)
```
rileyjmurray: Should move this up

pygsti/protocols/protocol.py
```diff
@@ -124,6 +124,218 @@ def run(self, data, memlimit=None, comm=None):
...
 
+    def run_mpi(self, data, num_ranks, *, mpiexec='auto', extra_mpi_args=None,
+                ranks_per_host=None, env=None, work_dir=None, dry_run=False,
+                blas_threads_per_rank=0, **run_kwargs):
+        """
+        Run this protocol in parallel using MPI workers launched as a subprocess.
+
+        Unlike :meth:`run`, this method can be called from anywhere (Jupyter notebooks,
+        interactive sessions, scripts) without requiring the caller to manage MPI
+        communicators or write launcher scripts manually.
```
rileyjmurray: NOTE: document that the MPI command will be launched in the same terminal environment as the Jupyter/iPython kernel.

rileyjmurray: That is, the subprocess inherits the environment that it comes from.

pygsti/protocols/protocol.py
```diff
@@ -124,6 +124,218 @@ def run(self, data, memlimit=None, comm=None):
...
+        These variables are applied before ``env``, so explicit entries in
+        ``env`` override them.
+
+        **Job schedulers.**
+        On HPC systems where jobs must be submitted via SLURM, PBS, or similar,
+        use ``dry_run=True`` together with ``work_dir`` to write the data,
+        protocol, and runner script to a permanent location and print the launch
+        command without executing it.  The printed command can then be embedded
+        in a batch script.
+        """
```
rileyjmurray, line 319: TODO: maybe have an option that generates the SLURM script as well

rileyjmurray: Done

rileyjmurray:
@coreyostrove this is ready for review. I've updated docs for ``Protocol.run_mpi``, added a new ``Protocol.stage_slurm`` function, and moved shared logic / small utilities to pygsti/tools/mpitools.py.

pygsti/protocols/protocol.py
```diff
@@ -124,6 +199,311 @@ def run(self, data, memlimit=None, comm=None):
...
+        location and print the launch command for embedding in your own batch
+        script.
+        """
+        import os as _os
+        import subprocess as _subprocess
+        import sys as _sys
+        import tempfile as _tempfile
+
+        if num_ranks == 1:
+            assert not dry_run, "dry_run=True is not meaningful when num_ranks=1"
```
coreyostrove, line 326: This warning should be something like "dry_run=True is incompatible with num_ranks=1, as in this case we revert back to the standard `run` method without any MPI overhead."

pygsti/protocols/protocol.py
```diff
@@ -124,6 +199,311 @@ def run(self, data, memlimit=None, comm=None):
...
+            return out
+
+        if dry_run and persistent_dir is None:
+            msg  = "persistent_dir must be set when dry_run=True, "
+            msg += "so that generated files persist after run_mpi returns."
+            raise ValueError(msg)
+
+        mpiexec  = _mpitools.resolve_mpiexec(mpiexec)
+        blas_tpr = _mpitools.compute_blas_threads(num_ranks, blas_threads_per_rank)
+        print(f"run_mpi: setting blas_threads_per_rank={blas_tpr}\n")
```
coreyostrove, line 337: This logging should be managed by VerbosityPrinter (and have a verbosity kwarg added).

pygsti/tools/mpitools.py
```diff
@@ -1086,3 +1088,300 @@ def closest_divisor(a, b):
...
+    artifact_dir = _pathlib.Path(artifact_dir)
+
+    # Write protocol so workers can restore it with the correct subclass.
+    protocol_dir = str(artifact_dir / 'protocol')
+    protocol_obj.write(protocol_dir)
+
+    # Pickle run_kwargs so arbitrary objects (checkpoints, simulators, etc.) pass through.
+    # Disable checkpointing by default — checkpoint files written to a subprocess CWD
+    # are useless, and the auto-generated checkpoint path can fail for non-trivial mode names.
+    run_kwargs.setdefault('disable_checkpointing', True)
```
coreyostrove, line 1253: Can you explain this decision in more detail? The design decision to have checkpointing be default was intentional and in part motivated by HPC use cases, so I would not support changing that behavior unless there was a compelling reason. What exactly breaks here?

coreyostrove (changes requested):
This is great work and will significantly reduce the overhead for users unfamiliar with MPI (many of users) getting access to those performance benefit

I have left a few requests and concerns in my review, but I think these should be fairly quick to address.

pygsti/protocols/protocol.py
rileyjmurray: I've revised the message but kept it as an error rather than a warning.

pygsti/protocols/protocol.py
rileyjmurray: Verbosity is usually an instance variable in the Protocol object, so I'm not sure if a kwarg is the best fit. 

I'm amenable to either of the following approaches.

1. We add a `verbose` kwarg that indicates if _any_ staging information is logged (boolean, all or nothing).
2. We infer `verbose = self.verbosity > 0` and default to verbose=True if the Protocol does not have a `verbosity` instance variable.

I'm less amenable to an approach that tried to create a VerbosityPrinter from a self.verbosity attribute, since then the natural thing to do would be to decrement self.verbosity, which might surprise a user in terms of getting less output about GST for given verbosity depending on if using protocol.run() or protocol.run_mpi().

Thoughts?

pygsti/tools/mpitools.py
rileyjmurray: I suppose this is over-zealous. What it's getting at is how checkpoints only "work" if your MPI is running with a persistent working directory. When I originally wrote this function there was no such option for a persistent directory--- everything was written to a temporary directory deleted before function return. I'll update.

pygsti/protocols/protocol.py
rileyjmurray: Ping @coreyostrove

coreyostrove: I like option #2, using the verbosity from the Protocol.

rileyjmurray: Done!

Marking this thread as resolved. Will merge once tests pass.

PR #730, opened 2026-03-24
Fix bug with fidelity constraint

coreyostrove:
Fix a bug in `tools.lindbladtools.random_CPTP_error_generator_rates` that was resulting in the wrong value of the fidelity or generator error when using the constraints but only one of 'H' or 'S' for the allowed error generator types. The bug arose because the code generates the S rates even when they are not returned in the final output (this should get changes, but would involve some more disentangling), meaning these were still entering into the computations to determine the relative contributions expected from each sector. To patch this we now simply check which sectors are allowed to gate the relevant parts of the computation.

Fixes #716.

Note 3/23: Failing unit tests on 3.12 are from numpy changes that @rileyjmurray has already patched on develop. We'll need to get those merged into bugfix at some point.

PR #731, opened 2026-03-25
pay down tech debt in lindbladtools.py

### LINKED ISSUE #716: Failure to constrain total error of random H-only or S-only model
**Describe the bug**
The random CPTP error generator tool does not properly enforce the overall channel error when using `error_metric_value` for H-only and S-only models. 

**To Reproduce**
MWE:
```
import pygsti
import numpy as np
import scipy as sp

random_Honly_errors = pygsti.tools.lindbladtools.random_CPTP_error_generator_rates(
    1, ('H',),
    error_metric="generator_infidelity",
    error_metric_value=1e-2,
    label_type="local")
errgen = pygsti.modelmembers.operations.LindbladErrorgen.from_elementary_errorgens(random_Honly_errors, state_space=["Q0"])
noisy_ptm = sp.linalg.expm(errgen.to_dense("HilbertSchmidt"))
assert np.isclose(pygsti.tools.generator_infidelity(noisy_ptm, np.eye(4)), 1e-2) # Will fail
```

**Expected behavior**
It is expected that the noisy PTM would have the same generator infidelity as specified by `error_metric_value`.

Interestingly, the following *DOES* work:
```
pygsti.tools.lindbladtools.random_CPTP_error_generator_rates(
    1, ('H','S'),
    error_metric="generator_infidelity",
    error_metric_value=1e-2,
    relative_HS_contribution=(1,0),
    label_type="local")
```

 - pyGSTi version v0.9.14.2
 - python version 3.11
 - RHEL 9

rileyjmurray:
This is targeting the branch Corey made to resolve #716; see his PR #730.

@coreyostrove please tell me if these changes look okay. I encourage you to just look at the code rather than the diff. The diff will be much harder to read than reading the code from scratch.

### DIFF
--- a/pygsti/tools/lindbladtools.py
+++ b/pygsti/tools/lindbladtools.py
@@ -527,66 +527,66 @@ def _filtered_leeldict_to_array(
 
 
 def _validate_random_CPTP_params(
-        errorgen_types     : tuple[str, ...],
-        H_params           : tuple[float, float],
-        SCA_params         : tuple[float, float],
-        error_metric       : str   | None,
-        error_metric_value : float | None,
-        rel_HS_contrib     : tuple[float, float] | None,
-        fixed_errorgen_rates: dict,
-        max_weights: dict[str, int] | None
+        errorgen_types       : tuple[str, ...],
+        H_params             : tuple[float, float],
+        SCA_params           : tuple[float, float],
+        error_metric         : Literal['generator_infidelity', 'total_generator_error'] | None,
+        error_metric_value   : float | None,
+        rel_HS_contrib       : tuple[float, float] | None,
+        fixed_errorgen_rates : dict,
+        max_weights          : dict[str, int] | None
     ) -> tuple[float, float]:
     """Validate inputs for random_CPTP_error_generator_rates. Returns (fixed_H_contrib, fixed_S_contrib)."""
-    if error_metric is None:
-        return 0, 0
+
+    if 'C' in errorgen_types or 'A' in errorgen_types:
+        msg = 'Must include S terms when C and A present. Cannot have a CP error generator otherwise.'
+        assert 'S' in errorgen_types, msg
+
+    if max_weights is not None:
+        msg = 'The maximum weight of the %s and A terms should be <= the maximum weight of S.'
+        assert max_weights.get('C', 0) <= max_weights.get('S', 0), msg % 'C'
+        assert max_weights.get('A', 0) <= max_weights.get('S', 0), msg % 'A'
+
+    if len(fixed_errorgen_rates) == 0 or error_metric is None:
+        return 0.0, 0.0
     
-    assert error_metric_value is not None, 'error_metric_value must be specified when error_metric is not None.'
-    assert H_params[0] == 0. and SCA_params[0] == 0., \
-        'Specifying non-zero HSCA means together with a target error metric is not supported.'
-    if error_metric not in ('generator_infidelity', 'total_generator_error'):
-        raise ValueError('Unsupported error metric type. Currently supported options are generator_infidelity and total_generator_error')
-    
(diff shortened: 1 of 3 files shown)


pygsti/tools/lindbladtools.py
```diff
@@ -306,33 +332,39 @@ def create_elementary_errorgen(typ, p, q=None, sparse=False):
             elem_errgen[:, i] = rho1.flatten()[:, None] if sparse else rho1.flatten()
     else:
         # Loop through the standard basis as all possible input density matrices
+        rho1 = _np.zeros((d,d), dtype=eeg_dtype)
```
coreyostrove, line 335: The only case that needs that actually uses this memory allocation is the H case, I think we should revert this back to happening only there.

pygsti/tools/lindbladtools.py
```diff
@@ -306,33 +332,39 @@ def create_elementary_errorgen(typ, p, q=None, sparse=False):
...
         for i in range(d): 
             for j in range(d):
                 # Only difference between H/S/C/A is how they transform input density matrices
                 if typ == 'H':
-                    rho1 = _np.zeros((d,d), dtype=_np.complex128)
-                    rho1[:, j] = -1j*p[:, i]
+                    # rho1 complex
+                    rho1[:] = 0
```
coreyostrove, line 341: Why re-initialize the array to zero?

pygsti/tools/lindbladtools.py
```diff
@@ -306,33 +332,39 @@ def create_elementary_errorgen(typ, p, q=None, sparse=False):
...
+                    rho1[:, j] -= 1j*p[:, i]
```
coreyostrove: Should keep this as a simple assignment

coreyostrove (changes requested):
Thank you very much for these refactors, Riley (and maybe your intern too?)! I have left a few minor requests to revert a couple changes to the elementary error generator construction code, but otherwise this looks good. Could you also take a look at the traceback on the python 3.9 tests related to type annotations? I'm guessing this just needs a `from __future__ import annotations` call.

pygsti/tools/lindbladtools.py
nkoskelo: rho1 is initialized outside the main loop and so it needs to be reset each loop iteration.

nkoskelo:
It appears as though the value of `typ` is constant throughout the execution of `create_elementary_errorgen_*`. However, the branches for the value of `typ` are checked within the loop. We can refactor to have the if statements outside the loop.

pygsti/tools/lindbladtools.py
rileyjmurray: This avoids allocating in the loop. The memory overhead is tiny (allocate d^2 space once or allocate d^2 space a total of d^4 times).

pygsti/tools/lindbladtools.py
rileyjmurray: good catch; fixed.

rileyjmurray:
> It appears as though the value of `typ` is constant throughout the execution of `create_elementary_errorgen_*`. However, the branches for the value of `typ` are checked within the loop. We can refactor to have the if statements outside the loop.

You're correct that `typ` is constant throughout the loop. Normally I'd agree that it's better to do the check out-of-loop. In this context I'd suggest we defer that to another PR.

rileyjmurray:
Failing tests for Python 3.12 are due to numpy 2.0 incompatibility on the `bugfix` branch. No such issue is on develop.

rileyjmurray:
Python 3.9 tests pass. @coreyostrove I believe this is ready to merge. Should I charge the target to bugfix and merge now, or stick with the current target (and also merge now :))?

coreyostrove:
Dealers choice

PR #735, opened 2026-04-19
Simultaneous GST experiment designs

rileyjmurray:
This PR adds *simultaneous GST* (SGST) experiment designs. Given a `QubitProcessorSpec`, a 1Q GST design, and a 2Q GST design, it produces one circuit list that exercises many disjoint 1Q and 2Q GST experiments simultaneously across the device. (This work was developed under the name "crosstalk-free GST" / XFGST — the branch name and the linked issues still use it — but everything shipped here is named for *simultaneous GST*.)

The packing of disjoint experiments onto the n-qudit device comes from an edge coloring of the 2Q connectivity graph. Each color class — a *patch* — is a set of mutually disjoint edges, so all of its 2Q experiments can run together, and any qubit not covered by an edge in that patch runs 1Q GST instead. The coloring algorithms are #848 and the lane-tensoring machinery is #849; this PR is the layer on top of both. It also carries changes to core code outside the feature, including one observable change to a public class, which has its own section and should be reviewed independently.

### The API

New module `pygsti/protocols/simultaneous_gst.py`, star-imported into `pygsti.protocols` with `__all__ = ['SimultaneousGSTDesign', 'make_simultaneous_gst_design']`.

`SimultaneousGSTDesign(GateSetTomographyDesign)` takes a processor spec, the two GST designs, and an edge coloring:

| Parameter | Default | Meaning |
|---|---|---|
| `edge_coloring` | *(required)* | `{color: [edge, ...]}`; one patch per color |
| `circuit_stitcher` | `assign_the_designs_with_mapping` | Pluggable; see the contract below |
| `nested` | `False` | A declaration about the stitcher's output, forwarded to GateSetTomographyDesign.__init__; the default stitcher always nests, so only set this to False when supplying one that does not |
| `debug_check` | `True` | Stitcher-agnostic verification of the result |
| `seed` | `None` | Seeds the `np.random.default_rng` handed to the stitcher |
| `verbosity` | `0` | 0 is silent; greater than 0 displays a progress bar showing how many germ powers we have worked through. Forwarded to the stitcher, so a custom one may ignore it |
| `**stitcher_kwargs` | — | Forwarded verbatim; takes precedence over the defaults above |

`make_simultaneous_gst_design(nq_pspec, oneq_gstdesign, twoq_gstdesign, seed=0, verbosity=0)` is the simplified constructor: it computes the connectivity graph and its maximum degree, obtains a coloring from `switchboard_find_edge_coloring("auto", ...)` — an optimal closed-form coloring for canonical topologies (line, ring, grid, torus, as produced by `ProcessorSpec(geometry=...))`, a generic `(deg + 1)`-color algorithm otherwise — and delegates. There are two independent consumers of randomness downstream of seed — the coloring and the stitcher — so seed is expanded with np.random.SeedSequence(seed).spawn(2) and each is handed its own stream; neither draws from the other's sequence, and neither depends on how much randomness the other consumed.

### The default stitcher

`assign_the_designs_with_mapping` holds essentially all of the feature's logic. It returns `circuit_lists[L]` for germ-power index `L`, and there are four decisions in it worth review.

*Patch-major ordering is a contract. Within each germ power, circuits follow the input ordering of `color_patches`, and downstream verification checks this rather than assuming it.

*The two designs must share a germ-power schedule*; the stitcher raises a NotImplementedError `if len(oneq_circuitlists) != len(twoq_circuitlists)`.

*List lengths are reconciled per lane slot. `random_index_schedule(n, max_len, randgen)` always includes `0..n-1`, so every circuit in the shorter list appears at least once; the remaining slots are drawn uniformly with repetition and the whole schedule is shuffled. Every circuit in the longer list is used exactly once, either list may be the longer one, and a fresh schedule per lane slot means different edges in a patch get different assignments. This is list-length equalization, distinct from the depth padding `batch_tensor` applies to shallower sub-circuits.

*A parameter `share_same_shape_schedules` in `assign_the_designs_with_mapping` defaults to `True` and controls whether the patches should have the same random schedule. If shared, patches are bucketed by `(num_edges, num_unused_qubits)`; one tensored circuit is built per group, and the rest obtain theirs via `map_state_space_labels` through `make_line_mapper`, which zips lane layouts positionally and raises on arity mismatch, inconsistent duplicates, and non-injective results. This is the stitcher's main optimization and also its main failure mode — copying instead of relabeling silently puts several patches' circuits on the representative's qubits — so MultiplePatchesSameShapeTester targets it directly.

Two other things to know. Idles are always explicit: build_layer_mappers maps the empty layer label Label(()) onto an explicit `Gi`/`Gii` idle and rewrites 1Q labels inside the 2Q design into parallel layers, which matters because the tensored circuits are later split back into lanes. And `seed` reproducibility is tied to the shape-group iteration order, so renaming or reordering patches changes the generated circuits. The default circuit stitcher nests the circuit lists. So, germ power `L+1` contain germ power `L` exactly, patch-wise; `NestingTester` covers the single-patch, same-shape, and three-germ-power transitivity cases, plus that nesting preserves the patch-major ordering above.

### Writing an alternative stitcher

```python
circuit_stitcher(oneq_gstdesign, twoq_gstdesign, vertices,
                             color_patches, randgen=..., verbosity=...,
                             **stitcher_kwargs) -> List[List[Circuit]]
```

`SimultaneousGSTDesign` supplies `randgen` and `verbosity` and merges `**stitcher_kwargs` over them, so an alternative can take its own options without a signature change anywhere; the built-in accepts and ignores `**kwargs` to stay compatible.

Validation is external by design. `assert_circuit_lists_match_color_patches`, run from `__init__` when `debug_check=True`, re-derives the patch layout and checks even patch-major chunking, absence of implicit idles, and that each circuit's line labels are exactly its patch's qubits and every multi-qubit op sits on one of that patch's own edges. Comparing only the set of line labels does not catch a mis-relabeled template, since `{0,1,2} == {0,1,2}` holds even when the gates are on the wrong lines. The per-circuit checks run against the denested content — each distinct circuit is checked once per patch and repeats are skipped — since both are pure functions of the circuit and its patch, and checking the same circuit twice does not yield any new information. Because the check runs against whatever was returned, it still applies to a swapped-in stitcher. One known non-behavior: patches are not deduplicated, so `[(0,1),(2,3)]` and `[(1,0),(3,2)]` both generate.

### Behavior changes outside the feature

**New `ImplicitOpModel.__getitem__(self, key)`. `ExplicitOpModel` has supported `model[label]` for a long time; `ImplicitOpModel` had no `__getitem__` at all, on this class or any base class, so this gives the implicit-model side the same indexing. The key is coerced with `Label(key)`, so a string, a tuple, or a Label all work `(m['Gx'], m['Gx:qb0'], m[('Gx', 'qb0')])`. It then searches the `OrderedMemberDicts` hanging off `prep_blks`, `povm_blks`, `operation_blks`, `instrument_blks`, and `factories`, returning the first hit and raising `KeyError` naming the key if there is none. As a heuristic it visits the non-empty dicts in ascending order of length, on the reasoning that SPAM dicts and the 1Q/2Q gate dicts that later get embedded into n-qubit layers are the short ones and the likely targets.

### Incidental changes

 -   `pygsti/models/memberdict.py`: new `LabelLike = Label | tuple | str` alias (used by the `__getitem__` above), applied as annotations to `OrderedMemberDict.__contains__`, `__getitem__`, `__setitem__`, and `__delitem__`. An unused `copy` import and a commented-out `_clean_paramvec` block go away.
 -   `pygsti/models/localnoisemodel.py`: eight unused imports dropped, plus a comment recording the open question of why `instrument_blks` is only ever keyed by 'layers' when the things in `operation_blks['gates']` have direct instrument analogs.
 -  `pygsti/baseobjs/label.py`: `has_prefix` gets a `prefix: str` annotation.
 -  `test/unit/objects/test_localnoisemodel.py` renamed to `test/unit/objects/test_implicitmodel.py` and restructured: the setup and a new `_test_getitem` check move onto an `ImplitOpModelMixin` so they can be reused as `CloudNoiseModel`'s tests migrate over from `test_nqnoiseconstruction.py` (there's a TODO to that effect), and `LocalNoiseModelInstanceTester` becomes `LocalNoiseModelTester`.
 -   `test/unit/tools/test_matrixtools.py`: `test_minweight_match` now asserts the returned weights and not only the pairs. Every matched pair in that fixture differs by exactly 0.1, so the weights are as diagnostic as the matching and were going unchecked.

### Testing

`test/unit/protocols/test_simultaneous_gst.py` (new) covers the stitcher against stub designs and the design class against a real 3-qubit line:

-   `AssignDesignsLengthPairingTester` — length reconciliation in both directions, plus idle padding of shallower sub-circuits.
-   `MultiplePatchesSameShapeTester` — same-shape patches are relabeled, not duplicated.
-   `NestingTester` — ensure the circuit lists are nested in germ power, patch major order.
-   `MakeSimultaneousGSTDesignTester` — the convenience entry point: graph derivation from the processor spec, that the coloring it produces is a proper edge coloring, that defaults land where they should, and that the two spawned seed streams reach the coloring and the stitcher separately and actually change the assignments.
-   `HelperRejectsMalformedInputTester` — the `debug_check` path, including that all patches are checked rather than just the first, and that a multi-qubit gate outside its patch's edges is caught.
-    `IndependentSchedulesTester` - when asked patches with the same number of edges can have either the same random schedule or independent ones. This tests the independent path.
-   `FindNeighborsTester`, `MakeLineMapperValidationTester`, `AssignDesignsDefaultRandgenTester` — the supporting helpers, including each of `make_line_mapper`'s four rejection cases and the omitted-`randgen` default.

`test/integration/test_simultaneous_gst.py` (new) runs design construction, data simulation, and GST on a 3-qubit line under four Lindblad noise configurations — H, S, H+S, and H+S+C+A, the last under `GLND` not `CPTPLND`, asserting a `2*deltaLogL` bound for each. It is reduced-scale by construction (`max_max_length=2`) and runs in about three minutes.


pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,185 @@
+import numpy as _np
```
nkoskelo: Should we separate out laned circuits into their own class rather than shoehorning them into `Circuit`?

pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,185 @@
...
+            assert starting_point in graph_qubits_to_neighbors
+            visited[starting_point] = graph_qubits_to_neighbors[starting_point]
+            output = set(visited[starting_point])
+            for child in graph_qubits_to_neighbors[starting_point]:
+                if child != starting_point:
+                    output.update(output, reachable_nodes(child, graph_qubits_to_neighbors, visited))
+            visited[starting_point] = output
+            return output
+
+    available_starting_points = list(sorted(qubits_to_potentially_entangled_others.keys()))
```
nkoskelo: Do we need to resort the qubits every call to reachable nodes?

pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,185 @@
...
+    return lanes_to_gates
+
+
+def batch_tensor(
+    circuits : Sequence[Circuit],
+    layer_mappers: Dict[int, Dict],
+    global_line_order: Optional[Tuple[int,...]] = None,
+    target_lines : Optional[Sequence[Tuple[int,...]]] = None
+    ) -> Circuit:
+    """
```
nkoskelo, line 165: This function will just tensor horizontally the circuits together into one big circuit with a lane structure. It is very similar to `cir.tensor_circuit(other)` chained together for every circuit in circuits.

pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,185 @@
...
+    c._static = False
+    c._labels = [layer_mappers[c.num_lines][ell] for ell in c._labels]
+    c.map_state_space_labels_inplace({k:v for k,v in zip(c.line_labels, target_lines[0])})
+    c.done_editing()
+    for i, c2 in enumerate(circuits[1:]):
+        c2 = c2.copy(editable=True)
+        c2._append_idling_layers_inplace(max_cir_len - len(c2))
+        c2.done_editing()
+        c2._static = False
+        c2._labels = [layer_mappers[c2.num_lines][ell] for ell in c2._labels]
```
nkoskelo: `layer_mappers` is a dictionary of integers to a dictionary. So, `layer_mappers[c2.num_lines]` may be the same as that of `layer_mappers[c1.num_lines]` resulting in the same output of labels. This works if you are changing gate names and you want to have a consistent naming convention for a gate. However, it means that for every single circuit with a unique length one needs to duplicate the rule set. This is not ideal. Also if, `[layer_mappers[c2.num_lines][ell] for ell in c2._labels]` returns a different length label then ell then we may have a different number of layers in the circuit `c2`. This could then result in `c.tensor_circuit(c2)` behaving incorrectly.

I think that `layer_mappers` should be renamed to `gate_name_canonicalizer` or something similar.

pygsti/circuits/split_circuits_into_lanes.py
rileyjmurray: Let's leave that to follow-up work. The point of this PR is to merge the useful stuff we already had on the other branch.

pygsti/circuits/split_circuits_into_lanes.py
rileyjmurray: Not sure. I suggest we not try to optimize this cost away unless we know it's a problem.

pygsti/circuits/split_circuits_into_lanes.py
rileyjmurray: Totally your call if you want to keep this function, drop this function, or modify it.

pygsti/circuits/split_circuits_into_lanes.py
rileyjmurray: This is a moot point if we drop this function. Please decide that first.

pygsti/tools/graphcoloring.py
```diff
@@ -0,0 +1,276 @@
+import numpy as np
+import copy
+
+
+
+def find_edge_coloring(deg: int, vertices: list, edges: list, neighbors: dict) -> dict:
```
nkoskelo: Can this be a randomized algorithm or are requiring it to be a deterministic algorithm? There is a paper which claims to beat Misra & Gries algorithm assymptotically.

rileyjmurray: Randomized is only okay if this function is a bottleneck or if the randomized algorithm has some additional robustness properties. I suggest we not consider this question now. Open a new github issue post PR merge instead.

pygsti/processors/processorspec.py
```diff
@@ -1341,8 +1341,17 @@ def compute_clifford_2Q_connectivity(self):
...
+                if len(avail) == 1 and avail[0] is None and gn == '{idle}':
+                    avail = [qubit_labels]
+                    # if qubit_labels.size == 2:
+                    #     avail = [qubit_labels]
+                    # else:
+                    #     raise ValueError('Availability of the idle gate has not been set.')
+                for sslbls in avail:
+                    i = qubit_labels.index(sslbls[0])
+                    j = qubit_labels.index(sslbls[1])
+                    CtwoQ_connectivity[i, j] = True
```
nkoskelo, line 1646: `CtwoQ_connectivity[j,i] = True`.

pygsti/models/localnoisemodel.py
```diff
@@ -403,6 +403,18 @@ def rescale(coeffs):
 
         return op_coeffs
 
+    def __getitem__(self, key):
```
nkoskelo: Lets add an example test case that uses this new functionality.

pygsti/processors/processorspec.py
nkoskelo: Yes, we do assume that it is symmetric. However, we assume it is an undirected graph and so do not want two copies of the same edge. Therefore, we will not add the "vu" edge unless explicitly told so.

pygsti/circuits/split_circuits_into_lanes.py
nkoskelo: This is a standardized and convenient way to build the tensor circuits so I am going to keep it.

pygsti/circuits/circuit.py
```diff
@@ -1343,13 +1345,27 @@ def get_sslbls(lbl): return lbl.sslbls
...
+                        components = [_Label('I', line) for line in self._line_labels]
+                    else:
+                        components = [layer_lbl]
+                else:
+                    components = list(layer_lbl.components)
+                    for line in self._line_labels:
+                        if line not in layer_lbl.sslbls:
+                            components.append(_Label('I', line))
+            else:
+                components = list(layer_lbl.components)
```
rileyjmurray: A label whose string representation starts with a `I` is assumed to be an Instrument. I'm not sure if we have a truly standard way of referring to an idle gate. `Gi` is the most common, but that's not universal. This seems like a deeply messy issue.

rileyjmurray: On a separate but related note, I don't like the depth of the nesting here. We've got a ``for``, then ``if``, then ``if``, then ``for``, then ``if``. That's very hard to follow.

pygsti/circuits/split_circuits_into_lanes.py
rileyjmurray, line 1: Probably worth moving to a second PR.

pygsti/tools/matrixtools.py
```diff
@@ -2504,49 +2619,91 @@ def zvals_int64_to_dense(zvals_int, nqubits, outvec=None, trust_outvec_sparsity=
...
-
     if outvec is None:
         outvec = _np.zeros(4**nqubits, 'd')
+        # `outvec` is guaranteed to be freshly all-zero here, so there is no need to
+        # (redundantly) reset it below.  This matters a lot in practice: `np.zeros`
+        # for a large array is backed by lazily-mapped, copy-on-write OS zero-pages
+        # and so is essentially free, but explicitly writing zeros into it (as
+        # `outvec[:] = 0` does) forces the *entire* O(4**nqubits)-sized array to be
+        # physically paged in and zeroed, even though we only ever touch O(2**nqubits)
+        # of its elements below.  Skipping this avoids that (otherwise dominant) cost.
```
rileyjmurray: This kind of exposition is entirely unnecessary. Remove.

pygsti/tools/matrixtools.py
```diff
@@ -2504,49 +2619,91 @@ def zvals_int64_to_dense(zvals_int, nqubits, outvec=None, trust_outvec_sparsity=
...
+
+    # The 2**N nonzero elements of the resulting vector are at indices that are
+    # base-4 numbers whose digits are only 0 or 3.  Rather than indexing these via
+    # the "natural" bit order of `finds` (which produces a bit-reversal permutation
+    # of indices), `_zvals_int64_indices_and_signs` indexes via `r = 0..2**N-1`
+    # directly, using order-preserving bit-spreading to get the corresponding
+    # index.  This means `final_indices` comes out already sorted, so the scatter-write
+    # below is sequential rather than jumping around unpredictably (the naive
+    # bit-indexing scheme is a textbook-bad memory access pattern once `outvec` no
+    # longer fits in cache).
```
rileyjmurray: It's extremely misleading for this function to suggest that efficiency here matters. It does not. I take it that this comment block was written by a robot. That's fine, but I expect you to remove these kinds of things.

pygsti/evotypes/densitymx_slow/effectreps.py
```diff
@@ -75,9 +75,12 @@ def __reduce__(self):
...
     def probability(self, state):
-        scratch = _np.empty(self.state_space.dim, 'd')
-        Edense = self.to_dense('HilbertSchmidt', scratch)
-        return _np.dot(Edense, state.data)  # not vdot b/c data is *real*
+        # Use the sparse inner-product routine rather than densifying this (mostly-zero)
+        # effect vector to length 4**nfactors and then taking a full dense dot product;
+        # this avoids O(4**nfactors) work (both the densification and the dot product)
+        # in favor of O(2**nfactors) work, mirroring the analogous fast C++ implementation
+        # used by the compiled evotypes.
+        return _mt.zvals_int64_probability(self.zvals_int, self.nfactors, state.data, self.abs_elval)
```
rileyjmurray: I appreciate the desire to have the Python implementation match the C++ implementation here. However, the savings is, at the end of the day, modest. It seems that the changes to matrixtools.py to support this change are over 200 lines long. This leads me to favor restoring the old implementation, and just leave a comment saying something like "the C++ densitymx EffectRepComputational evotype tries to be more clever."

rileyjmurray (commented):
It's occurring to me that much of PR's complexity can be attributed to the concept of "lanes" within a circuit. I recommend we move all of that functionality to a second PR.

pygsti/evotypes/densitymx_slow/effectreps.py
nkoskelo: I will also update the C++ version to do the clever trick on array access order.

rileyjmurray: @nkoskelo please don't. There's a good chance that all of pyGSTi's existing C++ code will be replaced within the next year. It's important to not devote effort to it unless it's about resolving a correctness issue.

rileyjmurray:
It seems possible to split out this PR's crosstalk-free GST feature code from its incidental changes. I'm putting together a PR for the incidental changes now. Once merged into develop we can return to the topic of (updating and) merging this PR.

rileyjmurray:
Ping @nkoskelo. I gave the robot the following prompt.

> I'm looking at the docstring of assign_the_designs_with_mapping, and I don't really follow. Specifically, I don't understand what's meant by "Patches with the same number of 2Q edges and unused qubits share randomized role-based schedules." Do you see well-posed meaning there, or can such meaning be derived from the code? 

I iterated with it a bit, explaining that edesign construction is __tiny__ compared to the cost of doing the GST model fitting. Its assessment is below. Please take a look. Let me know if you have a preference for how to move forward (and, if so, explain your reasoning). If you don't have a preference then I'll decide.

------------------------------------------------------

### What "share randomized role-based schedules" means

The sentence under review, from `assign_the_designs_with_mapping`'s docstring:

> Patches with the same number of 2Q edges and unused qubits share randomized role-based schedules. A representative tensored circuit is constructed once for each such group and then mapped onto equivalent patches.

Short answer: there *is* a well-posed meaning behind it, and the code implements exactly one thing, but the sentence is a poor description of that thing. "Role" is coined here and appears nowhere else in the module; "share" is ambiguous between the two readings that matter; and the sentence understates what is shared — it is not the schedules, it is the circuits. The grouping condition it states is also redundant. Everything below is derived from the code and checked against it.

#### 1. The mechanics, in the code's own terms

Three functions are involved. `build_patch_infos` buckets patches into `groups`, keyed by `(len(edge_set), len(unused_qubits))`. `build_group_schedules_and_mappers` is called once per group per germ power and returns `edge_perms` of shape `(num_edges, max_len)` and `oneq_perms` of shape `(num_unused_qubits, max_len)`. The generation loop then does, for each `j in range(max_len)`, one `batch_tensor` call producing a template for the group's representative, and one `map_state_space_labels` call per remaining patch in the group.

So a "role" is a **slot index**: edge slot `r` is row `r` of `edge_perms`, 1Q slot `r` is row `r` of `oneq_perms`. A "schedule" is that row — a length-`max_len` sequence of indices into the corresponding design's circuit list for this germ power, produced by `random_index_schedule`.

The slot-to-physical-line correspondence comes from `patch_lines`, which returns `sorted(edge_set) + [(q,) for q in unused_qubits]`. Edge slot `r` is therefore the `r`-th edge of a patch **in lexicographic order of the edge tuples**, and 1Q slot `r` is the `r`-th unused qubit in `vertices` order. `make_line_mapper` then zips the representative's `tensored_lines` against the target patch's positionally, which is what makes slot `r` of one patch correspond to slot `r` of another.

#### 2. What "share" actually means

The ambiguity worth resolving: does each patch draw *its own* schedule from the same distribution, or do all patches in a group get the *same realized* schedule? It is the latter, and in fact something stronger — the patches don't merely share indices, they share the fully constructed circuit. Only the representative's circuit is ever built; every other patch's circuit is that same object relabeled.

Verified on a 5-qubit line with patches `{0: [(0,1),(2,3)], 1: [(1,2),(3,4)]}` (both shape `(2 edges, 1 unused)`, so one group of two):

```
rep->patch1 mapper: {0: 1, 1: 2, 2: 3, 3: 4, 4: 0}
patch1 == relabel(patch0) elementwise: True
```

Within a group, different slots do get different schedules — the sharing is across patches, not across slots:

```
edge-slot-0 schedule (which 2Q circuit, by depth): [4, 3, 6, 5, 1, 2]
edge-slot-1 schedule:                              [5, 6, 2, 3, 1, 4]
```

An incidental consequence: `make_line_mapper`'s source and target line lists each cover every vertex exactly once (every qubit is either in a patch edge or unused), so the mapper produced from this call site is always a **permutation of the full vertex set**. All four of `make_line_mapper`'s rejection branches are therefore unreachable from `assign_the_designs_with_mapping` given well-formed patches; they are reachable only through malformed input, which is how `MakeLineMapperValidationTester` exercises them.

#### 3. What the randomization is for

This is the part the docstring never says, and it is the part a reader actually needs. Because `random_index_schedule(n, max_len, randgen)` always includes `0..n-1`, and `max_len = max(oneq_len, twoq_len) >= n` for both designs, **every slot's schedule contains every circuit of its design at least once**. So the randomization does not choose *which* circuits run: every edge runs the complete 2Q GST design and every idle qubit runs the complete 1Q GST design, at every germ power, unconditionally. Confirmed above: both edge slots' schedules cover all six 2Q circuits.

What the randomization chooses is the **co-occurrence pattern** — reading column `j` down `edge_perms` and `oneq_perms` gives the set of sub-circuits that get tensored together into tensored circuit `j`. The point is to avoid a systematic pairing (2Q circuit `k` always simultaneous with 1Q circuit `k`, or always with the same germ), which under real crosstalk would bias every patch's estimate in the same direction instead of averaging out.

Stating that would make the docstring's list-length paragraph and this one cohere: they are two aspects of the same schedule, one about coverage and one about pairing.

#### 4. Two problems with the sentence as written

**The grouping condition is redundant.** Under the function's own stated precondition — "Each color patch should contain mutually disjoint edges" — a patch's unused-qubit count is determined: `num_unused = len(vertices) - 2 * num_edges`. Verified for every patch in the example above. So the key `(num_edges, num_unused_qubits)` is effectively just `num_edges`, and "patches with the same number of 2Q edges and unused qubits" is really "patches with the same number of 2Q edges". The second component only becomes independent when the precondition is violated: a patch containing both `(0,1)` and `(1,0)` has `num_edges = 2` but, on 5 qubits, `num_unused = 3` rather than `1`. That is arguably a virtue of the key — it's defensive — but the docstring presents a redundant condition as a two-part one, which makes the grouping look more subtle than it is.

**"Share randomized role-based schedules" describes the wrong artifact.** The reader's operative question is "do two same-shape patches run the same circuits?", and the answer is yes. The sentence says something narrower and more abstract, and then the next sentence half-retracts it by mentioning the template. Leading with the template and deriving the schedule sharing from it would be both shorter and more accurate.

#### 5. Is the sharing well-motivated?

I do not think it currently is. The argument the branch implicitly makes is a cost argument, and the cost it reduces is the one cost in this pipeline that does not matter.

**The cost argument does not survive contact with the numbers.** Sharing reduces work at *design-construction* time only. It saves nothing anywhere else: the patches sit on different qubits, so their circuits are distinct circuits regardless — same shot count, same hardware wall-clock, same number of likelihood terms in the fit. A GST design that takes seconds to construct then runs for several minutes on a small example and for days on a cluster for an interesting one.

Measured directly, on a 5-qubit line with the full `smq2Q_XYICNOT` design at `max_max_length=8` (24,852 output circuits). The 6-qubit line is the control: its two color classes differ in size, so each patch is its own group and no template is ever reused, while the output circuit count is identical.

```
5q line: 2 patches in 1 group  [2]    | stitch 19.3s | debug_check 1.6s | 24852 circuits
6q line: 2 patches in 2 groups [1,1]  | stitch 43.7s | debug_check 1.1s | 24852 circuits
```

(The two aren't a perfect A/B — the 6-qubit patch shapes differ slightly, so the per-tensor cost isn't identical — but they bracket it.) The whole prize is roughly 24 seconds, once, on a design whose downstream consumption is measured in minutes to days. The underlying per-call ratio is real — `batch_tensor` at 4.750 ms against `map_state_space_labels` at 0.228 ms, 20.9x — but a 20x speedup on 0.1% of the pipeline is not a reason to do anything.

**That leaves the statistical argument carrying the whole thing, and it is at best a wash.** Two opposed readings:

*For sharing.* Patches in a group run identical content up to relabeling, so comparing patch A's slot-`r` fit against patch B's slot-`r` fit holds the design and the simultaneous context fixed and varies only physical location. Any difference is attributable to the hardware.

*Against sharing.* The premise of simultaneous GST is that the parallel experiments don't contaminate each other, and the main risk is that the premise is false. Detecting that requires seeing estimates move as context moves. Within a group there is no context variation at all, so any crosstalk-induced bias that depends on the shared schedule is **common-mode across the entire group** and invisible to every within-group comparison. Independently drawn schedules would make the patch-to-patch spread an estimator of that bias's magnitude; sharing removes it.

Note that both readings concern the *context* only. Coverage is unaffected either way: every edge runs the complete 2Q design at every germ power regardless, per the previous section. So nothing about identifiability or the per-edge fit turns on this.

And neither reading is cashed out by anything in the branch. No analysis here computes a matched patch-to-patch contrast, and none computes a patch-to-patch spread. Both arguments are about hypothetical future analyses, and neither is recoverable from the other after the fact — you cannot reconstruct matched contexts from independent draws, or context variation from shared ones. Between two unexercised options, independent randomization of a nuisance factor across units is the standard default; matched designs are something you adopt deliberately after identifying the contrast you want to protect. Nobody adopted it here deliberately. It fell out of a caching optimization.

**Two further wrinkles, both arguing the same way.** The matched-comparison property is partial: it holds only between edges at the *same* slot index in their respective patches, and slot index comes from the lexicographic sort of edge tuples inside `patch_lines`. Which physical edge gets which schedule is therefore a function of qubit labeling — relabel the device and the matched pairs change. And whether any of this is in play at all depends on parity (next section): on an even-length line every group is a singleton, so a property a future analysis might rely on silently switches off with the qubit count.

#### 6. Recommendation: drop the grouping

The grouping machinery is not free. It is `groups` in `build_patch_infos`, the mapper half of `build_group_schedules_and_mappers`, all of `make_line_mapper`, and the representative/relabel branch of the generation loop. It carries the module's most dangerous failure mode — copying instead of relabeling silently puts several patches' circuits on the representative's qubits, which is what `MultiplePatchesSameShapeTester` and the edge check in `assert_mapped_circuit_matches_patch` exist to catch. It is why reproducibility is tied to shape-group iteration order, and why the module's behavior changes qualitatively with the parity of the qubit count.

Against that: about 24 seconds of design-construction time, and a statistical side effect nobody chose.

The two objections that would normally weigh against removing it — that it changes the circuits a given seed generates, and that it is a larger diff than the branch wants — do not apply. There are no consumers of this code, even at prototype level, and seed-stability is not currently a requirement. This is therefore the cheapest moment this change will ever have, and it gets cheaper to *not* do only if someone starts depending on it.

**Shape of the change.** The per-germ-power loop body collapses from a group loop with a representative and a relabel pass to a straight loop over patches:

```python
patch_buffers = {}
for info in patch_infos:
    edge_perms, oneq_perms = build_patch_schedules(
        info["num_edges"], info["num_unused_qubits"],
        max_len, twoq_len, oneq_len, randgen)
    patch_buffers[info["patch"]] = [
        batch_tensor(
            [twoq_circuits[i] for i in edge_perms[:, j]]
            + [oneq_circuits[i] for i in oneq_perms[:, j]],
            layer_mappers, None, info["tensored_lines"])
        for j in range(max_len)
    ]
```

Consequences:

- `build_group_schedules_and_mappers` loses its `infos` / `representative` / `representative_lines` parameters and its `mappers` return value, becoming `build_patch_schedules` — the two `random_index_schedule` loops and nothing else.
- `build_patch_infos` stops building and returning `groups`. Both production call sites and both test call sites already discard it as `patch_infos, _ = ...`.
- `make_line_mapper` becomes dead. Confirmed by grep: its only production caller is `build_group_schedules_and_mappers`, and everything else referencing it is `MakeLineMapperValidationTester`, which goes with it. That is one function and five tests deleted — none of whose rejection paths were reachable from the stitcher anyway, since the mapper built from that call site is always a full-vertex permutation (see above).
- `template_circuit.copy()` goes away; the copy exists only because the template is reused.
- `MultiplePatchesSameShapeTester` no longer has a template-copy bug to guard. Its remaining content — distinct patches produce distinct circuits, each on its own edges and idle qubits — is still worth keeping, so rewrite it rather than delete it, dropping the "same shape" premise from its name and docstring.
- `assert_mapped_circuit_matches_patch`'s edge check becomes much less load-bearing, since `batch_tensor` now receives each patch's own lines directly and cannot land a gate on another patch's edge by construction. Keep it regardless: it is cheap, and its real purpose is as the stitcher-agnostic contract for swapped-in stitchers.
- The docstring sentence under review is deleted rather than reworded. The suggested replacement in the next section applies only if the grouping is kept.

Independent per-patch randomization also removes the parity discontinuity and the labeling-dependent slot assignment described above, since neither has anything left to attach to.

#### 7. How load-bearing the grouping actually is

Worth knowing, because the answer is "it depends on parity", which is not obvious from the docstring. Group structure for the colorings `switchboard_find_edge_coloring("auto", ...)` produces:

```
line n=3     2 patches, group sizes [2]
line n=4     2 patches, group sizes [1, 1]
line n=5     2 patches, group sizes [2]
line n=6     2 patches, group sizes [1, 1]
line n=7     2 patches, group sizes [2]
line n=8     2 patches, group sizes [1, 1]
ring n=3     3 patches, group sizes [3]
ring n=4     2 patches, group sizes [2]
ring n=5     3 patches, group sizes [1, 2]
ring n=6     2 patches, group sizes [2]
ring n=7     3 patches, group sizes [1, 2]
grid 3x3     4 patches, group sizes [4]
grid 4x4     4 patches, group sizes [2, 2]
```

On an even-length line the two color classes differ in size by one, so every group is a singleton: no template is ever reused, the optimization saves nothing, and the cross-patch correlation does not exist. On an odd-length line the classes are equal and sharing is fully active. So the same code path behaves qualitatively differently between a 6-qubit and a 7-qubit line, and the reproducibility caveat about shape-group iteration order has correspondingly variable reach. The 3-qubit line used throughout the unit tests is in the sharing-active case, which is the right default, but nothing currently tests the singleton-group case for a *multi-patch* device.

#### 8. Suggested replacement, if the grouping is kept anyway

Something along these lines, replacing the two sentences. This is the fallback to the recommendation above, not an alternative worth preferring: it documents the sharing as a deliberate design decision so that a future analysis doesn't unknowingly assume a patch-to-patch independence that isn't there.

> Patches that contain the same number of 2Q edges fall into one group (given disjoint edges, this fixes the number of unused 1Q qubits too). One tensored circuit is built per group per output slot, for the group's first patch, and the other patches in the group get that same circuit relabeled onto their own edges and qubits via `map_state_space_labels`. Patches in a group therefore run identical circuit content up to relabeling, including which sub-circuits are simultaneous with which.
>
> Within a patch, each edge and each unused qubit is a slot with its own random schedule, so slots do not run the same circuits as each other. Slot order is the patch's edges in sorted order followed by its unused qubits in `vertices` order, which is also what determines the relabeling between patches in a group.

And, in the list-length paragraph, one sentence making the coverage/pairing split explicit: every slot's schedule contains every circuit of its design at least once, so the randomization determines which sub-circuits run *simultaneously*, not which run at all.

rileyjmurray:
@nkoskelo I've marked my comment from earlier today as "outdated", so it doesn't show prominently in the history anymore. Please make the changes we discussed and update the PR description accordingly. We're very close to merging this. I'd love to get it done.

pygsti/models/localnoisemodel.py
```diff
@@ -177,12 +177,7 @@ def __init__(self, processor_spec, gatedict, prep_layers=None, povm_layers=None,
         self.operation_blks['gates'] = _OrderedMemberDict(self, None, None, flags)
         self.operation_blks['layers'] = _OrderedMemberDict(self, None, None, flags)
         self.instrument_blks['layers'] = _OrderedMemberDict(self, None, None, flags)
-        # ^ Unclear why instrument_blks should only be keyed by `layers`.
-        #
-        #     I'll grant that it seems weird to key by `gates`, but the things
-        #     stored in operation_blks['gates'] have direct instrument analogs
-        #     that aren't suitable for instrument_blks['layers'].
-        #
+        # TODO: Unclear why instrument_blks is only keyed by layers, as gates can have direct instrument analogs.
```
rileyjmurray: Revert this change. I wrote that comment in a PR earlier this week.

pygsti/protocols/simultaneous_gst.py
```diff
@@ -49,61 +49,34 @@
...
+    mapper_1q[empty_label] = oneq_idle_label
     assert empty_label not in mapper_2q.values()
     assert empty_label not in mapper_1q.values()
 
-    # Check for any labels in `mapper_2q` that imply a single-qubit target.
-    # For any such label, add an explicit single-qubit idle on the non-target
-    # qubit, and wrap the whole thing as a LabelTupTup.
-    m2q = mapper_2q.copy()
-    for k2 in mapper_2q:
+    # For any single-qubit targets in mapper_2q, add explicit single-qubit idle and wrap as Label
```
rileyjmurray: Restore the old comment. You can keep the change to in-place modification of `mapper_2q` and iterating over `list(mapper_2q.keys())`.

rileyjmurray (commented):
Two small comments. Ping me when resolved and I'll merge.

rileyjmurray:
Ping @nkoskelo for my two comments.

rileyjmurray:
Note for posterity: I'm _not_ marking this PR as `merged-without-review`. My reasoning is that while I opened the PR, the work belongs to Nick. All of our back-and-forth on this branch over many, many months was the result of several rounds of review.

PR #736, opened 2026-05-05
Carefully remove almost all files from the top-level `scripts/` directory 

rileyjmurray:
This PR's diff will make it look like I blindly deleted a huge number of files. That's not what happened. I applied a general rule to remove files that couldn't run successfully because they referenced files that don't exist (and couldn't be generated by other files elsewhere in the repo). Applying this removal rule recursively led me to delete almost the entirety of the `scripts/` directory.

My individual commit comments justify each removal.

This removal was important because coding agents insisted on exploring ~8000 lines of text within this directory, even though the contents of this directory had basically no value.

I made a couple of additive changes throughout this process.
 1. Resolved an issue related to the PYGSTI_NO_CUSTOMLM_SIGINT environment variable. It's no longer necessary to set this environment variable when running pytest with multiple workers. 
 2. Added a unit test for a parallel_apply utility function in mpitools.py.
 3. Updated pytest.ini so that it catches classes with the ``Test*`` name pattern in addition to the ``*Tester`` pattern. We need to support the former pattern because it's what coding agents will default to, and it's not worth trying to fight them on that.


scripts/profiling/pickling.py
coreyostrove, line 1: Looks empty, can we delete this?

coreyostrove (approved):
Minor question, but otherwise looks good. Thanks for the spring cleaning!

scripts/profiling/pickling.py
rileyjmurray: I don't know why git says this file still exists. Accessing it on the branch shows nothing.

https://github.com/sandialabs/pyGSTi/tree/audit-root-scripts-2/scripts/profiling/pickling.py

Issue #495, opened 2024-10-03
Extend Circuit parsing of qubit labels


sserita:
**Describe the bug**
Circuit parsing currently fails if the qubit labels are not integers or begin with 'Q'. This is both very restrictive, and the error message is uninformative.

**To Reproduce**
```
from pygsti.circuit import Circuit

c = Circuit([("Gxpi2", "A0")]) # This is fine
c.str # This is also fine

Circuit(c.str) # Errors with ValueError: invalid literal for int() with base 10: ''
```

**Expected behavior**
At the very least, a clearer error message. The current error message is because it assumes it is not a valid str since it doesn't start with Q, therefore it must be an int, and then the int cast fails. We should catch this and output a more useful message.

But we should also consider allowing arbitrary qubit labels (these can be used if not deserializing from string without issue, and can be serialized with no problem. It's just deserialization that is a problem). I'll note that there is _some_ importance assigned to state space label prefixes; specifically, I believe Q is used to automatically set the state space dimensions to those of a qubit. Others exist (i.e. L for leakage, etc), so the user would maybe have to specify the state space more explicitly in these cases.

**Environment (please complete the following information):**
 - pyGSTi 0.9.12.3
 - python 3.11
 - OSX 14

rileyjmurray:
Another option is to change the parser to be very flexible about the string representation of target qudits.

For example, if we wanted to allow any string compared of letters, integers, and underscores --- provided the string does not start with a reserved character in ``{'G','M','I','S'}`` --- then we could replace
https://github.com/sandialabs/pyGSTi/blob/9910707c9cec988ffe1bd02d97606bbca2d1862b/pygsti/circuits/circuitparser/fastcircuitparser.pyx#L260-L270
with
```python
        # --- reject any subsystem‐label starting with G, I, M or S ---
        if i < end and s[i] in (u'G', u'I', u'M', u'S'):
            raise ValueError(
                "Invalid subsystem label '%s': qudit labels may not start with G, I, M or S"
                % s[last:i+1]
            )
        # now consume the rest of the label (digits, letters or '_')
        while i < end:
            c = s[i]
            if u'0' <= c <= u'9':
                i += 1
            elif (u'a' <= c <= u'z') or (u'A' <= c <= u'Z') or c == u'_':
                i += 1; is_int = False
            else:
                break
```
If we made that change then we'd want to change the docstring of CircuitLexer.t_GATE from
```python
        r"""
[...]

PR #738, opened 2026-05-07
Resolve round-trip serializability issues with ModelMember labels and Circuit line labels

rileyjmurray:
This PR resolves issue #495, which was originally opened to address bugs of the form
```python
from pygsti.circuit import Circuit
c = Circuit([("Gxpi2", "A0")]) # This is fine
c.str # This is also fine
Circuit(c.str) # Errors with ValueError: invalid literal for int() with base 10: ''
```
The resolution is to have the constructor invocation above raise an error. The error is triggered whenever a Circuit object's line_labels field can't be round-tripped through pygsti.circuits.circuitparser.parse_label.

This PR also modifies the two implementations of parse_label (fast and slow) to allow labels where ...
 * the first character is a lowercase letter, or one of the letters Q, T, L, or a digit, or an underscore.
 * all remaining characters are lowercase letters, or numbers, or underscores.

The ability to start line labels with T or L is new. I chose those because they're mentioned in the StateSpace documentation.

Finally, this PR updates OrderedMemberDict to have a flag where it can check for round-trip-ability of a key provided to its `__setitem__` function. That flag is set to True for four members of ExplicitOpModel: preps, povms, instruments, and operations. This makes it impossible to set a ModelMember label that can't be round-trip serialized. In the process I found some code that used invalid labels (linear GST and one instrument test) and changed the labels to be valid.

### DIFF
--- a/pygsti/circuits/circuitparser/fastcircuitparser.pyx
+++ b/pygsti/circuits/circuitparser/fastcircuitparser.pyx
@@ -259,19 +259,35 @@ cdef get_next_simple_lbl(unicode s, INT start, INT end, bool integerize_sslbls, 
     sslbls = []
     while i < end and s[i] == u':':
         i += 1
-        last = i; is_int = True
+        last = i
+        is_int = True
         while i < end:
             c = s[i]
             if u'0' <= c <= u'9':
                 i += 1
-            elif u'a' <= c <= u'z' or c == u'_' or c == u'Q':
-                i += 1; is_int = False
+            elif u'a' <= c <= u'z' or c == u'_':
+                # Labels can contain any combination of lowercase letters and underscores.
+                i += 1
+                is_int = False
+            elif c in (u'G', u'M', u'I', u'S'):
+                # These reserved characters (all uppercase letters) indicate that we've already
+                # seen everything there is to see for the most recent/current label.
+                break
+            elif last == i and c in (u'Q', u'T', u'L'):
+                # Labels can start with reserved uppercase letters Q, T, and L, per the 
+                # StateSpace documentation.
+                i += 1
+                is_int = False
+            elif last == i and u'A' <= c <= u'Z':
+                msg = f"Invalid target label: {s[last:i + 1]!r}. Labels can't start with uppercase letters other than Q, T, and L."
+                raise ValueError(msg)
             else:
                 break
         if integerize_sslbls and is_int:
-            sslbls.append(int(s[last:i])); last = i
+            sslbls.append(int(s[last:i]))
         else:
-            sslbls.append(sys.intern(s[last:i])); last = i
+            sslbls.append(sys.intern(s[last:i]))
+        last = i
 
     if i < end and s[i] == u'!':
         i += 1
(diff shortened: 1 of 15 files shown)


rileyjmurray:
@coreyostrove and @sserita I'll want to merge this as soon as tests pass.

pygsti/circuits/circuitparser/fastcircuitparser.pyx
```diff
@@ -259,19 +259,35 @@ cdef get_next_simple_lbl(unicode s, INT start, INT end, bool integerize_sslbls,
...
             c = s[i]
             if u'0' <= c <= u'9':
                 i += 1
-            elif u'a' <= c <= u'z' or c == u'_' or c == u'Q':
-                i += 1; is_int = False
+            elif u'a' <= c <= u'z' or c == u'_':
+                # Labels can contain any combination of lowercase letters and underscores.
+                i += 1
+                is_int = False
+            elif c in (u'G', u'M', u'I', u'S'):
```
coreyostrove, line 272: What is "S" reserved for?

pygsti/circuits/circuitparser/slowcircuitparser.py
```diff
@@ -194,23 +195,37 @@ def _get_next_simple_lbl(s, start, end, integerize_sslbls, segment):
...
-            try:
-                val = int(s[last:i])
-            except:
-                val = s[last:i]
-            sslbls.append(val); last = i
+        if integerize_sslbls and is_int:
+            sslbls.append(int(s[last:i]))
         else:
-            sslbls.append(s[last:i]); last = i
+            sslbls.append(sys.intern(s[last:i]))
```
coreyostrove, line 227: Can you explain this line?

pygsti/models/explicitmodel.py
```diff
@@ -143,7 +144,9 @@ def flagfn(typ): return {'auto_embed': True, 'match_parent_statespace': True,
         self.povms = _OrderedMemberDict(self, default_povm_type, povm_prefix, flagfn("povm"))
         self.operations = _OrderedMemberDict(self, default_gate_type, gate_prefix, flagfn("operation"))
         self.instruments = _OrderedMemberDict(self, default_instrument_type, instrument_prefix, flagfn("instrument"))
-        self.factories = _OrderedMemberDict(self, default_gate_type, gate_prefix, flagfn("factory"))
+        self.factories = _OrderedMemberDict(self, default_gate_type, gate_prefix,
+                                            {'auto_embed': True, 'match_parent_statespace': True,
```
coreyostrove, line 148: This change looks disjoint from those described in the PR. What was the motivation for this change?

pygsti/circuits/circuitparser/fastcircuitparser.pyx
rileyjmurray: I forget if I added this or if I got it from elsewhere in pyGSTi. If I had to explain why it's there right now I'd say it's for state prep.

pygsti/circuits/circuitparser/slowcircuitparser.py
rileyjmurray: sys.intern adds a string to an internally-maintained list of common strings and then returns the string. It's useful when one string appears many times.

Issue #706, opened 2026-01-21
Meta-issue: resolving warnings triggered by test suite


rileyjmurray:
In a perfect world, people could run pyGSTi's test suite and not see a single warning. Of course, one way to do that is to suppress warnings writ-large, but that's a terrible idea. Instead, we can selectively suppress warnings when we know they're going to be triggered (e.g., we're intentionally testing a dangerous codepath) and we can resolve issues that are causing the warnings to be raised.

To do anything like that, we need to have an understanding of where warnings are coming from, and we need to be able to filter warnings based on more than string-matching in the warning message. PR #703 introduced custom UserWarning subclasses (living in pygsti.tools.exceptions) towards this end. That PR also resolved tons of warnings that were being raised, but many remain.

We should work on narrowing down the warnings that persist in our test suite. I did this for a while by using the following pytest.ini. Anyone interested in resolving warnings can copy this into their pytest.ini locally, temporarily. It'll cause warnings to trigger Exceptions (hence, test failures) except for the places where we indicate.
```
[pytest]
filterwarnings =
    error
    ignore:write_model(...) is unable to write:pygsti.tools.exceptions.pyGSTiDeprecationWarning
    ignore::pygsti.tools.exceptions.UnnamedReportWarning
    ignore::pygsti.tools.exceptions.ForwardSimulatorSuitabilityWarning
    # ^ TODO: remove the filter above; triggered everywhere in test/test_packages/report and test/test_packages/reportb.
    ignore::pygsti.tools.exceptions.OverparameterizationWarning
    ignore::pygsti.tools.exceptions.pyGSTiDeprecationWarning
    # ^ TODO: remove the filter above; triggered everywhere in /test/test_packages.
    ignore:numpy.core is deprecated and has been renamed to numpy._core:DeprecationWarning
    # ^ TODO: actually try and resolve the deprecation warning above.
    ignore::pygsti.tools.exceptions.UntouchedModelNoiseKey
    ignore::pygsti.tools.exceptions.UnknownGaugeSpaceDimension
    # ^ TODO: remove the filter above; triggered in test/unit/tools/test_leakage_gst_pipeline.py.
    ignore::pygsti.tools.exceptions.PrepareThyself
    ignore:Degrees of freedom <= 0 for slice:RuntimeWarning
    # ^ TODO: remove the filter above; triggered in test_rb.py.
    ignore:invalid value encountered in divide:RuntimeWarning
[...]

PR #739, opened 2026-05-08
Reduce warnings generated by test suite

rileyjmurray:
This gets the number of warnings down from over 500 to below 40. About 80 tests are skipped in my local environment. These likely trigger additional warnings if run.

Resolves the warning-burn-down half of meta-issue #706. This branch:

  - Flips `pytest.ini` from permissive to strict — 8 custom warning classes plus `scipy.optimize.OptimizeWarning` now raise as errors. Only two legacy string-matched ignores remain.
  - Migrates test callers off the deprecated APIs that those warnings flagged. Those deprecated functions include `find_sufficient_fiducial_pairs`, `create_standard_report`, `load_model`/`write_model`, `DataSet.save`/`load`, `make_lsgst_structs`, `stdmodule_to_smqmodule`, and `gaugeopt_to_target` positional kwargs, among other things.
  - Adds three new exception classes (`QiskitInteropWarning`, `ProbabilityClippingWarning`, `ForwardSimDiagnosticWarning`) and classifies the ~20+ bare `warnings.warn(...)` sites that fit them.
  - Pulls a few orthogonal cleanups along the way (see "Reviewer attention" below) because they were prerequisites for getting the suite quiet.

  ## What changed

  **Strict-warnings infrastructure (`pygsti/tools/exceptions.py`, `pytest.ini`)**
  - `pygsti.tools.exceptions`: add `QiskitInteropWarning`, `ProbabilityClippingWarning` (subclasses `NumericalDomainWarning`), `ForwardSimDiagnosticWarning` (with module-level `enabled = False` opt-in flag).
  - `pygsti.tools.exceptions`: **remove** `ForwardSimulatorSuitabilityWarning` — obsoleted by the `ForwardSimulator.product` refactor below.
  - `pytest.ini`: 8 warning classes flipped to `error`; 6 legacy ignores removed; only two pre-existing ignores retained (one for the evaluator-tree diagnostic, one for the `invalid escape sequence` deprecation).

  **Source-side changes**
 - Duplicate implementations of `{Matrix/Map}ForwardSimulator.product` have been moved to `ForwardSimulator.product`.
  - **`ForwardSimulator.copy(keep_model_attached=True)`** — new flag; default preserves old behavior. Passing `False` breaks the back-reference to the parent model so the test suite stops emitting `StolenResourceWarning`.
  - **`pygsti.tools.optools.relaxed_scalar_tolerance(...)`** — new context manager for temporarily loosening the scalar tolerance constant used by fidelity/Choi-trace checks. Used in `pygsti/report/reportables.py` and a handful of report-generating tests where gauge-optimization drift legitimately pushes us 1–2 orders of magnitude beyond the default tolerance.
  - Closed two file-handle leaks: binary `DataSet`/`MultiDataSet` I/O paths now close their files on the exception path (`pygsti/data/{dataset,multidataset}.py`).

  **Test-side changes**
  - ~20 commits with `fix(C2):` prefix migrating callers off the deprecated APIs named above. Where the migration was mechanical, the test calls the replacement directly. Where migration would have lost coverage of the deprecated API itself, the call is wrapped in `pytest.warns(...)` and a note was added to `issues/706/` explaining why.
  - `test/test_packages/testutils/cloudnoise.py` (new file): consolidates three duplicate `build_XYCNOT_cloudnoise_model` helpers that lived in three separate test files.
  - `test/unit/objects/test_subcirc_selection.py`: fixed a pre-existing duplicate-name bug (two `test_simple_subcirc_selection_full_connectivity` collided into one), renamed clearly, and added qiskit-side coverage.

  ## Known follow-ups (deferred — not blocking this PR)

  - `_create_xycnot_cloudnoise_circuits` is deprecated but used in tests. The suggested replacement (`create_cloudnoise_circuits`) takes a differently-shaped input and the test fixtures are frozen JSON files. Three test callers are wrapped in `pytest.warns(...)` pending a decision on whether the legacy helper truly goes away.
  - `find_sufficient_fiducial_pairs` (global FPR) is deprecated. It's a genuinely different algorithm from the per-germ replacement. Four test callers wrapped with `pytest.warns(...)` pending a decision on whether to delete or keep-deprecated.
  - `load_model` / `write_model` are deprecated. These functions can use different persistence formats. The `.txt` format appears is tests and has been wrapped with `pytest.warns(...)` . Tests that only need *some* model serialization have been migrated to JSON. The `write_model` call fires _two_ deprecation warnings. (Excessive.)

  ## Reviewer attention

  1. **`ForwardSimulatorSuitabilityWarning` was removed** from `pygsti/tools/exceptions.py`. Anyone importing it directly will hit an `ImportError`. That's okay because that warning never made it into a release. The warning is obsoleted by the `circuit_operator` refactor (we no longer need to nudge users between sim flavors).
  2. **`ForwardSimulator.copy` gained `keep_model_attached`** (commit `cb5e86579`). Default is backwards-compatible; opt-in `False` is what the test suite now uses to avoid `StolenResourceWarning`.


docs/markdown/simulation/ErrorGeneratorPropagation.md
```diff
@@ -223,7 +223,7 @@ from pygsti.baseobjs import Label
 def pauli_expectation_exact(error_propagator, target_model, circuit, pauli):
     #get the eoc error channel, and the process matrix for the ideal circuit:
     eoc_channel = error_propagator.eoc_error_channel(circuit, include_spam=True)
-    ideal_channel = target_model.sim.product(circuit)
+    ideal_channel = target_model.circuit_operator(circuit)
```
rileyjmurray, line 226: This needs to be reverted.

PR #748, opened 2026-06-01
Helper to convert ImplicitOpModel to ExplicitOpModel

rileyjmurray:
Adds the instance method `ImplicitOpModel.to_explicit_model`. This is used to deep-copy an ImplicitOpModel into an equivalent ExplicitModel.

Test check correctness for `LocalNoiseModel` objects created with the help of the `create_crosstalk_free_model` free-function.


rileyjmurray:
@coreyostrove this is ready for review

pygsti/models/implicitmodel.py
```diff
@@ -201,6 +201,65 @@ def compute_clifford_symplectic_reps(self, oplabel_filter=None):
...
+        # preserves parameter-sharing between primitives (e.g. one base gate
+        # embedded on multiple qubits): a sub-member encountered a second time is
+        # returned from the memo instead of being copied again.
+        memo = {}
+        for lbl, prep in self.prep_blks.get('layers', {}).items():
+            ret.preps[lbl] = prep.copy(ret, memo)
+        for lbl, povm in self.povm_blks.get('layers', {}).items():
+            ret.povms[lbl] = povm.copy(ret, memo)
+        for lbl, op in self.operation_blks.get('layers', {}).items():
+            ret.operations[lbl] = op.copy(ret, memo)
```
nkoskelo, line 257: What if we have a layer that has not yet been built from the set of `gate` building blocks?

rileyjmurray: I think there's nothing we can do about that. It's a limitation of the conversion to an ExplicitOpModel. 

If we wanted to reduce this limitation we could have a helper function that populated an ImplicitOpModel's `operation_blks['layers']` field with all possible layers given its operation model members.

nkoskelo: Can we have a flag which allows the user to toggle this behavior? I think it should be just the `layers` by default but if the user sets the flag then we can rebuild all of the layers, that do not exist in `layers`, from combinations of `gates`.

pygsti/models/implicitmodel.py
```diff
@@ -201,6 +201,65 @@ def compute_clifford_symplectic_reps(self, oplabel_filter=None):
...
+        """
+        from pygsti.models.explicitmodel import ExplicitOpModel as _ExplicitOpModel
+
+        factory_labels = ['%s:%s' % (blk_lbl, lbl) for blk_lbl, fdict in self.factories.items()
+                          for lbl in fdict]
+        if factory_labels:
+            raise ValueError(("Cannot convert an implicit model with op-factories to an explicit"
+                              " model: factories generate operations on-demand and have no fixed"
+                              " representation.  Offending factories: %s")
+                             % ", ".join(factory_labels))
```
coreyostrove, line 241: I am not sure why it is not possible to convert to an explicit model when there are factories. Can we not copy factories['layers'] into the explicit model's factories attribute?

pygsti/models/implicitmodel.py
coreyostrove: I like the simplicity of having the explicit model constructed snapshot the implicit model's cached layers. Adding this option without a clear user demand signal would add more clutter and maintenance overhead. There are other straightforward constructor codepaths for users who want an explicit model that has every combination of gates.

coreyostrove (changes requested):
One change requested regarding the factory handling.

pygsti/models/implicitmodel.py
nkoskelo: It could still be stale, since `model.factories` depends on the `LayerRules` we use. `CloudNoiseLayerRules` needs to check both `model.factories['layers']` and `model.factories['cloudnoise']` while the LocalNoiseModel typical LayerRule checks only `model.factories['layers']`. Now we could override this method for different ImplicitOpModels.

Note that ExplicitOpModels can actually generate gates with factories. If we just copy the factories over it may still feel like an `ImplicitOpModel` but it would technically be allowed. Currently ExplicitOpModels only check `model.factories[LayerLabel]` to get the factory. That factory may need to be updated depending on the original model.

nkoskelo: See localnoisemodel.py:531 and cloudnoisemodel.py:534 to see that the LayerRules for the LocalNoiseModel and the CloudNoiseModel build the operation using the factories at the individual factor level of each layer as opposed to the whole layer which the ExplicitLayerRules does.

coreyostrove: Ok, happy to defer to your assessment here, @nkoskelo. If we need factory conversions we can add that when requested.

coreyostrove (approved):
Discussion above on comment thread with @nkoskelo addressed my concern. Approving.

PR #750, opened 2026-06-03
Characterization tests for LindbladCoefficientBlock; small unrelated bugfixes

### LINKED ISSUE #751: `Polynomial(coeffs)` silently drops like monomials
I've been working on characterization tests for LindbladCoefficientBlock. In the process I found an issue in the Polynomial class (see pygsti.baseobjs.polynomials).

The issue is related to how `Polynomial(coeffs)` processes a dict of coeffs. While `coeffs` is a dict mapping `tuple[int,...]` to scalars, its keys _are not_ exponent vectors for the monomials. Instead, for any `n`, it associates the tuple `n * (0,)` with the degree-n univariate monomial $x_0^n$. What matters is the *number of times* a given integer appears in the tuple. So, for example, it associates `(0,1)` *and* `(1,0)` with the monomial $x_0 x_1$.

Example:
```python
from pygsti.baseobjs.polynomials import Polynomial

p = Polynomial({(0,1): 5})
q = Polynomial({(1,0): 5})

print(f'p = {str(p)}')  # '5.000x0x1'
print(f'q = {str(q)}')  # '5.000x0x1'
assert p.coeffs == q.coeffs  # passes
```

This creates problems when we pass the Polynomial constructor a dict where more than one key is associated with the same monomial. The existing behavior just clobbers one of the key-value pairs. Which key-value pair "wins" depends one's environment. The following snippet *should* show the issue on any machine.
```python
d1 = {(0,1): 5, (1,0): -5}
d2 = dict(reversed(d1.items()))

r = Polynomial(d1)
s = Polynomial(d2)

assert d1 == d2  # passes
print(f'r = {str(r)}')   # '-5.000x0x1', on my local machine
print(f's = {str(s)}')  # '5.000x0x1', on my local machine
assert r.coeffs == s.coeffs  # fails
```

There are a few ways forward. One option is to error if `Polynomial.__init__` is passed a dict whose keys get mapped to the same monomial. I think it's cleaner to just sum the provided coefficients associated to a given polynomial. This seems to be the behavior that LindbladCoefficientBlock expected, anyway.

NOTE: You're going to have a bad time if you want to find the part of the Polynomial constructor code where key-clobbering happens. The issue _does not_ arise in the `_vinds_to_int` helper function in `polynomial.py`. That will return distinct integers for tuples like `(0,1)` or `(1,0)`. The issue is in the C++ code that's wrapped with the Cython extension. Specifically, the issue is somewhere in here
https://github.com/sandialabs/pyGSTi/blob/b1195cba458c0067114491626dc8ee414981cdc7/pygsti/evotypes/basecreps.cpp#L148-L186
If I were a betting man, I'd say you can attribute the clobbering to the std::sort call on line 184.

### LINKED ISSUE #752: Report generation fails with plotly 6.8
The leakage GST integration test fails at report generation time when plotly 6.8 is installed. I first observed this on a CI run with unrelated changes. Verified locally on my machine. 

Example failing output is as follows. Note the assertion error at the end. Will follow up with proposed fix when I have one.
```
________________________________________________________________ TestLeakageGSTPipeline.test_pipeline_1Q_XYI _________________________________________________________________

self = <test_leakage_gst.TestLeakageGSTPipeline testMethod=test_pipeline_1Q_XYI>

    @pytest.mark.filterwarnings("ignore::pygsti.tools.exceptions.OverparameterizationWarning")
    def test_pipeline_1Q_XYI(self):
        # This is adapted from the Leakage-automagic ipython notebook.
        mp = smq1Q_XYI
        ed = mp.create_gst_experiment_design(max_max_length=4)
        # ^ The max length is small so we don't have to wait as long for the GST fit.
        #
        #       GST at L=2 has 168 data params and 161 effective model params;
        #       this is barely outside the overparameterized regime. GST at L=4
        #       (which we use here) has 285 data params and 160 model params.
        #
        tm3 = leaky_qubit_model_from_pspec(mp.processor_spec())
        # ^ Target model.
        dgm3, _ = with_leaky_gate(tm3, ('Gxpi2', 0), strength=0.125)
        # ^ Data generating model.
        num_samples = 100_000
        # ^ The number of samples is large to compensate for short circuit length.
        from pygsti.objectivefns import objectivefns
        OLD_MIN_PROB_CLIP = objectivefns.DEFAULT_MIN_PROB_CLIP  # restore before exit
        OLD_RADIUS        = objectivefns.DEFAULT_RADIUS         # restore before exit
        objectivefns.DEFAULT_MIN_PROB_CLIP = objectivefns.DEFAULT_RADIUS = 1e-12
        # ^ The lines above change numerical thresholding rules in objective evaluation
        #   to be appropriate when the number of shots/circuit is extremely large.
        ds = simulate_data(dgm3, ed.all_circuits_needing_data, num_samples=num_samples, seed=1997)
        gst = StandardGST(
            modes=('CPTPLND',), target_model=tm3, verbosity=4,
            badfit_options={'actions': ['wildcard1d'], 'threshold': 0.0}
        )
        pd = ProtocolData(ed, ds)
        res = gst.run(pd)
        kwargs_stdreport = {
            'advanced_options' : {
                'show_unmodeled_error' : True,
                'skip_sections'        : ['colorbox', 'input', 'meta', 'help']
            },
        }
        report, updated_res = construct_leakage_report(res, title='easy leakage analysis!', kwargs_stdreport=kwargs_stdreport)
        # ^ updated_res contains the leakage-aware gauge optimization model.
        with TemporaryDirectory() as dirname:
>           report.write_html(dirname)

test/integration/test_leakage_gst.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pygsti/report/report.py:192: in write_html
    _merge.merge_jinja_template_dir(
pygsti/report/merge_helpers.py:708: in merge_jinja_template_dir
    outfile.write(template.render(render_params))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/miniconda3/envs/pgdev313/lib/python3.13/site-packages/jinja2/environment.py:1295: in render
    self.environment.handle_exception()
/opt/miniconda3/envs/pgdev313/lib/python3.13/site-packages/jinja2/environment.py:942: in handle_exception
    raise rewrite_traceback_stack(source=source)
pygsti/report/templates/standard_html_report/main.html:154: in top-level template code
    {% include 'tabs/Summary.html' %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
pygsti/report/templates/standard_html_report/tabs/Summary.html:6: in top-level template code
    {{ final_fits_comparison_plot|render }}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
pygsti/report/merge_helpers.py:463: in render
    html = _render_as_html(value, render_options or {}, link_to)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pygsti/report/merge_helpers.py:305: in _render_as_html
    out = value.render('html')
          ^^^^^^^^^^^^^^^^^^^^
pygsti/report/workspace.py:2545: in render
    fig_dict = _plotly_ex.plot_ex(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

figure_or_data = Figure({
    'data': [{'hoverinfo': 'text',
              'marker': {'color': ['lightgreen']},
              'text': [...28153664],
                         'title': {'text': 'N<sub>sigma</sub>'},
                         'type': 'log'}}
})
show_link = False, link_text = 'Export to plot.ly', validate = False, resizable = True, lock_aspect_ratio = True, master = True, click_to_display = False, link_to = None
link_to_id = 'plot_796670_0', rel_figure_dir = None

    def plot_ex(figure_or_data, show_link=True, link_text='Export to plot.ly',
                validate=True, resizable=False, lock_aspect_ratio=False,
                master=True, click_to_display=False, link_to=None, link_to_id=False,
                rel_figure_dir="figures"):
        """
        Create a pyGSTi plotly graph locally, returning HTML & JS separately.
    
        Parameters
        ----------
        figure_or_data : plotly.graph_objs.Figure or .Data or dict or list
            object that describes a Plotly graph. See https://plot.ly/python/
            for examples of graph descriptions.
    
        show_link : bool, optional
            display a link in the bottom-right corner of
    
        link_text : str, optional
            the text of export link
    
        validate : bool, optional
            validate that all of the keys in the figure are valid? omit if you
            need to include extra, unnecessary keys in your figure.
    
        resizable : bool, optional
            Make the plot resizable by including a "resize" event handler and
            any additional initialization.
    
        lock_aspect_ratio : bool, optional
            Whether the aspect ratio of the plot should be allowed to change
            when it is sized based on it's container.
    
        master : bool, optional
            Whether this plot represents the "master" of a group of plots,
            all of the others which are "slaves".  The relative sizing of the
            master of a group will determine the relative sizing of the slaves,
            rather than the slave's containing element.  Useful for preserving the
            size of the features in a group of plots that may be different overall
            sizes.
    
        click_to_display : bool, optional
            Whether the plot should be rendered immediately or whether a "click"
            icon should be shown instead, which must be clicked on to render the
            plot.
    
        link_to : None or tuple of {"pdf", "pkl"}
            If not-None, the types of pre-rendered/computed versions of this plot
            that can be assumed to be present, and therefore linked to by additional
            items in the hover-over menu of the plotly plot.
    
        link_to_id : str, optional
            The base name (without extension) of the ".pdf" or ".pkl" files that are
            to be linked to by menu items.  For example, if `link_to` equals
            `("pdf",)` and `link_to_id` equals "plot1234", then a menu item linking
            to the file "plot1234.pdf" will be added to the renderd plot.
    
        rel_figure_dir : str, optional
            A relative path from the "current" path (the path of the generated
            html documents) to figure files.  Usually something like `"figures"`.
    
        Returns
        -------
        dict
            With 'html' and 'js' keys separately specifying the HTML and javascript
            needed to embed the plot.
        """
        from plotly import __version__ as _plotly_version
        from plotly import tools as _plotlytools
    
        #Processing to enable automatic-resizing & aspect ratio locking
        fig = _plotlytools.return_figure_from_figure_or_data(
            figure_or_data, False)
        orig_width = fig.get('layout', {}).get('width', None)
        orig_height = fig.get('layout', {}).get('height', None)
    
        if lock_aspect_ratio and orig_width and orig_height:
            aspect_ratio = orig_width / orig_height
        else: aspect_ratio = None
    
        #Special polar plot case - see below - add dummy width & height so
        # we can find/replace them with variables in generated javascript.
        if 'angularaxis' in fig['layout']:
            fig['layout']['width'] = 123
            fig['layout']['height'] = 123
    
        config = {}
        config['showLink'] = show_link
        config['linkText'] = link_text
    
        try:
            from plotly.io import to_html as _to_html
        except ImportError:
            print('Unable to import plotly.io.to_html. This likely means your version of plotly is incompatible (<3.8.0), please upgrade and try again.')
        import uuid as _uuid
        plot_html = _to_html(fig, config, auto_play=False, include_plotlyjs=False,
                                include_mathjax=False, post_script=None, full_html=False,
                                animation_opts=None, validate=validate)
>       assert(plot_html.startswith("<div>") and plot_html.endswith("</div>"))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError

pygsti/report/plotly_plot_ex.py:122: AssertionError
```

rileyjmurray:
This PR adds a dedicated unit-test module, `test/unit/modelmembers/test_lindbladcoefficients.py`, that pins the input/output behavior of LindbladCoefficientBlock across the full `(block_type, param_mode)` range of configurations.

Two small bugfixes:
 - *Tangentially related.* In the process I uncovered the bug in the Polynomial class described in [#751](#751). One of the characterization tests in this PR requires the correction to the `Polynomial.coeffs` property implementation. So this PR also resolves #751.
 - *Unrelated*. CI tests failed with plotly 6.8. I updated the relevant function on our end to be robust to that change in plotly. So this PR resolves #752.

## Motivation

LindbladCoefficientBlock is the object that parameterizes the Hamiltonian, Pauli-stochastic, and Pauli-correlation/active ("other") blocks of a LindbladErrorgen, and is therefore central to CPTP-friendly noise modeling (`CPTPLND`, `GLND`, `H+S`, …). Despite that, it had no dedicated unit-test coverage: it was only exercised incidentally, as a building block for the error-generator construction in `test/unit/tools/test_optools.py`.

The class is also intricate — three block types, five parameterization modes (with a non-trivial validity matrix between them), Cholesky/depolarizing reparameterizations, analytic first/second derivatives, an elementary-error-generator interface, and Term-simulator polynomial construction. That combination (high importance, high internal complexity, no targeted tests) makes it a good candidate for a characterization suite: a regression safety net that locks in today's behavior so that any future change — refactoring or otherwise — surfaces as an explicit test failure rather than a silent regression.

## What's covered

The suite parameterizes over `block_type ∈ {ham, other_diagonal, other}`, the valid `param_mode`s for each, and three bases — `('pp', 4)` (qubit), `('gm', 9)` (qutrit), and `('pp', 16)` (two qubits) — so the relationships below are checked on 1-qudit, qutrit, and 2-qubit instances.

| Area | What is pinned |
|---|---|
| **Parameter count & round-trips** | `num_params` formula per `(block_type, param_mode)`; `to_vector`/`from_vector` round-trip, anchored on `block_data`. |
| **Analytic derivatives** | `deriv_wrt_params` vs. central finite differences of `block_data`; `superop_deriv_wrt_params` and `superop_hessian_wrt_params` vs. finite differences; the linearity / chain-rule identities relating them to `create_lindblad_term_superoperators`; the elementary-error-generator Jacobian; the matrix-structured (`superops_are_flat=False`) path; zero-parameter (`static`) derivative shapes. |
| **Elementary error generators** | Values (including the tricky C/A non-Hermitian combinations for the `other` block); round-trip through `set_elementary_errorgens`; `on_missing ∈ {ignore, warn, raise}`. |
| **Labels** | `coefficient_labels` / `param_labels` snapshots, including the `n×n` `other` case. |
| **Serialization & conversion** | `NicelySerializable` round-trip; `convert` across param modes preserving `block_data`; `is_similar`. |
| **Term-simulator construction** | A platform-independent snapshot of the polynomial coefficients produced by `create_lindblad_term_objects` (1-qubit). |
| **Edge cases & validation** | Near-singular `other`/`cholesky` truncation; rejection of an invalid `block_type` and of a `param_mode` invalid for its `block_type`. |

## Design notes

- **Anchored on `block_data`.** Parameter vectors are *not* unique under `cholesky` and `depol`/`reldepol` (multiple parameter assignments map to the same coefficients), so round-trips and derivative comparisons are anchored on `block_data`, which is the unique invariant.
- **Platform-independent term snapshot.** The Term-simulator test pins each term's polynomial coefficient *evaluated at a fixed generic parameter point*, compared as a sorted multiset with a tolerance — rather than a hash of the serialized polynomial. Evaluated values are deterministic IEEE arithmetic (identical across Linux/Windows to ~1e-15), whereas the serialized form (packed monomial keys, float `repr`, `+0.0` vs. `-0.0`) is platform-dependent and previously caused spurious cross-platform failures.
- **Conditioning of the `other`/`cholesky` path.** For `other`/`cholesky`, the parameters (a Cholesky factor) are ill-conditioned near rank-deficiency even though `block_data` still round-trips cleanly. The finite-difference tests therefore use well-conditioned positive-definite data, and the rank-deficient/truncation path is exercised by a dedicated test.

## Notes for reviewers

These are characterization tests, not a specification. Where they encode a quirk of the current implementation, that is intentional — the point is to detect *changes*. A few behaviors are worth calling out explicitly because they are surprising and are pinned deliberately:

- **`param_mode` validation is lazy.** A `param_mode` that is invalid for a given `block_type` is *not* rejected at construction; `InvalidParamModeError` is raised on first use (e.g. on the `num_params` property). An unrecognized `block_type`, by contrast, *is* rejected at construction (with a `ValueError`).
- **`convert` does not re-validate feasibility.** `convert(param_mode)` changes the parameterization without re-checking that the `(block_type, param_mode)` pair is feasible.
- **`to_vector` aliasing.** In `elements` mode, `to_vector()` returns a *view* of `block_data`; the derivative tests `.copy()` before perturbing to avoid mutating the block underneath the finite-difference loop.
- **`other`-block superop derivative reads cached state.** `superop_deriv_wrt_params` for the `other` block reads its Cholesky factor from an internal cache populated by a prior `from_vector`, so the relevant tests call `from_vector` first.

## Testing

```bash
python -m pytest test/unit/modelmembers/test_lindbladcoefficients.py
# 247 passed in ~0.3s
```


coreyostrove:
The fix for the monomial collision issue here is not a complete one, for a more complete fix see https://github.com/sandialabs/pyGSTi/commit/33ea9e604a82cb47274410eabe759e82ada7336f. I'd recommend reverting this change and cherry picking in this one if it is important that this is working on develop right meow.

rileyjmurray:
Thanks for the pointer. I'm going to keep this change as-is for now. I'll open a PR cherry-picking your commit in the (very) near future.

PR #755, opened 2026-06-08
Refactoring LindbladCoefficientBlock: parameterization by composition

rileyjmurray:
I've split the `LindbladCoefficientBlock` changes from the Issue 607 work into two PRs: this one is the structural refactor, and a follow-up adds the Torchable (PyTorch forward-sim) support on top of it.

The refactor reorganizes the class so that each parameterization mode is a first-class object instead of a `param_mode` string switched on throughout the body. Behavior is pinned by the characterization tests added in #750; that suite is why I wrote those tests first, and it still passes unchanged here apart from a single deliberate update.

## What changed

Everything is in `pygsti/modelmembers/operations/lindbladcoefficients.py`:

- **Parameterization is now a strategy object.** Each `param_mode` (`static`, unconstrained `elements`, `cholesky`, `depol`/`reldepol`) is a `_BlockParameterization` subclass that owns the `params ↔ block_data` map, its Jacobian, the parameter labels, and the superop Hessian. A block holds one instance in `self._parameterization` and delegates all `param_mode`-dependent numeric work to it. Adding a parameterization is now "write one subclass and register it."
- **One subclass per block type.** `_HamCoeffBlock`, `_OtherDiagonalCoeffBlock`, and `_OtherCoeffBlock` each declare the block-type-specific pieces (the valid `param_mode`s via a `_PARAM_CLASSES` map, term-generator construction, coefficient shape/dtype). `LindbladCoefficientBlock.__new__` dispatches on `block_type` to the right subclass, so `LindbladCoefficientBlock('ham', ...)` still returns the right object — construction is unchanged for callers.
- **On-disk format is unchanged.** Instances still serialize under the base `LindbladCoefficientBlock` name; the concrete subclass is reselected from `block_type` on load via the same `__new__` dispatch. Objects round-trip across the refactor, so existing serialized blocks load as-is.

## Notes for reviewers

- **One intentional behavior change.** An invalid `(block_type, param_mode)` pair is now rejected at construction (`InvalidParamModeError`) rather than lazily on first use of `num_params`. #750 pinned the old lazy behavior as a deliberately-documented quirk; this PR flips it and updates that one test to match. (An invalid `block_type` is likewise rejected at construction, now via a dedicated `InvalidBlockTypeError` — still a `ValueError` subclass, so `except ValueError` callers are unaffected.)
- **The diff is large but mostly mechanical.** ~1500 lines move, but the NumPy code paths (`params_to_block_data`, `block_data_jacobian`, the analytic derivatives, term construction) are the same computations relocated into the strategy or subclass that owns them. The characterization suite is what gives confidence that nothing changed numerically.

## Testing

```bash
python -m pytest test/unit/modelmembers/test_lindbladcoefficients.py
# 247 passed

python -m pytest test/unit/modelmembers
# 673 passed, 3 skipped
```


rileyjmurray:
@coreyostrove I've merged in functionality for your unconstrained black-box coefficient block.

coreyostrove:
Ok, I'll give this PR a review tomorrow.

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -92,6 +92,413 @@ def __init__(self, *args: object) -> None:
...
+    def num_params(self, blk):
+        return 1
+
+    def params_to_block_data(self, blk, v):
+        blk.block_data[:] = v.item() * _np.ones(len(blk._bel_labels), 'd')
+
+    def block_data_to_params(self, blk, ttol):
+        first = blk.block_data[0]
+        all_equal = [_np.isclose(val, first, atol=1e-6) for val in blk.block_data]
+        assert all(all_equal), f"Diagonal lindblad coefficients are not equal! (tol={ttol})"
```
coreyostrove, line 290: This list comprehension can be replaced with a numpy one-liner: 
`np.allclose(blk.block_data, first, atol=1e-6)`

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -92,6 +92,413 @@ def __init__(self, *args: object) -> None:
...
+
+    def params_to_block_data(self, blk, v):
+        blk.block_data[:] = (v.item()**2) * _np.ones(len(blk._bel_labels), 'd')
+
+    def block_data_to_params(self, blk, ttol):
+        nonneg = [val >= ttol for val in blk.block_data]
+        assert all(nonneg), f"Lindblad stochastic coefficients are not positive! (tol={ttol})"
+        first = blk.block_data[0]
+        all_equal = [_np.isclose(val, first, atol=1e-6) for val in blk.block_data]
+        assert all(all_equal), f"Diagonal lindblad coefficients are not equal! (tol={ttol})"
```
coreyostrove, line 259: This list comprehension can be replaced with a numpy one-liner:
np.allclose(blk.block_data, first, atol=1e-6)

pygsti/modelmembers/operations/lindbladerrorgen.py
```diff
@@ -1619,6 +1625,10 @@ def cast(cls, obj):
                 block_types = ['ham', 'other']; param_modes = ['elements', 'cholesky']
             elif abbrev == "GLND":
                 block_types = ['ham', 'other']; param_modes = ['elements', 'elements']
+            elif abbrev == "GLNDU":
+                # like GLND, but the non-Hamiltonian block uses the flat, per-elementary-errorgen
+                # 'other_unconstrained' representation (supports reduced/flexible parameterizations).
+                block_types = ['ham', 'other_unconstrained']; param_modes = ['elements', 'elements']
```
coreyostrove, line 1651: We can make this change in a follow-up PR once we've played around with this, but my preference would be to have this block composition replace that of "GLND".

pygsti/tools/lindbladtools.py
```diff
@@ -449,7 +458,7 @@ def create_elementary_errorgen_pauli(typ, p, q=None, sparse=False):
             for i in range(d): 
                 for j in range(d):
                     rho1 = p[:,i].reshape((d,1))@p[j,:].reshape((1,d))
-                    rho1[i,j] += -1
+                    rho1[i,j] += -rho_scale
```
coreyostrove: -=

coreyostrove (approved):
A few non-blocking changes and suggestion. Approving now, and you can decide when/how/if to address these. Thank you very much for integrating the changes from my other branch with this one, @rileyjmurray.

pygsti/tools/lindbladtools.py
nkoskelo: There may also be a faster way to do the outer product that does not do the reshape operation.

pygsti/modelmembers/operations/lindbladcoefficients.py
nkoskelo, line 1: There is an unused local variable "num_bels" on line 915.

pygsti/tools/lindbladtools.py
coreyostrove: Agreed, I left a note to myself about this as a TODO in the code a while back. But the idea I had is that you actually shouldn't need to do any vector multiplication at all if we take advantage of the fact that the paulis are generalized permutation matrices. I.e. we should be able to just read off the location of the nonzero entries and assign them directly.

This all is out-of-scope for the current PR, though.

nkoskelo: Most of the constructors it seems could be refactored to use einsums.

rileyjmurray:
Test failures are unrelated to this PR. Numpy 2.5 is raising deprecation warnings when we try to set array shapes directly. I'm going to open a new PR resolving those deprecation warnings momentarily.

pygsti/modelmembers/operations/lindbladcoefficients.py
coreyostrove: 258 and 259 can still be condensed down one more level using allclose, fwiw.

PR #765, opened 2026-06-10
Fix cataloged bugs in the Circuit class

### LINKED ISSUE #756: `Circuit.extract_labels(strict=False)` produces garbage line labels and ignores the `lines` filter
This issue brought to you by a robot, written from the robot's perspective.

Found during a systematic audit of `pygsti/circuits/circuit.py` (static analysis plus executed
repros) in preparation for a planned Circuit-internals refactor. All line numbers and outputs
are from the head of `develop` (47b3dcae5).

### Defect 1: line labels become `('a', 'u', 't', 'o')`

```python
c = Circuit("[Gx:0Gy:1][Gz:0]", line_labels=(0,1))
sub = c.extract_labels(layers=slice(0,2), lines=None, strict=False)
print(sub.line_labels)   # ('a', 'u', 't', 'o')
```

When `layers` is non-integer and `strict=False`, line 1360 assigns the *string sentinel*
`lines = "auto"`, which then flows into `Circuit._fastinit(..., tuple(lines), ...)` at
lines 1363–1365. `tuple("auto")` is `('a','u','t','o')`, and `_fastinit` performs no
validation, so the result is a structurally corrupt circuit whose line labels are the
characters of the word "auto".

### Defect 2: the `strict=False` filter is a no-op

Line 1355 reads:

```python
(not strict and len(sslbls.intersection(lines)) >= 0)
```

`len(...) >= 0` is always true, so with `strict=False` every label in the layer is included
regardless of the requested `lines`:

```python
c = Circuit("[Gx:0Gy:1]@(0,1)", line_labels=(0,1))
sub = c.extract_labels(layers=slice(0,1), lines=(0,), strict=False)
print(sub.layertup)   # (Label((Label(('Gx', 0)), Label(('Gy', 1)))),)  — Gy:1 included
```

`Gy:1` has zero overlap with `lines=(0,)` but is returned anyway. The docstring's stated
intent ("include labels that are *not* contained within the layers and lines" boundary
behavior) suggests the test should be a nonempty-intersection check.

### Notes

- The non-draft PR #735 branch (`upstream-xfgst-core`) already fixes Defect 2 (it changes the
  test to `sslbls.intersection(lines)`) and reworks the `lines=None` path to insert explicit
  idles — but it does **not** touch the `lines = "auto"` path, so Defect 1 survives there.
  Coordinating the fix with that branch would avoid conflicting semantics.
- `extract_labels` backs `Circuit.__getitem__`, so these paths are reachable from ordinary
  slicing with line filters.

### Suggested fix

Replace the `"auto"` sentinel with the actual union of sslbls retained by the extraction (or
the parent's line labels restricted to those), and change the non-strict test to
`sslbls.intersection(lines)`. Characterization tests pinning the current (buggy) behavior are
planned to land first, so the behavior change in the fix is a visible one-line test flip.

### LINKED ISSUE #760: `sort_layer_labels_inplace` is the only mutator without a static guard — calling it on a static circuit leaves stale hash/str caches
This issue brought to you by a robot, written from the robot's perspective.

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5).

Every in-place editing method on `Circuit` begins with `assert(not self._static)` — except
`sort_layer_labels_inplace` (circuit.py:1150–1162). That is by design in one sense:
`done_editing` (circuit.py:5036–5057) calls it at line 5050 *while the circuit is still
formally static-transitioning*. But the method is public, and calling it directly on a static
circuit rewrites `_labels` while leaving the eagerly-cached `_hashable_tup`/`_hash`/`_str`
stale.

### Repro

```python

# build a static circuit whose layer is NOT in canonical order (possible via _fastinit,
# which skips sorting — see the companion construction-path issue)
c = Circuit._fastinit((Label((('Gy',1),('Gx',0))),), (0,1), False)
h_before = hash(c)
c.sort_layer_labels_inplace()        # runs without error on a static circuit
print(hash(c) == h_before)           # True  — cached hash unchanged...
rebuilt = Circuit._fastinit(c._labels, (0,1), False)
print(hash(c) == hash(rebuilt))      # False — ...but disagrees with the actual labels
```

After the call, `hash(c)` (cached) and `c.tup` (recomputed from `_labels`) disagree — the
worst kind of inconsistency for an object used as a dict key throughout `DataSet.cirIndex` and
the layout caches.

### Suggested fix

Split the method: a private unguarded `_sort_layer_labels_inplace` for `done_editing`'s use,
and either remove the public name or give it the standard `assert(not self._static)` guard.
(Cross-corpus usage analysis shows no external callers of the public name; it is
INTERNAL-ONLY.) Alternatively, if it must remain callable on static circuits, recompute
`_hashable_tup`/`_hash`/`_str` after sorting.

Related: the editable/static cache discipline was recently reworked in #692; this is a
leftover hole in that discipline.

### LINKED ISSUE #762: `tensor_circuit(_inplace)`: fails on implicit-line-label gates; crashes later when `line_order` is a list
This issue brought to you by a robot, written from the robot's perspective.

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5). Note: PR #735's branch already rewrites `tensor_circuit_inplace`
(via a new `Label.concate` API), which fixes Defect 1 below — but Defect 2 survives the
rewrite. Filing so the fix has a citable home and so the two branches don't diverge silently.

### Defect 1: tensoring circuits whose gates have implicit (None) sslbls fails with a cryptic error

```python
a = Circuit("Gx@(0)")              # bare 'Gx' — sslbls is None (acts on its line implicitly)
b = Circuit("Gy@(1)")
a.tensor_circuit(b)
# ValueError: Cannot operate on a block that is straddled by Gx!
```

The simplest imaginable tensor product — one 1-qubit circuit beside another — fails. Root
cause chain: `tensor_circuit_inplace` calls `insert_labels_as_lines_inplace`
(circuit.py:2520), which reaches `_clear_labels`; there, a label with `sslbls is None` is
treated as acting on **all** lines (`sslbls = set(self._line_labels)`, circuit.py:1885), and
since the new line has already been added to the circuit, every implicit-sslbls gate
"straddles" the block and trips the ValueError at circuit.py:1887.

With explicit sslbls (`Circuit("Gx:0@(0)")`), the same operation works fine — so the failure
depends on a representation detail (`Gx` vs `Gx:0`) that users reasonably consider equivalent
and that the existing tests happen to avoid (they use explicit `:Q0`-style labels throughout).
This is the known "implicit idle / None-sslbls means all lines" ambiguity biting; arguably
refusing to tensor an all-lines gate is *correct*, but then the error should be raised
up-front in `tensor_circuit_inplace` with an actionable message, not deep in `_clear_labels`.

### Defect 2: `line_order` given as a list is stored un-tupled, crashing later

```python
a = Circuit("Gx:0@(0)", editable=True)
a.tensor_circuit_inplace(Circuit("Gy:1@(1)"), line_order=[1,0])
print(type(a._line_labels))   # list
a.done_editing()
# TypeError: can only concatenate tuple (not "list") to tuple
```

`new_line_labels = line_order` (circuit.py:2515) and the assignment at :2521 never coerce to
tuple, violating the `_line_labels`-is-a-tuple invariant; the crash surfaces later in
`.tup` (circuit.py:748) during `done_editing` — far from the cause. The signature annotation
explicitly invites lists (`Union[None, List[...], Tuple[...]]`). One-character fix:
`self._line_labels = tuple(new_line_labels)`. The PR #735 rewrite retains the same un-tupled
assignment and would benefit from the same fix.

### LINKED ISSUE #763: Callable mappers are broken in `map_names_inplace` and `map_state_space_labels_inplace`
This issue brought to you by a robot, written from the robot's perspective.

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5).

Both in-place mapping methods document accepting "a dictionary or function", but their
callable branch returns the function object itself instead of calling it:

```python
# circuit.py:2911-2912 (map_names_inplace) and :2945-2946 (map_state_space_labels_inplace)
def mapper_func(gatename): return mapper[gatename] \
    if isinstance(mapper, dict) else mapper        # <-- should be mapper(gatename)
```

### Repro

```python

c = Circuit("Gx:0Gy:1@(0,1)", editable=True)
c.map_state_space_labels_inplace(lambda l: l + 10)
# AssertionError: State space label '<function <lambda> at 0x...>' must be a string or integer!

c2 = Circuit("Gx:0@(0)", editable=True)
c2.map_names_inplace(lambda n: n.lower())
# AssertionError: `name` must be a string, but it's '<function <lambda> at 0x...>'
```

The function object is passed through as the new sslbl/name and trips Label validation (so at
least it fails loudly rather than corrupting — credit to the Label asserts). Dict mappers work
fine; the **non**-inplace `map_state_space_labels` has the correct branch
(`else mapper(line_label)`, circuit.py:2975–2976).

### Suggested fix

`else mapper(gatename)` / `else mapper(line_label)` — two one-word changes, plus tests for the
callable path on both inplace variants (its absence from the suite is how this shipped; the
cross-corpus usage analysis found no in-package callers of the callable+inplace combination,
which also means the fix is low-risk).

### LINKED ISSUE #764: circuit.py papercuts: error-handling and dead-code inventory
This issue brought to you by a robot, written from the robot's perspective.

Small defects in `pygsti/circuits/circuit.py` found during a systematic audit (head of
`develop`, 47b3dcae5). None warrants its own issue; together they make a tidy small PR.
Substantive bugs found in the same audit are filed separately (#756–#763).

### 1. `_proc_key_arg` returns an exception instead of raising it (circuit.py:1205)

```python
if len(key) != 2:
    return IndexError("Index must be of the form <layerIndex>,<lineIndex>")
```

The `IndexError` instance is *returned* as if it were a `(layers, lines)` tuple, so
`c[1,2,3]` surfaces as a confusing unpacking `TypeError` downstream instead of the intended
message. Should be `raise`.

### 2. `validate_line_labels` error message is a non-f-string (circuit.py:218)

```python
msg = "Line label '{line_lbl}' could not round-trip through the parse_label function."
```

Missing the `f` prefix — users see the literal text `{line_lbl}` instead of the offending
label. (Added in #738.)

### 3. `replace_gatename_inplace` warns about the wrong variable (circuit.py:2619)

```python
_warnings.warn(f'Casting `old_gatename` of type {type(new_gatename)} ...')
```

The second warning casts `new_gatename` but the message says `old_gatename` (copy-paste from
the line above).

### 4. Dead branch in `delete_idling_lines_inplace` (circuit.py:3103–3106)

The method opens with `assert(not self._static)` (:3098), then immediately has an
`if self._static:` branch — unreachable.

### 5. Dead conditionals in `extract_labels`'s zero-area path (circuit.py:1331–1335)

Inside the `else` of an `if self._static:` split, the expressions
`() if self._static else []` and `tuple(lines) if self._static else lines` are constants
(`self._static` is known False there). Harmless but misleading.

### 6. ~45 dead lines in `_write_q_circuit_tex` (circuit.py:3836+)

The method raises `NotImplementedError("TODO: need to upgrade this method")` on its first
statement; everything after it is unreachable. Either delete the body (keeping the stub) or
delete the method — the cross-corpus usage analysis found zero references anywhere.

### Suggested course

One small PR fixing 1–3 (behavioral, each trivially testable) and deleting 4–6 (pure dead
code; the flake8-critical gate plus a grep confirm no stragglers).

rileyjmurray:
This PR brought to you by a robot. I reviewed every line before opening it.

## Summary

Five independent bug fixes in `pygsti/circuits/circuit.py`, **one commit per issue** (plus three review-response commits, see below) so the PR can be reviewed bug-by-bug, batched into a single PR to economize on CI time.

Fixes #756, fixes #760, fixes #762, fixes #763, fixes #764.

Every defect was confirmed before fixing (see the linked issues), and every fix ships with regression tests that fail on `develop` without it. No fix touches the circuit identity core: the `tup` wire format, hash values, string grammar, etc..

## The fixes, commit by commit

1. #763: the callable branch returned the mapper function object instead of calling it, so any non-dict mapper hit a Label-validation `AssertionError`. Mirrors the correct branch that the non-inplace `map_state_space_labels` already had. Adds the first direct unit tests for `map_names_inplace` (it had none, dict path included).

2. #760: the only mutator without `assert(not self._static)`; on a static circuit it rewrote `_labels` while leaving the cached `_hashable_tup`/`_hash`/`_str` stale, so `hash(c)` silently disagreed with `c.tup`. It also re-encoded *editable* circuits into the static tuple-of-Labels format, breaking subsequent in-place edits. Now it has the standard guard and preserves the editable nested-list encoding. `done_editing` — its only in-repo caller, which flips `_static` *before* sorting — calls the module-level `_sort_layer_labels` helper directly; its behavior is unchanged (pinned by a new sorting+hash test, and benchmarked identical).

3. #762. This has two parts.
    * tensoring onto a circuit whose gates have implicit (`None`) state-space labels failed deep inside `_clear_labels` with the cryptic *"Cannot operate on a block that is straddled by Gx!"* — now an up-front, actionable `ValueError`. The guard matches develop's failure surface exactly: it only considers implicit labels **in the layers the insertion writes to**, so implicit labels in deeper layers, implicit-sslbls *insertees*, depth-0 or zero-line insertees, and empty idle layers all keep working — each pinned by a test.
    * A list-valued `line_order` (which the signature has always advertised) was stored un-tupled, violating the `_line_labels`-is-a-tuple invariant and crashing much later in `done_editing`. Also adds the first tests ever for the non-inplace `tensor_circuit`.

5. #756: the non-strict membership test `len(sslbls.intersection(lines)) >= 0` was always true (no filtering happened at all), and the result's line labels were assigned the sentinel *string* `'auto'`, which `_bare_init` blindly tuples into `('a', 'u', 't', 'o')`. Now the filter keeps exactly the labels that act on or straddle the requested lines, and the result's line labels are computed for real: the requested lines, extended (in the parent's line-label order) by any extra lines covered by straddling labels. **`strict=False` has zero callers inside pyGSTi and had zero test coverage prior to this PR** — both defects were reachable only through the public API. Bonus: non-strict extraction got ~17% *faster* (it no longer builds oversized layers).

6. #764 misc fixes.
   * `_proc_key_arg` *returned* an `IndexError` instead of raising it (so `c[0,1,2]` surfaced as an unpacking `TypeError`).
   * a non-f-string error message showed users the literal `{line_lbl}`.
   * a cast warning blamed `old_gatename` while casting `new_gatename`.
   * two dead branches (`delete_idling_lines_inplace`'s `if self._static:` after an assert-not-static, constant ternaries in `extract_labels`' zero-area path).
   * `_write_q_circuit_tex` deleted outright — it raised `NotImplementedError` as its first statement and its 43 dead lines referenced attributes that no longer exist (`self.line_items`, `self.identity`). Zero references repo-wide.

## Adversarial self-review (commits 6–8)

Before opening this PR, an adversarial multi-agent review tried to refute each fix. It confirmed four of five outright and caught three real problems, fixed in dedicated commits:

- **Guard too broad** (`bb3f8a08c`): the first version of the 762 guard rejected hosts with implicit labels in layers the insertion never touches (e.g. trailing global idles), which develop tensored successfully. Narrowed to `self._labels[:circuit.num_layers]` and gated on the insertee actually adding lines.

- **CircuitLabel flattening** (`2aff09799`): `CircuitLabel.IS_SIMPLE` is `True`, so the inplace mappers rebuilt CircuitLabels as bare Labels, silently discarding the subcircuit — a path the callable fix made newly reachable (it used to crash). Both mappers now preserve CircuitLabels; the sslbls variant delegates to `CircuitLabel.map_state_space_labels`, which also repairs the pre-existing dict-mapper flattening there.

- **Hot-path nit** (`4fcccd5ed`): a genexpr closing over a local added a `MAKE_CELL` prologue to every `extract_labels` call (~+7% on the `c[i]` fast path that `prefixtable`/`model.py` hit in hot loops). Replaced with closure-free `filter()`; bytecode verified clean.

## Relation to PR #735

Two of these bugs are ones [#735](#735)'s branch currently works around, per the plan that the bug-shaped fixes land on `develop` first and `develop` is then merged back into [#735](#735):

- The [#756](#756) filter fix **matches [#735](#735)'s branch verbatim** (`not strict and sslbls.intersection(lines)`), so no semantic divergence on merge. ([#735](#735) does not fix the `'auto'` line-labels defect; after the merge it inherits this fix.)

- For [#762](#762), this PR makes the implicit-sslbls host case a clear *error*; [#735](#735)'s rewrite later makes it *work* via `Label.concate` — error → works is a clean forward path, no conflict. The `tuple(line_order)` coercion also repairs the same latent un-tupled assignment that [#735](#735)'s rewrite retains (`new_line_labels = line_order`), so merging `develop` into [#735](#735) fixes that branch's copy of the bug for free.

## Testing

26 new unit tests in `test/unit/objects/test_circuit.py::CircuitBugfixRegressionTester`: regression tests verified (in a develop worktree) to **fail on `develop`** for exactly the catalogued reasons, plus explicitly-labeled context pins for neighboring behavior that must *not* change (strict extraction, implicit-sslbls insertees, trailing global idles, empty idle layers, `done_editing` sorting, dict mappers).

```
pytest test/unit/objects/test_circuit.py -q     # 98 passed
pytest test/unit -q                             # full unit suite: green
```

(Environment note: 4 tests fail identically on a clean `develop` checkout in the dev container — 3× `test_gst.py::*::test_write_and_read_to_dir` on a sandbox `PermissionError` and 1× `test_mpi` — i.e. pre-existing environment failures, not branch-caused. CI should be authoritative.)

Indirect-caller paths exercised explicitly: `test_edesigntools.py` / `test_protocols.py`, `test_rb.py` (`map_names_inplace` via MirrorRB), `test_compilers.py` (`map_state_space_labels_inplace`) — all pass.

## Observations out of scope for this PR (no action here)

- `pygsti/extras/crosstalk/core.py:755,898,1039` call the copy-returning `tensor_circuit` and discard the result — silent no-ops since 2020-09 (c47c1c55d), zero test coverage. Worth its own issue.
- `extract_labels` on an *editable* circuit returns layers stored as immutable `Label` objects rather than the editable nested-list encoding (same corruption class as [#760](#760); pre-existing, unchanged by this PR). Candidate for the rewrite or a future issue.
- `insert_circuit_inplace` (circuit.py:2392) is safe under the new `strict=False` semantics only because of its `_is_line_idling` assertion — noted for future maintainers.

### DIFF
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -2500,12 +2500,17 @@ class Circuit(object):
         assert(not self._static), "Cannot edit a read-only circuit!"
         #assert(self.identity == circuit.identity), "The identity labels must be the same!"
 
-        if circuit.num_layers > 0 and len(self._line_labels) > 0 \
-           and _sslbls_of_nested_lists_of_simple_labels(self._labels) is None:
+        # Implicit (None) state-space labels are interpreted as acting on *all* of a circuit's
+        # lines, so a gate with implicit sslbls cannot sit in a layer that the insertion below
+        # writes to (the first circuit.num_layers layers): it would straddle the new lines.
+        # Layers beyond the insertion region are unaffected (e.g. a trailing global idle).
+        if circuit.num_layers > 0 and len(circuit._line_labels) > 0 and len(self._line_labels) > 0 \
+           and _sslbls_of_nested_lists_of_simple_labels(self._labels[:circuit.num_layers]) is None:
             raise ValueError("Cannot tensor: this circuit contains gates with implicit (None) state-space "
-                             "labels, which are interpreted as acting on *all* of the circuit's lines and "
-                             "so cannot coexist with the new lines being added.  Give these gates explicit "
-                             "state-space labels (e.g. 'Gx:0' rather than 'Gx') before tensoring.")
+                             "labels in the layers being tensored, which are interpreted as acting on *all* "
+                             "of the circuit's lines and so cannot coexist with the new lines being added.  "
+                             "Give these gates explicit state-space labels (e.g. 'Gx:0' rather than 'Gx') "
+                             "before tensoring.")
 
         #Construct new line labels (of final circuit)
         overlap = set(self._line_labels).intersection(circuit._line_labels)
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -1337,6 +1337,7 @@ class Circuit(object):
                                         not self._static)  # zero-area region
 
         ret = []
+        observed_sslbls = set()  # the sslbls of all labels kept when strict=False (unused when strict=True)
         if self._static:
             def get_sslbls(lbl): return lbl.sslbls
         else:
@@ -1354,12 +1355,19 @@ class Circuit(object):
                 else:
                     sslbls = set(sslbls)
                 if (strict and sslbls.issubset(lines)) or \
-                   (not strict and len(sslbls.intersection(lines)) >= 0):
+                   (not strict and sslbls.intersection(lines)):
                     ret_layer.append(l)
+                    if not strict:
+                        observed_sslbls.update(sslbls)
             ret.append(_Label(ret_layer) if len(ret_layer) != 1 else ret_layer[0])  # Labels b/c we use _fastinit
 
         if nonint_layers:
-            if not strict: lines = "auto"  # since we may have included lbls on other lines
+            if not strict:
+                # Included labels may straddle the boundary of the requested lines,
+                # so the result can act on more lines than requested -- extend the
+                # line labels to cover them (in this circuit's line-label order).
+                extra_lines = observed_sslbls - set(lines)
+                lines = tuple(lines) + tuple(ll for ll in self._line_labels if ll in extra_lines)
             # don't worry about string rep for now...
 
             return Circuit._fastinit(tuple(ret) if self._static else ret,
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -1365,7 +1365,9 @@ class Circuit(object):
                 # so the result can act on more lines than requested -- extend the
                 # line labels to cover them (in this circuit's line-label order).
                 extra_lines = observed_sslbls - set(lines)
-                lines = tuple(lines) + tuple(ll for ll in self._line_labels if ll in extra_lines)
+                # (filter, not a genexpr: a genexpr would close over extra_lines and put a
+                #  MAKE_CELL prologue on every call, measurably slowing the int fast path)
+                lines = tuple(lines) + tuple(filter(extra_lines.__contains__, self._line_labels))
             # don't worry about string rep for now...
 
             return Circuit._fastinit(tuple(ret) if self._static else ret,
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -2932,7 +2932,16 @@ class Circuit(object):
             if isinstance(mapper, dict) else mapper(gatename)
 
         def map_names(obj):  # obj is either a simple label or a list
-            if isinstance(obj, _Label):
+            if isinstance(obj, _CircuitLabel):
+                # CircuitLabel.IS_SIMPLE is True, but rebuilding it through the
+                # simple-label branch below would silently discard its subcircuit.
+                # Map only the label's own (box) name; the component labels inside
+                # the subcircuit are not renamed (matching the dict-mapper behavior
+                # of leaving non-matching CircuitLabels untouched).
+                new_name = mapper_func(obj.name)
+                newobj = _CircuitLabel(new_name, obj.components, obj.sslbls, obj.reps, obj.time) \
+                    if (new_name is not None and new_name != obj.name) else obj
+            elif isinstance(obj, _Label):
                 if obj.IS_SIMPLE:  # *simple* label
                     new_name = mapper_func(obj.name)
                     newobj = _Label(new_name, obj.sslbls) \
@@ -2968,7 +2977,12 @@ class Circuit(object):
         self._line_labels = tuple((mapper_func(l) for l in self._line_labels))
 
         def map_sslbls(obj):  # obj is either a simple label or a list
-            if isinstance(obj, _Label):
+            if isinstance(obj, _CircuitLabel):
+                # CircuitLabel.IS_SIMPLE is True, but rebuilding it through the
+                # branch below would silently discard its subcircuit; delegate to
+                # the Label-level method, which maps the subcircuit's sslbls too.
+                newobj = obj.map_state_space_labels(mapper_func)
+            elif isinstance(obj, _Label):
                 new_sslbls = [mapper_func(l) for l in obj.sslbls] \
                     if (obj.sslbls is not None) else None
                 newobj = _Label(obj.name, new_sslbls)
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -2494,6 +2494,13 @@ class Circuit(object):
         assert(not self._static), "Cannot edit a read-only circuit!"
         #assert(self.identity == circuit.identity), "The identity labels must be the same!"
 
+        if circuit.num_layers > 0 and len(self._line_labels) > 0 \
+           and _sslbls_of_nested_lists_of_simple_labels(self._labels) is None:
+            raise ValueError("Cannot tensor: this circuit contains gates with implicit (None) state-space "
+                             "labels, which are interpreted as acting on *all* of the circuit's lines and "
+                             "so cannot coexist with the new lines being added.  Give these gates explicit "
+                             "state-space labels (e.g. 'Gx:0' rather than 'Gx') before tensoring.")
+
         #Construct new line labels (of final circuit)
         overlap = set(self._line_labels).intersection(circuit._line_labels)
         if len(overlap) > 0:
@@ -2514,7 +2521,7 @@ class Circuit(object):
             if len(extra) > 0:
                 raise ValueError("`line_order` had nonpresent line labels %s." % str(extra))
 
-            new_line_labels = line_order
+            new_line_labels = tuple(line_order)
         else:
             new_line_labels = self._line_labels + circuit._line_labels
 
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -215,7 +215,7 @@ def validate_line_labels(linelabels):
         test_str = f'Gi:{str(line_lbl)}'
         parsed = str(parse_label(test_str))  # can trigger a ValueError.
         if parsed != test_str:
-            msg = "Line label '{line_lbl}' could not round-trip through the parse_label function."
+            msg = f"Line label '{line_lbl}' could not round-trip through the parse_label function."
             raise ValueError(msg)
     return
 
@@ -1204,7 +1204,7 @@ class Circuit(object):
         """ Pre-process the key argument used by many methods """
         if isinstance(key, tuple):
             if len(key) != 2:
-                return IndexError("Index must be of the form <layerIndex>,<lineIndex>")
+                raise IndexError("Index must be of the form <layerIndex>,<lineIndex>")
             else:
                 return key[0], key[1]
         else:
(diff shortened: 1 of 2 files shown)


pygsti/circuits/circuit.py
```diff
@@ -2909,10 +2931,19 @@ def map_names_inplace(self, mapper):
 
         # If the mapper is a dict, turn it into a function
         def mapper_func(gatename): return mapper.get(gatename, None) \
```
coreyostrove, line 2933: Move to new line.

PR #767, opened 2026-06-10
Tensor Symmetry failing

nkoskelo:
Consider two `Circuit` objects, `c1` and `c2`. There are some checks in place to confirm one can tensor those two circuits together via a call `c1.tensor_circuit(c2)`. For instance, say `c2` contains qubits that `c1` acts on. Then, tensoring the two would be ill-defined. So we block it. This PR ensures that both `c1.tensor_circuit(c2)` and `c2.tensor_circuit(c1)` are blocked if either is blocked.

The example in this PR is:

```
c1 = Circuit("Gx:0")
c2 = Circuit("Gy@1")
c1.tensor_circuit(c2) # works.
c2.tensor_circuit(c1) # fails prior to the PR because c2 has an implicit qubit label for the gate "Gy"
```

Now, both will fail.

Note this PR does not mean that `c1.tensor_circuit(c2) == c2.tensor_circuit(c1)` (the order can change the way the tensored circuit is displayed by changing the `line_labels` order) merely that one successfully executes if and only if the other one also successfully executes.

### DIFF
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -2506,9 +2506,11 @@ class Circuit(object):
         # lines, so a gate with implicit sslbls cannot sit in a layer that the insertion below
         # writes to (the first circuit.num_layers layers): it would straddle the new lines.
         # Layers beyond the insertion region are unaffected (e.g. a trailing global idle).
-        if circuit.num_layers > 0 and len(circuit._line_labels) > 0 and len(self._line_labels) > 0 \
-           and (_sslbls_of_nested_lists_of_simple_labels(self._labels[:circuit.num_layers]) is None \
-            or _sslbls_of_nested_lists_of_simple_labels(circuit._labels[:self.num_layers]) is None):
+        circ_nonempty: bool = circuit.num_layers > 0 and len(circuit._line_labels) > 0
+        self_nonempty: bool = self.num_layers > 0 and len(self._line_labels) > 0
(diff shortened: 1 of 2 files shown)


pygsti/circuits/circuit.py
```diff
@@ -2507,7 +2507,8 @@ def tensor_circuit_inplace(self, circuit: Circuit, line_order: Union[None, List[
         # writes to (the first circuit.num_layers layers): it would straddle the new lines.
         # Layers beyond the insertion region are unaffected (e.g. a trailing global idle).
         if circuit.num_layers > 0 and len(circuit._line_labels) > 0 and len(self._line_labels) > 0 \
-           and _sslbls_of_nested_lists_of_simple_labels(self._labels[:circuit.num_layers]) is None:
+           and (_sslbls_of_nested_lists_of_simple_labels(self._labels[:circuit.num_layers]) is None \
+            or _sslbls_of_nested_lists_of_simple_labels(circuit._labels[:self.num_layers]) is None):
```
rileyjmurray: This condition was already hard to read. Now it's basically impossible. Please define temporary bool variables for each of these conditions before the if-statement and have the conditional for the if-statement  rest comfortably on a single line. (I recognize there is a performance penalty by preventing short-circuit conditional evaluation. The price of that penalty is well-worth the improved readability.)

test/unit/objects/test_circuit.py
rileyjmurray, line 1: I love this new test!

coreyostrove (approved):
This looks good to me, but please flesh out the explanation in the PR slightly more for posterity before merging.

PR #10, opened 2016-06-29
Merge python3 into develop (Once MPI tests pass)

LSaldyt:

PR #769, opened 2026-06-12
Modernize Label.py

### LINKED ISSUE #783: Label to_native() Return trip does not create same label.
On develop:

```
from pygsti.baseobjs.label import Label as L

lbl = L((("Gx",None), ("Gy", None)))
l2 = L(lbl.to_native())
print(lbl.to_native())
assert l2 == lbl # returns true
print(type(l2)) # Gives a LabelTup
print(type(lbl)) # Gives a LabelTupTup
assert type(l2) == type(lbl) # Fails
```

Problems:
1: A LabelTupTup is equivalent to a LabelTup despite representing a different number of gates.
2: `l2` should be a LabelTupTup as it holds 2 distinct gates.
3: The `lbl.to_native()` flat data structure is human readable but can lose the meaning when brought back through the `Label` dispatcher.

nkoskelo:
This PR modernizes label.py

Highlights:
-	**Adds** regression tests for label.py
-	**Ensures** the following paths that are round-trippable including metadata
o	`Label(lbl.to_native()) == lbl`
o	`Label(str(lbl)) == lbl`
o	`jsoncodec.encode_obj(lbl, False)` and `jsoncodec.decode_obj(coded, False)`
o	`s = pickle.dumps(lbl)` ->`l2 = pickle.loads(s)`
-	**Updates** to_native() serialization format to make it round-trippable.
o	Maintains backwards readability with old `to_native()` tuples/strs in `Label()`.
-	**Removes** the `time` field from the default subclasses.
o	Needs to be a WithTime, WithArgs, `CircuitLabel` or `LabelStr` to have a time field.
o	This was done because it is unclear if time means the start of the execution of the Label or the duration of time that the gate takes to complete. An issue has been added for that specific detail.
o	We have fixed the internal callers of this field. However, you may also need to fix it if you have Labels that were not types `{WithX}`.

To fix this in python code one can use `getattr(lbl, “time”, 0.0)`. However, in cython code the `getattr` triple does not work for `cdef` functions. In those cases one needs to use `if hasattr(lbl, “time”): lbl.time else: 0.0` or something similar.

-	**Adds** the ability to concat two labels, lbl1, lbl2, together into a `LabelTupTup`.
o	Asymmetric lbl1.concat(lbl2) != lbl2.concat(lbl1)
-- Need to sort if you want symmetry
o	One cannot concat together anything with a `LabelStr`
o	The depth of each label must match
o	If only one of the labels is a depth 1 CircuitLabel then that label will be converted into a non-CircuitLabel Label and then concatenated with the other label.
o	Introduces ClobberingWarning issued when trying to concatenate two Labels with different non-default time values. The code will continue past this warning and build a `LabelTupTup`.
o	This feature is intended to support tensoring circuits together.
-	**Clarifies** `expand_subcircuits()`
o	If a `LabelTupTupWithArgs` contains a `CircuitLabel` then those arguments will be propagated to each of the `LabelTupTupWithArgs` containing the layers of the expanded CircuitLabel. This is done because a CircuitLabel can be used as a shorthand for a circuit. We are assuming that the arguments are not dependent on time/duration. It does not propagate the `time` field due to the unknown nature of the `time` field.
o	LabelTupTupWithTime will not propagate the `time` field to the expanded subcircuit layers due to the `time` - duration discussion.
-	**Type change** (`Label.sslbls`) must now either be a `tuple[int, …]` or a `tuple[str, …]` you are not allowed to mix the types for the identifiers within the space.

**Inheritance Structure:**

LabelStr -> Label
LabelTupWithArgs -> LabelTupWithTime -> LabelTup -> Label
LabelTupTupWithArgs -> LabelTupTupWithTime -> LabelTupTup -> Label
CircuitLabel -> LabelTupTupWithTime -> LabelTupTup -> Label


rileyjmurray:
@nkoskelo the following is a review by Opus 4.8. I've asked it for minimal breaking examples.

-------------------------------------------------

The PR's existing 170 new tests pass — but several real bugs slip through because the tests don't exercise the paths
that break. **Every code example below was re-run against the current branch HEAD and reproduces as shown.**

> **Update — remote pull on 2026-06-25 (HEAD `7e1f49678`).** New commits from Nick have already
> resolved several of the "🟢 safe to push" items, so they are **no longer action items** for you:
> - **Finding 3** (`LabelStr` non-zero-time `to_native` round-trip) — **fixed** (`0e813b6b7`); now
>   serializes as `('rho0', '!2.5')` and round-trips with time preserved.
> - **Finding 5** (`mapforwardsim_calc_generic.py:312` unguarded `rholabel.time`) — **fixed**; the line
>   is now `t += getattr(rholabel, 'time', 0.0)` (and the `.pyx` twin switched to the `hasattr` form,
>   `da033bf1f`).
> - **Finding 6** (`tensor_circuits` silently dropping circuits on mismatched-length lists) — **fixed**;
>   there is now an explicit `len(...) != len(...)` → `ValueError` check plus a covering test.
> - **Finding 7** (duplicate / shadowed test method names) — **fixed** (`3c1cf6dfe`); no duplicate
>   `def test_*` names remain.
> - The method was **renamed `concate` → `concat`** (`7e1f49678`), resolving the naming nit.
>
> The six **🟡 discuss-with-Nick** findings below are all **still open** and were re-confirmed against
> this HEAD. Remaining 🟢 items: the `"UNMUTABLE"` docstring typo (`circuitlist.py:25`, still present)
> and the test-strengthening in finding 12.


## Overall assessment

This is a genuinely good direction. Pulling the `time` property out of the label types that
don't carry time, factoring the duplicated `_integerize_sslbls` / `_create_concatenated_label`
helpers, fixing the long-standing "`to_native()` drops time" bug for the WithTime types, and adding
real test coverage are all net improvements. The class-hierarchy cleanup (`WithArgs` now extends
`WithTime`) is sensible.

However, the two **new** features — `concat()` and the changed `to_native()` serialization format —
are **incomplete and ship with confirmed bugs that the new tests do not catch**. The simple paths
work; the richer paths (anything involving a `CircuitLabel`, `args`, or a non-zero-time `LabelStr`)
break. I'd recommend not merging until at least the correctness items below are resolved.

---

## Ownership summary

**🟡 Discuss with Nick** — touches the intended semantics/design of his new features or a back-compat policy decision:

| # | Finding | Where |
|---|---------|-------|
| 1 | `concat()` crashes on **any** `CircuitLabel` with sslbls (core feature broken) | `label.py` `_check_concate_compatibility` ~71, `CircuitLabel.concat` ~1699 |
| 2 | `concat()` double-counts args in `collect_args()` (feeds the op factory) | `label.py` `LabelTup/LabelTupTup.concat` ~647/1277 |
| 4 | `CircuitLabel` non-zero time fails `to_native()` round-trip (read side missing) | `label.py` `CircuitLabel.to_native` ~1641 |
| 8 | `LabelTupTup.init` emits spurious `RuntimeWarning` on ordinary construction | `label.py` ~1044–1072 |
| 9 | `LabelTupTupWithArgs.expand_subcircuits` duplicates `self.args` onto every layer | `label.py` ~2092 |
| 10 | Back-compat: old-format `to_native` on disk is unreadable / silently misread | `label.py` constructor |

**🟢 Reasonable for you (Riley) to fix and push** — mechanical, unambiguous, low risk of clashing with Nick's intent:

| # | Finding | Where |
|---|---------|-------|
| 12 | Test-strengthening (loose `assertIsInstance`, tautological assert, brittle `repr`) | `test_label.py` |

Detail on every item follows.

---

## 🔴 Correctness bugs

### 1. `concat()` crashes on any `CircuitLabel` that has state-space labels — *discuss with Nick*

Both the helper `_check_concate_compatibility` (~line 71) and `CircuitLabel.concat` (~line 1699)
"unwrap" a depth-1 `CircuitLabel` with:

```python
Label(label2.components, label2.sslbls, label2.time, label2.args)
```

When `name` is a tuple **and** `state_space_labels` is non-`None`, the constructor skips the
tuple-of-tuples branch and falls through to `LabelTup(WithArgs).init`, which asserts `name` is a
string.

**Minimal breaking example** (verified in this checkout):

```python
from pygsti.baseobjs import Label, CircuitLabel

g  = Label('Gy', (1,))                                  # LabelTup on qubit 1
cl = CircuitLabel('c', [Label('Gz', (2,))], (2,), 1)    # CircuitLabel on qubit 2

g.concat(cl)
# AssertionError: `name` must be a string, but it's '(Label(('Gz', 2)),)'

cl.concat(g)
# AssertionError: `name` must be a string, but it's '(Label(('Gz', 2)),)'
```

So `concat` works for `LabelTup`/`LabelTupTup`/`WithTime`/`WithArgs` with disjoint qubits, but
**every path through a `CircuitLabel` raises**. The existing `concat` tests only build CircuitLabels
with `sslbls=None` — the *one* combination that happens to dodge the bug — which is why CI is green:

```python
from pygsti.baseobjs import Label, CircuitLabel

cl0 = CircuitLabel('c', [Label('Gz', (2,))], None, 1)   # sslbls=None
cl0.concat(Label('Gy', (1,)))
# Label((Label(('Gz', 2)), Label(('Gy', 1))))           # no crash
```

(See finding 7: the dead duplicate test at `test_label.py:834` actually does exercise a
CircuitLabel unwrap; un-shadowing it will turn this into a failing test.)

Why discuss: the fix is to construct the unwrapped label correctly, but the right construction
depends on what Nick intends the unwrapped form to be (a `LabelTupTup`/`WithTime`/`WithArgs` built
from the components). A reasonable candidate is to route through `_create_concatenated_label(self.components, self.time, self.args)`
instead of `Label(...)`, but please confirm intent with him.

### 2. `concat()` double-counts arguments — *discuss with Nick*

`LabelTup.concat` / `LabelTupTup.concat` do:

```python
all_args = self.collect_args() + other.collect_args()
return _create_concatenated_label(components, new_time, all_args)
```

This puts the collected args at the **top level** of the resulting `LabelTupTupWithArgs` while the
component sub-labels **still carry their own args**. `collect_args()` then sums both levels.

**Minimal breaking example** (verified in this checkout):

```python
from pygsti.baseobjs.label import LabelTupWithArgs

a = LabelTupWithArgs.init('Gx', (0,), args=(0.5,))
b = LabelTupWithArgs.init('Gy', (1,), args=(0.7,))
r = a.concat(b)

r.args            # (0.5, 0.7)
r.collect_args()  # (0.5, 0.7, 0.5, 0.7)   <-- each arg appears twice
```

`collect_args()` is consumed by `modelmembers/operations/opfactory.py` to build ops, so a
concat-built label would hand doubled args to the factory. Why discuss: the fix is a design choice
(don't set top-level args and let `collect_args` gather from components, or strip args off the
components) — Nick should decide where args are meant to live on a combined label.

### 4. `CircuitLabel` with non-zero time does not round-trip through `to_native()` — *discuss with Nick*

`CircuitLabel.to_native` (~line 1641) inserts the time as `'!<time>'` right after `reps`:

```python
tup = self[0:3]
if self.time is not None and self.time != 0.0:
    tup += ('!' + str(self.time),)
tup += tuple(x.to_native() for x in self.components)
```

But the reconstruction branch in `Label.__new__` treats everything in `name[3:]` as sub-labels — it
never looks for the `'!time'` marker — so it injects a bogus `Label('!3.0')` component and loses the
time.

**Minimal breaking example** (verified in this checkout):

```python
from pygsti.baseobjs import Label, CircuitLabel

cl = CircuitLabel('', (Label('Gx', (0,)),), (0,), 1, time=3.0)

cl.to_native()
# ('', (0,), 1, '!3.0', ('Gx', 0))

rt = Label(cl.to_native())
rt
# CircuitLabel('',(Label('!3.0'), Label(('Gx', 0))),(0,),1,0.0)
rt.time        # 0.0                                  <-- time lost
rt.components  # (Label('!3.0'), Label(('Gx', 0)))    <-- spurious '!3.0' sub-label
rt == cl       # False
```

This is the same class of defect as #3 but the fix needs the **read** side updated too, so it's
coupled to the serialization-format design — discuss with Nick (and it overlaps with the back-compat
question in #10).




## 🟡 Serialization / backward-compatibility

### 10 Old on-disk `to_native` format may be unreadable or silently misread — *discuss with Nick*

The only **live** persistence path for `to_native()` is `models/fogistore.py`: `_GetState` does
`op.to_native()` for every op label, the result is JSON-serialized, and `_SetState` rebuilds each one
with `_Label(op)` (file lines ~310–345). A FOGIStore written by a released pyGSTi therefore holds
**old-format** native tuples that this PR's `Label()` constructor must read back. With the format
change, three concrete cases break.

**Minimal breaking examples** (verified in this checkout — these are the exact tuples old `to_native()`
produced, fed to the new `Label()`):

```python
# (a) old LabelTupWithArgs native was (name, 2+len(args), *args, *sslbls).
#     i.e. LabelTupWithArgs.init('Gx', (0,), args=('foo',)).to_native() == ('Gx', 3, 'foo', 0)
from pygsti.baseobjs.label import LabelTupWithArgs

new = Label(('Gx', 3, 'foo', 0))
new                # Label(('Gx', 3, 'foo', 0))   <-- a plain LabelTup, SILENTLY wrong
type(new).__name__ # 'LabelTup'
new.args           # ()                 <-- 'foo' arg dropped
new.sslbls         # (3, 'foo', 0)      <-- K-index and arg parsed as qubits

# ...and the corruption hides behind ==, so a load looks "successful":
new == LabelTupWithArgs.init('Gx', (0,), args=('foo',))   # True
```

```python
# (b) old LabelTupTupWithArgs native was (1+len(args), *args, *components).
#     i.e. ...init((('Gx',0),('Gy',1)), args=('foo',)).to_native() == (2, 'foo', ('Gx',0), ('Gy',1))

Label((2, 'foo', ('Gx', 0), ('Gy', 1)))
# AssertionError: `name` must be a string, but it's '2'
```

```python
# (c) CircuitLabel native after a JSON round-trip: FOGIStore stores via JSON, so the
#     tuple sslbls (0,) come back as the list [0], which the constructor rejects.
import json
from pygsti.baseobjs import Label, CircuitLabel

cl = CircuitLabel('', (Label('Gx', (0,)),), (0,), 1)
native = cl.to_native()                            # ('', (0,), 1, ('Gx', 0))
from_disk = json.loads(json.dumps(native))         # ['', [0], 1, ['Gx', 0]]
Label(from_disk)
# AssertionError: State space label '[0]' must be a string or integer!
```

Case (a) is the most dangerous: it raises no error, returns a `LabelTup`, and (as the snippet above
shows) `==` against the correct `LabelTupWithArgs` still returns `True` — so a silently-corrupted
model loads "successfully" and even compares equal to the real thing.

Real-world exposure is **bounded**: FOGIStore op labels are normally plain `LabelStr`/`LabelTup`/
`LabelTupTup` (no args/time), and the main model/circuit IO uses `jsoncodec`, not `to_native()`. New
→ new round-trips, `__reduce__`/pickle, circuit `.str` parsing, and full model write/read all verified
working. But a stored args- or time-bearing FOGIStore from a released pyGSTi is a concrete
exception/data-loss path. Decision for Nick: do we need to read the legacy format (add a
compatibility branch), or is a clean break acceptable for this format? Either way, the JSON
list-vs-tuple sslbls issue in the `CircuitLabel` reconstruction branch is worth hardening.

---

## 🟡 Behavior / noise

### 8. `LabelTupTup.init` emits a spurious `RuntimeWarning` on normal construction — *discuss with Nick*

The new `__debug__` heuristics (~lines 1044–1072) fire on ordinary inputs. The final `elif`
(all components share one time value, no args) triggers for the very common all-zero-time case and
is **already firing inside the test suite** (`test_to_native[label6]/[label8]`). Note this warning is
a plain `RuntimeWarning` — it is *not* the new `ClobberingWarning` that `da033bf1f`/`8d02caca5`
introduced for the `_create_concatenated_label` overwrite case (label.py:65–69); that's a separate path.

**Minimal reproducing example** (verified in this checkout):

```python
import warnings
from pygsti.baseobjs.label import LabelTupTup, LabelTupWithTime

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    LabelTupTup.init((LabelTupWithTime.init('Gx', (0,), 0.0),
                      LabelTupWithTime.init('Gy', (1,), 0.0)))
    print(w[0].category.__name__, "->", w[0].message)
# RuntimeWarning -> You may want to consider a LabelTupTupWithTime as all of the
# entries have the same time which was previously the default time.
```

The intent (nudging users toward `LabelTupTupWithTime`) is reasonable, but as written it will spam
warnings in normal use. Discuss with Nick whether to narrow the conditions, downgrade to a debug log,
or drop the zero-time branch. (Note its behavior also depends on the PR's removal of `.time` from
`LabelTup` — whether it fires changes based on whether components are `LabelTup` vs `WithTime`, which
is itself a smell.)

### 9. `LabelTupTupWithArgs.expand_subcircuits` replicates `self.args` onto every layer — *discuss with Nick*

The new override (~line 2092) does `LabelTupTupWithArgs.init(ec, args=self.args)` inside the
per-layer loop, so a single args-bearing label that expands into N layers gets its args duplicated N
times. This is inconsistent with `LabelTupTupWithTime.expand_subcircuits`, which does **not** push
`time` onto the expanded layers.

**Minimal reproducing example** (verified in this checkout):

```python
from pygsti.baseobjs import Label, CircuitLabel
from pygsti.baseobjs.label import LabelTupTupWithArgs

# a 2-layer sub-circuit, wrapped in a CircuitLabel, inside an args-bearing tuple-of-tuples
sub = CircuitLabel('', [Label('Gx', (0,)), Label('Gy', (0,))], (0,), 1)
top = LabelTupTupWithArgs.init((sub,), args=(0.5,))

top.args                                       # (0.5,)
expanded = top.expand_subcircuits()
len(expanded)                                  # 2
[getattr(l, 'args', None) for l in expanded]   # [(0.5,), (0.5,)]  <-- args on EVERY layer
```

Confirm with Nick whether args should propagate to every expanded layer (semantically that passes
the args to N operations instead of one).

nkoskelo:
@rileyjmurray or @coreyostrove. This is ready for your review.

rileyjmurray:
@nkoskelo please comment on if and how you addressed points 8 and 9 in my earlier comment.

rileyjmurray (approved):
In-person discussion answers my remaining questions.

Please revise the opening PR comment with the whole picture in mind. Then we can merge.

nkoskelo:
In response to [Riley's Comment](https://github.com/sandialabs/pyGSTi/pull/769#issuecomment-4803500902, part 8, I have explicitly removed that warning.

For part 9 we are explicitly keeping the copying of `args` to all applicable layers.
`expand_subcircuits()` on a `LabelTupTupWithX` will not propagate the time value to any layers beyond the first since we do not know if the time value is a duration or the time at which this layer is executed.
`expand_subcircuits()` on a `LabelTupTupWithArgs` will propagate the argument list to all layers since a `CircuitLabel` can be viewed as a way to contract the writing of a Label with repeated structure.

rileyjmurray (approved):

PR #772, opened 2026-06-12
Be slightly more permissive with implicit state space labels when tensoring two circuits.

nkoskelo:
We are judging whether implicit state space labels are bad on a layer by layer basis. Basically both Labels need to have either a sslbl or are both the implicit idle

pseudocode of check
```
for i in range(min(len(c), len(c2)):
    if c[i] != c2[i] and c2 == implicit idle:
       raise ValueError
    elif c[i] == implicit_idle and c2[i] != c[i]:
       raise ValueError
    elif c[i] != implicit_idle and c2[i] != implicit_idle and (c[i].sslbls is None or c2[i].sslbls is None):
       raise ValueError
```


rileyjmurray (approved):
LGTM. I've locally verified that tests in `test_packages` pass after this change, while they fail on develop.

PR #780, opened 2026-06-15
Reduce CI wall-clock latency, plus shift to Python 3.10 -- 3.13

### LINKED ISSUE #606: Tests in test/unit/ take way too long
I work with a Macbook Pro with a 12-core M2 Pro CPU. If I install pytest-xdist (to enable parallel tests) and then run ``pytest -n auto test/unit --durations=20``, I get the following. There's an insane amount of time spent on the 10 or so slowest tests.
```
=================================================================================================== slowest 20 durations ====================================================================================================
1592.01s call     test/unit/tools/test_errgenproptools.py::ErrgenCompositionCommutationTester::test_errorgen_composition
1505.58s call     test/unit/tools/test_errgenproptools.py::ApproxStabilizerMethodTester::test_alpha
467.73s call     test/unit/protocols/test_rb.py::TestCliffordRBDesign::test_parallel_design_construction
391.56s call     test/unit/objects/test_circuit.py::CircuitMethodTester::test_convert_to_cirq
275.94s call     test/unit/objects/test_forwardsim.py::ForwardSimConsistencyTester::test_consistent_jacs
267.55s call     test/unit/tools/test_errgenproptools.py::ApproxStabilizerMethodTester::test_stabilizer_probability_correction
263.35s call     test/unit/tools/test_errgenproptools.py::ApproxStabilizerMethodTester::test_approximate_stabilizer_probabilities
203.66s call     test/unit/tools/test_errgenproptools.py::ApproxStabilizerMethodTester::test_phi
167.68s call     test/unit/tools/test_errgenproptools.py::ApproxStabilizerMethodTester::test_approximate_stabilizer_probability
148.52s call     test/unit/tools/test_edesigntools.py::FisherInformationTester::test_calculate_fisher_information_matrix_with_instrument
117.19s call     test/unit/tools/test_errgenproptools.py::ErrgenCompositionCommutationTester::test_bch_approximation
87.17s call     test/unit/modelmembers/test_operation.py::StaticOpTester::test_diamonddist
86.65s call     test/unit/modelmembers/test_operation.py::LinearlyParamOpTester::test_diamonddist
82.30s call     test/unit/modelmembers/test_operation.py::DiagonalCPTPLindbladDenseOpTester::test_diamonddist
79.26s call     test/unit/modelmembers/test_operation.py::AffineShiftOpTester::test_diamonddist
78.29s call     test/unit/modelmembers/test_operation.py::FullOpTester::test_diamonddist
73.89s call     test/unit/tools/test_errgenproptools.py::ApproxStabilizerMethodTester::test_approximate_stabilizer_pauli_expectation
58.17s call     test/unit/report/test_colormaps.py::BasicColormapTester::test_get_matplotlib_norm_and_cmap
58.16s call     test/unit/report/test_colormaps.py::PiecewiseLinearColormapTester::test_get_matplotlib_norm_and_cmap
56.92s call     test/unit/protocols/test_rb.py::TestInterleavedRBProtocol::test_interleavedrb_protocol_noisy
```
I suggest extremely expensive unit tests be moved into a different folder. Thoughts, @coreyostrove ?

rileyjmurray:
This branch substantially cuts the wall-clock **time-to-green** of a CI run (40+ minutes on the slowest leg today) through a sequence of independent changes to the reusable test workflow and packaging. The net effect is that CI now takes less than 20 minutes. Once merged, this PR will resolve #606.

The error-generator unit tests that dominated the Python-3.12 job (`test_errgenproptools.py::…::{test_errorgen_composition, test_iterative_error_generator_composition}`, ~920 s combined and pinned to one `xdist` worker by `--dist loadscope`) are slow because **coverage.py's legacy C trace function is de-optimized by Python 3.12's specializing adaptive interpreter** (PEP 659). Switching coverage to its `sys.monitoring` core (PEP 669) restores near-native speed: the same two tests drop from ~595 s / ~325 s to a few seconds each.

The only remotely controversial change in this PR is dropping support for Python 3.9. That version reached end-of-life in October 2025 according to the Python Software Foundation. Dropping support for it here shaves almost 10 minutes off build time (one of our dependencies, `stim`, has no Python 3.9 wheels and so has an expensive build-from-source). We could add Python 3.9 back if we pinned the `stim` version in the 3.9 build to something that has a wheel.

### 1. Faster installs (foundation)
- Move the install path to **`uv`** with dependency caching (`actions/cache` over the uv cache), and resolve PyTorch — a transitive test dependency — from the **CPU-only** wheel index (`UV_TORCH_BACKEND=cpu`) instead of pip's default CUDA build, which otherwise drags in ~2.5 GB of unused `nvidia-*` wheels per job.
- `setup.py` build speed-ups, and disable parallel extension builds on Windows (a known MSVC parallel `build_ext` failure mode).

### 2. Coverage via `sys.monitoring` — the dominant latency lever
- Set `COVERAGE_CORE: sysmon` at the workflow level so `--cov` runs use coverage.py's PEP 669 core. On Python 3.12 the legacy C tracer made coverage of call-heavy tests dramatically slower (the errgen composition tests ran ~595 s / ~325 s under coverage vs a few seconds without); `sysmon` removes essentially all of that overhead. It falls back gracefully (with a `CoverageWarning`) on Python < 3.12. Coverage *data* is unchanged.

### 3. Test Python 3.10–3.13 (drop 3.9, add 3.13)
- `requires-python` → `>=3.10`; every workflow matrix moves to `3.10–3.13` (feature branches keep an extremal pair, now `3.10`/`3.13`).
- Dropping 3.9 removes the slowest jobs (3.9 lacked stim/numpy-2 wheels and built more from source). Adding 3.13 brings the test set up to the current Python.

### 4. Measure coverage on a single job
- CI computed `--cov` on **every** matrix job but never uploaded or consumed the result, so it was pure overhead. A new `run-coverage` input (default `'false'`) gates `--cov=pygsti`, and each caller enables it for **exactly one** job — `ubuntu-latest` + Python 3.13 — where `sysmon` keeps the tracer cheap. The steps use `shell: bash` so the conditional flag expands on Windows runners too.

### 5. Drop the redundant explicit `build_ext`
- The Cython install step ran `uv pip install -e .[testing]` **and then** `python setup.py build_ext --inplace`, recompiling all extensions a second time (~40 s/job). The PEP 660 editable install already compiles them in-place via build isolation — verified on a clean checkout: the editable install alone yields importable `.so` for all 19 extensions.

### 6. Consistency
- Build cp310–cp313 wheels (drop cp39, add cp313) so the release matrix matches the supported set, and update the README's stated minimum to "python 3.10 or higher".

## Why the Python version changes compose

`sysmon` alone is necessary but not sufficient: time-to-green was gated by the **3.9** jobs, which paid big overhead due to building `stim` from source.

Once 3.9 is gone, the tests for `beta` and `master` would hit bottlenecks with **3.11**. That version has the specializing interpreter (the performance hit) but no `sys.monitoring` (3.12+ only). 

Assuming we proceed with the 3.9 removal it's natural to bump the highest-tested Python version to 3.13. Our develop tests only run the lowest-version and highest-version supported Python, which become 3.10 and 3.13 after this PR. Strictly speaking, we could run efficient coverage tests with Python 3.12.

## Verification

- All six workflow YAML files parse.
- `COVERAGE_CORE=sysmon`: the two errgen tests run in ~24 s combined under the exact CI invocation (vs ~920 s on the legacy tracer); full `test/unit/tools` green.
- **Python 3.13, fresh environment:** the full dependency closure (numpy 2.4, scipy, stim, qiskit, torch, cirq, cvxpy, jupyter, …) installs cleanly; the editable install compiles all 19 Cython extensions in-place; **full `test/unit` = 2335 passed, 0 failed** (one MPI-gated test skips only because that local env was built without MPI; CI has it).
- Coverage gating traced by hand and by an independent review: exactly one coverage job per workflow, always `ubuntu` + `3.13` + cython.

## Notes / follow-ups (not in this PR)

- The cp313 wheel build in `autodeploy.yml` only runs on release, so it is **not exercised by this PR** — first validated at the next release.
- Coverage is computed but **never uploaded** anywhere (no Codecov step, no `coverage xml`). This PR keeps one human-readable coverage report on the 3.13 job at near-zero cost; dropping CI coverage entirely, or wiring a real upload, are both reasonable separate decisions.
- Pre-existing items noticed while here, left for separate PRs: `develop.yml`/`bugfix.yml` declare a `use-cython: ['true','false']` matrix but force `'true'`, so every job builds twice; the no-cython install step sets `PYGSTI_CYTHON_SKIP` while `setup.py` checks `PYGSTI_SKIP_CYTHON`; and Python 3.13 surfaces `SyntaxWarning: invalid escape sequence` for LaTeX in two `errgenproptools.py` docstrings (harmless today — raw-string them).

### DIFF
--- a/.github/workflows/beta-master.yml
+++ b/.github/workflows/beta-master.yml
@@ -21,6 +21,7 @@ jobs:
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
+      run-coverage: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == '3.13' && matrix.use-cython == 'true' }} # cover exactly one job
       use-cython: ${{ matrix.use-cython }}
       run-unit-tests: 'true'
       run-integration-tests: 'true'
--- a/.github/workflows/beta-master.yml
+++ b/.github/workflows/beta-master.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: [3.9, '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
         use-cython: ['true', 'false']
     uses: ./.github/workflows/reuseable-main.yml
     name: Run pyGSTi tests
--- a/.github/workflows/beta-master.yml
+++ b/.github/workflows/beta-master.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
         use-cython: ['true', 'false']
     uses: ./.github/workflows/reuseable-main.yml
     name: Run pyGSTi tests
--- a/.github/workflows/autodeploy.yml
+++ b/.github/workflows/autodeploy.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.0.0
         env:
-          CIBW_BUILD: cp39-* cp310-* cp311-* cp312-*
+          CIBW_BUILD: cp310-* cp311-* cp312-* cp313-*
           CIBW_BUILD_VERBOSITY: 1
           CIBW_BEFORE_ALL_LINUX: ./.github/ci-scripts/before_install.sh
           CIBW_BEFORE_ALL_MACOS: ./.github/ci-scripts/before_install_macos.sh
--- a/.github/workflows/bugfix.yml
+++ b/.github/workflows/bugfix.yml
@@ -22,6 +22,7 @@ jobs:
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
+      run-coverage: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == '3.13' && matrix.use-cython == 'true' }} # cover exactly one job
       use-cython: 'true' # Cython only
       run-unit-tests: 'true'
       run-integration-tests: 'true'
--- a/.github/workflows/bugfix.yml
+++ b/.github/workflows/bugfix.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest] # No Mac
-        python-version: [3.9, '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
         use-cython: ['true', 'false']
     uses: ./.github/workflows/reuseable-main.yml
     name: Run pyGSTi tests
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -22,6 +22,7 @@ jobs:
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
+      run-coverage: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == '3.13' && matrix.use-cython == 'true' }} # cover exactly one job
       use-cython: 'true' # Cython only
       run-unit-tests: 'true'
       run-integration-tests: 'true'
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest] # No Mac
-        python-version: [3.9, '3.10', '3.11', '3.12']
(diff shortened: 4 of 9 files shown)


rileyjmurray:
@coreyostrove please LMK if this is okay to merge.

.github/workflows/beta-master.yml
```diff
@@ -14,13 +14,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: [3.9, '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
```
nkoskelo: Can we have a test that runs on python 3.14? This would be so that we may be able to anticipate any version related bugs earlier. Maybe this test would only run once every month / quarter. We could call it the $\pi$ test.

rileyjmurray: I'd be okay with having it run every time. We can just keep the coverage test to Python 3.13.

Will wait for comments from Corey before making changes.

Issue #787, opened 2026-06-18
`compile_cnot_circuit` can error or return wrong output on processors with restricted connectivity


rileyjmurray:
The connectivity-adjusted CNOT compiler algorithms `COiCAGE` and `OiCAGE` are used by `compile_cnot_circuit`, and indirectly by `compile_conditional_symplectic` / `compile_stabilizer_state` / `compile_clifford`.

These algorithms frequently produce CNOT circuits whose symplectic representation does **not** match the requested `s` when compiling for a device with restricted connectivity (line/ring/grid). The error surfaces if the caller passes `check=True` (which is the default!), which raises `AssertionError: Algorithm has failed!`. If `check=False`, then the algorithm silently returns a wrong circuit.

## Impact

- The DirectRB.ipynb example notebook errors.
- Any code path that calls `compile_cnot_circuit`/`compile_conditional_symplectic` with `check=False` on restricted geometry can silently get an incorrect circuit.

## Reproduction

4-qubit ring, 300 random CNOT circuits, counting how many compiled outputs are wrong (caught by `check=True`):

```python
import numpy as np
from pygsti.algorithms import compilers
from pygsti.processors import QubitProcessorSpec as QPS
from pygsti.tools import symplectic
from pygsti.circuits import Circuit
from pygsti.baseobjs import Label

ql = ['Q0', 'Q1', 'Q2', 'Q3']
ring = QPS(4, ['Gxpi2', 'Gxmpi2', 'Gypi2', 'Gympi2', 'Gcphase'], qubit_labels=ql, geometry='ring')
rng = np.random.RandomState(0)

def random_cnot_sym(nlayers=12):
    layers = [Label('CNOT', tuple(ql[i] for i in rng.choice(4, size=2, replace=False)))
              for _ in range(nlayers)]
    s, _ = symplectic.symplectic_rep_of_clifford_circuit(
        Circuit(layer_labels=layers, line_labels=ql).parallelize())
    return s

for alg in ['COiCAGE', 'OiCAGE', 'BGE']:
    fails, rng2 = 0, np.random.RandomState(0)
    for _ in range(300):
        s = random_cnot_sym()
        aargs = [ql] if alg == 'OiCAGE' else []
        try:
            compilers.compile_cnot_circuit(s, ring, compilation=None, qubit_labels=ql,
                                           algorithm=alg, compile_to_native=False,
                                           check=True, aargs=aargs, rand_state=rng2)
        except AssertionError:
            fails += 1
[...]

PR #788, opened 2026-06-18
Drop Torch from `complete` build. Update notebooks so they can run. Bugfixes.

rileyjmurray:
This branch makes every tutorial and example notebook under `docs/markdown/` execute cleanly again, and wires the notebook regression suite into CI so they *stay* runnable. Getting there required three kinds of work:

1. **Test infrastructure** — a `docs/conftest.py` that materializes the data files some notebooks load from others, plus the CI plumbing to actually run the notebook suite (it has been disabled across all workflows since "end of May 2024").
2. **Notebook content fixes** — updating tutorials/examples that had drifted out of sync with the current pyGSTi API (gauge-optimization keywords, renamed functions, removed deprecated kwargs, numpy-2 compatibility, etc.).
3. **Library bugfixes** — a number of genuine bugs in `pygsti` that were only surfaced by running the notebooks end to end. Each of these comes with the tutorial it unblocks, and several come with new unit tests.

The notebook suite runs as an unordered, parallel [`nbval`](https://github.com/computationalmodelling/nbval) job (`pytest -n auto --nbval-lax --dist loadscope docs/markdown`). It is enabled on the `develop` and `beta`/`master` workflows on a single Linux job; the lightweight feature-branch workflow is left untouched.

### `docs/conftest.py` — shared cross-notebook fixtures

A handful of tutorials analyze data that *other* tutorials generate and save under `docs/tutorial_files`:

- `objects/DataSet` writes `Example_Dataset.txt`, `Example_Dataset_LowCnts.txt` and the `Example_GST_Data` protocol-data directory, which `gst/Driverfunctions`, `gst/LowLevel`, `gst/Protocols` and `reporting/ReportGeneration` read back.
- `gst/Protocols` runs GST and writes `Example_GST_Data/results`, which `objects/Results` reads back.

When the notebooks run as an unordered, parallel suite there is no guarantee a generator runs before its consumers. `conftest.py` solves this with a `pytest_sessionstart` hook that materializes exactly these shared inputs once, before any notebook runs, using the same model pack and seed as the source tutorials so the generated files match what a reader running the notebooks in order would produce. It is xdist-safe (a cross-worker file lock elects one worker to generate the fixtures; the rest wait on a `.fixtures_ready` sentinel) and writes everything into the gitignored `docs/tutorial_files`, so nothing it creates is ever committed.

This is the mechanism that lets any single notebook be run on its own — which is also what makes the parallel CI run reliable.

### CI: actually run the notebook regression

The `Run notebook regression` step already existed in `reuseable-main.yml` but every caller passed `run-notebook-tests: 'false'`, so it never ran. This branch:

- Enables it on **`develop.yml`** and **`beta-master.yml`**, on exactly one job (Linux + Python 3.13 + Cython) — the same single-job selector already used for coverage. The suite is slow, so it is deliberately not fanned out across the full OS/Python matrix. `feature-branches.yml` and `bugfix.yml` remain notebook-free to keep their light suites light.
- Makes the step **Linux-only at the reusable-workflow level** (`&& inputs.os == 'ubuntu-latest'`), since it compiles CHP with `gcc` and installs it under `/usr/local/bin` — it can't run on the Windows/macOS runners even if a caller asked it to. The simple way to extend to Windows/macOS would be to `pip install chp_sim`, which is a pure-Python implementation.
- Points the suite at `docs/markdown` (the canonical MyST sources) rather than the old `docs/notebooks` path, materializing the paired `.ipynb` in-place via `jupytext --sync` (see `docs/jupytext.toml`).
- Removes a stale `TODO: address compilation failure in beta-master builds` comment, now that those builds are green.

### Packaging / ignore-rule changes

- **`pyproject.toml`**: `torch` is pulled out of the `testing*` extras and made an explicit, opt-in `pytorch` extra. Its default Linux wheels drag in ~2.5 GB of `nvidia-*` CUDA wheels that CI never uses, so the install steps now request `.[testing,pytorch]` / `.[testing_no_cython,pytorch]` explicitly (resolved CPU-only via `UV_TORCH_BACKEND=cpu`). This keeps `torch` available so the `TorchForwardSimulator` paths are exercised rather than skipped, without forcing the CUDA payload on everyone who installs `pygsti[complete]`.
- **`.gitignore`**: the `docs/tutorial_files` and `docs/example_files` scratch directories are now tracked via a `.keep` placeholder (so they exist on a fresh checkout) while everything the notebooks write inside them stays ignored. The offline JS/CSS asset bundle that report-writing drops next to a running notebook (`docs/markdown/**/offline/`) is also ignored.

## Notebook content fixes (API drift)

These notebooks executed against an older API and were updated to the current one:

- **Gauge optimization**: switched tutorials from the old `'g0'` / `go0` keys to `'stdgaugeopt'`, and from passing `gauge_opt_params` alongside the now-default `gauge_opt_suite_name` to the correct mutually-exclusive usage — `Driverfunctions`, `GaugeOpt`, `CustomOperator`, `WorkspaceSwitchboards`, `WorkspaceExamples`, `ReportGeneration`, plus the `FAQ`.
- **Renamed / removed API**: `ErrorGeneratorPropagation` uses `random_CPTP_error_generator_rates`; `Protocols` drops the deprecated `wildcard1d_reference` kwarg; `LowLevel` updates `run_iterative_gst` result unpacking; `WorkspaceExamples` uses the `mx_basis_x` kwarg.
- **numpy 2**: `1QGST-InterpolatedOps` and the `InterpolatedOperators` example updated (the latter also drops the deprecated `vec`/`unvec` helpers from the `interpygate` namespace).
- **Self-contained / robustness**: `DriftCharacterization` made self-contained; `MPI-RunningGST`, `MPI-RunningGST-Automagic`, `MCFE-Example`, and `CircuitSimulation-CHP` made runnable; `TimeDependent` switched to an `ObjectiveFunctionBuilder` and a cache-free simulator; `ParityBenchmarking` defaults to the Clarabel solver instead of SCS.
- **`IBMQExperiment`**: the four cells that build a `ProcessorSpec` / experiment design from a live IBM Quantum backend (`device`, `pspec`, `widths`, `compilations`, …) are now tagged `nbval-skip`, matching the credentialed cells around them. Previously these cells were left runnable and cascaded into `NameError`s under `nbval` because the cell that defines `device` (which requires `service.least_busy()` — i.e. real credentials and network) is itself skipped. The notebook's narrative is unchanged; it simply no longer fails CI for want of an IBMQ account.
- **Removed `2QGST-ErrorBars`**: complicated to test and maintain, not well developed, and obviated by the `ProceduralErrorBars` example. The TOC and FAQ now point at `ProceduralErrorBars` instead.

## Library bugfixes (surfaced by the notebooks)

Each item below was a real bug found by running a tutorial; the affected tutorial is named.

- **`pygsti/algorithms/compilers.py`** — several fixes in the conditional-symplectic / CNOT compilation path so the `DirectRB` example runs: a `gatesring=` → `layer_labels=` typo, a stale-variable bug (`s` → `sout`), `.copy()` → `.copy(editable=True)`, and a `try/except` fallback from the connectivity-adjusted (CO)iCAGE compiler to the all-to-all Gaussian-elimination compiler when the former silently produces an incorrect circuit on restricted geometries. Filed as [#787](#787); covered by new tests in `test/unit/algorithms/test_compilers.py` and `test/unit/protocols/test_rb.py`.
- **`pygsti/algorithms/core.py`** — preserve op-label aliases in the precomputed layout cache, so circuits containing aliased labels don't raise a `KeyError` during dataset lookup (fixes `ContextDependence`).
- **`pygsti/circuits/circuit.py`** — allow tensoring empty idle layers in `tensor_circuit_inplace`; an empty idle layer (`Label(())`) carries no gates and is always safe to tensor, but it was tripping the "implicit sslbls matter" guard that `SimultaneousExperimentDesign` relies on (fixes `VolumetricBenchmarks`, `MultiRBExperiments`).
- **`pygsti/models/modelconstruction.py`** — canonicalize non-string op labels through `Label` before parsing (fixes `CircuitSimulation`, `ExplicitModel`), and normalize string Lindblad error-generator keys like `"HX"` to the tuple form `("H", "X")` in the local-noise path (fixes `IdleTomography`).
- **`pygsti/models/model.py`** — `split_circuits` now returns the ops circuit (rather than `None`) for prep-less circuits (fixes `DataSimulation`).
- **`pygsti/models/explicitmodel.py`** — restore the `num_modeltest_params` setter that was dropped when the getter was overridden (fixes `ModelTesting`).
- **`pygsti/modelmembers/povms/__init__.py`** — in CPTP POVM conversion, iterate over the constructed `base_povm` rather than `base_items`, which is left unset in the `ComputationalBasisPOVM` "easy case" branch (fixes `ModelParameterization`).
- **`pygsti/drivers/longsequence.py`** — replace the `assert` guarding mutually-exclusive `gauge_opt_suite_name` / `gauge_opt_params` with a `ValueError` whose message explains the common cause (overriding `gauge_opt_params` while leaving `gauge_opt_suite_name` at its `'stdgaugeopt'` default), and type `gauge_opt_suite_name` as `Optional[str]`.
- **`pygsti/evotypes/chp/__init__.py` / `povmreps.py`** — make the `chp` evotype auto-detect the `chp_sim` Python package (used by `CircuitSimulation-CHP`) and improve the "no chp executable configured" error message; the CHP test was also made robust to `chp_sim` being absent.
- **`pygsti/extras/idletomography/idtreport.py`** — pass `title="auto"` rather than a `{'text': "auto"}` dict to the observed-rate plot (fixes the `IdleTomography` report).

## Verification

- **Cross-notebook fixtures are sufficient.** With `docs/tutorial_files`/`docs/example_files` cleaned to empty, the pure *consumer* notebooks — `objects/Results`, `gst/Driverfunctions`, `gst/LowLevel`, `gst/Protocols`, `reporting/ReportGeneration` — were run together under the exact CI harness (`pytest -n 2 --nbval-lax --dist loadscope --nbval-current-env`) with no generator notebook present. All passed (**67 passed, 1 skipped**), confirming `conftest.py` alone provides every input they need.
- nbval was confirmed to execute each notebook with its working directory set to the notebook's own directory, so the `../../tutorial_files` / `../../example_files` relative paths resolve correctly even though `pytest` is invoked from the repo root.
- **Full suite.** The complete CI command (`pytest -n 2 --nbval-lax --dist loadscope --nbval-current-env docs/markdown`, all 91 notebooks) was run end to end on a memory-constrained dev box (~4 GB free). Result: **1055 passed, 44 skipped, 5 failed**. The five failures fell into exactly two buckets, both since resolved or explained:
  - **`IBMQExperiment` (4 cells)** — the orphaned-cell `NameError`s described above. Fixed by the `nbval-skip` tags; the notebook now reports `3 passed, 24 skipped`.
  - **`MirrorRB-Universal-Gate-Sets` (1 cell)** — the RB-fit cell hit nbval's default 2000 s cell timeout under memory contention (two heavy GST/RB notebooks sharing ~4 GB with no swap). Run in isolation the same cell completes in **1.2 s** (whole notebook: 6 passed in 28 s), so this is a resource-starvation artifact of the constrained 2-worker local run, not a code or notebook problem. CI runs the suite on a single Linux job with `-n auto` on a runner with ample RAM, where this does not occur.

  Net: with the IBMQ fix in place, the suite is green; the only non-pass is an environment artifact that does not reproduce with adequate memory.

Environment notes for reproducing locally: the suite needs `pygsti[testing,pytorch,docs]`, an MPI runtime (`mpiexec` + `mpi4py`) for the MPI examples, and outbound network access (the CHP notebook `%pip install`s `chp_sim`; the MPI `!mpiexec` cell is `nbval-skip`-tagged). When verifying a fix, remember that almost everything under `docs/tutorial_files` and `docs/example_files` is gitignored — delete *all* untracked files there between runs, not just the ones `git status` shows.

## Follow-ups / notes for reviewers

- The `compilers.py` fallback is a safety net, not a root-cause fix: the underlying (CO)iCAGE bug on restricted connectivity ([#787](#787)) still deserves its own fix. The inline comment there also references a workspace-local audit note (`active-project/direct_rb_detour_audit.md`) that won't exist for other contributors — worth rewording before merge.
- CI still compiles the external CHP binary, but the `CircuitSimulation-CHP` notebook now relies on the `chp_sim` Python package (auto-detected), so the binary is effectively a fallback. It can stay or be dropped in a later cleanup.


rileyjmurray:
Failing build is due to data race in compiling Cython extensions.

docs/markdown/gst/Driverfunctions.md
rileyjmurray, line 1: Claim in this file needs validation

docs/markdown/gst/Protocols.md
rileyjmurray, line 1: Need to special-case whether this file is run / in what order.

docs/markdown/gst/TimeDependent.md
rileyjmurray, line 1: Need to validate claim here.

docs/markdown/objects/CustomOperator.md
rileyjmurray, line 1: Need to validate claim.

docs/markdown/objects/Results.md
rileyjmurray, line 1: Need to special-case whether this file is run / in what order.

docs/markdown/protocols/DriftCharacterization.md
rileyjmurray, line 1: Changes here should be scrutinized.

docs/markdown/reporting/ReportGeneration.md
rileyjmurray, line 1: Need to validate changes.

docs/markdown/utilities/GaugeOpt.md
rileyjmurray, line 1: Need to validate changes.

docs/conftest.py
rileyjmurray, line 1: Key infrastructure introduced in this PR.

pygsti/algorithms/compilers.py
rileyjmurray, line 1: Changes here are partial fixes for issue #787.

pygsti/algorithms/core.py
rileyjmurray, line 1: Change needed for ContextDependence tutorial.

pygsti/circuits/circuit.py
rileyjmurray, line 1: Needed to fix either VolumetricBenchmarks or MultiRBExperiments tutorial (possibly inclusive)

pygsti/evotypes/chp/__init__.py
rileyjmurray, line 1: Need to verify claim that hitting this function means `chp_sim` is not installed.

pygsti/evotypes/chp/povmreps.py
rileyjmurray, line 1: One of this PR's main code changes.

pygsti/extras/idletomography/idtreport.py
rileyjmurray, line 1: Change needed for IdleTomography tutorial to run

pygsti/extras/interpygate/__init__.py
rileyjmurray, line 1: This PR deliberately removes the deprecated `vec`/`unvec` functions.

pygsti/extras/interpygate/process_tomography.py
rileyjmurray, line 1: This PR deliberately removes these functions. Equivalent code has been in-lined for the Interpygate example notebook.

pygsti/extras/paritybenchmarking/disturbancecalc.py
rileyjmurray, line 1: Change default convex optimization solver from SCS to Clarabel.

^Just a note.

pygsti/modelmembers/povms/__init__.py
rileyjmurray, line 1: This is a plain and simple bugfix. Iterating over a bare `base_items` only caught its _keys_, while we need to iterate over (key, value) pairs.

pygsti/models/model.py
rileyjmurray, line 1: Change needed for DataSimulation tutorial

pygsti/models/modelconstruction.py
rileyjmurray, line 1: Change needed for IdleTomography tutorial

docs/markdown/gst/Driverfunctions.md
rileyjmurray: Validated. Updated error message in commit 1ed676fa4b27cc03b26042d1ea09337ca869c0a3.

docs/markdown/gst/TimeDependent.md
rileyjmurray: Upon further review the claim is obviously true.

docs/markdown/objects/CustomOperator.md
rileyjmurray: Validated. Added a clarifying comment in e2bad0b46c6a4394e3a14eb87769ca8ad8b45333.

docs/markdown/protocols/DriftCharacterization.md
rileyjmurray: The changes are significant in order to make the notebook standalone. It previously relied on running the `objects/TimestampedDataSet` tutorial first.

docs/markdown/reporting/ReportGeneration.md
rileyjmurray: Confirmed the changes are appropriate. The thing that stood out was using the estimate_label part of advanced_options.

docs/markdown/utilities/GaugeOpt.md
rileyjmurray: Validated.

pygsti/evotypes/chp/__init__.py
rileyjmurray: I've decided it's just a bad error message. I've changed it in 35dbcdd82986aef92f8c391cb25b8555af447bc5.

rileyjmurray:
I've audited carefully. Will merge once tests pass. Note that the tests run in a PR don't include notebooks. If any such tests fail (unlikely), they'll do so on the beta-master runner.

PR #789, opened 2026-06-18
Error Generator Polynomials

coreyostrove:
### Summary

This PR adds new methods for performing error generator propagation with symbolic error generator rates, outputting polynomial representations of most of the core objects produced by the existing error generator propagation code.

### Major Changes

1. A new module, `errgenpolytools`, which contains the bulk of the functionality for producing error generator polynomials. Support includes:
  1. Symbolic magnus expansions up to second-order.
  2. Symbolic taylor expansion to arbitrary order.
  3.  Probability and observable expectation value corrections as polynomials.
2. Helper functions in `errgenpolytools` for mapping between the internal data structures used by the existing `Polynomial` class we're leveraging and those of the error generator propagation module.
3. `errorgen_gate_contributors` relocated from a method of `ErrorGeneratorPropagator` to a function in errgenpolytools.py. This will break some user's code (why I am marking it as major), but should be very easy for users to patch. 

### Minor Changes

1. Full bugfix for issue #751, an issue where related coefficients that should have been considered equivalent under commutativity were being silently dropped. 
2. New `Polynomial` methods:
  1.  `to_string` method that allows for overriding variable labels with custom user-specified ones.
  2.  `sum` method for more efficiently adding multiple polynomials together.
  4. New `from_variable_and_coefficient_lists` for instantiating Polynomials from a list of variable monomial terms and coefficients.
  5. Basic `__eq__` method.
3. New option for constructing `ErrorGeneratorPropagator`s using a fixed error generator layer structure rather than pulling one in from a model. 

In addition, a new set of unit tests have been added covering the error generator polynomial code. A new tutorial has been added (and tested) covering this functionality, and the `ErrorGeneratorPropagation.md` tutorial has been updated to demonstrate the new fixed error generator layer instantiation.

The github diff for the ErrorGeneratorPropagator is kind of wonky looking, sorry about that. 

### TODO
I need to add type annotations. I pinky promise these will get added before a merge and hope the review gods will overlook this in the meantime.

### Things to ignore

If you're reading the diff before PR #755 is merged in (and develop subsequently merged into this branch), please ignore the changes to `lindbladtools.py`, `lindbladcoefficients.py`, `lindbladerrorgen.py`, `test_lindbladcoefficients.py`, `test_unconstrained_coefficients.py`. These are all either related to PR #755, or a small PR that I will be following that one up with based on the changes on the `lindbladblock-simplification` branch.


docs/markdown/simulation/ErrorGeneratorPolynomials.md
rileyjmurray, line 1: Please update `docs/_toc.yml` to include an entry for this file.

rileyjmurray:
From Corey's initial comment
> If you're reading the diff before PR https://github.com/sandialabs/pyGSTi/pull/755 is merged in (and develop subsequently merged into this branch), please ignore the changes to lindbladtools.py, lindbladcoefficients.py, lindbladerrorgen.py, test_lindbladcoefficients.py, test_unconstrained_coefficients.py. These are all either related to PR https://github.com/sandialabs/pyGSTi/pull/755, or a small PR that I will be following that one up with based on the changes on the lindbladblock-simplification branch.

There are still changes to two of these files.
 * tools/lindbladtools.py. Type annotations and docstring updates. Clearly worth keeping.
 * tools/test_lindbladcoefficients.py. Remove some assertions and change from parameterizations from "static" to "elements." Could use scrutiny for whether these changes are in fact what we want.

rileyjmurray:
@coreyostrove, in case you're curious, here's the review from codex (GPT 5.5). It was based on an earlier review by Claude Opus 4.8.

--------------------------------

The branch's current net contribution is now focused on the **error-generator to polynomial** feature:

- new `pygsti/tools/errgenpolytools.py`,
- `Polynomial` extensions and representation canonicalization,
- fixed-layer support in `errorgenpropagation/errorpropagator.py`,
- two simulation tutorials, and
- new unit tests for `errgenpolytools`.

The core symbolic math still looks solid: the targeted tests compare first/second-order symbolic Magnus, Taylor, probability-correction, and Pauli-expectation results against the numeric `errgenproptools` references and pass. The remaining concerns are API-contract regressions, executable tutorial breaks, fixed-layer coverage/limitations, and a few cleanup/test gaps.

**Overall recommendation:** approve with changes. The approach is fine; the branch should fix the tutorial execution issue and make explicit decisions about `Polynomial.__eq__`/hashability, fixed-layer matrix support, and rep parity before merge.

## Triage at a glance

### Riley pushes directly to the branch

| # | Item | File | Sev |
|---|------|------|-----|
| P1 | Propagation tutorial still calls removed instance method `errorgen_propagator.errorgen_gate_contributors(...)`; change it to `epoly.errorgen_gate_contributors(error_model, ..., include_spam=True)` and import/use `errgenpolytools` in that tutorial. | `docs/markdown/simulation/ErrorGeneratorPropagation.md:126` | major |
| P2 | `r"""` the Magnus docstring. With `-W error::SyntaxWarning`, `py_compile` fails on `invalid escape sequence '\s'`. | `pygsti/tools/errgenpolytools.py:457` | minor |
| P3 | Remove unused `typing` imports (`Literal`, `Optional`, `Union`, `Callable`, `Iterable`). | `pygsti/tools/errgenpolytools.py:36` | minor |
| P4 | Fix wrong/placeholder return docstrings. `_second_order_magnus_term_symbolic_polynomial` still says `TBD`; `_error_generator_layer_pairwise_commutator_symbolic_polynomial` documents intermediate lists rather than the returned dict. | `pygsti/tools/errgenpolytools.py:478,541-551` | minor |
| P5 | `callable` -> `typing.Callable` in `Polynomial` type hints. | `pygsti/baseobjs/polynomial.py:539,572` | minor |
| P6 | Delete stale commented-out `map_indices_inplace` code. | `pygsti/baseobjs/polynomial.py:413-434` | nit |
| P7 | `fixed_rate` truthiness -> `is not None`, matching the model-backed path and preserving `fixed_rate=0`. | `pygsti/errorgenpropagation/errorpropagator.py:619` | minor |
| P8 | Constructor docstring/comment typos: `fixed_errorgen_layers` plural and "an if it is a GEEL". | `pygsti/errorgenpropagation/errorpropagator.py:54,78` | nit |
| P9 | Fix tutorial prose/import nits: `_LSE` is unused in `ErrorGeneratorPolynomials.md`; prose says observable `ZZZZ` but code uses `IIXZ`; several setup paragraphs are garbled. | `docs/markdown/simulation/ErrorGeneratorPolynomials.md:43,57,90,97,348` | minor/nit |
| P10 | Remove or correct stale TODO saying the fixed-layer `.copy()` is probably unnecessary; the copies avoid cross-layer dictionary aliasing. | `pygsti/errorgenpropagation/errorpropagator.py:618` | nit |

### Leave as review comments for Corey

| # | Item | File | Sev |
|---|------|------|-----|
| C1 | `Polynomial.__eq__` makes the class unhashable and crashes on non-`Polynomial` operands. | `pygsti/baseobjs/polynomial.py:674` | **major** |
| C2 | Python fallback vs C++/Cython `PolynomialRep.add_inplace` still have different zero-pruning semantics; this is now load-bearing through `Polynomial.sum` and exact `__eq__`. | `pygsti/evotypes/basereps.py:300` vs `pygsti/evotypes/basecreps.cpp:77` | major |
| C3 | Fixed-layer propagator is silently unusable with matrix output paths because `errorgen_layer_dict_to_errorgen` assumes `self.model` exists. | `pygsti/errorgenpropagation/errorpropagator.py:756` | major |
| C4 | No tests construct `ErrorGeneratorPropagator(fixed_errorgen_layer=...)`; the tutorial documents it, but constructor casting, width validation, SPAM layer counts, and global-label support are untested. | `test/unit/objects/test_errorgenpropagation.py` | major |
| C5 | `errorgen_gate_contributors` was removed from `ErrorGeneratorPropagator`'s public API and reintroduced as `errgenpolytools.errorgen_gate_contributors(model, ...)` with a new first positional argument. | `pygsti/errorgenpropagation/errorpropagator.py`, `pygsti/tools/errgenpolytools.py:248` | major |
| C6 | `max_num_vars = len(errorgen_to_var_map)` oversizes polynomials in gate-aggregated paths. In the reviewed fixture this is 222 map entries but 36 distinct variable indices, reducing packing density from 12 to 8 variable indices per integer. | `pygsti/tools/errgenpolytools.py` | minor/perf |
| C7 | Test gaps remain: no second-order aggregated Magnus check, no quadratic-polynomial Taylor input, bulk-vs-scalar tests do not independently compare bulk to numeric references, and no direct `Polynomial` unit tests for `sum`/`to_string`/`__eq__`/canonicalization. | `test/unit/tools/test_errgenpolytools.py`, `test/unit/objects/test_polynomial.py` | minor |
| C8 | Order-1 fast paths alias the input dicts. No downstream mutation currently observed, but the behavior is surprising and the TODO is unresolved. | `pygsti/tools/errgenpolytools.py:614,864` | nit |
| C9 | Parallel-dict ordering coupling remains in the commutator construction (`zip(d.items(), other.values())`). Correct today because the dicts are built together, but fragile. | `pygsti/tools/errgenpolytools.py:579` | nit |
| C10 | User-facing validation uses `assert`, which is stripped under `python -O`. This matches some house style, so it is a policy call. | `pygsti/errorgenpropagation/errorpropagator.py:69-91` | nit |

---

## Detailed findings

### 1. Polynomial core and reps

The branch makes polynomial monomials commutative by sorting variable indices during integer encoding and adds `Polynomial.sum`, `from_variable_and_coefficient_lists`, `to_string`, and `__eq__`. The canonicalization is present in the Python object layer, Python fallback rep, Cython rep, and C++ rep.

- **[C1 - major, verified] `__eq__` breaks hashability and crashes on foreign operands** - `pygsti/baseobjs/polynomial.py:674`

  ```python
  def __eq__(self, other):
      return self.coeffs == other.coeffs
  ```

  Defining `__eq__` without `__hash__` sets `__hash__ = None`. Reproduced in `py313`:

  ```text
  hash(p) -> TypeError: unhashable type: 'Polynomial'
  p == 5 -> AttributeError: 'int' object has no attribute 'coeffs'
  p == None -> AttributeError: 'NoneType' object has no attribute 'coeffs'
  ```

  On `develop`, `Polynomial` had identity hashability. If unhashability is intended, it should be explicit and foreign comparisons should return `NotImplemented`. If hashability should be preserved, define an appropriate `__hash__` policy.

- **[C2 - major] Pure-Python and compiled reps still differ on zero pruning** - `pygsti/evotypes/basereps.py:300` vs `pygsti/evotypes/basecreps.cpp:77`

  The Python fallback `add_inplace` retains exact cancellation as a `0j` coefficient; the C++ path erases keys with `abs(newval) <= 1e-12` and skips near-zero additions. In the active `py313` compiled path, `Polynomial.sum([x, -x])` returns `{}` and compares equal to the zero polynomial. The source fallback still has different behavior, and the branch now exposes that difference through `Polynomial.sum` plus exact `__eq__`.

- **[P5/P6] Minor cleanup:** use `typing.Callable` rather than builtin `callable` in type hints, and remove the stale commented-out `map_indices_inplace` block.

- **[C7] Test gap:** `test/unit/objects/test_polynomial.py` is not updated. The new equality/hash/canonicalization behavior deserves direct tests, ideally with a forced Python fallback or rep-level parity test because C2 is exactly a compiled-vs-fallback risk.

### 2. `errgenpolytools`

This is the main new module. The implemented symbolic constructions match the numeric reference in the targeted tests. The current suite covers unaggregated first/second-order Magnus, first/second-order Taylor from linear inputs, scalar probability and Pauli-expectation corrections, and bulk-vs-scalar consistency.

- **[P2 - minor, verified] Syntax warning promoted to error** - `pygsti/tools/errgenpolytools.py:457`

  `conda run -n py313 python -W error::SyntaxWarning -m py_compile pygsti/tools/errgenpolytools.py` fails on the docstring line containing `\sum`. Make the docstring raw.

- **[C6 - minor/perf, verified by probe] Aggregated paths oversize `max_num_vars`.**

  Several polynomial constructions use `max_num_vars=len(errorgen_to_var_map)`. In gate-aggregated maps, many keys share the same variable index. In the reviewed fixture, the map has 222 entries but only 36 distinct variable indices:

  ```text
  Polynomial._vindices_per_int(36)  -> 12
  Polynomial._vindices_per_int(222) -> 8
  ```

  That is a correctness-safe packing-density loss. A canonical helper such as `max(errorgen_to_var_map.values()) + 1` should be used consistently anywhere polynomials built from the same map need to add/multiply.

- **[P3/P4] Cleanup:** remove unused typing imports and repair return docstrings.

- **[C8/C9] Latent maintainability hazards:** the order-1 fast paths return/alias input dicts, and `_error_generator_layer_pairwise_commutator_symbolic_polynomial` relies on parallel dict insertion order via `zip(commuted_errorgen_var_dict.items(), commuted_errorgen_coeff_dict.values())`.

- **[C7] Test gaps:** the suite is good where present, but it still does not cover second-order aggregated Magnus, quadratic polynomial Taylor inputs, independent numeric references for the bulk APIs, `magnus_order=3` validation, or the aggregated non-local `RuntimeError` branch.

### 3. `ErrorGeneratorPropagator`

The branch adds fixed-layer construction and moves `errorgen_gate_contributors` out of the class into `errgenpolytools`. Existing model-backed tests still pass.

- **[C3 - major] Fixed-layer matrix paths are broken** - `pygsti/errorgenpropagation/errorpropagator.py:756`

  In fixed-layer mode, `self.model is None`, but `errorgen_layer_dict_to_errorgen` does:

  ```python
  num_qubits = len(self.model.state_space.qubit_labels)
  ```

  Consequently `eoc_error_channel` and direct matrix conversion are unusable in fixed-layer mode. Either derive `num_qubits` from the labels/circuit, or explicitly guard and document fixed-layer mode as symbolic/propagation-map only.

- **[C4 - major] No unit tests for fixed-layer mode.**

  `test/unit/objects/test_errorgenpropagation.py` still never constructs `ErrorGeneratorPropagator(fixed_errorgen_layer=...)`. The tutorial exercises the path informally, but CI does not check local-label construction, global-label casting via `state_space_labels`, width validation, empty-layer behavior, SPAM layer counts, or copy/no-aliasing behavior.

- **[C5 - major, verified] Public API removal.**

  `hasattr(ErrorGeneratorPropagator, "errorgen_gate_contributors")` is `False` in `py313`. The replacement free function is:

  ```text
  errgenpolytools.errorgen_gate_contributors(model, errorgen, circuit, layer_idx, include_spam=True, return_operators=False)
  ```

  All in-repo tests were migrated, but external callers and the propagation tutorial still break. Consider a deprecated forwarding shim, or at least release-note the API change.

- **[P7/P8/P10] Small cleanup:** `fixed_rate=0` should be honored via `is not None`; fix constructor typos; remove or correct the misleading `.copy()` TODO.

### 4. Tutorials

The docs live in `docs/markdown/simulation/` as jupytext sources. The earlier stale `random_error_generator_rates` issue is fixed: both tutorials now import `random_CPTP_error_generator_rates`.

- **[P1 - major, verified] `ErrorGeneratorPropagation.md` still will not execute** - line 126 calls:

  ```python
  errorgen_propagator.errorgen_gate_contributors(...)
  ```

  That method no longer exists. The tutorial should import `errgenpolytools` and call `epoly.errorgen_gate_contributors(error_model, ..., include_spam=True)`.

- **[P9 - minor/nit] `ErrorGeneratorPolynomials.md` doc polish:** `_LSE` is imported but unused; the prose says observable `ZZZZ` while the code uses `IIXZ`; the setup paragraphs around lines 57, 90, and 97 have grammar/wording problems.


## Suggested next actions

1. Land P1-P10 as a small review-cleanup commit.
2. Ask Corey for explicit decisions on C1, C2, C3, and C5.
3. Add focused tests for fixed-layer propagation and direct `Polynomial` behavior before merge.

test/unit/objects/test_errorgenpropagation.py
rileyjmurray, line 1: Claude's description of these tests.

**Fixed-layer `ErrorGeneratorPropagator` (new `FixedLayerErrorgenPropTester`)**

- local-label construction → keys cast to `LocalStimErrorgenLabel`, rates preserved.
- global-label construction with `state_space_labels` as int and as list.
- constructor validation: both model+fixed → AssertionError; neither → AssertionError;
  GEEL without `state_space_labels` → AssertionError; mixed-width labels → AssertionError.
- circuit width mismatch in `construct_errorgen_layers` → AssertionError.
- SPAM layer counts: `include_spam=True` → `len(circuit)+2`, `False` → `len(circuit)`
  (empty circuit: 2 / 0; depth-2 circuit: 4 / 2).
- each layer equals the stored fixed layer; per-layer `.copy()` keeps layers and the stored
  fixed layer from aliasing (mutating one layer leaves the others + original untouched).
- empty fixed layer (`{}`) constructs and yields `[{}, {}, ...]`.
- `fixed_rate` override: `None` keeps rates {0.01,0.02}; `1` → all rates 1; `0` → all rates 0.

**Fixed-layer matrix path is BROKEN (confirmed, captured as xfail)**

`errorgen_layer_dict_to_errorgen` (errorpropagator.py:756) does
`num_qubits = len(self.model.state_space.qubit_labels)`, but `self.model is None` in
fixed-layer mode. Empirically both `errorgen_layer_dict_to_errorgen` and `eoc_error_channel`
raise `AttributeError: 'NoneType' object has no attribute 'state_space'`.
Captured by `test_fixed_layer_matrix_path_should_work_C3` marked `@unittest.expectedFailure`
(asserts the desired `4**n x 4**n` channel). It flips to "unexpected success" when C3 is fixed,
prompting removal of the marker. This is an author decision (derive num_qubits from labels/circuit,
or guard + document fixed-layer as symbolic/propagation-map only) — not fixed here.

test/unit/objects/test_polynomial.py
rileyjmurray, line 1: Claude's description of changes here.

**Polynomial direct tests (new `PolynomialBranchFeatureTester`)**

- monomial canonicalization: `{(1,0):2}` → key `(0,1)`; equality invariant to var order.
- `sum`: accumulation, empty→zero, single→equal, exact cancellation (asserted on the *value*,
  not coeff dict, to stay agnostic to the C2 compiled-vs-fallback zero-pruning divergence),
  mismatched `max_num_vars` → AssertionError.
- `from_variable_and_coefficient_lists`: repeated monomials summed; length mismatch → AssertionError.
- `__eq__`: unequal distinguished; foreign operands (int, None) return False (old C1 crash gone).
- `__hash__`: raises RuntimeError; unusable as dict key.
- `to_string`: default labels, zero poly, complex coeff, powers — all pass.

test/unit/tools/test_errgenpolytools.py
rileyjmurray, line 1: Claude's description of changes here.

**errgenpolytools gaps (added to `ErrgenPolyToolsTester`)**

- bulk probability correction (order 1 & 2) and bulk Pauli-expectation (order 1) now compared
  *independently* to the numeric `errgenproptools` references (the existing bulk tests only
  checked bulk == scalar-symbolic).
- second-order aggregated Magnus vs numeric BCH-2 on the shared-parameter model
  (only first-order aggregated was covered).
- order-2 Taylor from a genuinely *quadratic* polynomial input (square two first-order Magnus
  polys → degree-2 values → degree-4 composed terms) vs the numeric reference.
  NOTE: the obvious approach (feed the full `second_order_magnus_polys`) OOMs — order-2 Taylor
  builds `product(keys, repeat=2)`, so the large key set blows up combinatorially. Restricting
  to 2 error generators with quadratic values exercises the higher-degree encoding without
  the explosion. (Real scaling hazard worth a separate note.)
- `magnus_order=3` and Taylor `order=3` both → AssertionError (validation).

pygsti/tools/errgenpolytools.py
```diff
@@ -0,0 +1,871 @@
...
+                aggregated_error_generator_indices_by_gate[gate_name_model_indices_tup] = dict()
+            
+            #break the error generators into equivalence classes based on truncations, enforcing locality.
+            try:
+                truncated_errorgen = _truncate_lse_support(errorgen, gate_sslbls, validate_locality=True)
+            except RuntimeError:
+                msg = "An error generator with nonlocal support relative to the gate with which it is associated has been detected."
+                msg += " Aggregation of error generators by gates and shared model parameters together is only supported models with local errors "
+                msg += "(those with support restricted to the target qubits of the gate in question) at present."
+                raise RuntimeError(msg)
```
rileyjmurray, line 189: Claude is under the impression that the codepath with the re-raise is unreachable. Reasoning as follows.

> Implicit non-LocalNoise models are rejected earlier (line 114), and LocalNoiseModel / Explicit models have gate-local errors by construction, so no errorgen is non-local relative to its contributing gate. The underlying locality check (`_truncate_lse_support(..., validate_locality=True)`) is already tested (`test_truncate_lse_support_with_locality_validation_raises`).

rileyjmurray (commented):
I've pushed several commits with minor fixes. See individual file comments below. I've flagged files where Claude wrote additional tests. Please verify those tests for appropriateness.

There's a test in [test/unit/objects/test_errorgenpropagation.py](https://github.com/sandialabs/pyGSTi/pull/789/files/be30a9c6708296746b95955c0d9e57445e5da354#diff-ae6a4db9673ddc99fb81e7994a120d1058481bc3a852b98bc3d77108bc6288b9) that's especially important, since it's currently marked as expecting a failure. If you agree that this is a problem, please fix the source. If you disagree that it's a problem, feel free to indicate as much.

jordanh6:
FYI I'm taking a read of the code today for correctness/conceptual completeness.

jordanh6 (approved):
I didn't find any major gaps in the code; looks conceptually correct/complete to me. Nice work, Corey! :)

rileyjmurray:
Pushed changes to resolve a merge conflict.

docs/markdown/simulation/ErrorGeneratorPolynomials.md
rileyjmurray: Nevermind; I've pushed a suitable change.

pygsti/tools/errgenpolytools.py
coreyostrove: I agree with Claude's reasoning. It would not be possible to hit the except block. As it is currently innocuous I'm going to leave it as is so that the next time I start messing with this code and breaking contracts (which I plan to do when I add cloud noise model support) it is loud about where I need to patch things.

coreyostrove: I will add a note, though.

coreyostrove:
Waiting on tests to pass, but I think I have all of your feedback addressed and have reviewed the proposed changes you added on top of the original PR. Specific changes:

- Added type annotations, as promised.
- Reviewed new unit tests. Kept most of these but removed the following:
  -  The new bulk correction polynomial tests didn't exercise any new code paths compared to the original tests, so removed for efficiency.
  - Unit test related to taylor series polynomials greater than order 3 raising an exception was deleted. This was a correct parsing of a mistaken/outdated assertion that was left over from early development. order>3 works with the current implementation.
- Fixed bug in construction EOC channel numpy arrays, and removed xfail marking from the related test.
- Added note about not being able to hit except block of on a codepath that you called out above. Keeping this as is (see above), but comment *was* correct.

Let me know if you want to take another pass, @rileyjmurray. Otherwise I will merge once the tests are passing (I'll wait a bit so you have time to interject).

PR #790, opened 2026-06-18
Add regression tests for basis.py

nkoskelo:
Add regression tests for basis.py.

- Ensure that any basis object is recoverable and equivalent to original basis object when going through the serialization-deserialization pipeline.
- Add a regression test for equality that checks for equivalence up to a numerical tolerance.
o The hash should only be equal if they match exactly.
- Add a regression test to not copy the basis if the sparsity flags match on a call to `with_sparsity`.
- Add regression test for lazy build of the basis.


nkoskelo:
@coreyostrove or @rileyjmurray this is ready for your review when you get the chance.

rileyjmurray (approved):
Looks awesome!

PR #792, opened 2026-06-22
Support for numpy 2.5: address deprecation warnings

rileyjmurray:
Our CI builds are setup to error if we have an un-suppressed deprecation warnings. Numpy 2.5 deprecated manually setting the `shape` field of an array (a-la `x.shape = newshape`). We've used this pattern in many places within pyGSTi, and so CI builds started to fail after the numpy 2.5 release.

This PR replaces the explicit shape-assignment with the pattern `x = reshape_no_copy(x, newshape)`, where `reshape_no_copy` is a helper function that accounts for our support of numpy 1.x in earlier versions of Python.

The location of the helper function is in `baseobjs/_compatibility.py`. This is a pre-existing file. I'd have put it elsewhere if this file didn't already exist.

### DIFF
--- a/pygsti/algorithms/gaugeopt.py
+++ b/pygsti/algorithms/gaugeopt.py
@@ -16,6 +16,7 @@ import warnings as _warnings
 import numpy as _np
 
 from pygsti import baseobjs as _baseobjs
+from pygsti.baseobjs import _compatibility as _compat
 from pygsti import optimize as _opt
 from pygsti import tools as _tools
 from pygsti.tools.optools import relaxed_scalar_tolerance as _relaxed_tol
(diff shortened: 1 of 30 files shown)


rileyjmurray (commented):
These changes made by an agent. Commenting here to note that I've reviewed all of them.

coreyostrove:
Low context question (on phone so all I am seeing is the PR text). Why the
helper? Can't we just use the reshape method in these instances?

rileyjmurray:
@coreyostrove preserving old semantics requires calling reshape with `copy=False`. However, the `copy` kwarg was only added to numpy reshape in numpy 2.1. So in order to support numpy 2.0 and 1.x we use the old behavior: just assign to the `shape` field of the ndarray.

coreyostrove:
The issue is error handling? Did setting the shape directly raise a value
error when a copy was required? Otherwise reshape pre-2.1 is like ravel and
only makes a copy when needed.

rileyjmurray:
@coreyostrove, yes, setting the shape directly raised errors if a copy was required. Example
```python
import numpy as np
t = np.arange(12).reshape(3, 4).T 
try:
    t.shape = (12,)
except AttributeError as e:
    print(e)
```
Output
```
Incompatible shape for in-place modification. Use `.reshape()` to make a copy with the desired shape.
```

rileyjmurray:
Going to merge now since the need for the helper is clear. There are a few other PRs that need to be merged today so having CI unblocked is important.

PR #793, opened 2026-06-22
Fix the typos within comments

nkoskelo:
This is part of the effort to clean up the code base for ease of maintenance by humans and robots.

I am using the following tool to find the typos.

```
pip install typos
typos pygsti/*/*.py >> my_file.txt
```

To confirm that none of the changes are done within any active piece of code and only the comments I have the following script. Then, to confirm that the changes do not modify anything that might be in the specific dialect of pygsti (i.e. a comment referencing a variable and that variable looks like it might be misspelled ('fpr' for fiducial pair reduction)) see the other script.
[confirm_changes_in_comments.py](https://github.com/user-attachments/files/29219709/confirm_changes_in_comments.py)
[confirm_no_forbidden_changes.py](https://github.com/user-attachments/files/29219710/confirm_no_forbidden_changes.py)


rileyjmurray (approved):
While reviewing files one-by-one I found that a typo of "proportion" had been incorrectly mapped to "proposition" in a way that didn't respect context. This led me to run the diff through a robot and ask it to make changes as needed while considering context. It found 9 instances of misapplied corrections. I've pushed those back up to this branch. Approving now, then merging.

PR #803, opened 2026-06-27
Avoid eager cvxpy import during package import

basnijholt:
## Summary
- avoid importing the optional `cvxpy` dependency while importing `pygsti`
- lazy-load `cvxpy` only when SDP/diamond-norm helpers need it
- add a subprocess import test covering the optional-dependency side effect

## Example
```
python -c "import pygsti; import grpc"
# aborts: SetTimeZone() has already been called

python -c "import grpc; import pygsti"
# ok
```

## Motivation
`cvxpy` can initialize protobuf/Abseil-related native code through its solver stack. Importing it during `import pygsti` can make later imports of other packages that initialize the same native stack order-dependent. Since `cvxpy` is only needed for SDP/diamond-norm functionality, package import should not eagerly load it.

## Test
- `PYGSTI_NO_CYTHON_WARNING=1 python -m pytest test/test_packages/tools/test_imports.py -q`

PR #807, opened 2026-06-29
Bind lazy cvxpy import in SDP solver

basnijholt:
Follow-up to #803.

`solve_sdp` catches `cp.SolverError`, so it still needs a local `cvxpy` binding after the eager module-level import was removed. Without this, environments where the first SDP solver is unavailable fail with `NameError: name 'cp' is not defined` before trying the next solver.

Checks run:

```text
python -m pytest test/unit/tools/test_optools.py::GateOpsTester::test_diamond_distance test/unit/tools/test_optools.py::GateOpsTester::test_diamond_norm_epigraph test/unit/protocols/test_gst.py::GateSetTomographyTester::test_run_with_no_gaugeoptsuite test/test_packages/tools/test_imports.py -q
# 4 passed
```

Also verified `import pygsti` still leaves `cvxpy` unloaded before importing `grpc`.

### DIFF
--- a/pygsti/tools/sdptools.py
+++ b/pygsti/tools/sdptools.py
@@ -46,6 +46,7 @@ SDP_SOLVER_PRIORITY = ['MOSEK', 'CLARABEL', 'CVXOPT']
 
 
 def solve_sdp(prob: cp.Problem, **kwargs) -> tuple[np.floating, dict[str, np.ndarray]]:
+    cp = _get_cvxpy()
 
     objective_val : np.floating = np.array(np.nan).item()
     varvals : dict[str, np.ndarray] = dict()


rileyjmurray:
Hi @basnijholt. Thanks for the follow-up. As you noticed, I merged the PR without trying to trigger our CI runs. That would have required I make a copy of your branch on the pyGSTi repo, which seemed overkill for this. Of course, the egg is on my face since a bug slipped through.

I've done the right thing now and have triggered the full test suite. If it passes then I'll merge.

basnijholt:
Seems to all pass now at least 😅

rileyjmurray (approved):

basnijholt:
@rileyjmurray, thanks for merging. 

Any idea when to expect the next release?

rileyjmurray:
We're targeting a release of 0.10.0 on Friday. If we miss that mark then 0.10.0 will be delayed by a month. There are options for a bugfix release in that scenario. Let's discuss if that becomes relevant.

Issue #809, opened 2026-07-01
Stricter label prefixes on circuit parsing - desired behavior for QEC futureproofing?


sserita:
#738 (commit d96b2c3) further restricted the starts of state space labels to Q, T, and L. This actually breaks a chunk of my LoQS code where I have circuits with labels such as A* for auxiliary qubits and D* for data qubits.

I see two solutions:
1. We relax this new restriction back to users can use whatever sslbl prefix they want. However, pyGSTi only automatically knows what to do in cases like Q, where it assumes it is a qubit. Non-standard prefixes would need to explicitly define the appropriate state space, which was the old behavior.
2. You tell me, "Well, the old behavior was an oversight and this really is what the API should be, QTL or bust." In which case, I can implement a LoQS side fix where I have aliases from LoQS user labels (arbitrary) to pyGSTi labels (that only use Q).

Option 1 is obviously less work for me, but admittedly does not match StateSpace documentation. Option 2 is no work on pyGSTi side and some annoying bookkeeping on LoQS side. I guess the question is how often do we foresee LoQS-like use cases, where we want to denote the qubits by something more descriptive than Q?

As an aside, pyGSTi L is almost certainly for a leakage level. But it is common in LoQS to use L to denote a logical qubit (whereas A and D denoted auxiliary and data physical qubits in my above example). If we are futureproofing for QEC, this may be something we run into in the future. It could be of interest to do a multiscale model on a logical qubit where the physical qubits have a leakage level - what does L mean in that case?

**Action items:**
- [ ] Decide between Options 1 and 2 so that I can fix LoQS without pinning to before d96b2c3.
- [ ] Longer term discussion on what prefixes mean/make sense in the case where pyGSTi circuits are used to describe logical qubits with other conventions (e.g. A,D,L).

rileyjmurray:
This is a very good point. I support expanding allowed prefixes to include A and D. We can even do this before the 0.10 release if @coreyostrove is okay with it.

Our usage of L is weird. It's mentioned in the StateSpace tutorial: https://pygsti.readthedocs.io/en/develop-with-notebooks/markdown/objects/StateSpace.html as referring to a single level. But a single level can't define a meaningful (addressable) register-- quantum or classical. The labels used in circuit parsing are **line labels**, work only [need to] work for addressing. We don't have a first-class abstraction for line labels. In fact, there are many places in the code where we use term "state space labels" or the variable name "sslbls" in reference to line labels.

[...]

PR #810, opened 2026-07-01
Enable A/D line label prefixes.

sserita:
Short fix for #809 that enables the use of A and D prefixes on line labels for Circuits, primarily for use with labeling qubits as "auxiliary" or "data", respectively.

Also fixes a few unescaped \ characters in docstrings since these warn out in Python 3.14.


rileyjmurray (changes requested):
Thanks for the PR! Please update the slow circuit parser as well.

rileyjmurray (approved):

rileyjmurray:
Looks good to me! Please feel free to merge once tests pass.

PR #813, opened 2026-07-02
Handle type annotations in metaprogramming 

nkoskelo:
Calling a function cannot have type annotations. In `workspace.py:_makefactory` we do just that when we create a function which just calls the `__init__()` method of the tables in `workspace_tables.py`.


nkoskelo:
@rileyjmurray or @coreyostrove This should be good to merge after the tests pass.

rileyjmurray (approved):
My god this is terrifying code. Good on you for finding it and taking on the revisions. Love the new tests!

rileyjmurray:
Actually, @nkoskelo, please run comprehensive tests before merging. In addition to `test/unit`, that means

 * test/test_packages. Not sure if reports are exercised here.
 * test/integration. This generates some reports, so might trigger failures.
 * notebook tests. How to do this involves some funky command that I think is in the README.md. It's in our continuous integration scripts, so you can ask an agent to extract the command from there.

Notebook tests can take up to 1 hour.

nkoskelo:
> Actually, @nkoskelo, please run comprehensive tests before merging. In addition to `test/unit`, that means
> 
>     * test/test_packages. Not sure if reports are exercised here.
> 
>     * test/integration. This generates some reports, so might trigger failures.
> 
>     * notebook tests. How to do this involves some funky command that I think is in the README.md. It's in our continuous integration scripts, so you can ask an agent to extract the command from there.
> 
> 
> Notebook tests can take up to 1 hour.

- test/test_packages passes 228 tests, 1 skipped and 17 warnings. None of which appear to be new warnings.
- test/integration passes 7 tests, with 7 warnings. These appear to be the same warnings as one develop.
- The notebook tests pass now. Note that we need to run them in an environment with MPI.

PR #822, opened 2026-07-05
Declare import-time dependencies

basnijholt:
## Summary

Adds two dependencies to the core `pygsti` package metadata:

- `typing_extensions`, imported unconditionally by `pygsti.baseobjs.label`
- `matplotlib`, imported at package import time through `pygsti.report.fogidiagram`

This matches the current behavior of `import pygsti` and avoids downstream packaging workarounds for import-time `ModuleNotFoundError`s.

## Context

While updating the conda-forge feedstock to `pygsti 0.10`, the import test failed with:

```text
ModuleNotFoundError: No module named 'typing_extensions'
```

The feedstock now has a downstream workaround, but the root cause is that the upstream metadata does not declare all packages required by the top-level import path.

## Verification

- Confirmed current `pyproject.toml` metadata was missing `matplotlib` and `typing_extensions`.
- Ran a metadata check after the change confirming both dependencies are present.
- Ran `pip install --dry-run --ignore-installed --report /tmp/pygsti-pip-report.json .` in an isolated Python 3.10 build environment and verified the generated metadata includes both `matplotlib` and `typing_extensions` in `requires_dist`.


rileyjmurray:
Thanks for catching this, @basnijholt. I've changed the target branch from `master` to `develop`. If tests pass then I'll merge tomorrow. I expect to make a bugfix release 0.10.1 tomorrow by EOD, unless discussions with other pyGSTi devs advocate otherwise in internal discussions.

rileyjmurray (approved):
Note: the change to the macos deployment runner is because this branch has `master` as its base, which happened to be one commit ahead of `develop`. The PR now targets `develop`.

PR #827, opened 2026-07-06
Improve the test coverage for the verbosity printer.

nkoskelo:
There was an example of how to use the verbosity printer. I moved that to an integration test.


pygsti/baseobjs/verbosityprinter.py
```diff
@@ -568,27 +571,23 @@ def show_progress(self, iteration, total, bar_length=50, num_decimals=2, fill_ch
...
             if self.verbosity == self._progressStack.pop():
                 if last_progress_params is not None:
-                    (iteration, total, barLength,
-                     numDecimals, fillChar, emptyChar,
-                     prefix, suffix, indent) = last_progress_params
-                    progress = self._progress_bar(iteration + 1, total, barLength,
-                                                  numDecimals, fillChar, emptyChar,
-                                                  prefix, suffix, indent)
+                    iteration, *rest = last_progress_params
+                    progress = self._progress_bar(iteration + 1, *rest)
```
rileyjmurray, line 580: Slick; love this

test/integration/test_verbosityprinter_demo.py
rileyjmurray, line 1: I would have put this in test/unit. I can see how it's more of an integration test as far as VerbosityPrinter objects go. However, things in test/integration are notable because they integrate multiple (usually expensive) pyGSTi subsystems. This is cheap enough and small-enough scope that I think test/unit is a better home. You can just put it in the existing file within that directory.

rileyjmurray (changes requested):

PR #829, opened 2026-07-07
Update Advanced Options for Speed.

nkoskelo:
- Add test cases for AdvancedOptions.py
- Use a frozenset for the collection of valid keys instead of a tuple so that checking for membership can be done in O(1) time versus O(num_keys) time.


rileyjmurray:
From our discussion, I'm closing since the potential benefit of this change is not even worth thinking about.

PR #831, opened 2026-07-08
Add default on_space argument to rep-level to_dense (#806)

### LINKED ISSUE #806: to_dense() does not always have a default operator
Sometimes we call the `to_dense(self, on_space)` function without an argument. This can lead to problems if there is not a default argument for `on_space`.

Files to correct:

[pygsti/evotypes/statevec/opreps.pyx](https://github.com/sandialabs/pyGSTi/blob/60008609bde9c07d9386e25d2b6a8a1287d2a695/pygsti/evotypes/statevec/opreps.pyx#L107)
[pygsti/evotypes/stabilizer/effectreps.pyx](https://github.com/sandialabs/pyGSTi/blob/60008609bde9c07d9386e25d2b6a8a1287d2a695/pygsti/evotypes/stabilizer/effectreps.pyx#L68)
[pygsti/evotypes/densitymx/statereps.pyx](https://github.com/sandialabs/pyGSTi/blob/60008609bde9c07d9386e25d2b6a8a1287d2a695/pygsti/evotypes/densitymx/statereps.pyx#L55)
[pygsti/evotypes/densitymx_slow/statereps.py](https://github.com/sandialabs/pyGSTi/blob/60008609bde9c07d9386e25d2b6a8a1287d2a695/pygsti/evotypes/densitymx_slow/statereps.py#L54)
[pygsti/evotypes/densitymx/opreps.pyx](https://github.com/sandialabs/pyGSTi/blob/60008609bde9c07d9386e25d2b6a8a1287d2a695/pygsti/evotypes/densitymx/opreps.pyx#L95)

There may be others as well.

arpitjain099:
Resolves #806.

Several evotype representation classes declared `to_dense(self, on_space)` without a default value, so the fairly common no-argument call `rep.to_dense()` raised `TypeError`. This adds the `'minimal'` default that the modelmember-level `to_dense` API already uses, applied to the rep-level methods across the densitymx, statevec, stabilizer, and chp evotypes in both the compiled Cython and pure-Python variants. That covers the files pointed out in the issue (statevec/opreps.pyx, stabilizer/effectreps.pyx, densitymx/statereps.pyx, densitymx_slow/statereps.py, densitymx/opreps.pyx) plus the other reps that shared the same signature, as the issue anticipated. The single method that takes a required `outvec` positional after `on_space` is left unchanged, since it cannot be called with no arguments.

I verified against a fresh Cython build that the affected state, operation, and effect reps now return the same dense array from `to_dense()` as from `to_dense('minimal')`, where the no-argument form previously raised. A small regression test under test/unit/evotypes covers this, and the existing modelmember operation and spamvec unit tests still pass.


rileyjmurray:
Thanks for this contribution, @arpitjain099! Two requests for you.

First, can you please provide a statement here (just as a github comment) indicating that this contribution is under the terms of an Apache 2.0 license?

Second, can you please update changed lines of code to use the `SpaceT` type annotation in appropriate places?

arpitjain099:
Thanks @rileyjmurray. I've added the `SpaceT` annotation to the changed `to_dense` signatures in the Cython reps (the `.pyx` files under `pygsti/evotypes/`), so they now read `on_space: SpaceT = 'minimal'` and match what the pure-Python reps and the modelmember-level API already do. Each of those modules imports `SpaceT` from `pygsti.pgtypes`. The pure-Python (`.py`) reps in this PR already carried the annotation, so those were left as-is.

I rebuilt the extensions locally and the evotype tests pass (`test/unit/evotypes/`, plus the new regression test in test_rep_to_dense_default.py). The earlier red check looks like a transient setuptools_scm version-parse hiccup in the build env rather than anything from this change; the subsequent run came back green, and this push kicks off a fresh one.

rileyjmurray (approved):
This looks good to me. Can you please add a comment in this GitHub PR thread indicating that your contributions are under the terms of Apache 2.0? Once you leave such a comment I can merge.

arpitjain099:
My contributions are under the terms of Apache 2.0

@rileyjmurray comment added

rileyjmurray:
Merged! Thanks, @arpitjain099 :)

PR #833, opened 2026-07-08
Document no-op local array cleanup

### LINKED ISSUE #698: free_local_array not implemented, yet used within pyGSTi
**Describe the bug**
The function layouts/copalayouts.py:free_local_array does not do what is described in its docstring. To be more specific, it does not do anything, as its only line of code is "pass". I have verified that the function is in fact called. Two examples that run this function are the unit tests

objects/test_forwardsim.py::ForwardSimConsistencyTester::test_consistent_jacs
objects/test_forwardsim.py::ForwardSimConsistencyTester::test_consistent_probs

We either find that the function is not needed anymore and remove it, or implement it. 

For context, here is the function's docstring:

        """
        Frees an array allocated by :meth:`allocate_local_array`.

        This method should always be paired with a call to
        :meth:`allocate_local_array`, since the allocated array
        may utilize shared memory, which must be explicitly de-allocated.

        Parameters
        ----------
        local_array : numpy.ndarray or LocalNumpyArray
            The array to free, as returned from `allocate_local_array`.

        Returns
        -------
        None
        """

yrasool:
Fixes #698.

`CircuitOutcomeProbabilityArrayLayout.allocate_local_array` returns a plain NumPy array, so `free_local_array` has no shared-memory handle to release. The method is still useful as part of the layout API because callers can pair every allocation with a free call without checking the concrete layout type.

This updates the stale docstring, adds the issue pointer requested in the discussion, and adds a small test for the no-op local cleanup contract.

Testing, Python 3.10 local no-Cython install:
- `python -m py_compile pygsti\layouts\copalayout.py test\unit\objects\test_forwardsim.py`
- `git diff --check`
- `python -m flake8 . --count --show-source --statistics --config=.flake8-critical`
- `python -m pytest -q test\unit\objects\test_forwardsim.py -k "free_local_array_is_noop_for_numpy_array or dprobs_hprobs_empty_for_zero_param_model"`
- `python -m pytest -q test\unit\objects\test_forwardsim.py`

### DIFF
--- a/pygsti/layouts/copalayout.py
+++ b/pygsti/layouts/copalayout.py
@@ -364,19 +364,23 @@ class CircuitOutcomeProbabilityArrayLayout(_NicelySerializable):
         """
         Frees an array allocated by :meth:`allocate_local_array`.
 
-        This method should always be paired with a call to
-        :meth:`allocate_local_array`, since the allocated array
-        may utilize shared memory, which must be explicitly de-allocated.
+        This layout allocates plain NumPy arrays, so there are no external
+        shared-memory handles to release.  This method is intentionally a
+        no-op, but exists so callers can pair every
+        :meth:`allocate_local_array` call with a matching free call without
+        checking the concrete layout type.
 
         Parameters
         ----------
-        local_array : numpy.ndarray or LocalNumpyArray
-            The array to free, as returned from `allocate_local_array`.
+        local_array : numpy.ndarray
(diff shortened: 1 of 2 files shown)


yrasool:
I am making this contribution under the terms of the Apache 2.0 license. I also tested this on my own locally with the commands listed in the PR body.

rileyjmurray (approved):
Looks great!

Issue #607, opened 2025-07-15
Forward simulation on the GPU


rileyjmurray:
PR #390 added preliminary support for PyTorch-backed forward simulation. There's pending follow-up work in PR #613. All of this work so far has been to explore PyTorch's automatic-differentiation features. In principle, this work could be extended to do forward simulation on the GPU. This GitHub issue outlines changes we'd need for such forward simulation to be useful.

TorchForwardSimulator is currently limited to ExplicitOpModels whose members are all in the "full TP" parameterization. In order to make PyTorch-enabled GPU computations truly useful we'd need to extend TorchForwardSimulator to ImplicitOpModels and, we'd need to update several ModelMember classes to implement the [Torchable API](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/torchable.py). 

Here are the classes that need to implement the Torchable API.

POVMs
 * [ComputationalBasisPOVM](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/povms/computationalpovm.py)
 * [ComposedPOVM](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/povms/composedpovm.py)

States
 * [StaticState](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/states/staticstate.py)
 * [ComposedState](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/states/composedstate.py)

Gates
 * Various "static" operation classes (of which there's not a particularly meaningful base class): [StaticArbitraryOp](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/operations/staticarbitraryop.py), [StaticCliffordOp](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/operations/staticcliffordop.py), [StaticStandardOp](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/operations/staticstdop.py), and [StaticUnitaryOp](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/operations/staticunitaryop.py).
 * [ComposedOp](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/operations/composedop.py)
 * [EmbeddedOp](https://github.com/sandialabs/pyGSTi/blob/master/pygsti/modelmembers/operations/embeddedop.py)
 
Error processes
[...]

PR #834, opened 2026-07-08
Extend pytorch support beyond plain "full TP" models

rileyjmurray:
PR #390 added TorchForwardSimulator, but it only worked on ExplicitOpModel objects whose members were all in the `full TP` parameterization — `Torchable.stateless_data`/`torch_base` were implemented for a handful of classes. Issue #607 laid out what it would take to make PyTorch-backed (and eventually GPU) forward simulation useful in general: implement the Torchable API across the rest of the ModelMember hierarchy. This PR targets that issue.

After this PR, TorchForwardSimulator runs end-to-end on Lindblad-parameterized models (Lindblad SPAM and gates), models with Instrument/TPInstrument layers (including ones built via `Instrument.from_effects` or `Instrument.from_cptr_superops`), and models built from EmbeddedOps.

The main limitation of this PR is that TorchForwardSimulator only supports ExplicitOpModel objects. Extending to ImplicitOpModel objects will be important, later work.

### What changed

**Error generators, the main payoff.** LindbladCoefficientBlock (all three block types, all five `param_mode`s), LindbladErrorgen, ExpErrorgenOp, and EmbeddedErrorgen are now Torchable, and ComposedOp composes them. The chain mirrors the numpy path at each level: each block supplies a differentiable `block_data_torch`, `LindbladErrorgen.torch_base` contracts those against the (dense, precomputed) elementary-generator tensor, ExpErrorgenOp exponentiates via `torch.linalg.matrix_exp`, and ComposedOp multiplies factors in `to_dense` order. This is dense-only — `LindbladErrorgen.stateless_data` asserts a non-sparse representation.

**Lindblad SPAM.** ComposedState and ComposedPOVM / ComposedPOVMEffect are now Torchable (prep = `error_map_superop @ static_ket`; POVM = `base_effect_rows @ error_map_superop`, rows in `self.keys()` order to match how TorchForwardSimulator multiplies against the super-ket). A `CPTPLND`/`GLND` model with Lindblad prep, POVM, *and* gates is now entirely Torchable.

**Static and composite ops.** A new StaticTorchable mixin (`torchable.py`) bakes a static op's constant `to_dense('HilbertSchmidt')` superop into `stateless_data`; StaticUnitaryOp / StaticStandardOp / StaticArbitraryOp subclass it. EmbeddedOp and FullArbitraryOp are Torchable now too (the latter with a realness assert — the torch path would otherwise silently discard an imaginary part `to_dense` doesn't have).

**Instruments.** Instrument, TPInstrument, and TPInstrumentOp are Torchable, including instruments built by the CPTR-based `Instrument.from_effects`/`from_cptr_superops` constructors — which meant making `RootConjOperator` (the ρ ↦ E^½ ρ E^½ operator those constructors produce) Torchable too. That one required `torch.linalg.eigh`, whose AD rules are unstable at degenerate/boundary spectra in *both* forward and reverse mode; see the docstring warning on `RootConjOperator.torch_base` for specifics (and note the numpy reference warns/raises on out-of-range eigenvalues where the torch path clamps silently — a known, documented divergence, not a bug).

**`ComputationalBasisPOVM`** is Torchable (0 parameters — `stateless_data` just bakes the dense POVM matrix).

**TorchForwardSimulator/`torchfwdsim.py`** — substantially reworked to support all of the above:
- StatelessModelCircuitStore now expands instrument layers per circuit (via `model.expand_instruments_and_separate_povm`) and stacks each instrument's members under their `simplify_operations`-derived expanded labels, so instrument-containing circuits work at all.
- POVM rows are selected per-circuit by the expanded circuit's effect labels, so dataset-built layouts — which can reorder a circuit's outcomes relative to `povm.keys()`, or observe only a subset of them — get probabilities written into the right slots instead of silently swapped or shape-mismatched.
- The constructor takes explicit `dtype=`/`device=` arguments; the default dtype is now `torch.float64`.
  `device=` bypasses the `use_gpu` {True, False, None} auto-detection entirely when given.
- `use_gpu=None` no longer clobbers a detected CUDA/XPU device — it used to compute `device = 'cpu' if not use_gpu else ...`, which is wrong for the `None` (auto-detect) case; only a deliberate `use_gpu=False` forces CPU now. MPS is still skipped by default (no float64 support, and matrix exponential can only be handled by temporarily transferring to the CPU).
- A non-Torchable model member now fails with a clear, explanatory `assert`, not an opaque `AttributeError` from a missing `stateless_data`.
- Instrument/TPInstrument's `stateless_data`/`torch_base` functions are shared in a new `StackedMemberDictTorchable` mixin.

**`pyproject.toml`** pins `torch>=2.4` (first version with both `torch.mps.device_count` and `torch.xpu`, which `torchfwdsim.get_device` needs).

### Tests

Per-member unit tests, each checking that `torch_base` reproduces the numpy `to_dense` and that `torch.func.jacrev(torch_base)` reproduces the numpy `deriv_wrt_params`:

- `test_torchable_members.py` (new) — FullArbitraryOp, EmbeddedOp/EmbeddedErrorgen, ComputationalBasisPOVM, ComposedPOVMEffect, RootConjOperator, Instrument, TPInstrument.
- `test_experrorgen_torch.py` — the Lindblad chain: blocks → LindbladErrorgen → ExpErrorgenOp → ComposedOp → ComposedState/ComposedPOVM.
- `test_lindbladcoefficients_torch.py` — per-block `block_data_torch`.
- These share their `torch_base`/Jacobian-checking helpers via `torchable_test_utils.py` (new) rather than each file defining its own copy.

End-to-end coverage, in `test_forwardsim.py`: several new `ForwardSimConsistencyTester` subclasses check that `TorchForwardSimulator` agrees with the four other forward simulators (probability and Jacobian colinearity) on:

- a full CPTPLND/GLNDU model,
- Instrument- and TPInstrument-layered circuits, including a two-layer circuit (so the layout expands to the *product* of both layers' outcomes),
- a CPTR-built instrument,
- a `ComputationalBasisPOVM` model, and
- embedded 2-qubit ops,

plus targeted tests for dataset-ordered/subset layouts and instrument-expansion bookkeeping.

### Design notes and gotchas

- **Type-dispatch pattern.** Composite Torchables (ComposedOp, EmbeddedOp, ExpErrorgenOp, the Instrument/TPInstrument mixin) store `(type(submember), submember.stateless_data(...))` rather than assuming a fixed submember type, and dispatch via `type_handle.torch_base(...)` at eval time. This is what let the ExpErrorgenOp/EmbeddedErrorgen crash get fixed with a one-line change (dispatch instead of hardcoding LindbladErrorgen) rather than a new code path.
- **MRO gotcha.** Adding the `Torchable` (a ModelMember) mixin to a class that *also* has a non-ModelMember mixin with its own `__init__` (ExpErrorgenOp's `_ErrorGeneratorContainer`) can reorder the C3 MRO and pull that mixin into the cooperative `super().__init__()` chain, breaking construction. Fix: list `Torchable` before the non-ModelMember mixin in the bases.
- **`DataSet.add_count_dict` sorts plain-dict keys.** Reproducing a dataset-ordered (non-`povm.keys()`-order) layout in a test requires an `OrderedDict`/`OutcomeLabelDict` input, or real time-series data — a plain dict gets its keys sorted before the layout is built.
- **Realness asserts.** `ComposedPOVMEffect.torch_base` uses `emap.mT`, not `emap.conj().mT` — correct only because error maps are real on Hilbert-Schmidt space today. `ComposedPOVM.torch_base` carries the same assumption. Both now assert `not emap.is_complex()`, so a future complex-valued error map Torchable fails loudly here instead of silently multiplying by an unconjugated tensor.

### Known follow-ups (deferred, not blocking)

- **Hot-path efficiency.** `_bulk_fill_dprobs` rebuilds `StatelessModelCircuitStore` — including all `stateless_data` device uploads — on every call; `TPPOVM.torch_base` converts an identity in the hot path; `LindbladErrorgen.torch_base` contracts in complex and takes `.real`. None of this blocks correctness.
- **Real→complex dtype selection** is implemented separately in `cptrop.py`, `lindbladerrorgen.py`, and `torchable.py`. Small enough in each place that a shared helper wasn't judged worth the indirection.
- **Sparse-superop Lindblad representations** are out of scope — the torch path is dense-only.
- **`StatelessModelCircuitStore.free_param_sizes`** is currently write-only; kept as-is rather than deleted.


rileyjmurray (commented):
The text below is out of date as of July 13, 2026.

------------------------------

The following are not yet Torchable. We need to update so they're Torchable, and then add appropriate tests.

 * GLNDU-parameterized LindbladCoefficientBlocks
 * EmbeddedOp
 * ComputationalBasisPOVM (might automatically work since it's static)
 * RootConjOperator
 * Instrument (might automatically work once RootConjOperator works)
 * TPInstrument

We'll also want to add a new simulation tutorial that addresses how/when to use TorchForwardSimulator. Relevant decisions include precision, targeting CPU or GPU, how reasonable it is to run on Apple Silicon GPUs, etc..

rileyjmurray:
### TODOs

Algorithm changes in TorchForwardSimulator:
 * sort circuits by length, collect process matrices for circuits of a given length in a tensor (3d or 4d), evaluate the matrix-chain multiplication by ``reduce(torch.matmul, [prep] + list(torch.unbind(optensor, dim=0) + [povm])``

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -202,62 +262,112 @@ def circuit_probs_from_free_params(self, *free_params: Tuple[torch.Tensor], enab
...
     """
     A forward simulator that leverages automatic differentiation in PyTorch.
     """
 
     ENABLED = TORCH_ENABLED
+    model   : ExplicitOpModel | None
 
-    def __init__(self, model : Optional[ExplicitOpModel] = None):
+    def __init__(self, model : Optional[ExplicitOpModel] = None, use_gpu: Optional[bool] = None,
+                 dtype: Optional[torch.dtype] = None, device: Optional[str] = None):
```
sserita: A more common pyGSTi pattern is doing something like `device: Optional[str] = "auto"` for the autodetect functionality instead of having one flag and one param to use in case the flag is False. Thoughts?

I guess the one complication is it wouldn't provide the same way to override `"mps"` with `"cpu"`, but maybe we do something like `"mps"` requires `dtype` to explicitly be `torch.float32` for user-aware opt-in (and either warn or override if `dtype` anything else with `"mps"`)?

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -202,62 +262,112 @@ def circuit_probs_from_free_params(self, *free_params: Tuple[torch.Tensor], enab
...
+        if smcs.default_to_reverse_ad:
+            J_func = torch.func.jacrev(smcs.circuit_probs_from_free_params, argnums=argnums) # type: ignore
         else:
-            # Then slm.circuit_probs_from_free_params will automatically skip the extra
-            # steps needed for torch_base to support reverse-mode AD.
-            J_func = torch.func.jacfwd(slm.circuit_probs_from_free_params, argnums=argnums)
-        # ^ Note that this _bulk_fill_dprobs function doesn't accept parameters that
-        #   could be used to override the default behavior of the StatelessModel. If we
-        #   have a need to override the default in the future then we'd need to override
-        #   the ForwardSimulator function(s) that call self._bulk_fill_dprobs(...).
```
sserita, line 260: Are these comments removed because this has changed? Or just because it is not worth worrying about? I'm not seeing any edits yet that address this directly.

pygsti/modelmembers/operations/cptrop.py
sserita, line 1: I didn't even know this was a thing... 😅

pygsti/modelmembers/operations/cptrop.py
```diff
@@ -144,6 +152,57 @@ def to_dense(self, on_space: SpaceT = 'HilbertSchmidt') -> _np.ndarray:
...
+            e.g. ideal projective effects (spectrum of 0s and 1s) yield entirely-NaN Jacobians.  Use
+            torch AD for this operator, in either mode, only for effects whose spectra are interior to
+            ``(0, 1)`` and non-degenerate; only the value path has no such restriction.
+
+        Unlike the numpy reference :func:`~pygsti.tools.optools.rootconj_superop` -- which warns
+        when an eigenvalue of ``E`` strays outside ``[0, 1]`` by more than :attr:`EIGTOL_WARNING`
+        and raises a ``ValueError`` past :attr:`EIGTOL_ERROR` -- this path clamps silently at any
+        magnitude of violation.  So under torch-driven optimization it can keep returning
+        (clamped) values in regions where the numpy path would refuse to evaluate.
+        """
```
sserita, line 195: Is there any way to detect this and warn the user more explicitly as it occurs? This seems subtle and potentially problematic, but doubtful the user will read this docstring.

pygsti/modelmembers/operations/lindbladcoefficients.py
```diff
@@ -1110,6 +1199,9 @@ def _block_data_indices_impl(self):
             block_data_indices[i] = [(1.0, _LEEL('H', (lbl,)))]
         return block_data_indices
 
+    def _coefficient_labels_impl(self):
+        return ["%s Hamiltonian error coefficient" % lbl for lbl in self._bel_labels]
+
```
sserita, line 1204: Thanks for doing these at the same time.

pygsti/modelmembers/povms/composedeffect.py
```diff
@@ -11,36 +11,49 @@
...
 from pygsti.baseobjs import _compatibility as _compat
 from pygsti import SpaceT
 
-class ComposedPOVMEffect(_POVMEffect):  # , _ErrorMapContainer
+
+class ComposedPOVMEffect(_POVMEffect, _Torchable):  # , _ErrorMapContainer
     """
-    TODO: update docstring
-    A Lindblad-parameterized POVMEffect (that is also expandable into terms).
+    A POVM effect formed by composing a static "ideal" effect vector with a parameterized error
```
sserita, line 35: Thanks for this update!

test/unit/objects/test_forwardsim.py
```diff
@@ -339,13 +365,229 @@ def test_consistent_jacs(self):
...
+
+
+class LindbladForwardSimConsistencyTester(ForwardSimConsistencyTester):
+    """End-to-end check that TorchForwardSimulator runs on a full CPTPLND (Lindblad) model and agrees
+    with the other forward simulators (GitHub issue 607).  Every model member is a Lindblad-composed
+    object: the prep is a ComposedState, the POVM a ComposedPOVM, and each gate a
+    ComposedOp([static ideal, ExpErrorgenOp]) -- exercising the whole Torchable op + SPAM chain."""
+
+    PROBS_TOL = 1e-12
+    JACS_TOL = 1e-9
```
sserita, line 393: Worth throwing these in the base class, and then overriding for the looser numerical deriv below?

sserita (changes requested):
A few minor changes requested, otherwise great work on a huge feature!

pygsti/modelmembers/operations/cptrop.py
rileyjmurray: I added it for instrument modeling just a couple of months ago :)

pygsti/forwardsims/torchfwdsim.py
rileyjmurray: Interesting point about the `auto` flag. Normally I dislike because it causes a union between a string and non-strings. In this case there's no such issue with a type union. Will (likely) change and report back.

pygsti/forwardsims/torchfwdsim.py
rileyjmurray: Just not worth worrying about

pygsti/modelmembers/operations/cptrop.py
rileyjmurray: That's a good question. I agree that this failure case is potentially problematic. My actual desire is to figure out a way to sidestep this by falling back on finite-difference methods, which shouldn't have this problem. I'd like to open a github issue about this rather than try to address here.

pygsti/modelmembers/operations/lindbladcoefficients.py
rileyjmurray: Thank the robot! lol

pygsti/modelmembers/povms/composedeffect.py
rileyjmurray: This one I _do_ take credit for :)

test/unit/objects/test_forwardsim.py
rileyjmurray: Turns out that 7 of 8 definitions of PROBS_TOL use 1e-12. The only definition that's different is the base class definition of 1e-14. 

JACS_TOL is 1e-10 in the base class, then 1e-9 every other class but CPTRInstrumentForwardSimConsistencyTester (which uses 1e-6).

Since the base class uses the most strict tolerances I think it makes sense to keep the per-class tolerances as they are.

sserita: That's fair, I missed the tighter base class tols, agree it's better to keep those tight.

sserita (approved):
Good to go!

PR #835, opened 2026-07-09
Add `pygsti.extras.ml`: quantum physics-aware neural networks (QPANNs)

nonohuff:
This PR adds **`pygsti.extras.ml`**, a new optional sub-package implementing **quantum physics-aware neural networks (QPANNs)** for *quantum capability learning* — predicting how well a noisy quantum processor will execute a given circuit, without the exponential cost of full circuit simulation.

It's a from-scratch rewrite of pyGSTi's earlier, unreleased neural-network code, directly implementing **Hothem, Miller & Proctor, "What is my quantum computer good for? Quantum capability learning with physics-aware neural networks"** ([arXiv:2406.05636](https://arxiv.org/abs/2406.05636), NeurIPS 2024), plus one notable extension beyond the paper: **all four elementary error generator sectors (H, S, C, A) are supported**, where the paper demonstrated only H and S.

A QPANN factors into two pieces:
1. A small, heavily-localized, trainable network `N` predicting the rate of each modelled elementary error generator, per circuit layer.
2. A fixed, non-trainable physics layer `f` that propagates those generators through the (Clifford) circuit in polynomial time — via [arXiv:2504.15128](https://arxiv.org/abs/2504.15128)'s efficient propagation algorithm — and combines them into a first-order prediction of the circuit's outcome probabilities.

This factorization is what buys generalization to circuits deeper/wider than the training set, sample efficiency, and physically-interpretable learned weights (they're literal error-generator rates, not black-box weights). Core pyGSTi does not depend on this sub-package, and it is **not imported automatically** — `from pygsti.extras import ml` is required.

This has been through two rounds of review from @rileyjmurray (40+ inline comments, all addressed individually in-thread) plus a `develop` merge along the way; this description reflects the current, final state of the branch. **One unrelated core bugfix that was riding along has been split out into #876** (merge that first) — see "Changes outside `extras/ml`" below.

### What's in this PR

- **`errgentools.py`** (1228 lines) — canonical indexing/enumeration for elementary error generators. H/S are indexed by a single Pauli; C/A by an unordered pair of distinct non-identity Paulis, with explicit sign-canonicalization helpers handling the antisymmetry of `A_{P,Q} = −A_{Q,P}`. Arbitrary weight-`k` enumeration, with graph-locality-restricted variants (for weight ≥ 2 C/A pairs, the union of both Paulis' supports must fit within the locality radius).
- **`encoding.py`** (811 lines) — `StandardCircuitEncoder` (circuit → multi-hot tensor), Clifford error propagation (`circuit_error_propagation_matrices`, built on `pygsti.errorgenpropagation.ErrorGeneratorPropagator`), and first-order sensitivity (α) tensors following arXiv:2504.15128, in a `'concise'` (default, sparse, multiprocessed, all four sectors) and a `'matrix'`/`'expanded'` (dense, H/S-only by design — the C/A range grows as `O(16ⁿ)`, and raises a descriptive `NotImplementedError` if asked for more) representation.
- **`qpanns.py`** (440 lines) **+ `customlayers.py`** (294 lines) — the `QPANN` Keras model. `CustomDense` evaluates every per-generator sub-network simultaneously via one `tf.einsum` rather than a Python loop over generators; `ProbabilitiesLayerConcise` implements the physics layer `f` (`probabilities = probabilities_ideal + sum(corrections * error_rates)`).
- **`snippers.py`** (145 lines) — the paper's space-local filter `W_j`: for each modelled error generator, which circuit-encoding indices its sub-network is allowed to see, via BFS distance on the qubit graph. Takes the union of qubits touched by every Pauli in a generator's tuple (matters for C/A 2-tuples).
- **`graphtools.py`** (513 lines, new module) — lets every graph-consuming function above accept `networkx`/`igraph`/`graph-tool`/`pygsti.baseobjs.QubitGraph`/`pygsti.processors.QubitProcessorSpec`/a plain adjacency matrix, rather than requiring a hand-built graph Laplacian. A bare matrix must be an adjacency matrix; one with a negative off-diagonal entry (i.e., looks like a Laplacian) raises a clear `ValueError` explaining the mismatch and how to convert it, rather than silently guessing. igraph/graph-tool are recognized by duck-typing on `type(obj).__module__` and are never imported, so both stay fully optional.

### Changes outside `extras/ml`

Small and purely additive/infrastructural:
- `pyproject.toml` — new opt-in `ml` extra (`tensorflow>=2.16`, `keras>=3.0`; see "Design notes" for why Keras 3 specifically). Deliberately excluded from `testing`/`no_mpi`/`complete` since TensorFlow is heavy and platform-specific.
- `.github/workflows/reuseable-main.yml` — installs `.[ml]` for the `run-extra-tests` job and the notebook-regression job, so `test/test_packages/extras/ml` and the QPANN tutorial notebooks actually execute in CI (gated behind `run-extra-tests: 'true'`, i.e. on `beta`/`master`) instead of failing collection or being silently skipped.
- `docs/_toc.yml`, `MANIFEST.in` — wire in the new docs pages and test package.

**A previously-included, unrelated core bugfix has been split out into #876.** Two `UnboundLocalError` typos in `pygsti/tools/errgenproptools.py`'s 'A'-type-generator code (found while testing this PR's C/A support, on code paths this subpackage doesn't itself call) are now their own PR at @rileyjmurray's request, the same way `0bdbb2830` was split into #865 earlier in this PR's review. `pygsti/tools/errgenproptools.py` and `test/unit/tools/test_errgenproptools.py` are byte-identical to `develop` again in this branch; `develop` will be merged back in once #876 lands.

### Documentation

A 4-part MyST tutorial series under `docs/markdown/ml/` (`Overview`, `CircuitEncoding`, `ErrorPropagation`, `Training`), jupytext-paired and `nbval`-tested in CI. `Training.md` builds a `CloudNoiseModel` with known H/S/C/A ground-truth rates on a 3-qubit ring, samples `edgegrab` circuits, trains a QPANN, and recovers all nine injected rates with the correct sign and order of magnitude. I re-executed it fresh for this write-up (CI's own sized-down config: 200 circuits/150 epochs, ~90s) and got 4.8%-14.5% relative error across the nine rates (the page's own note quotes ~5-11% for a larger, slower 600-circuit/400-epoch configuration on this same problem — consistent given normal run-to-run training variance; the page explains why more data/epochs doesn't move the needle much at this problem size). Also includes callouts on the physicality bound (`c² + a² ≤ s_P·s_Q`, checked post-hoc in the notebook) and a pyGSTi model-builder quirk (silently dropping reversed C/A pair keys) that the tutorial adds a defensive assertion against.

### Testing

```
pytest -v test/test_packages/extras/ml/test_ml.py
# 30 passed in 10.24s  (MLSubpackageTester: 19, GraphToolsTester: 11)

mypy pygsti/extras/ml --ignore-missing-imports --follow-imports=silent
# Success: no issues found in 7 source files

flake8 pygsti/extras/ml --config=.flake8-critical --count
# 0

(cd docs && jupytext --sync 'markdown/ml/*.md')
pytest -q --nbval-lax --nbval-current-env docs/markdown/ml
# 16 passed in 92.53s  (the 4 tutorial notebooks, exactly as CI's notebook-regression job runs them)
```

Coverage highlights in `test_ml.py`: exhaustive Pauli-pair index bijection round-trips and full-coverage checks for n = 1…3; H/S index backward-compatibility, C/A order-invariance, and a full `[0, num_error_generators(n))` round-trip with no overlap; the `A`-type canonicalization sign cross-checked against a hand-derived example *and* an independent `ErrorGeneratorPropagator`-based computation; end-to-end `error_generator_tensors` with a full H/S/C/A mix, every α entry cross-validated against `alpha_coefficient` + `random_support`; full `QPANN` forward-pass-and-`.fit()` (including with weight-2 C/A generators), asserting weights actually change; a definite-outcome Hamiltonian-α short-circuit test that monkeypatches `alpha_coefficient` with a call counter to prove the fast path is genuinely taken; an isolated-qubit snipper regression; and an all-graph-backends-agree test asserting every graph representation yields byte-identical generator lists and snippers.

### Design notes for reviewers

- **Keras 3 compatibility required two non-obvious fixes**, both regression-tested: `CustomDense` subclasses `keras.layers.Layer` rather than `Dense` (`Dense.kernel` became a read-only `@property` in Keras 3 to support LoRA, breaking `self.kernel = self.add_weight(...)`), and the model's boolean masks (`stochastic_mask`) are plain numpy arrays rather than eager `tf.constant`s, which get captured into a stale `FuncGraph` and raise `InaccessibleTensorError` during `.fit()`.
- **`'S'`-sector network outputs are squared** to enforce non-negative stochastic rates; `'H'`/`'C'`/`'A'` are left unconstrained (they can legitimately be negative).
- **The locality-filter (snipper) rule changed**: `layer_snipper_from_qubit_graph` now takes the union of qubits touched by *every* Pauli in a generator's tuple (a no-op for H/S 1-tuples, but needed for C/A 2-tuples) and correctly includes a qubit with no graph edges as touching only itself, replacing a previous Laplacian-power heuristic that dropped isolated qubits entirely (`test_isolated_qubit_snipper_fix`).
- **`qpann.summary()` prints "Total params: 0"** despite `qpann.trainable_variables` being correctly populated and training working — a Keras-3-subclassed-model introspection quirk, likely related to `QPANN.build(self)`'s non-standard zero-argument signature. Cosmetic only; not something this PR fixes.

### Known follow-ups (deferred, not blocking)

- The Pauli-expectation-value / process-fidelity prediction pathway (the paper's F(c), Eq. 5) is not implemented here — an earlier half-built attempt at it was removed as dead code (zero callers) rather than kept as misleading scaffolding. Tracked in #869; the underlying core math (`errgenproptools.alpha_pauli`/`stabilizer_pauli_expectation`) is untouched, tested pygsti-core code that a real implementation can build on.
- The paper's joint C/A-vs-S physicality bound (`c² + a² ≤ s_P·s_Q`) isn't architecturally enforced in the network — same level of rigor as the existing per-generator-only `'S'` squaring heuristic. `docs/markdown/ml/Training.md` validates it post-hoc instead.
- `pygsti.extras.ml.graphtools` overlaps conceptually with `pygsti.tools.graphcoloring`; unifying them into a `pygsti.tools.graphs` subpackage before 0.11.x is tracked in #866.
- Importing anything from `pygsti.extras.ml` requires TensorFlow/Keras, even the pure-numpy `errgentools`/`encoding`/`snippers` helpers, because `__init__.py` imports `qpanns` first.
- `test/test_packages/extras/ml` only runs in CI when `run-extra-tests` is `'true'` (`beta`/`master`), not on feature-branch pushes — a pre-existing CI-scoping choice for the whole `test_packages` tree, not specific to this PR.

### Backward compatibility

No breaking changes to existing APIs — `pygsti.extras.ml` is new, optional, and not auto-imported; core pyGSTi is untouched by this PR now that #876 is split out.


rileyjmurray:
Thanks for opening the WIP PR, Noah!

You scared me with the number of lines added, but then I saw this.
image

I trust that won't need to be merged :)

nonohuff:
C and A error integration finished!

nonohuff:
# Pre-PR cleanup checklist

The description below assumes these are done first. Verified via a trial `git merge-tree origin/develop feature-ml-rewrite`:

| # | Item | Why |
|---|---|---|
| 1 | Remove `jupyter_notebooks/Tutorials/tutorial_files/myFirstGSTReport/main.html` | 14,792-line generated GST report, committed by accident in `a7e71eb78` |
| 2 | Remove `jupyter_notebooks/Tutorials/tutorial_files/Example_Short_Dataset.txt` | Same; generated tutorial artifact |
| 3 | Revert `jupyter_notebooks/Tutorials/algorithms/ErrorGeneratorPropagation.ipynb` | Deleted on `develop` by the JupyterBook migration (#673); currently a **modify/delete conflict**. If its 2 real source edits are worth keeping, port them to `docs/markdown/simulation/ErrorGeneratorPropagation.md` in a separate PR |
| 4 | Revert whitespace-only churn in `pygsti/tools/internalgates.py` and `pygsti/modelmembers/operations/lindbladerrorgen.py` | Net effect is +1 trailing-whitespace line, −2 blank lines. Pure noise |
| 5 | Resolve `pygsti/modelmembers/operations/experrorgenop.py` conflict | `develop` added `stateless_data`/`torch_base` immediately after `to_dense`; keep both |
| 6 | Consider `nbstripout` on the 4 QPANN notebooks | `QPANN-Training.ipynb` is 165 KB, of which ~110 KB is two base64 PNGs and a 32 KB ANSI-coloured Keras log. `execution_count`s are non-sequential (cells run out of order) |
| 7 | Consider squash-merge | 153 commits including `...`, `typo`, `still bug hunting...`, `might have found the bug...` |

---

# PR description

## Add `pygsti.extras.ml`: quantum physics-aware neural networks (QPANNs)

### Summary

This PR merges `feature-ml-rewrite` into `develop`, adding **`pygsti.extras.ml`** — a new optional sub-package implementing **quantum physics-aware neural networks (QPANNs)** for *quantum capability learning*: predicting how well a noisy quantum processor will execute a given circuit.

It is a from-scratch rewrite of pyGSTi's earlier, unreleased neural-network code, and is a direct implementation of the architecture in **Hothem, Miller & Proctor, "What is my quantum computer good for? Quantum capability learning with physics-aware neural networks"** ([arXiv:2406.05636](https://arxiv.org/abs/2406.05636), NeurIPS 2024), plus subsequent evolution — most notably **full support for all four elementary error generator sectors** (`H`, `S`, `C`, `A`), where the paper demonstrated only `H` and `S`.

A QPANN factors into two components:

1. **A trainable, heavily-localized MLP `N`** that predicts, for each circuit layer, the rate of each modelled local elementary error generator.
2. **A fixed, non-trainable physics layer `f`** that propagates those generators through the (Clifford) circuit in polynomial time and combines them into a first-order prediction of the circuit's outcome probabilities.

This factorization is what buys generalization to deeper/wider circuits than seen in training, high sample efficiency, and physical interpretability of the learned weights.

Core pyGSTi does not depend on this sub-package, and `pygsti.extras.ml` is **not** imported automatically.

---

### Feature set

#### 1. Error generator bookkeeping — `errgentools.py`

Canonical, stable indexing and enumeration for the *Taxonomy of Small Errors* elementary error generators.

- **All four sectors supported.** `H` (Hamiltonian) and `S` (stochastic Pauli) are indexed by a single Pauli; `C` (stochastic Pauli-correlation) and `A` (active) by an **unordered pair of distinct non-identity Paulis**.
- **Index scheme** (`error_generator_index` / `index_to_error_gen`): `H → [0, 4ⁿ)`, `S → [4ⁿ, 2·4ⁿ)`, `C → [2·4ⁿ, 2·4ⁿ+M)`, `A → [2·4ⁿ+M, 2·4ⁿ+2M)` where `M = num_pauli_pairs(n) = (4ⁿ−1)(4ⁿ−2)/2` (= 105 for n=2, matching the taxonomy paper). **H/S indices are unchanged from the previous scheme**; C/A are appended after.
- **Pauli-pair bijection** (`pauli_pair_to_index` / `index_to_pauli_pair`): unordered-pair lexicographic rank, with an **integer-only binary search** inverse to avoid the floating-point precision loss of the quadratic-formula inverse. Exhaustively round-trip tested for n = 1…3.
- **Antisymmetry handled explicitly.** Because `C_{P,Q} = C_{Q,P}` but `A_{P,Q} = −A_{Q,P}`, `canonical_pauli_pair` and `error_generator_canonicalization_sign` separate *index* canonicalization from the ±1 *sign* correction that consumers must apply. This sign is folded into the propagation tensors by `encoding.circuit_error_propagation_matrices`.
- **Arbitrary weight-`k` enumeration**, generalizing the previous `k ≤ 2`-only implementation: `up_to_weight_k_paulis`, `up_to_weight_k_pauli_pairs`, and graph-restricted variants that keep only supports whose induced subgraph is connected within `num_hops`. For C/A, the *union* of the two Paulis' supports is the connected blob.

#### 2. Circuit encoding and physics tensors — `encoding.py`

- **`StandardCircuitEncoder`**: multi-hot binary encoding over every available `(gate name, qubit placement)` pair in a `QubitProcessorSpec`. `circuits_to_tensor` batches and zero-pads to `(num_circuits, encoding_depth, encoder.length)`.
- **Clifford error propagation** (`circuit_error_propagation_matrices`, `error_propagation_tensors`): for each circuit, `(depth, num_error_gens)` `indices` and `signs` tensors mapping "generator `j` inserted after layer `l`" to the canonical end-of-circuit generator it becomes, with a ±1 sign. Supports incrementally appending new generators to previously-computed tensors.
- **First-order sensitivity (α) tensors**, following [arXiv:2504.15128](https://arxiv.org/abs/2504.15128), in two representations:
  - **`'concise'` (default)** — `alphas` of shape `(num_circuits, 2ⁿ, depth, num_error_gens)`, with signs pre-folded. **Supports all four sectors.** Multiprocessed.
  - **`'matrix'`/`'expanded'`** — dense `(num_circuits, 2ⁿ, 2·4ⁿ)` α matrix. **H/S only by design** (the C/A range grows as `O(16ⁿ)`); raises a descriptive `NotImplementedError` otherwise.
- **`error_generator_tensors(...)`** is the one-shot user-facing entry point returning `{'indices', 'signs', 'probabilities', 'alphas'}`.
- Also supports Pauli-expectation-value targets (`measurements='paulis'`) on the data-generation side, via `stabilizer_pauli_expectation` / `alpha_pauli`.

#### 3. The model — `qpanns.py` + `customlayers.py`

- **`QPANN`** — a subclassed `keras.Model`. Input is `[circuits_tensor, alphas, probabilities_ideal]`; output is a `2ⁿ` probability vector. Fully `get_config()`-serializable.
- **`CircuitToErrorRatesEinSum` → `EinsumSubNetwork` → `CustomDense`** — the trainable branch. `CustomDense` replicates Dense's parameters over a leading error-generator axis and evaluates all per-generator sub-networks `N_j` **simultaneously via a single `tf.einsum`**, rather than a Python loop over generators. Ragged snipper index lists are `-1`-padded and masked. Final layer initialized to `RandomUniform(-1e-4, 1e-4)` so the network starts near zero output — appropriate for predicting small error rates.
- **`ProbabilitiesLayerConcise`** (default) / **`ProbabilitiesLayer`** — the non-trainable physics layer `f`.
- **Physicality heuristic**: `'S'`-sector outputs are squared to enforce non-negative stochastic rates; `'H'`, `'C'`, `'A'` are left unconstrained (they can legitimately be negative).
- **`SelectiveDense`** — a second custom layer providing per-group input-subset kernels. Currently exercised only by the test suite.

#### 4. Locality filters ("snippers") — `snippers.py`

A *snipper* is a `list[list[int]]` — one index list per modelled error generator — specifying which entries of the circuit encoder's per-layer feature vector that generator's sub-network is allowed to see. This is the code realization of the paper's space-local filter `W_j`, and is what makes the trainable component small and resistant to overfitting.

`layer_snipper_from_qubit_graph` takes the union of the generator's Pauli supports (correct for C/A 2-tuples), expands by `hops` **BFS shortest-path steps**, and maps through `encoder.indices_for_qubits`.

> **Behaviour change**: this replaces the previous `L**hops` Laplacian-power heuristic, which produced wrong results for isolated (degree-0) qubits — they would fail to include themselves. Covered by `test_isolated_qubit_snipper_fix`.

#### 5. Graph interoperability — `graphtools.py` (new module)

Callers no longer have to hand-construct a Laplacian. `up_to_weight_k_*_from_qubit_graph` and `layer_snipper_from_qubit_graph` now accept any of:

`networkx.Graph/DiGraph/MultiGraph/MultiDiGraph` · `igraph.Graph` · `graph_tool.Graph` · `pygsti.baseobjs.QubitGraph` · `pygsti.processors.QubitProcessorSpec` · a raw Laplacian or adjacency matrix (numpy / nested list / `scipy.sparse`) · an explicit edge list (via `qubit_graph_from_edges`).

- **networkx is the single canonical internal representation** (already a hard pyGSTi dependency). **igraph and graph-tool are never imported** — they're recognized by duck-typing on `type(obj).__module__`, so both remain fully optional.
- Laplacian-vs-adjacency auto-detection (`input_is='auto'`) with clear `ValueError`s on genuinely ambiguous input, overridable via explicit `input_is=`.
- Node-ordering reconciliation against `qubit_labels`, with a positional fallback.
- A `QubitProcessorSpec` is converted via `compute_2Q_connectivity()`, **not** `pspec.qubit_graph` (which is edgeless unless an explicit `geometry` was supplied).

**Backward compatible**: the old `qubit_graph_laplacian=` / `adjacency_matrix=` keyword arguments still work, now emitting a `DeprecationWarning`.

#### 6. Legacy helpers — `legacy.py`

Pre-QPANN helpers (`extract_success_fail`, `create_spec_model`, `create_mle_model`, `batch_prediction`) preserved for backwards compatibility. Unrelated to the neural networks; no TensorFlow dependency.

---

### Changes to core pyGSTi

These are small but **not** confined to `extras/` — please review them specifically.

#### Bug fixes in `pygsti/tools/errgenproptools.py`

1. **`error_generator_commutator`, `S`–`A` branch** — two references to an undefined local `ptup` where `ptup1` was intended, causing `UnboundLocalError`. Reachable only for a degenerate `A(Q,Q)` label, which `LocalStimErrorgenLabel` does not currently forbid.
2. **`slow_bulk_alpha`, "A1" branch** — assigned to `sensitivities_by_bitstring` (the whole return array) instead of the per-iteration local `sensitivity`, which was then read on the next line → `UnboundLocalError`. This affects **essentially every weight-1 `A` generator** (any two distinct single-qubit Paulis anticommute).

   ⚠️ This bug is **masked in most environments**, because `bulk_alpha` is aliased to the Cython `fast_bulk_alpha` whenever `fasterrgencalc` is built. The new regression test calls `slow_bulk_alpha` directly to force the pure-Python path.

Both fixes are covered by new tests in `test/unit/tools/test_errgenproptools.py`, which **do** run in CI on all branches. Each is validated against an independent numerical computation (`error_generator_commutator_numerical`; per-label `alpha`), and the second is additionally swept over 20 randomly-sampled 3-qubit `A`-type labels against a random Clifford tableau.

#### `ExpErrorgenOp.to_dense` — cast to unitary

`to_dense(on_space='Hilbert')` (and `'minimal'` when the evotype's minimal space is Hilbert) now attempts `optools.superop_to_unitary(...)` rather than returning a superoperator, raising a descriptive `ValueError` if the error generator is not purely Hamiltonian. Needed to build Hamiltonian-only noise models in the QPANN pipeline.

> **Merge note:** this conflicts with `develop`, which added `stateless_data` / `torch_base` immediately after `to_dense`. Both changes should be kept.

---

### Packaging

A new opt-in extra:

```toml
ml = ['tensorflow>=2.16', 'keras>=3.0']
```

Pinned to `tensorflow>=2.16` (which brings Keras 3) because the custom layers were **fixed for the Keras 3 API**:

- `keras.layers.Dense.kernel` became a read-only `@property` in Keras 3 (for LoRA), so `self.kernel = self.add_weight(...)` raised `AttributeError`. `CustomDense` therefore subclasses `Layer`, not `Dense`.
- Eager `tf.constant` masks stored as model attributes get captured into a stale `FuncGraph` and raise `InaccessibleTensorError` during Keras 3 `.fit()`. The `stochastic_mask` / `hamiltonian_mask` attributes are deliberately plain numpy arrays.

To reproduce the original paper's environment (TF/Keras 2.12), install those pins directly instead — note that requires Python ≤ 3.11.

The extra is **deliberately excluded** from `testing` / `no_mpi` / `complete`: TensorFlow is a heavy, platform-specific dependency, and the ML tests do not currently run in CI regardless (see Known limitations).

---

### Testing

**`test/test_packages/extras/ml/test_ml.py` — 32 new tests in 2 classes.**

`MLSubpackageTester` (20) covers every module. Highlights:

- Exhaustive Pauli-pair index bijection round-trips and full-coverage checks for n = 1…3; `up_to_weight_k_pauli_pairs` cross-validated against an inline naive reference.
- H/S index backward-compatibility; C/A order-invariance; full `[0, num_error_generators(n))` round-trip with no overlap.
- The `A`-type canonicalization sign, cross-checked against a hand-derived `Gzpi2` example *and* an independent computation via `ErrorGeneratorPropagator` + `LocalStimErrorgenLabel.propagate_error_gen_tableau` + `errgenproptools.alpha`.
- End-to-end `error_generator_tensors` with a full H/S/C/A mix, with **every** α entry cross-validated against `alpha_coefficient` + `random_support`.
- Full `QPANN` **forward pass and `.fit()`** (including with weight-2 C/A generators), asserting weights actually change — regression coverage for both Keras 3 bugs above.
- A definite-outcome Hamiltonian-α short-circuit test that monkeypatches `alpha_coefficient` with a call counter to prove the fast path is genuinely taken.
- Four named regression tests for prior-axis handling, reversed qubit indexing, a `get_config()` key typo, and `padded_depth` validation.

`GraphToolsTester` (12) covers the new graph plumbing: Laplacian↔adjacency round-trips, `input_is='auto'` detection and its `ValueError`s, networkx/DiGraph/string-label inputs, node-ordering reconciliation, `QubitGraph`/`QubitProcessorSpec`/`scipy.sparse` inputs, the deprecated-alias `DeprecationWarning`s, the isolated-qubit snipper fix, and an **all-backends-agree** test asserting every input representation yields byte-identical generator lists and snippers. igraph and graph-tool tests `skipUnless` those packages are importable.

**`test/unit/tools/test_errgenproptools.py`** — 2 new regression tests for the core fixes above.

Run locally with:
```bash
pip install -e ".[ml]"
pytest -v test/test_packages/extras/ml/test_ml.py
pytest -v test/unit/tools/test_errgenproptools.py
```

---

### Documentation

A **4-part tutorial series** in `pygsti/extras/ml/`:

| Notebook | Content |
|---|---|
| `QPANN-Overview.ipynb` | What QPANNs are, why physics-awareness matters, series roadmap, environment check |
| `QPANN-CircuitEncoding.ipynb` | Pauli/error-generator indexing, graph-restricted enumeration, C/A errors, circuit→tensor encoding |
| `QPANN-ErrorPropagation.ipynb` | Clifford propagation with a worked row-by-row example, α tensors, snippers at `hops=0` vs `hops=1` |
| `QPANN-Training.ipynb` | End-to-end: build a `CloudNoiseModel` with known ground-truth H/S/C/A rates, generate 600 `edgegrab` circuits on a 3-qubit ring, train, then **recover the true rates to 4.5–11% relative error** |

`QPANN-Training.ipynb` includes a prominent warning block on physicality and sorting pitfalls when constructing C/A ground truth:
- the CP bound `c² + a² ≤ s_P·s_Q` (violation triggers a hard `Lindblad coefficients are not CPTP!` assertion);
- pyGSTi's model builder **silently drops** reversed C/A pair keys — the notebook adds a defensive `.coefficients()` read-back assertion;
- sign consistency under `A_{P,Q} = −A_{Q,P}`.

---

### Backward compatibility

No breaking changes to existing APIs.

- `pygsti.extras.ml` is a new, optional, non-auto-imported sub-package; core pyGSTi does not depend on it.
- H/S error generator indices are unchanged.
- Deprecated graph kwargs still work, with a `DeprecationWarning`.
- The only pre-existing files touched are three core modules, all covered above.

---

### Known limitations and follow-up work

Called out explicitly so reviewers can scope the review; none of these are regressions.

**Coverage / infrastructure**
- **The ML test suite does not run in CI.** `feature-branches.yml` and `develop.yml` set `run-extra-tests: 'false'`. On `beta`/`master`, where it's `'true'`, `test_ml.py`'s unguarded module-level `import tensorflow` would produce a **collection error**, not a skip, because `ml` is not in the `testing` extra. Needs either a `pytest.importorskip` guard or a CI job that installs `.[ml]`. *(The two `test/unit` fixes above are unaffected — those do run everywhere.)*
- The QPANN notebooks live under `pygsti/extras/ml/`, so they aren't covered by the `nbval` job (which targets `jupyter_notebooks/` only, and is disabled in every workflow anyway).

**Functional gaps**
- **The Pauli-expectation / process-fidelity pathway is unwired.** `encoding` can produce φ-sensitivity tensors (`measurements='paulis'`, `alpha_coefficient_pauli`) and `qpanns.FidelityLayer` exists, but no `QPANN` mode consumes them and `FidelityLayer` is never instantiated. This is the main remaining architectural gap.
- The `'matrix'`/`'expanded'` α path is **H/S-only** (deliberate — `O(16ⁿ)` width otherwise), with an explicit `NotImplementedError`.
- **The CP bound `c² + a² ≤ s_P·s_Q` is not enforced architecturally.** The network's output layer squares only `S` rates; joint physicality across a C/A pair is left to the loss/data. The training notebook validates it post-hoc.
- `encoding.dense_dataset_encoding` is a non-functional stub (`nbit_strings = []  # TO DO`).

**Code quality**
- ~445 lines of commented-out legacy model classes remain at the end of `qpanns.py`, plus a commented-out `sparse_alpha_matrix` in `encoding.py`. Kept intentionally for historical reference; candidates for deletion.
- `QPANN` is registered for Keras serialization under the placeholder package name `'Blah1'`.
- `QPANN.build(self)` takes no `input_shape`, deviating from the Keras convention.
- `CustomDense.call`'s docstring transposes the batch and error-generator axes relative to its own einsum.
- A documented convention inconsistency remains between `up_to_weight_k_paulis` (docstring claims reverse qubit indexing, body uses direct) and `up_to_weight_k_paulis_from_qubit_graph` (direct). Both call sites use direct; the docstring should be corrected.
- `customlayers.py` doesn't follow the package's `import x as _x` underscore convention.
- Assorted unused imports (`ErrorGeneratorPropagator`, `copy`, `warnings`, `itertools.starmap`).

---

### Merge notes

Merging into current `develop` produces **two conflicts**:

1. `pygsti/modelmembers/operations/experrorgenop.py` — content conflict; keep both the unitary cast and `develop`'s `stateless_data`/`torch_base`.
2. `jupyter_notebooks/Tutorials/algorithms/ErrorGeneratorPropagation.ipynb` — modify/delete; `develop` removed `jupyter_notebooks/` entirely in the JupyterBook migration (#673). Resolve by accepting the deletion.

Given the 153-commit, multi-year, 9-author history, a **squash merge** is recommended.

pygsti/extras/ml/graphtools.py
rileyjmurray, line 1: The develop branch now has a `pygsti.tools.graphcoloring` subpackage. In view of your graphtools.py module I suggest we rename the subpackage to `pygsti.tools.graphs` and have it include both graphcoloring and the contents of your new graphtools.py. I don't think we need to do that reorganization as part of this PR. Instead I just ask that you open a github issue pointing out the benefits of a `pygsti.tools.graphs` subpackage, linking to this comment, and saying we should restructure before the release of pygsti 0.11.x.

pygsti/extras/ml/QPANN-CircuitEncoding.ipynb
rileyjmurray, line 1: Notebooks should be turned into jupyterbook-compatible markdown and live in `/docs/markdown/`.

(This applies to all notebooks. Only leaving the comment here.)

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,996 @@
...
+            # go out of [0, 1] for one sign of eps). This does not hold for 'S'-type (stochastic)
+            # errors, whose rates are constrained to eps >= 0, so being at a boundary doesn't
+            # force a one-sided derivative to vanish. Skipping the (otherwise unnecessary)
+            # alpha_coefficient call here is purely a performance optimization: it was checked
+            # empirically (across many random stabilizer states/qubit counts) that
+            # alpha_coefficient already evaluates to exactly 0 in this regime, so this does not
+            # change any computed alphas -- but computing it anyway costs real time, and for
+            # high-fidelity (near-deterministic, i.e. low "random support") circuits -- the
+            # regime this whole codebase targets -- the vast majority of bitstrings can be at
+            # p_bs = 0, so skipping them here is a substantial speedup in this hot loop.
```
rileyjmurray: Please reformat for improved readability.

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,996 @@
...
+#     contributing_error_generators = [[] for bs in bitstrings]
+#     alphas = [[] for bs in bitstrings]
+#     for i, bs in enumerate(bitstrings):
+#         for j in error_generators:
+#             alpha = alpha_coefficient(j, num_qubits, tableau, bs)
+#             if not _np.isclose(alpha, 0.):
+#                 contributing_error_generators[i].append(j)
+#                 alphas[i].append(alpha)
+
+#     return contributing_error_generators, alphas
```
rileyjmurray: Please remove.

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,996 @@
...
+    # generator's Pauli(s) in whatever order Clifford propagation happened to produce them --
+    # NOT necessarily the canonical (lexicographically sorted) order that `error_generator_index`
+    # uses internally to assign a stable index to a 'C'/'A' pair (regardless of input order).
+    # For 'C'-type generators this canonicalization is "free" (C_{P,Q} = C_{Q,P}, so reindexing
+    # into canonical order changes nothing). But for 'A'-type generators, A_{P,Q} = -A_{Q,P}
+    # (antisymmetric under swapping its two indexing Paulis -- see "A Taxonomy of Small Errors",
+    # Sec. V.D, Eq. 16), so silently reindexing a non-canonically-ordered propagated 'A' label
+    # into its canonical index WITHOUT also flipping the sign of its propagated rate would use
+    # the wrong sign. `error_generator_canonicalization_sign` returns exactly the compensating
+    # +-1 factor needed (and is a no-op, always returning +1, for 'H'/'S'/'C').
```
rileyjmurray: Please make more readable.

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,996 @@
...
+        Circuit to compute propagation data for.
+    error_generators : list
+        List of error generator labels. Each element should be a tuple:
+            (errorgen_type, paulis_tuple)
+        where `errorgen_type` is the error generator type ('H'', 'S', 'C', or 'A') and `paulis_tuple` is a tuple of Pauli strings.
+        For 'H' and 'S', this is a single Pauli string. The Pauli strings should be length-n strings
+        where n is the number of qubits on which `circuit` acts.        
+
+
+
```
rileyjmurray: Excessive number of blank lines.

jupyter_notebooks/Tutorials/tutorial_files/myFirstGSTReport/main.html
rileyjmurray, line 1: Please delete.

jupyter_notebooks/Tutorials/algorithms/ErrorGeneratorPropagation.ipynb
rileyjmurray, line 1: You can safely delete this file. Its changes are already reflected in the superseding file within `docs/markdown`.

jupyter_notebooks/Tutorials/tutorial_files/Example_Short_Dataset.txt
rileyjmurray, line 1: Please delete.

pygsti/extras/ml/qpanns.py
rileyjmurray, line 1: Please remove the hundreds of lines of commented-out code in the bottom half of this file.

pyproject.toml
```diff
@@ -74,6 +67,21 @@ linting = [
...
+# a.k.a. QPANNs, sub-package). Pinned to `tensorflow>=2.16` (which brings in Keras 3, via
+# tensorflow's own `keras>=3.0.0` dependency) because pygsti.extras.ml's custom Keras layers
+# were fixed to work against Keras 3's API (Keras 3 made `keras.layers.Dense.kernel` a
+# read-only property, among other changes). If you instead want to match the original QPANN
+# paper's environment (TensorFlow/Keras 2.12), install those older pins directly rather than
+# using this extra -- note that requires Python <=3.11, since `tensorflow<2.16` has no wheels
+# for Python 3.12+. This extra is deliberately NOT included in `testing`/`no_mpi`/`complete`
+# below: TensorFlow is a heavy, platform-specific dependency, and pygsti.extras.ml's tests are
+# not currently run in CI regardless of whether it's installed (see
+# `.github/workflows/feature-branches.yml`'s `run-extra-tests: 'false'`).
```
rileyjmurray: This file doesn't need to hold justifications of version number bounds. Tutorials or release notes can include this level of detail if need be.

rileyjmurray (changes requested):
Thanks for the hard work, @nonohuff! This is super exciting. I've mostly left comments at the file level. I also provide some comments below that aren't naturally attached to a an individual file you've changed.

Requests:
 * Please change dates on new file copyright notices to 2026.
 * Please add a note about this feature to the CHANGELOG file (only after you've pulled in changes from develop).

General Python style:
 * Two blank lines between class definitions.
 * One blank line between return statements and the next function definition within a class.
 * Two blank lines between module-level free functions (i.e., `def` statements that don't have a leading indentation).

rileyjmurray:
Oh, one more comment I left out of the review.

> **Coverage / infrastructure**
> 
> * **The ML test suite does not run in CI.** `feature-branches.yml` and `develop.yml` set `run-extra-tests: 'false'`. On `beta`/`master`, where it's `'true'`, `test_ml.py`'s unguarded module-level `import tensorflow` would produce a **collection error**, not a skip, because `ml` is not in the `testing` extra. Needs either a `pytest.importorskip` guard or a CI job that installs `.[ml]`. _(The two `test/unit` fixes above are unaffected — those do run everywhere.)_

It would be bad if test collection failed just because we have a missing dependency. For example, if someone runs `pytest tests/*` they shouldn't encounter an error. Please add a `pytest.importorskip` to make sure collection doesn't break. 

Tests for this subpackage should actually run (i.e., not get skipped by a `pytest.importorskip` call) when the CI build flag `run-extra-tests` is true. This will involve some changes to our CI scripts.

nonohuff:
Thanks for the thorough review, @rileyjmurray — all of it is addressed. Summary below; each item is its own commit so they can be read independently.

**First, the big one: `develop` is merged in** (249 commits, merge base was `4983cc1e8`). Only two conflicts. Since `develop` replaced `jupyter_notebooks/` with `docs/markdown/`, three of your comments resolved as part of that merge.

### Review comments

| # | Comment | Done in |
|---|---|---|
| 1 | `pygsti.tools.graphs` subpackage | Issue #866 filed, linking your comment, targeted before 0.11.x |
| 2 | Notebooks → jupyterbook markdown in `/docs/markdown/` | `15130a0d2` |
| 3 | `encoding.py` H-sensitivity comment: "reformat for readability" | `9757f13b9` — 15 lines → 6 |
| 4 | `encoding.py` `sparse_alpha_matrix`: "please remove" | `b29722068` |
| 5 | `encoding.py` canonicalization NOTE: "make more readable" | `9757f13b9` — 12 lines → 4 |
| 6 | `encoding.py`: "excessive number of blank lines" | `9757f13b9` |
| 7 | `myFirstGSTReport/main.html`: "please delete" | merge `5917ef547` |
| 8 | `ErrorGeneratorPropagation.ipynb`: safe to delete | merge `5917ef547` |
| 9 | `Example_Short_Dataset.txt`: "please delete" | merge `5917ef547` |
| 10 | `qpanns.py`: "hundreds of lines of commented-out code" | `b29722068` — 444 lines |
| 11 | `pyproject.toml`: doesn't need version-bound justifications | `9757f13b9` — 11 lines → 3 |
| — | Copyright notices → 2026 | `d9818ed3b` |
| — | CHANGELOG note (after merging develop) | `7a27195d7` |
| — | Blank-line conventions | `d9818ed3b` — 164 flake8 violations → 0 |
| — | `pytest.importorskip` + make the tests actually run in CI | `9d6beb962` |

On #8 — I confirmed before dropping it that this branch's only real edit to that notebook (`random_error_generator_rates` → `random_CPTP_error_generator_rates`) is already present in `docs/markdown/simulation/ErrorGeneratorPropagation.md`.

### Three things worth your attention

**1. The `ExpErrorgenOp.to_dense` change was not ours, and it's a real bug on develop.** You didn't flag it, but it was unrelated core code riding along here, so I went looking. It turns out to be @sserita's commit `0bdbb2830`, *"Convert ExpErrorgenOp, not LindbladErrorgen, to unitary"* (Jul 2024) — an ancestor of the `feature-ml-*` branches only, not of develop/beta/master. So develop currently has neither half of that move, and:

```python
eg = LindbladErrorgen.from_elementary_errorgens({('H','X'): 0.1}, evotype='statevec', state_space=1)
op = ExpErrorgenOp(eg)
op.evotype.minimal_space   # 'Hilbert'
op.to_dense().shape        # (4, 4)  <- superoperator, not a 2x2 unitary, silently
op.to_dense('Hilbert')     # bare AssertionError from inside LindbladErrorgen.to_dense
```

Extracted to **#865** off develop, with unit tests and a docstring. Removed from this branch in `2011a15e6`, so `experrorgenop.py` is byte-identical to develop here again.

**2. A silent semantic conflict from the merge.** `ErrorGeneratorPropagator.__init__` now asserts that one of `model`/`fixed_errorgen_layer` is supplied. `encoding.py` passed `None` deliberately — it only wants the stateless propagation helpers, and builds its error generator layers explicitly. The merge was textually clean, so nothing flagged it; it broke 7 of 32 tests and only showed up on running them. Now passes `fixed_errorgen_layer={}` (`200c9e12c`).

**3. Your CI point had a second half I want to check with you.** `importorskip` is in, verified against a simulated TensorFlow-less interpreter: `pytest test/test_packages/extras/ml/` goes from *"1 error during collection"* to *"1 skipped"*, and `--collect-only test/test_packages` collects all 228 tests instead of aborting.

For making them actually run, I added an `uv pip install --system -e .[ml]` step gated on `run-extra-tests == 'true'`, rather than folding `ml` into the `testing` extra — TensorFlow would otherwise be installed for all 30 jobs in the beta-master matrix to benefit one directory. The notebook-regression step installs `.[docs,ml]` for the same reason: I deliberately did **not** add TensorFlow to the `docs` extra, because ReadTheDocs installs `.[docs]` but never executes a notebook (`execute_notebooks: 'off'`), so it would download ~600 MB purely to render Markdown. Happy to change either if you'd rather.

### On the notebooks

They're now `docs/markdown/ml/{Overview,CircuitEncoding,ErrorPropagation,Training}.md`, registered in `_toc.yml` after the Simulation chapter. Three things needed fixing beyond a plain conversion:

- The markdown already used `{{num_qubits}}`-style substitutions, but **none of them were ever defined** — they'd have rendered literally. Now defined in each page's frontmatter.
- `jupytext --sync` silently drops metadata it doesn't recognise, which deleted those definitions again. Since CI runs exactly that command before nbval, they'd have been destroyed on every docs build. Fixed with `notebook_metadata_filter: substitutions`.
- `{{epochs}}`/`{{learning_rate}}` were inside inline-code backticks, where MyST doesn't substitute.

`Training.md` is resized (600 circuits → 200, 400 epochs → 150) so nbval can afford it: 3m26s → ~90s. This costs nothing — `EarlyStopping` halted the original at **89 of its 400 epochs**, and the nine recovered rates land at the same 5–11% relative error either way. There's a `{note}` telling readers exactly that, and what does warrant scaling up (more qubits, more modelled generators, or shot-noisy real data).

Verified with the real CI commands: `jupytext --sync` then `pytest --nbval-lax --nbval-current-env docs/markdown/ml` → 16 cells, all pass. Also ran a full `jb build docs/` (exit 0, no warnings on these pages, no unresolved `{{...}}` in the built HTML).

### Verification

- `test/test_packages/extras/ml/test_ml.py` — 32 passed
- `mypy pygsti/extras/ml` — clean; `flake8 --config=.flake8-critical` — 0
- PEP8 blank-line/whitespace debt in `extras/ml` — 164 → 0; also fixed 26 `W605` invalid escapes that were emitting `SyntaxWarning` on every import
- full `test/unit` — the only failures are pre-existing on develop (`test_gst.py::test_run_with_no_gaugeoptsuite`, reproduced on a pristine develop worktree, plus the `@pytest.mark.slow` graphcoloring timing tests, which CI's `-m "not slow"` deselects anyway)

**This branch's diff against develop is now 21 files, +5800/−6** — down from 23 files and +22,043/−74, and essentially just `pygsti/extras/ml`, its tests and docs, the two `errgenproptools` bugfixes, and the packaging/CI touches.

pygsti/extras/ml/__init__.py
rileyjmurray, line 1: Need blank line at EOF

pygsti/extras/ml/customlayers.py
```diff
@@ -0,0 +1,309 @@
...
+#     list(range(1, 100, 2)),  # Neuron 2: odd indices
+#     list(range(0, 50))       # Neuron 3: indices 0 to 49
+# ]
+
+# inputs = tf.keras.Input(shape=(100,))
+# selective_dense_layer = SelectiveDense(units=3, input_indices=input_indices, activation='relu')
+# outputs = selective_dense_layer(inputs)
+
+# model = tf.keras.Model(inputs=inputs, outputs=outputs)
+# model.summary()
```
rileyjmurray: Delete or move elsewhere.

pygsti/extras/ml/encoding.py
```diff
@@ -465,7 +470,13 @@ def circuit_error_propagation_matrices(circuit: _Circuit, error_generators: list
...
+    # An empty `fixed_errorgen_layer` (rather than no arguments at all) is required: the
+    # constructor asserts that exactly one of `model`/`fixed_errorgen_layer` is given. We need
+    # neither -- this instance is only used for its stateless propagation helpers
+    # (`construct_stim_layers`/`construct_propagation_layers`/`_propagate_errorgen_layers`, none
+    # of which read `self.model` or `self.fixed_errorgen_layer`), because the error generator
+    # layers are built explicitly below from `error_generators`.
+    error_propagator = _ep.ErrorGeneratorPropagator(fixed_errorgen_layer={})
     stim_layers = error_propagator.construct_stim_layers(circuit, drop_first_layer=True)
     propagation_layers = error_propagator.construct_propagation_layers(stim_layers)
     # TODO : It's the list of error gens in every layer of the circuit.
```
rileyjmurray: 1. Please shorten this explanation. 
2. What's the TODO about?

pygsti/extras/ml/encoding.py
```diff
@@ -474,24 +485,17 @@ def circuit_error_propagation_matrices(circuit: _Circuit, error_generators: list
...
-    # into canonical order changes nothing). But for 'A'-type generators, A_{P,Q} = -A_{Q,P}
-    # (antisymmetric under swapping its two indexing Paulis -- see "A Taxonomy of Small Errors",
-    # Sec. V.D, Eq. 16), so silently reindexing a non-canonically-ordered propagated 'A' label
-    # into its canonical index WITHOUT also flipping the sign of its propagated rate would use
-    # the wrong sign. `error_generator_canonicalization_sign` returns exactly the compensating
-    # +-1 factor needed (and is a no-op, always returning +1, for 'H'/'S'/'C').
+    # Clifford propagation can return an 'A' label whose two Paulis are not in the canonical
+    # order `error_generator_index` indexes by. Since A_{P,Q} = -A_{Q,P} ("A Taxonomy of Small
+    # Errors", Sec. V.D, Eq. 16), reindexing must be accompanied by a compensating sign.
+    # `error_generator_canonicalization_sign` supplies it (always +1 for 'H'/'S'/'C').
```
rileyjmurray: Excellent simplification.

pygsti/extras/ml/graphtools.py
rileyjmurray: Since you opened issue #866 I'm marking this thread as resolved.

pygsti/extras/ml/legacy.py
```diff
@@ -152,5 +156,5 @@ def create_mle_model(df: "pandas.DataFrame", indices: dict, pspec: "ProcessorSpe
     gst_protocol = GST(specmodel, gaugeopt_suite=cast(Any, None),  badfit_options={}, verbosity=1) #
     gst_results = gst_protocol.run(gst_protocol_data) #fits an mle model to the gst_protocol_data
     mlemodel = gst_results.estimates['GateSetTomography'].models['final iteration estimate'] # extracts the mle model parameters
-    
+
     return mlemodel
```
rileyjmurray: Missing blank line at EOF

pygsti/extras/ml/qpanns.py
```diff
@@ -173,22 +174,24 @@ def circuit_to_probability(self, inputs: list | tuple) -> _tf.Tensor:
         else:
             raise ValueError("Invalid probability_computation choice: " + str(self.probability_computation))
 
-        # Old code that was written by someone that attempted to find a correction to the first-order approximation. 
+        # Old code that was written by someone that attempted to find a correction to the first-order approximation.
         # C = _tf.reshape(self.dense_correction(_tf.reshape(circuit_encoding, [1, -1])), [-1])
         # probabilities = probabilities +C
-        
+
```
rileyjmurray: Probably worth removing. But since it's short you can keep if you see conceptual value in doing so.

pygsti/modelmembers/operations/experrorgenop.py
rileyjmurray, line 1: Now that PR #865 has been merged into develop you should merge develop back into this branch. That'll eliminate the diff here.

rileyjmurray (changes requested):
Very small comments. I'll do a more thorough review that gets at content later.

Edit: some of my comments seem to have been on outdated files? Not sure how that happened. If a comment doesn't make sense then please just ignore and mark as resolved.

pygsti/extras/ml/__init__.py
nonohuff: Done.

pygsti/extras/ml/customlayers.py
nonohuff: Done.

pygsti/extras/ml/encoding.py
nonohuff: Done.

pygsti/extras/ml/legacy.py
nonohuff: Done

pygsti/modelmembers/operations/experrorgenop.py
nonohuff: Done.

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+    (the default in `error_generator_tensors`) for circuits/models involving 'C'/'A' generators.
+
+    Parameters
+    ----------
+    circuits : list[pygsti.circuits.Circuit]
+        Circuits to process.
+    error_generators : list
+        Currently unused directly (kept for API symmetry); alphas are computed in the
+        canonical indexing over all H/S end-of-circuit generators, optionally restricted
+        by `indices`.
```
rileyjmurray: API symmetry with what? I'm skeptical that the symmetry is valuable enough to warrant an unused function parameter.

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+            for k in range(indices.shape[1]):
+                first_order_coefficients[l, j, k] = alphas_dict[l, indices[j,k]]
+
+    return (probabilities, first_order_coefficients)
+
+
+def _circuit_loop_paulis(circuit: _Circuit, indices: _np.ndarray, num_qubits: int, paulis: list[str]) -> tuple[_np.ndarray, _np.ndarray]:
+
+    unique_indices = set(indices.flatten())
+
```
rileyjmurray: Rouge blank lines.  (Irrelevant if you remove this function.)

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+    else:
+        raise ValueError(f"Unknown measurements type: {measurements}")
+
+    return measurements_array, first_order_coefficients
+
+
+def _circuit_loop_probs(circuit: _Circuit, indices: _np.ndarray, nbit_strings: list[str], num_qubits: int) -> tuple[_np.ndarray, _np.ndarray]:
+
+    unique_indices = set(indices.flatten())
+
```
rileyjmurray: Rouge blank lines.  (Irrelevant if you remove this function.)

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+    for l, bs in enumerate(nbit_strings):
+        for error_generator_index in unique_indices:
+            egtype = cast(Any, _tools.index_to_error_gen(error_generator_index, num_qubits))[0]
+
+            # A definite-outcome bitstring (p_bs exactly 0 or 1) has zero first-order
+            # sensitivity to an 'H'-type error, since p_bs is bounded to [0, 1] and the rate can
+            # take either sign. `alpha_coefficient` already returns exactly 0 here, so this is
+            # purely a speedup -- but a worthwhile one, as most bitstrings sit at p_bs = 0 for
+            # the near-deterministic circuits this code targets. ('S' rates are one-sided, so
+            # the argument does not apply to them.)
```
rileyjmurray: Please move this inside the if-statement.  (Irrelevant if you remove this function.)

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+
+
+def _circuit_loop_paulis(circuit: _Circuit, indices: _np.ndarray, num_qubits: int, paulis: list[str]) -> tuple[_np.ndarray, _np.ndarray]:
+
+    unique_indices = set(indices.flatten())
+
+    tableau = _get_tableau(circuit)
+    shape = (len(paulis), indices.shape[0], indices.shape[1])
+    first_order_coefficients = _np.zeros(shape, float)
+    #scale = 1 / 2 ** _egptools.random_support(tableau) #TODO: This might overflow
```
rileyjmurray: Remove commented-out code. (Irrelevant if you remove this function.)

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+    measurements = _np.array([_egptools.stabilizer_pauli_expectation(tableau, p) for p in paulis]).T
+    alphas_dict = {}
+    for l, p in enumerate(paulis):
+        for error_generator_index in unique_indices:
+            egtype = cast(Any, _tools.index_to_error_gen(error_generator_index, num_qubits))[0]
+            #print(egtype)
+            # TIM THINKS THIS IS CORRECT BUT COMMENTING OUT WHILE BUG FIXING
+            #if egtype == 'H' and (_np.isclose(measurements[l], -1.) or _np.isclose(measurements[l], 1.)):
+            #    alpha = 0
```
rileyjmurray: Presumably you're no longer bug fixing. (Irrelevant if you remove this function.)

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+
+    measurements: measurement to calculate the sensitivity tensor for, currently supports 'probabilties' (bitstring probabilties)
+    and 'paulis' (pauli outcomes)
+
+    pauli_params: describes the measurements to use in the calculation, recieves a two component list, the first argument in the max weight
+    of any pauli observables, the second element is the list of allowed paulis
+
+    process_num: number of processes to use for parallel computation
+
+    """
```
rileyjmurray, line 759: unfinished docstring

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+
+    elif measurements == 'paulis':
+
+        assert(measurement_paulis is not None), "Must provided the measurement Pauli operators!"
+        shape = (indices.shape[0], len(measurement_paulis), indices.shape[1], indices.shape[2])
+        first_order_coefficients = _np.zeros(shape, float)
+        measurements_array = _np.zeros((len(circuits),len(measurement_paulis)))
+        circ_indices_tuples=[]
+
+        # TIM COPIED OUT WHILE BUG HUNTING
```
rileyjmurray: What?

pygsti/extras/ml/snippers.py
```diff
@@ -0,0 +1,164 @@
...
+        predicting the rate of the corresponding error generator. This list is in the correct
+        format to be passed to an initialization of a QPANN, as the `snipper` argument.
+
+    Notes
+    -----
+    "Within `hops` steps" is determined by true (unweighted) shortest-path graph distance
+    (computed via breadth-first search). Earlier versions of this function instead computed a
+    graph Laplacian `L = D - A` and used `L**hops` to infer which nodes are within `hops` steps
+    (via nonzero entries); that was a heuristic (and, for an isolated qubit with no edges, an
+    inaccurate one -- it dropped the qubit's own index from its own list for any `hops >= 1`).
```
rileyjmurray: Do you have a specific reason to document what the earlier version of the function did? That sort of comment is typical of LLMs but often of little to no value.

pygsti/extras/ml/snippers.py
```diff
@@ -0,0 +1,164 @@
...
+        # for 'C'/'A' error generators (see "A Taxonomy of Small Errors", Sec. V.C-V.D). In
+        # either case, we take the UNION of the qubits acted on by every Pauli in the tuple,
+        # since (per the same paper, Sec. VIII) the "support"/"weight" of a 'C'/'A' generator
+        # C_{P,Q}/A_{P,Q} is defined as the union of P's and Q's individual qubit supports.
+        pauli_strings = error_generator[1]
+        # The following commented-out line is *wrong* but it used to be in the code, so leaving it here
+        # but commented out for now. It is unclear if somehow this was the correct thing to do in older
+        # versions of the QPANN code before my (Tim's) rewrite.
+        # pauli_string = pauli_string[::-1] # for reverse indexing
+        #
```
rileyjmurray: Please figure out what to do with this.

pygsti/extras/ml/snippers.py
```diff
@@ -0,0 +1,164 @@
...
+        positional and are matched to `encoder.pspec.qubit_labels` by position. See
+        `pygsti.extras.ml.graphtools.qubit_graph_to_networkx` for the full list of accepted
+        types and exactly how they're interpreted.
+
+    hops : int
+        The number of steps on the qubit graph to take.
+
+    input_is : {'auto', 'laplacian', 'adjacency'}, optional
+        Only consulted when `qubit_graph` is a bare matrix; see
+        `pygsti.extras.ml.graphtools.qubit_graph_to_networkx`.
```
rileyjmurray: I think it costs basically nothing to decide if you're handed a Laplacian or an adjacency matrix. I suggest this option be removed.

pygsti/extras/ml/snippers.py
```diff
@@ -0,0 +1,164 @@
...
+
+    adjacency_matrix : numpy.ndarray, optional
+        Deprecated alias for `qubit_graph`. Specify only one of the two.
```
rileyjmurray: Probably not worth a deprecated argument alias for a function that's never appeared in a pygsti release.

pygsti/extras/ml/legacy.py
rileyjmurray, line 1: Is it really worth keeping this file? If reproducibility is a concern then we can (1) tag a commit that has this file, (2) reference that tag in appropriate places, and (3) remove this file from the PR.

pygsti/extras/ml/encoding.py
rileyjmurray, line 1: It's not clear to me that `_circuit_loop_paulis`, `make_pauli`, and `make_paulis_of_weight` are needed. If you comment them out (and comment out the last two lines in `test/test_packages/extras/ml/test_ml.py::MLSubpackageTester::test_encoding) then all tests pass.

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+        for j in range(indices.shape[0]):
+            for k in range(indices.shape[1]):
+                first_order_coefficients[l, j, k] = alphas_dict[l, indices[j,k]]
+
+    return (measurements, first_order_coefficients)
+
+
+def make_paulis(num_qubits: int, maximum_weight: int) -> list[Any]:
+    """
+    """
```
rileyjmurray: remove, or write, or remove the entire function

pygsti/extras/ml/encoding.py
```diff
@@ -0,0 +1,957 @@
...
+    for w in range(1, maximum_weight+1):
+        paulis += make_paulis_of_weight(num_qubits, w)
+    return paulis
+
+
+def make_paulis_of_weight(num_qubits: int, weight: int) -> list[Any]:
+    """
+    num_qubits : number of qubits
+    weight : the weight of the Pauli operators
+    """
```
rileyjmurray: remove, or write properly, or remove entire function

pygsti/extras/ml/errgentools.py
```diff
@@ -0,0 +1,1288 @@
...
+#         for i in range(n):
+#             for j in range(i + 1, n):
+#                 for p in ['X', 'Y', 'Z']:
+#                     for q in ['X', 'Y', 'Z']:
+#                         nq_pauli = n * ['I']
+#                         nq_pauli[n - 1 - i] = p # reversed index
+#                         nq_pauli[n - 1 - j] = q # reversed index
+#                         paulis.append(''.join(nq_pauli))
+
+#     return paulis
```
rileyjmurray: Remove

pygsti/extras/ml/errgentools.py
```diff
@@ -0,0 +1,1288 @@
...
+    all-Q-only (which would make P the identity, disallowed), and all-'both'-with-matching-
+    letters (which would make P == Q, disallowed) -- these three cases are disjoint for `w >= 1`
+    (each forces the *other* two Pauli(s) to be manifestly non-identity/distinct), so by
+    inclusion-exclusion there are `15**w - 3*3**w = 15**w - 3**(w+1)` valid *ordered* pairs, and
+    half that many *unordered* (canonically-ordered) pairs, since swapping the roles of P and Q
+    in the per-qubit choices always produces another valid combination representing the same
+    unordered pair. This was cross-checked (during development) against
+    `num_pauli_pairs(n)` (the *unrestricted* total, obtained here by summing over all supports
+    of all sizes `1..n`) for small `n`, and against the paper's own stated total of 105 for
+    `n=2`; see the corresponding unit tests.
```
rileyjmurray: Please structure this note in a more readable way.

pygsti/extras/ml/graphtools.py
rileyjmurray, line 1: We seem to be paying a material price by supporting graph specification by adjacency matrix _or_ Laplacian. What's the rationale for Laplacian support? It's rarely used as a way of exchanging data structures since it contains redundant information (the diagonal). Also, if someone has a Laplacian, they most definitely know how to convert it to an adjacency matrix. The converse is not true.

pygsti/extras/ml/qpanns.py
```diff
@@ -0,0 +1,516 @@
...
+import tensorflow as _tf
+import keras as _keras
+import numpy as _np
+import copy as _copy
+import warnings as _warnings
+from pygsti.extras.ml import customlayers as _cl
+
+
+@_keras.utils.register_keras_serializable(package='Blah1')
```
rileyjmurray: `package='Blah1'`?

pygsti/extras/ml/qpanns.py
```diff
@@ -0,0 +1,516 @@
...
+        elif self.probability_computation == 'concise':
+            corrections_coefficients = inputs[1]  # The alpha coefficients in a (circuit depth, self.modelled_error_generators) array
+            probabilities_ideal = inputs[2]  # ideal (no error) probabilities
+            probabilities = self.probability_approximation_layer([error_rates, corrections_coefficients, probabilities_ideal])
+        else:
+            raise ValueError("Invalid probability_computation choice: " + str(self.probability_computation))
+
+        # Old code that was written by someone that attempted to find a correction to the first-order approximation.
+        # C = _tf.reshape(self.dense_correction(_tf.reshape(circuit_encoding, [1, -1])), [-1])
+        # probabilities = probabilities +C
```
rileyjmurray: Please confirm that you actually want to keep this commented-out code.

pygsti/extras/ml/qpanns.py
```diff
@@ -0,0 +1,516 @@
...
+
+        return probabilities
+
+    # @_tf.function
```
rileyjmurray: Remove commented-out decorator.

pygsti/extras/ml/qpanns.py
```diff
@@ -0,0 +1,516 @@
...
+    def call(self, inputs: list | tuple | _tf.Tensor) -> _tf.Tensor:
+        """Vectorize `circuit_to_probability` over a batch using `tf.map_fn`."""
+        return _tf.map_fn(self.circuit_to_probability, inputs, fn_output_signature=_tf.float32)
+
+# ------------------------------------------------------------------- #
+#        Main part of the QPANNs (input circuit --> error rates matrix)
+# ------------------------------------------------------------------- #
+
+# @_keras.utils.register_keras_serializable()
```
rileyjmurray: Remove

pygsti/extras/ml/qpanns.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+class FidelityLayer(_keras.layers.Layer):
+    """(Incomplete/experimental) layer intended to map error rates to process fidelity predictions.
+
+    Warning
+    -------
+    This class appears incomplete/duplicated (two `call` methods exist) and references
+    undefined variables (e.g., `Px_ideal` in one call). It is preserved as-is; docstrings
+    here reflect intent rather than guaranteed behavior.
```
rileyjmurray: Do whatever you need to so that this is complete/non-experimental. Consider removing and opening a github issue to add this feature if you don't want to do that in this PR.

rileyjmurray (changes requested):
Plenty of comments for ya!

pygsti/extras/ml/encoding.py
nonohuff: It wasn't kept for symmetry with anything in particular — on inspection the parameter had no effect on the output (which end-of-circuit generator indices get computed is controlled entirely by `indices`, not by `error_generators`). Removed in e083b3108, along with folding `prior_error_generators`/`prior_alphas` into a single `prior_alphas` for the same reason (this function's alpha matrices are always full H/S-width regardless of which generators are asked for, so there was never really a separate "prior generators" concept here either).

pygsti/extras/ml/encoding.py
nonohuff: Moot — `_circuit_loop_paulis` (and the whole `measurements='paulis'` pathway) is removed in c4801a77e; see my reply on the `FidelityLayer` thread below for the full rationale.

pygsti/extras/ml/encoding.py
nonohuff: Done, in c4801a77e.

pygsti/extras/ml/encoding.py
nonohuff: Done, in c4801a77e.

pygsti/extras/ml/encoding.py
nonohuff: Moot — this function is removed in c4801a77e; see my reply on the `FidelityLayer` thread below.

pygsti/extras/ml/encoding.py
nonohuff: Moot — this function is removed in c4801a77e; see my reply on the `FidelityLayer` thread below.

pygsti/extras/ml/encoding.py
nonohuff: Done, in c4801a77e (this function's docstring needed a full rewrite anyway once the `measurements='paulis'` branch came out).

pygsti/extras/ml/encoding.py
nonohuff: This whole branch (`measurements='paulis'`) is removed in c4801a77e — see my reply on the `FidelityLayer` thread below for the full rationale.

pygsti/extras/ml/snippers.py
nonohuff: No specific reason — you're right that it isn't valuable. Removed in 0991c42ba.

pygsti/extras/ml/snippers.py
nonohuff: Deleted it. The current code is correct and covered by existing tests (Pauli-string position `i` always maps to qubit `i` throughout this package, per graphtools.py's module docstring); the reversed-indexing convention it was hedging against belonged to a pre-rewrite version of this code that no longer exists anywhere in the package. Done in 0991c42ba.

pygsti/extras/ml/snippers.py
nonohuff: Agreed — removed entirely (not just the option, the whole Laplacian-vs-adjacency distinction) in 913436c23. See my reply on the graphtools.py thread below for the full rationale, since it's the same fix answering both comments.

pygsti/extras/ml/snippers.py
nonohuff: Removed in 913436c23.

pygsti/extras/ml/legacy.py
nonohuff: Agreed, removed in e2f88251c. `13d7a6e0a` (the last commit on this branch before that one) is the reference point if this code is ever needed again — didn't cut a separate tag for it, but happy to if you'd prefer that over pointing at the commit directly.

pygsti/extras/ml/encoding.py
nonohuff: Confirmed: none of `_circuit_loop_paulis`/`make_paulis`/`make_paulis_of_weight` (nor `alpha_coefficient_pauli`, nor the `measurements='paulis'` branch that calls them) has any caller anywhere in the repo, so removed all of it in c4801a77e rather than commenting it out. Opened #869 to track finishing this pathway properly — the underlying math (`errgenproptools.alpha_pauli`/`stabilizer_pauli_expectation`) is untouched, tested pygsti-core code, so nothing here is actually lost.

pygsti/extras/ml/encoding.py
nonohuff: Removed, in c4801a77e — see my reply on the `FidelityLayer` thread below for the full rationale.

pygsti/extras/ml/encoding.py
nonohuff: Removed, in c4801a77e — see my reply on the `FidelityLayer` thread below.

pygsti/extras/ml/errgentools.py
nonohuff: Done, in 667a28c2a.

pygsti/extras/ml/errgentools.py
nonohuff: Restructured into a short bulleted derivation in 667a28c2a.

pygsti/extras/ml/graphtools.py
nonohuff: No good rationale in retrospect — you're right on both counts (the diagonal is redundant information, and anyone holding a Laplacian can trivially negate its off-diagonal). Removed Laplacian support entirely (not just `input_is`) in 913436c23: bare matrices are adjacency-only now, and a matrix with a negative off-diagonal entry raises a `ValueError` explaining it looks like a Laplacian and how to convert it, so old Laplacian-passing code fails loudly instead of silently working for the wrong reason.

pygsti/extras/ml/qpanns.py
nonohuff: Leftover placeholder, fixed in d485c24d3 (`package='pygsti.extras.ml'`).

pygsti/extras/ml/qpanns.py
nonohuff: No, don't need it — removed in d485c24d3.

pygsti/extras/ml/qpanns.py
nonohuff: Done, in d485c24d3.

pygsti/extras/ml/qpanns.py
nonohuff: Done, in d485c24d3.

pygsti/extras/ml/qpanns.py
nonohuff: Traced its actual dependencies before deciding what to do here: `FidelityLayer` is never instantiated by `QPANN.build()`, and consumes 'expanded'-style tensors (`P`/`S`/masks) that don't match what the encoding-side half-attempt at this (`_circuit_loop_paulis`, `alpha_coefficient_pauli`, `make_paulis`/`make_paulis_of_weight`, the `measurements='paulis'` branch) actually produces — these are two independent, unfinished sketches, not two halves of one working pipeline, and nothing in the live QPANN forward path reaches either of them. Removed all of it in c4801a77e, including the `hamiltonian_mask` attribute that only `FidelityLayer` read. Opened #869 to track doing this properly — the underlying math this would need (`errgenproptools.alpha_pauli`/`stabilizer_pauli_expectation`, paper 1 Eq. 5, paper 3 Eq. 35) is untouched, tested pygsti-core code, so a real implementation has a clean foundation to build on.

nonohuff:
Separate from the review comments above (all 24 replied to individually): this branch also carries a small, already-merged-into-`feature-ml-rewrite` fix to core `pygsti/tools/errgenproptools.py` -- commit `6d56c0dc5` ("Fix two UnboundLocalError bugs in errgenproptools.py's A-type-generator code"), two unrelated variable-name typos in `error_generator_commutator`'s S-A branch and `slow_bulk_alpha`'s A1 branch, found while adding 'C'/'A' support to `extras/ml`. It's fully test-covered (+65 lines in `test/unit/tools/test_errgenproptools.py`) and unrelated to `extras/ml` itself.

This is the same situation as the `ExpErrorgenOp.to_dense` change that became #865 off `develop`. Happy to split it out into its own small PR the same way, if you'd rather it not ride along in this one -- just say the word.

rileyjmurray (changes requested):
Thanks for responding to all my comments! This looks good. However, your last in-thread comment indicated that some changes here constitute a bugfix. If that's the case, please split them off into a dedicated PR that we can land before this one. (Ordinarily I'm okay with landing bugfixes in feature PRs. In this case I'd prefer to avoid since I anticipate a bugfix release imminently.)

nonohuff:
Done — split the `errgenproptools.py` A-type `UnboundLocalError` fixes out into #876 (branched from `develop`, same pattern as #865). Removed them from this branch in d474e2f97 (`pygsti/tools/errgenproptools.py` and `test/unit/tools/test_errgenproptools.py` are byte-identical to `develop` again here), so this PR no longer carries a bugfix. Will merge `develop` back into this branch once #876 lands, the same way #865 came back in.

Also gave the PR description at the top of this thread a full rewrite reflecting the branch's current state (post both review rounds, post this split), per your request — including fresh test/mypy/flake8/notebook-regression output.

rileyjmurray (approved):
MWAHAHAHAHA! You have fallen into my trap, @nonohuff! With this approval, and the subsequent PR merge, you will have become responsible for 5000 lines of code in a python package that's central to the field of QCVV. May the code be maintained eternally!!!

PR #837, opened 2026-07-13
Support label-aware Clifford symplectic reps

enielse:
**Summary**
Adds ability to resolve Clifford symplectic representations using full `Label` objects, including arguments, before falling back to gate-name lookup.  Previously we only used the `name` of a Label, which is fine as long as all the gates being compiled have different names but can be insufficient when using parameterized gates.

**Changes**
- Adds support for label-aware symplectic rep factories so parameterized gates can participate in Clifford compilation.
- Preserved/remapd label-like gate arguments during processor-spec cloning, subsetting, and qubit-label mapping.  This is necessary when a parameterized gate's arguments are themselves qubit indices/labels.
- Updates model and processor-spec symplectic-rep handling to retain label-specific reps when same-name gates differ.  Otherwise falls back to old behavior to avoid backward compatibility problems.
- Add unit coverage for label-specific lookup, serialization, argument remapping, symplectic circuit evaluation, and compiler template search.


pygsti/tools/symplectic.py
```diff
@@ -1058,6 +1060,28 @@ def compute_internal_gate_symplectic_representations(gllist=None):
     return srep_dict
 
 
+def _evaluate_clifford_srep(value: _Any, gate_label: _Label):
+    """Return a symplectic representation from a static value or label-aware callable."""
+    return value(gate_label) if callable(value) else value
```
rileyjmurray: I'd rather that we in-line this code rather than define it and only use it twice.  It doesn't seem that meaningful when separated from the contexts where it's currently used.

test/unit/objects/test_processorspec.py
rileyjmurray: While we're here, can you remove this, Erik?

pygsti/processors/processorspec.py
```diff
@@ -994,8 +1017,18 @@ def _to_nice_serialization(self):
...
-        state['symplectic_reps'] = {k: (self._encodemx(s), self._encodemx(p))
-                                    for k, (s, p) in self._symplectic_reps.items()}
+        static_symplectic_reps = {k: v for k, v in self._symplectic_reps.items() if v is not None}
+        if any((isinstance(k, _Lbl) for k in static_symplectic_reps)):
+            state['symplectic_reps'] = [
+                {'key_type': 'label' if isinstance(k, _Lbl) else 'name',
+                 'key': k.to_native() if isinstance(k, _Lbl) else k,
+                 'srep': (self._encodemx(s), self._encodemx(p))}
+                for k, (s, p) in static_symplectic_reps.items()
+            ]
```
rileyjmurray: Please flatten (de-nest) this statement.

pygsti/processors/processorspec.py
```diff
@@ -1067,6 +1109,49 @@ def compute_ops_on_qubits(self):
...
+        return {gn: indices for gn, indices in self.gate_arg_label_indices.items() if gn in gate_name_set}
+
+    def _nonstd_gate_symplecticreps_for_clone(self, gate_names=None, qubit_labels=None, mapper=None):
+        """Return explicit symplectic reps and factories for a cloned processor spec.
+
+        Exact-label entries are filtered when a subset removes labels and are
+        remapped when a clone changes state-space labels.  Name-keyed entries
+        are copied unchanged because they remain applicable to all labels of
+        that gate name.
+        """
```
rileyjmurray, line 1136: Could you add a small concrete example to the docstring?

pygsti/processors/processorspec.py
rileyjmurray: While we're here, can you look into if this TODO is still valid? We should either make the change or remove the TODO.

rileyjmurray (changes requested):
Changes to explicitopmodel.py and implicitopmodel.py are duplicated almost verbatim. Is there a way to recycle this logic by putting it somewhere else? If not a base class implementation, there could be private free functions in the file holding the base class.

I left a few individual comments too. LMK when you'd like another review.

pygsti/tools/symplectic.py
enielse: Done.

test/unit/objects/test_processorspec.py
enielse: Yes, fine to remove.  I removed it.

pygsti/processors/processorspec.py
enielse: Done.

pygsti/processors/processorspec.py
enielse: Done.

pygsti/processors/processorspec.py
enielse: Yeah, that's an old outdated comment.  I removed it.

enielse:
I think I addressed all the items.  I refactored the common implicit and explicit model code into a single function in the common `OpModel` base class (`OpModel.compute_clifford_symplectic_reps()`).  When you have a chance could you give this PR another review?

pygsti/processors/processorspec.py
```diff
@@ -1195,6 +1327,71 @@ def compute_clifford_symplectic_reps(self, gatename_filter=None):
...
+        and then rebase those templates onto real processor labels.  Most label
+        arguments are physical parameters and should not be remapped, but some
+        argumented gates use arguments to identify state-space labels.  The
+        `gate_arg_label_indices` constructor argument identifies those positions.
+        """
+        gate_label = _Lbl(gate_label)
+        if isinstance(mapper, dict):
+            map_label = mapper.__getitem__
+        else:
+            map_label = mapper
```
rileyjmurray, line 1380: I'd never seen this way of normalizing a dict input into a callable input! I love it.

rileyjmurray (approved):
Looks great. TYSM!

PR #839, opened 2026-07-14
Qiskit interop: warning hygiene, 3.0-proofing, honest dependency floors

rileyjmurray:
This makes pyGSTi's qiskit interop clean under `error::pygsti.tools.exceptions.QiskitInteropWarning`, which this PR promotes from a defined-but-inert exception class to a `pytest.ini` error. Getting there required fixing several real bugs that were hiding behind warnings nobody could see (or that were misclassified so they wouldn't fire at all), removing qiskit APIs slated for removal in 3.0, and correcting the `ibmq` extra's dependency floors, which were stale enough to break on qiskit versions we claim to support.

Five commits, one per phase, each leaving the suite green under the strict filter: `7e39339dc` → `73a4f54bc` → `f4178a8c3` → `962c67a1e` → `1bb2ae93f`.

### What changed

**Warning hygiene (`pygsti/tools/_qiskit_interop.py`, new)**
- Eleven copy-pasted exact-version checks against `qiskit==2.1.1`, scattered across `circuit.py`, `internalgates.py`, `subcircuit_selection.py`, `mirror_edesign.py`, and `scarab.py`, are replaced by one helper, `check_qiskit_version(context, required=True)`, checked against a tested *range* (`TESTED_QISKIT_RANGE = ((1, 4), (3,))`) instead of a single pin.
- That consolidation incidentally fixed real bugs the old pattern was causing: a warn-inside-`try`/bare-`except` that turned the promoted warning into a false "qiskit does not appear to be installed" `RuntimeError`; a `MissingDependencyWarning` misclassification in `subcircuit_selection.py` where a genuine version mismatch was reported as a missing dependency and execution silently continued; and a `NameError` fall-through in `internalgates.py` when qiskit is truly absent.
- `from_qiskit` now warns about discarded classical data only when the circuit actually has clbits — previously it fired on `len(circuit.cregs)`, so a pure pygsti→qiskit→pygsti round trip warned about discarding the empty `cr` register that `convert_to_qiskit` always attaches, even when nothing was measured.
- The plain-`UserWarning` `'skipping measure'` is now a `QiskitInteropWarning` with an actual message.
- `from_qiskit` gains `lossy={'warn', 'ignore', 'raise'}`; the mirror-edesign pipeline passes `lossy='ignore'` at call sites where the loss is intentional.
- `mirror_edesign`'s fullstack path no longer forwards `coupling_map` to `transpile()` alongside a backend — the warning claimed it was ignored, but an explicit `coupling_map` actually overrides the backend's map, so the old behavior silently used the wrong topology.
- New `test/unit/tools/test_qiskit_interop.py` exercises the warning paths directly.

**Qiskit 3.0-proofing**
- Removed the last uses of APIs slated for removal in qiskit 3.0: iterating a `CircuitInstruction` as a tuple (→ `.operation`/`.qubits`), `qubit._index` and `layout.initial_layout[idx]._register.name == 'ancilla'` (→ `circuit.find_bit(qubit).index` and `isinstance(..., AncillaQubit)`), and `circuit._qbit_argument_conversion(i)[0]` (→ the public, stable-since-1.0 `circuit.qubits[i]`).
- Test files: the deprecated `QFT` class (built on the also-removed `BlueprintCircuit`) is replaced with `QFTGate`.
- Considered and rejected: a global pytest filter promoting qiskit `DeprecationWarning`s to errors. Warning attribution follows `stacklevel`, so such a filter would promote *all* pygsti-attributed `DeprecationWarning`s, not just qiskit-related ones — too much blast radius on the full suite. Documented in the phase-2 commit message.

**`extras/ibmq` guard split + honest floors**
- `ibmqexperiment.py` imported `ConvertToMidCircuitMeasure` (added in `qiskit-ibm-runtime` 0.44) in the same `try`/`except` as the core runtime names. On any older runtime the whole import failed and `_Sampler` was set to `None`, disabling *all* IBMQ submission — even though nothing else needed the new pass. It's now guarded separately, with the assertion only in the code path that actually converts mid-circuit measurements (message names the real fix: upgrade runtime, or pass `use_ibm_mcm=False`).
- `pyproject.toml`'s `ibmq` extra goes from the stale, unbounded `qiskit>1, qiskit-ibm-runtime>=0.17.1` to `qiskit>=1.4,<3, qiskit-ibm-runtime>=0.40`, matching `TESTED_QISKIT_RANGE`.
- Validated: all 11 ibmq tests now pass on both qiskit 2.5.0/runtime 0.47 and qiskit 2.1.1/runtime 0.41.1 — the latter previously failed solely because of the coupled import.

**Coverage**
- `test/unit/protocols/test_scarab.py` (new): smoke tests for all three scarab benchmark wrappers (`lowlevel`/`fullstack`/`subcircuit_mirror_benchmark`), which are public API in `pygsti.protocols.scarab` but previously had zero coverage — these would have caught the phase-1 bare-except bug.
- `test_circuit.py`: a `convert_to_qiskit` argument matrix (`qubits_to_measure` in `None`/`'all'`/`'active'`/list form, oversized `num_qubits`, `ordered_data_indices` metadata, the invalid-option error path).
- `test_subcirc_selection.py`: `greedy_growth_subcirc_selection` coverage (direct call and via `sample_subcircuits(strategy='greedy')`), plus a `stochastic_2q_drops` test — see below. `subcircuit_selection.py` coverage goes from 42% to 80%.
- `extras/devices` gets no new tests, per plan: its calibration formats predate the current runtime API, and deprecation is proposed to maintainers instead of investing in coverage.

**CI**
- New `.github/workflows/qiskit-interop.yml`, weekly: a bare qiskit-1.4-floor matrix leg (runtime uninstalled), a runtime-0.40-floor leg, and a latest leg. Keeps `TESTED_QISKIT_RANGE` honest and should flag qiskit 3.0 breakage the week it ships rather than at user-report time.

### Reviewer attention

1. **New bug found, not fixed: `stochastic_2q_drops=True` is broken.** Writing the `greedy`/`stochastic` coverage in phase 4 (previously 0% covered) surfaced that `simple_weighted_subcirc_selection(..., stochastic_2q_drops=True)` re-adds dangling two-qubit gates whose support, by definition, includes at least one qubit outside the selected subset — and the subcircuit is then built with `line_labels=sorted(qubit_subset)`, so `Circuit.__init__` raises `ValueError` on essentially any connected circuit. The new test documents the intended behavior and is marked `@unittest.expectedFailure`. Fixing it needs a maintainer call: should retained dangling gates expand the subcircuit's line labels (changing its effective width), or should they be truncated to their in-subset qubits? Whoever picks a direction should also remove the `expectedFailure` marker.
2. **`ibmq` extra's dependency floor moved** from `qiskit>1` to `qiskit>=1.4,<3` and from `qiskit-ibm-runtime>=0.17.1` to `>=0.40`. This is a correction, not a new restriction — anything outside the old, unbounded range was already broken (that's how finding 5 was found), but it's technically a tightening of the declared floor that downstream pins might notice.
3. **No global `DeprecationWarning` promotion.** See the qiskit-3.0-proofing section above — flagging so it isn't independently proposed and then hit the same blast-radius problem.

### Testing

Cross-version validation, all with the strict `error::QiskitInteropWarning` filter, 141 passed + 2 xfailed on each:
- qiskit 2.5.0 + qiskit-ibm-runtime 0.47 (py313 conda env)
- qiskit 2.1.1 + qiskit-ibm-runtime 0.41.1
- qiskit 1.4.6 + qiskit-ibm-runtime 0.37.0

### DIFF
--- a/.github/workflows/qiskit-interop.yml
+++ b/.github/workflows/qiskit-interop.yml
@@ -1,0 +1,64 @@
+# Weekly qiskit interop version-matrix job.
+#
+# The qiskit-facing test files run against the floor and the latest of the tested
+# qiskit range (TESTED_QISKIT_RANGE in pygsti/tools/_qiskit_interop.py, mirrored by
+# the 'ibmq' extra in pyproject.toml), so the range stays continuously honest: when
+# a new qiskit release breaks the interop -- most notably qiskit 3.0 -- this flags
+# it when it ships instead of at user-report time.
+
+name: Qiskit interop version matrix
+
+on:
+  schedule:
+    - cron: '17 6 * * 1' # weekly, Mondays 06:17 UTC
+  workflow_dispatch: # Allow manual running from GitHub
+
+jobs:
+  qiskit-interop:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Floor of the tested conversion range, without qiskit-ibm-runtime
+          # (no runtime release is compatible with both qiskit 1.4 and our
+          # declared runtime floor): the circuit-conversion/mirroring interop
+          # supports bare qiskit 1.4; runtime-dependent tests skip themselves.
+          - label: qiskit-floor-no-runtime
+            pins: 'qiskit==1.4.*'
+            uninstall-runtime: true
+          # Floor of the declared ibmq extra (runtime 0.44 adds the optional
+          # ConvertToMidCircuitMeasure pass; 0.40 is the supported floor without it).
+          - label: runtime-floor
+            pins: 'qiskit-ibm-runtime==0.40.* qiskit<3'
+          # Whatever pip resolves today.
+          - label: latest
+            pins: 'qiskit qiskit-ibm-runtime'
+    name: Qiskit interop (${{ matrix.label }})
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.13'
+    - name: Install pyGSTi and test dependencies
+      run: |
+        python -m pip install --upgrade pip
+        # testing_no_cython_mpi already pulls in the ibmq extra; the matrix pins
+        # then override the resolved qiskit/runtime versions.
+        python -m pip install -e .[testing_no_cython_mpi]
+        if [ "${{ matrix.uninstall-runtime }}" = "true" ]; then
+          python -m pip uninstall -y qiskit-ibm-runtime
+        fi
+        python -m pip install ${{ matrix.pins }}
+        python -c "import qiskit; print('qiskit', qiskit.__version__)"
+        python -c "import qiskit_ibm_runtime; print('runtime', qiskit_ibm_runtime.__version__)" || echo "qiskit-ibm-runtime not installed"
+    - name: Run qiskit interop tests
+      run: |
+        python -m pytest -v \
+          test/unit/tools/test_qiskit_interop.py \
+          test/unit/circuits/test_circuit.py \
+          test/unit/objects/test_subcirc_selection.py \
+          test/unit/protocols/test_mirror_edesign.py \
+          test/unit/protocols/test_scarab.py \
+          test/unit/extras/ibmq/test_ibmqexperiment.py
--- a/test/unit/objects/test_subcirc_selection.py
+++ b/test/unit/objects/test_subcirc_selection.py
@@ -6,6 +6,7 @@ from pygsti.tools.exceptions import MissingDependencyWarning
 
 import numpy as _np
 import networkx as _nx
+import unittest
 import warnings
 
 from collections import defaultdict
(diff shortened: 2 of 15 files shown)


.github/workflows/qiskit-interop.yml
```diff
@@ -0,0 +1,64 @@
...
+        python -m pip install -e .[testing_no_cython_mpi]
+        if [ "${{ matrix.uninstall-runtime }}" = "true" ]; then
+          python -m pip uninstall -y qiskit-ibm-runtime
+        fi
+        python -m pip install ${{ matrix.pins }}
+        python -c "import qiskit; print('qiskit', qiskit.__version__)"
+        python -c "import qiskit_ibm_runtime; print('runtime', qiskit_ibm_runtime.__version__)" || echo "qiskit-ibm-runtime not installed"
+    - name: Run qiskit interop tests
+      run: |
+        python -m pytest -v \
```
nkoskelo, line 58: Are we planning on excluding these tests from `reuseable-main.yml's` unit tests?

If this is intended to be a small test case for just interoperability with qiskit, we should perhaps name the tests with a common subsequence like `qiskit_interop` and then use `-k qiskit_interop` to reduce the tests to those which match that sequence in their name.

sserita: A good point, the common subsequence would be easier to maintain than remembering where all the tests are.

I personally don't see the problem with having it run in the main unit tests too, i.e. no need to ignore it there. I suspect the biggest use of this cron will be to catch qiskit problems as they ship, regardless of what is passing/failing in unit tests on feature branches/develop, but having it tested to make sure commits don't break qiskit interop is also valuable.

test/unit/objects/test_subcirc_selection.py
```diff
@@ -452,3 +454,139 @@ def get(self, *args):
...
+                self.assertEqual(aux['width'], 3)
+                self.assertEqual(aux['depth'], 4)
+
+
+    # The stochastic-2q-drops path is currently broken: it re-adds dangling gates
+    # whose support extends outside the selected qubit subset, then constructs the
+    # subcircuit with line_labels=sorted(qubit_subset), which raises
+    # "ValueError: line labels must contain at least {...}" whenever any dangling
+    # gate is retained. This test describes the intended behavior and should start
+    # passing once the line-label handling for retained dangling gates is fixed.
```
sserita, line 552: This one is probably for either me or Noah S. I'll put it on my plate for now, presuming that Noah has better things to do this summer prepping for the future :)

sserita (approved):
+1 on Nick's suggestion for the subseq for easier selection of tests, but not blocking. Other than that, looks great, merge as you will. I'll make an issue for the `2q_drops` problem, which will be on my plate most likely.

.github/workflows/qiskit-interop.yml
rileyjmurray: The point of this workflow is to test with a wider and explicit range of qiskit versions. The same tests run on PRs and the beta/master suite, just with whatever version of qiskit we happen to pick up from pypi.

That in mind, I'd like to merge as-is.

Issue #843, opened 2026-07-15
Expand Cirq-to-pyGSTi conversion to handle MCMs/parameterized gates


jordanh6:
Problem:
pygsti.circuits.circuit.Circuit.from_cirq allows you to convert a Cirq circuit into a pyGSTi circuit, but it currently isn't able to handle some important circuit components, namely:
-Parameterized gates (in particular, Z(theta) gates, but it might make sense to also at Y(theta) and X(theta)
-Mid-circuit measurements 

Reason to care about this: Recent Google experiments have released their circuits in Cirq form, and it'd be nice to be able to easily convert these circuits to pyGSTi (and I currently have a strong desire to do this).

Proposed solution:
Add logic to handle Z-basis MCMs and Z(theta) gates. I believe this requires the following based on a quick glance at the code:
pygsti/circuits/circuit.py: edit from_cirq to have logic to handle instruments
pygsti/tools/internalgates.py: Add new gates into pyGSTi/Cirq conversions

rileyjmurray:
@jordanh6 this is the proposal drafted by Fable in response. I've stripped out content about its planned course of action but kept the material about the planned outcomes.

Please review and let me know if you'd like material changes. If not, I'll kick things off.

--------------------------

The goal is for `pygsti.circuits.Circuit.from_cirq` to correctly convert Cirq circuits containing (a) parameterized rotation gates — Z(θ) primarily, plus X(θ)/Y(θ) — and (b) Z-basis mid-circuit measurements. The motivating use case is converting circuits from recently released Google experiments, which are dominated by `cirq.PhasedXZGate` (arbitrary angles), `cirq.ZPowGate` (virtual Z, arbitrary exponents), 2-qubit entanglers, and `cirq.MeasurementGate`.

## Findings from code exploration (current state)

### Conversion machinery

- `Circuit.from_cirq` (`pygsti/circuits/circuit.py:4102`) converts moment-by-moment. Per operation it does an exact dictionary lookup `cirq_to_gate_name_mapping[op.gate]` (works because Cirq gates implement value equality/hashing), and on `KeyError` falls back to `unitary_to_standard_gatename(op.gate._unitary_(), up_to_phase=True)` with a warning. The op-to-name resolution code is duplicated verbatim in the single-op branch (`circuit.py:4201`) and the multi-op branch (`circuit.py:4228`).
[...]

PR #846, opened 2026-07-15
Expand Cirq-to-pyGSTi conversion for MCMs and parameterized gates

rileyjmurray:
This PR targets #843.

`Circuit.from_cirq`/`convert_to_cirq` previously only handled a fixed dictionary of exact-match gates, and had no support at all for `cirq.MeasurementGate`. That's a problem for the motivating use case: recently released Google experiment circuits are dominated by `cirq.PhasedXZGate` (arbitrary angles), `cirq.ZPowGate` (virtual Z, arbitrary exponent), and `cirq.MeasurementGate` (mid-circuit measurement) — none of which converted before this PR (`ZPowGate(exponent=0.37)` died with an `AssertionError`; `MeasurementGate` died with a `TypeError`, since it has no unitary).

This PR adds: 
 1. two new standard parameterized gates, `Gxr`/`Gyr` (closing a pre-existing gap where the Qiskit conversion maps already referenced them but `standard_gatename_unitaries()` didn't define them);
 2. an isinstance-based dispatch table so `from_cirq` converts any `ZPowGate` / `XPowGate` / `YPowGate` / `CZPowGate` / `PhasedXZGate` (including the `Rx`/`Ry`/`Rz` subclasses) at an arbitrary angle, not just the pre-existing fixed set;
 3. mid-circuit-measurement support via pyGSTi's existing per-qubit `'Iz'` instrument-label convention, with terminal measurements dropped by default since they duplicate pyGSTi's implicit end-of-circuit POVM;
 4. the `convert_to_cirq` side of all of the above, so everything round-trips;
 5. warning hygiene for `from_cirq` that follows the convention #839 just established for `from_qiskit` — a typed `CirqInteropWarning`, a `lossy={'warn', 'ignore', 'raise'}` kwarg, and promotion to an error in `pytest.ini`.

### What changed

**New standard gates + conversion tables (`pygsti/tools/internalgates.py`)**
- `Gxr(θ) := expm(-iθX/2)`, `Gyr(θ) := expm(-iθY/2)` (matching the Qiskit `RXGate`/`RYGate` convention), registered in `standard_gatename_unitaries()` next to the existing `Gzr`/`Gczr`.
- New `cirq_parameterized_gatenames_standard_conversions()`: an ordered list of `(cirq gate class, pygsti name, gate -> args)` tuples — `ZPowGate`→`Gzr`, `XPowGate`→`Gxr`, `YPowGate`→`Gyr`, `CZPowGate`→`Gczr`, `PhasedXZGate`→`Gu3` — checked via `isinstance` only after the existing exact-match dictionary lookup fails, so all pre-existing exact conversions (Cliffords, etc.) are unaffected. Every angle-conversion formula was verified numerically against `cirq.unitary(...)` (not just derived symbolically) before being committed; `Gzr`/`Gczr` are exact matches (`global_shift=0` pins the identity-basis-state phase), `Gxr`/`Gyr`/`Gu3` match up to global phase.
- `standard_gatenames_cirq_conversions()` gains callable entries (`angle(s) -> cirq gate`) for `Gzr`/`Gxr`/`Gyr`/`Gczr`/`Gu3`, so `convert_to_cirq` can render them back. Fixing the reversal in `cirq_gatenames_standard_conversions()` to skip these required filtering on `isinstance(value, cirq.Gate)` rather than the more obvious `callable(value)` — `cirq.Gate` instances are themselves callable (that's how `.on(qubits)`-less application like `cirq.X(q0)` works), so the naive filter silently collapsed the ~41-entry exact-match table down to 9 entries. Caught by a regression test asserting a minimum table size, not by inspection.

**`from_cirq` (`pygsti/circuits/circuit.py`)**
- Refactored the duplicated single-op/multi-op label-resolution logic into one `op_to_labels()` helper, followed by a single unified layer-assembly block (global-idle/implied-idle handling no longer branches on raw operation count). Preserves the existing 4-variant `global_idle_replacement_label` behavior exactly (verified against the pre-existing `test_from_cirq`), including a subtle `not all_idles` guard in the implied-idle-stripping condition that would otherwise strip a legitimate all-idle, replacement-disabled layer down to nothing.
- Per-operation dispatch order: gate-less operations (e.g. `ClassicallyControlledOperation`) raise `NotImplementedError`; `MeasurementGate` goes through the new MCM path; exact dictionary lookup (unchanged); symbolic (sympy) parameters raise a `ValueError` *before* the parameterized-family lookup is attempted (so a symbolic exponent gets a clear message instead of a `float(sympy.Symbol)` `TypeError`); the parameterized-family lookup; the existing unitary-matching fallback, now using the more robust `cirq.unitary(op.gate)` instead of `op.gate._unitary_()`.
- New kwargs: `cirq_parameterized_gate_conversion` (override/extend hook, mirrors the existing `cirq_gate_conversion`); `drop_terminal_measurements=True`; `mcm_label='Iz'`; `lossy='warn'` (see Warning hygiene below).
- MCM handling: one `Label(mcm_label, (qubit,))` per measured qubit (a multi-qubit `cirq.measure` splits within the same layer); a nontrivial `invert_mask` raises `NotImplementedError` unconditionally, even for a measurement that will end up dropped, so a semantically-important invert mask is never silently discarded. Terminality is a **per-operation** lookahead (`not any(later moment touches any of this op's qubits)`), so staggered per-qubit final measurements are each handled independently — verified with a test where a 2-qubit measure op has one qubit reused later and the whole op is correctly kept.

**Warning hygiene (`pygsti/tools/exceptions.py`, `pygsti/circuits/circuit.py`, `pytest.ini`)**
- New `CirqInteropWarning(UserWarning)`, a direct parallel to `QiskitInteropWarning`, and promoted to an error in `pytest.ini` right alongside it.
- `from_cirq` gains `lossy={'warn', 'ignore', 'raise'}`, mirroring `from_qiskit`'s kwarg and `_lossy_notice` closure exactly. The two genuinely lossy notices (a dropped terminal measurement, a discarded measurement key) route through it; the unitary-fallback notice stays unconditional (nothing is actually lost when the fallback succeeds), matching how #839 keeps its version-check warnings outside the `lossy` mechanism.
- Fixed the measurement-key warning firing even when the only measurement in a circuit is terminal (and therefore dropped) — the most common case, a circuit ending in a final measurement, previously got two warnings, one of them moot. It now fires only when a measurement is actually kept as an `'Iz'` label.
- `drop_terminal_measurements=True` + `lossy='raise'` means any circuit containing a terminal measurement raises — that's the intended semantics of opting into `'raise'` ("tell me loudly about any loss"), documented as such in the docstring.

**`convert_to_cirq` (`pygsti/circuits/circuit.py`)**
- Distinguishes callable `gatename_conversion` entries from fixed `cirq.Gate` objects (same `isinstance(x, cirq.Gate)` pattern, same reasoning as the `from_cirq` reversal fix) and invokes them with the label's args.
- New `mcm_label` kwarg (default `'Iz'`, matching `from_cirq`): converts to `cirq.measure` with an auto-generated key unique to qubit and layer.

**Validation (`test/unit/circuits/test_circuit.py`, `test/unit/tools/test_internalgates.py`)**
- Unit tests for every dispatch branch: arbitrary-angle round trip for all five gate families, exact-match Cliffords unaffected, `Rx`/`Ry`/`Rz` subclass dispatch, symbolic-gate and classically-controlled-op errors, mid-circuit vs. terminal measurement (default drop + warning, `drop_terminal_measurements=False`), multi-qubit measure split, custom `mcm_label`, staggered terminal measurement, `invert_mask` error, the `lossy` kwarg (`'ignore'` silent, `'raise'` raises, invalid value rejected), the measurement-key warning firing only when a measurement is kept, and both round-trip directions (pyGSTi→Cirq→pyGSTi label equality; Cirq→pyGSTi→Cirq via `cirq.allclose_up_to_global_phase`).
- One genuinely new end-to-end regression test, `test_convert_to_cirq_matches_simulator_for_crosstalk_free_model`: random circuits from the full parameterized-gate family, run through a crosstalk-free `Model.probabilities()` and independently through `cirq.Simulator`, fixed seed for determinism. Nothing exercised the full `Circuit → Model → probabilities()` pipeline against an independent oracle before.
- A broader (non-deterministic, larger-sweep, statistical) validation script was used during development but isn't part of this PR, since it demonstrates rather than regresses: a larger unitary-only comparison, a mid-circuit-measurement joint-probability comparison against `cirq.Simulator` sampling (using a 1-qubit explicit model with a per-qubit `Iz` `TPInstrument`, the existing `test_edesigntools.py` pattern), and a Google-experiment-style smoke test (`PhasedXZ` + arbitrary-angle `CZPowGate` + MCM + terminal measurement).

**Docs (`docs/markdown/interfaces/CirqIntegration.md`)**
- `from_cirq`'s docstring now lists the supported parameterized-gate table and MCM/error semantics, including the `lossy` kwarg.
- Three new demonstrated conversions in the Cirq-interop tutorial: arbitrary-angle parameterized rotations, mid-circuit measurements (including the default terminal-measurement-dropping behavior and the round trip back through `convert_to_cirq`), and a note that the lossy notices are `CirqInteropWarning`s controllable via `lossy`. Verified via `jupytext --sync` + `pytest --nbval-lax` that the regenerated notebook — including the pre-existing cells — still executes cleanly end-to-end.

### Notes

1. **`create_explicit_model` requires `embed_gates=True` even for a single qubit** once the pspec contains a factory-backed gate like `Gxr`/`Gyr`/`Gzr` — the default `embed_gates=False` path asserts against factories unconditionally, regardless of qubit count. This is pre-existing behavior (not changed by this PR), but it's the reason the MCM validation used a non-default kwarg; flagging in case it's worth a friendlier error message or a doc note elsewhere.
6. **Not addressed (explicit non-goals):** embedded/per-qubit instrument support for multi-qubit implicit/explicit models (pre-existing "FUTURE" debt — crosstalk-free models already handle it fine, so this only limits explicit-model multi-qubit MCM circuits); arbitrary-exponent `ISwapPowGate`/`FSimGate` (Google's native sqrt-iSWAP-family gates have no standard pyGSTi 2-qubit equivalent — would mean inventing a new standard gate, flagging as a possible follow-up); classical feedforward (`ClassicallyControlledOperation`) is rejected with a clear error, not supported; measurement-key preservation isn't supported (would need an API change to `from_cirq`'s return type, which currently returns a bare `Circuit`).
7. **Terminal-measurement dropping defaults to on** (`drop_terminal_measurements=True`). This changes behavior for any *hypothetical* pre-existing caller who fed `from_cirq` a circuit with `cirq.measure` calls — previously that would have raised a `TypeError` unconditionally, so there's no real behavior change for anyone whose code worked before, but flagging the default since it's an opinionated choice (motivated by avoiding double-counting against pyGSTi's implicit end-of-circuit POVM) rather than a neutral one. Combined with `lossy='raise'`, this means any circuit with a terminal measurement raises by design (see Warning hygiene above).

### Testing

`pytest test/unit/circuits/ test/unit/tools/ test/unit/construction/ test/unit/objects/test_processorspec.py` (py313 conda env, cirq 1.7.0) — 432 passed, 22 skipped, 1 xfailed. The skips are guards for optional dependencies not installed in this test environment, plus pre-existing skips unrelated to this PR.

`pytest --nbval-lax` on the regenerated `CirqIntegration.ipynb` (32 cells) — all pass.

A workspace-local validation script (not part of this PR) cross-checked the conversion against independent Cirq execution: a unitary-only random-circuit sweep against `cirq.Simulator` (max error ~2e-7, floating-point noise only), an MCM joint-probability comparison against 200k-shot Cirq sampling (agreement well within statistical tolerance), and a Google-experiment-style smoke test combining `PhasedXZ`, an arbitrary-angle `CZPowGate`, and a mixed mid-circuit/terminal measurement.


rileyjmurray:
@jordanh6 please pull down this branch locally and give it a spin. I'd appreciate a review that speaks to the three items flagged for "reviewer attention," in addition to whatever else you'd like to comment on.

jordanh6:
This takes care of the issues I stated in my request...but upon testing it out, I ran into something I missed. Cirq also has a reset gate cirq.ResetChannel() that's used fairy heavily. Admittedly, I've never used a reset in pyGSTi, so I don't know how we should be handling this---and if we have a typical strategy for handling mid-circuit hard reset, I'm happy to defer to that. In the absence of that, here is an approach that would be useful in the context of our DEM/QEC work: For a reset of qubit i, add to the pyGSTi circuit a new fresh qubit i'. and do all later operations on qubit i in the cirq circuit on qubit i' in the pyGSTi circuit. If we do this, the output should also contain a dictionary with the mapping between the labels of the added qubits and the labels of the qubits they originated from.

I'm including an example circuit here for reference (originates from Google's cultivation work https://arxiv.org/abs/2512.13908)[
[example_cultivation_c.json](https://github.com/user-attachments/files/30102593/example_cultivation_c.json)
](url)

rileyjmurray:
Thanks for the feedback, Jordan. I'll separate the two parts of what you raised. First, there's the question of modeling a mid-circuit hard-reset in pygsti. Second, there's the following operation.

> For a reset of qubit i, add to the pyGSTi circuit a new fresh qubit i'. and do all later operations on qubit i in the cirq circuit on qubit i' in the pyGSTi circuit.

It sounds like you only need that second operation if you don't have access to the hard reset abstraction. Is that the case? I ask because the second operation isn't going to scale well with the number of qubits. We don't have a notion of adding a qubit mid-circuit, so what we'd have to do is have the candidate resett**er**-qubit sitting in waiting, subject to a perfect idle until such time as a measurement outcome needs to swap it out with a resett**ee**-qubit. Every reset operation would consume a distinct resett*er* qubit.

I'd like to get thoughts from @pcwysoc and @sserita on this situation.

Piper and Stefan --- would you agree that the hard-reset abstraction isn't compatible with what pyGSTi circuits currently represent? If so, do you think that extending our circuits to support hard resets is less of an "ask" than support for general feed-forward operations, like those in LoQS?

jordanh6:
For my particular use case, I need the second functionality (adding qubits) even if we implement a reset operation in some way. But IMO a solution of the first form is fine---in that case, I'd just write my own code to process circuits with resets into a form where I can use them.

Elaboration: I am using this in the context of generating DEMs from error generator models. QEC circuits typically contain MCMs+reset, and the error generator propagation workflow we use requires a circuit with only Cliffords (+initial state preparation and terminating measurement). It cannot easily handle a hard reset operation, even if it were to exist in pyGSTi. The current code we have to generate DEMs (e.g., in feature-dems-from-errorgens) uses this exact trick when processing stim circuit inputs. This sort of simulation/analysis is all done with our error generator propagation pipeline, so very large numbers of qubits aren't an issue. 

One *could* remove the resets at the Cirq level using this trick. So I guess reset handling isn't a strictly necessary addition, but IMO it could be useful. If y'all decide it'd be too much of a mess, I think this is workable for now as-is!

sserita:
We can do a hard reset with an instrument, no? The effect in the outer product always maps to 0, i.e. regardless of the input state, the output state is always mapped to 0?
Unless I misunderstand the question

jordanh6:
Yeah, that should work just fine.

rileyjmurray:
>  The effect in the outer product always maps to 0, i.e. regardless of the input state, the output state is always mapped to 0?

That would require that the conditional maps in our instruments are affine, not just linear.

sserita:
@rileyjmurray and I have talked it over, our conclusion is that while the current Instrument framework may not allow us to do what I initially suggested, i.e. have the maps be something like |0><0| and |0><1|, that is probably the best route to getting a reset-like operation into pyGSTi rather than any circuit manipulation.
As such, we are punting that from this PR and I'll make a related Instrument issue shortly to make that happen, if that sounds good to everyone?

sserita:
[07-16_InstrumentReset.ipynb](https://github.com/user-attachments/files/30106983/07-16_InstrumentReset.ipynb)

@jordanh6 you man have already figured it out, but here's a test notebook showing a reset with an Instrument. To both Riley and my surprise, it just kinda works. Won't open another issue after all :P

sserita (approved):
Straightforward, looks good to me!

PR #848, opened 2026-07-15
Feature graphcoloring

nkoskelo:
This PR should be merged before https://github.com/sandialabs/pyGSTi/pull/735 as we are splitting that PR to make it simplier to review.

## Overview
This PR adds a robust, high-performance, and mathematically rigorous **graph edge-coloring toolkit** inside `pygsti/tools/graphcoloring`. This toolkit is designed to support scheduling and routing operations on quantum processoring unit's coupling maps/interaction graphs by partitioning the edges into colored sets so that no two adjacent edges share a color class.

All algorithms are wrapped behind a single, unified public API switchboard: `switchboard_find_edge_coloring`. If you would like to check whether a returned coloring is valid, you can call `check_valid_edge_coloring(color_patches, ret_false_on_error=False)`. Note that it verifies **properness** (no two edges sharing a vertex are assigned the same color, and no edge is duplicated across color classes) -- it does *not* independently check **completeness** (that every input edge received a color); all four algorithms are documented/tested to always return a complete coloring, so this is primarily a sanity check against accidental omissions or bugs.

---

### 1. `auto` (Recommended Default)
- **Approach**: First structurally inspects the graph to see if it matches a canonical processor geometry ("line", "ring", "grid", or "torus"). If matched, it uses an optimal closed-form edge-coloring requiring the exact minimum chromatic index (at most 2 colors for lines, 2–3 for rings, and 4 for grids/even-tori) in $O(|V|)$ time.
- **Bipartite Fallback**: If the graph is not a canonical topology but is bipartite (detected via `networkx.is_bipartite` in $O(|V| + |E|)$), it runs a randomized bipartite-optimal algorithm (`_BipartiteEdgeColoring`) which frequently colors bipartite graphs using exactly $\Delta$ colors.
- **General Fallback**: For all other arbitrary non-bipartite graphs, it falls back to the deterministic `vizing_edge_coloring`.

### 2. `vizing` & `misra_gries` (Exact constructive Vizing-chain algorithms)
- **Guarantees**: Proper, complete edge coloring using at most $\Delta + 1$ colors (Vizing's theorem) on all simple graphs.
- **Worst-case Complexity**: $O(|V| \cdot |E|)$ time.
- **Implementation**: Robust constructive implementation utilizing Vizing fans and alternating color-recoloring chains.
- **References**:
  - V. G. Vizing, "On an estimate of the chromatic class of a p-graph," *Diskret. Analiz*, vol. 3, pp. 25-30, 1964 (in Russian; no stable DOI/URL -- the original chromatic-index theorem this module's "Vizing" and "Misra-Gries" algorithms both constructively prove).
  - J. Misra and D. Gries, "A constructive proof of Vizing's theorem," *Information Processing Letters*, vol. 41, no. 3, pp. 131-133, 1992. https://doi.org/10.1016/0020-0190(92)90041-S

### 3. `sinnamon` (Deterministic Greedy-Euler-Color)
- **Guarantees**: Proper, complete coloring using at most $2\Delta - 1$ colors.
- **Worst-case Complexity**: $O(|E| \log(\Delta))$ time.
- **Reference**: C. Sinnamon, "Fast and Simple Edge-Coloring Algorithms," [arXiv:1907.03201, 2019](https://arxiv.org/abs/1907.03201).

### 4. `random_euler_color` (Randomized Random-Euler-Color)
- **Guarantees**: Proper, complete coloring using at most $\Delta + 1$ colors.
- **Worst-case Complexity**: Runs in expected $O(|E| \sqrt{|V|})$ time with high probability.
- **Reference**: C. Sinnamon, "Fast and Simple Edge-Coloring Algorithms," [arXiv:1907.03201, 2019](https://arxiv.org/abs/1907.03201).

- Branch [graphcoloring-algorithms-saved](https://github.com/sandialabs/pyGSTi/tree/graphcoloring-algorithms-saved) holds implementations of other graph algorithms if one wants to use those in the future.
---

## 🧪 Testing & Verification

The tests are implemented in `test/unit/tools/test_graphcoloring.py`:
- **Correctness & Completeness**: Every algorithm is tested on standard families (complete graphs, cycles, grids, paths, random regular, hubs) to ensure it produces a proper (no adjacent edges share a color) and complete (all edges colored) coloring. If there is a specific family of graphs you would like to ensure are tested you can add them here.
- **Seed-Reproducibility**: Randomized algorithms (and the `"auto"` bipartite fallback) are verified to be fully deterministic when passing a fixed `seed`.
- **Scaling Benchmarks**: Includes a robust scaling test class (`GraphColoringScalingTester`) guarded by hard wall-clock timeouts that prints comparative performance and color-quality tables when run locally (`pytest -m slow`).
-- Added a workflow test to run the graphcoloring scaling tests if there is a change detected to the subpackage: tools/graphcoloring or its test, (test/unit/tools/test_graphcoloring.py). This will run for your PR as long as there is a difference in those files compared to the target branch. Note it should not run for the beta/master/bugfix branches as one should now the scaling of each of the potentially new algorithms before it makes it to that step in the workflow.


pygsti/tools/graphcoloring/__init__.py
```diff
@@ -0,0 +1,31 @@
+"""
+Edge-coloring toolkit: a name-based switchboard (`switchboard_find_edge_coloring`)
+over several edge-coloring algorithms, plus topology-aware closed-form
+colorings (`auto_edge_coloring`) for the canonical topologies produced by
+`ProcessorSpec(geometry=...)`.
+
+This is a package (split by algorithm family for readability); everything
+below is re-exported here so `from pygsti.tools.graphcoloring import X`
+keeps working exactly as it did when this was a single module.
```
rileyjmurray: We don't care about the "keeps working," since this PR is proposing this functionality for the first time. Please re-export only what's needed.

pygsti/tools/graphcoloring/_dispatch.py
```diff
@@ -0,0 +1,65 @@
...
+    Dispatches to different edge coloring algorithms based on the provided name.
+
+    Parameters:
+    algorithm_name (str): The name of the algorithm to use ('misra_gries', 'vizing',
+        'sinnamon', 'random_euler_color', 'auto').
+        'sinnamon' is Sinnamon (2019)'s deterministic Greedy-Euler-Color, a
+        (2*deg-1)-edge-coloring running in O(m log(deg)) time. 'random_euler_color' is
+        Sinnamon (2019)'s randomized Random-Euler-Color, a (deg+1)-edge-coloring running
+        in O(m*sqrt(n)) time with high probability -- see `sinnamon_2d_minus_1_edge_coloring`
+        and `sinnamon_euler_color_edge_coloring`, respectively, for details.
```
sserita: Wouldn't having the options be "greedy_euler_color" and "random_euler_color" make this distinction clear even without digging into docstring? Anything tying us to "sinnamon"? Or is the greedy version so synonymous with "sinnamon" that a user that knows the basics of graph coloring would assume the correct relation?

pygsti/tools/graphcoloring/_sinnamon.py
```diff
@@ -0,0 +1,527 @@
+"""
+Sinnamon (2019)'s Euler-Template edge-coloring algorithms: the deterministic
+"Greedy-Euler-Color" (using ≤ 2*deg-1 colors in O(m log(deg)) worst-case time)
```
sserita, line 3: Complementary to my first comment: If we keep "sinnamon" as the option, worth mention here that greedy is dispatched by the plain "sinnamon", IMO

test/unit/tools/test_graphcoloring.py
```diff
@@ -0,0 +1,886 @@
...
+# automatically. The benchmark-style scaling suite in this module is meant to be
+# run locally, so we skip it in CI via this environment-variable ``skipif`` -- the
+# same pattern used elsewhere in the test suite (e.g. the ``PYGSTI_NO_CUSTOMLM_SIGINT``
+# check in test/unit/optimize/test_sigint.py and the ``needs_*`` helpers in
+# test/unit/util.py). This keeps the slow tests local-only without any change to
+# the CI workflow. To run them locally, simply invoke pytest outside of CI (or
+# unset ``CI``); to also exclude them locally, use ``-m "not slow"``.
+skip_in_ci = pytest.mark.skipif(
+    'CI' in os.environ,
+    reason="benchmark-style scaling test; run locally only (skipped in CI)")
```
sserita: Ideally maybe we would implement something in the Action that detects if there are any changes in pygsti.tools.graphcoloring and sets a variable to turn these tests ON (and default OFF).

If you have any ideas on how to do that, feel free to do so - if not, maybe leave a TODO about doing that in this comment so that we remember to come back to it.

test/unit/tools/test_graphcoloring.py
```diff
@@ -0,0 +1,886 @@
...
+
+# Deterministic algorithms that always terminate and produce a proper, complete
+# coloring with at most deg+1 colors (Vizing's theorem) on every family tested.
+DETERMINISTIC_EXACT_ALGORITHMS = ["vizing", "misra_gries"]
+
+# Randomized algorithms that terminate and produce a proper, complete coloring.
+# ('sinnamon' -- Sinnamon (2019)'s deterministic Greedy-Euler-Color -- ignores
+# `seed` and is included here only because the reproducibility/validity tests
+# below are written to be seed-agnostic; it is not actually randomized.)
+RANDOMIZED_ALGORITHMS = ["sinnamon", "random_euler_color"]
```
sserita: Maybe then SEED_AGNOSTIC would be a better and more accurate term

pytest.ini
```diff
@@ -19,3 +18,6 @@ filterwarnings =
 python_classes = 
     *Tester
     Test*
+
+markers =
+    slow: marks tests as slow / benchmark-style (deselect with '-m "not slow"')
```
sserita, line 23: Neat trick, I'll keep it in mind for other slow tests that we may want to toggle on and off

sserita (changes requested):
One discussion topic, one test change request, and one "fix or put a TODO". Very clean code!

pygsti/tools/graphcoloring/_dispatch.py
nkoskelo: I will have it dispatched by "deterministic_euler_color" and in the comments state that we are using an implementation of Sinnamon's deterministic algorithm with the reference to the paper.

test/unit/tools/test_graphcoloring.py
nkoskelo: The `reusable-main.yml` will run a check to see if there is a change to the graphcoloring subpackage or its unit tests. If so then it will trigger the scaling tests as well in the CI.

sserita (approved):
Looks good! Nice work

PR #849, opened 2026-07-15
Xfgst circuit lanes

nkoskelo:
Split off from PR #735

This PR focuses on the ability to lane circuits. It is not dependent on #848.

A circuit can be split into lanes, which are sets of non-interacting qubits. Once we have those lanes, we can potentially simulate the circuit more quickly depending on the error model.

The current way to build a circuit with a known lane structure is to use the function `batch_tensor`, which requires two or more circuits (it will raise an error if given only one). Note that the lane structure will not be generated until you either build the circuit with `batch_tensor` or call `compute_subcircuits(circuit, cache_lanes_in_circuit=True)`. This cache is only used for static (read-only) circuits, so it cannot go stale from an in-place edit; converting a circuit back to editable and calling `done_editing()` clears the cache. A future PR may update the cache based upon the action done to the circuit, but that is not available at this time.

When calling `compute_subcircuits`, the caller does not need to specify the lane partition (`qubit_to_lanes`/`lane_to_qubits`); if omitted, it will be automatically detected from the circuit. However, if the partition is already known, passing it in explicitly avoids the detection step and will be faster.

Each circuit's lane cache is independent: copying a circuit deep-copies `saved_auxinfo`, so mutating one circuit's cache cannot corrupt another's.

This PR is purely standalone infrastructure in `pygsti/circuits/` (`split_circuits_into_lanes.py` plus supporting cache hooks in `Circuit`). It is not yet wired into any forward simulator, protocol, or layout code — that integration is left for a future PR.

Tests are in `test/unit/circuits/test_circuit_splitting.py` and cover: basic lane splitting and caching, cache invalidation/miss detection, automatic lane detection, `batch_tensor` with mismatched-depth circuits, custom line ordering (including multi-qubit gates and string labels), the single-circuit error case, and regression cases for non-contiguous/interleaved lanes.

### DIFF
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from cirq.circuits.circuit import Circuit as CirqCircuit
     from pygsti.baseobjs.label import ConcreteLabel
 
+import copy as _copy
 import itertools as _itertools
 import warnings as _warnings
 
@@ -622,13 +623,7 @@ class Circuit(object):
         self._name = name  # can be None
         #self._times = None  # for FUTURE expansion
         self.auxinfo = {}  # for FUTURE expansion / user metadata
-        if saved_aux:
-            self.saved_auxinfo = {
-                k: (v.copy() if isinstance(v, dict) else v)
-                for k, v in saved_aux.items()
-            }
-        else:
-            self.saved_auxinfo = {}
+        self.saved_auxinfo = _copy.deepcopy(saved_aux) if saved_aux else {}
         return self
 
     #pickle management functions
--- a/pygsti/circuits/circuit.py
+++ b/pygsti/circuits/circuit.py
@@ -598,8 +598,9 @@ class Circuit(object):
         self._name = name  # can be None
         #self._times = None  # for FUTURE expansion
         self.auxinfo = {}  # for FUTURE expansion / user metadata
+        # `saved_auxinfo["lanes"]` is a cache used by pygsti.circuits.split_circuits_into_lanes.
+        # It is populated lazily (see Circuit._cache_tensor_lanes)
         self.saved_auxinfo = {}
-        self.saved_auxinfo["lanes"] = {tuple(line_labels): list(self._labels)}
 
     #Note: If editing _copy_init one should also check _bare_init in case changes must be propagated.
     #specialized codepath for copying
@@ -1351,7 +1352,9 @@ class Circuit(object):
 
         for i in layers:
             ret_layer = []
-            layer_lbl = self.layertup[i]
+            # Equivalent to `layer_lbl = self.layertup[i]`
+            raw_layer = self._labels[i]
+            layer_lbl = raw_layer if isinstance(raw_layer, _Label) else _Label(raw_layer)
 
             components = list(layer_lbl.components)
 
@@ -2548,9 +2551,9 @@ class Circuit(object):
         if len(overlap) > 0:
             raise ValueError(
                 "The line labels of `circuit` and this Circuit must be distinct, but overlap = %s!" % str(overlap))
-        if self._line_labels == '*':
+        if self._line_labels == ('*',):
             raise ValueError("The line labels of 'self', are underspecified and are currently the placeholder labels.")
-        if circuit._line_labels == '*':
+        if circuit._line_labels == ('*',):
             raise ValueError("The line labels of 'circuit', are underspecified and are currently the placeholder labels.")
 
         all_line_labels = set(self._line_labels + circuit._line_labels)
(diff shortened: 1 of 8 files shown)


pygsti/circuits/circuit.py
```diff
@@ -620,7 +622,13 @@ def _copy_init(self, labels, line_labels, editable, name='', stringrep=None, occ
         self._name = name  # can be None
         #self._times = None  # for FUTURE expansion
         self.auxinfo = {}  # for FUTURE expansion / user metadata
-
+        if saved_aux:
+            self.saved_auxinfo = {
```
nkoskelo: copy.deepcopy() instead.

pygsti/circuits/circuit.py
```diff
@@ -2560,10 +2570,48 @@ def tensor_circuit_inplace(self, circuit: Circuit, line_order: Union[None, List[
...
+            if isinstance(lbl, _LabelStr):
+                # These lbls have no qubits associated with them because of reasons.
+                # So we will assume that each line in the circuit of each has that gate applied.
+                lbl = _Label(lbl.name, self.line_labels)
+            if isinstance(other_lbl, _LabelStr):
+                other_lbl = _Label(other_lbl.name, circuit.line_labels)
+
+            lbl = lbl.concat(other_lbl)
+            new_layers.append(lbl)
+            ind += 1
```
sserita, line 2589: Why do we do this instead of the inplace edit as before? I can see padding below as needing a separate pass (if we don't have that functionality somewhere already), but not seeing the advantage of above.

sserita: Also, the core of the feature seems to be the _cache_tensor_lanes below, which doesn't use idle gate logic. Why was that inserted here?

pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,265 @@
...
+        out = {}
+        for key, val in lanes_to_qubits.items():
+            for qb in val:
+                out[qb] = key
+        return out
+
+    return compute_qubits_to_lanes(lanes), lanes
+
+
+def compute_subcircuits(circuit: Circuit,
```
sserita: I would rather call this split_subcircuits or something like that. "compute" to me would have implied that this is figuring out what the subcircuits are, but I think that is happening in the cache function and this is just making subcircuits from the main circ.

Also, alternate approach: It is not more efficient, but it would be cleaner to do something like 
```
subcirc = circ.copy(editable=True)
subcirc.delete_lines(<qubits not in my subcirc set>)
```
With the advantage that delete lines acts as an extra safety check that will warn if there are straddlers (i.e. two lanes are actually connected).

pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,265 @@
...
+            sorted_group = sorted(group, key=lambda x: x.qubits[0])
+            lanes_to_gates[lane].append(Label(tuple(sorted_group))) # Go through the label factory.
+
+    if cache_lanes_in_circuit:
+        circuit = circuit._cache_tensor_lanes(lanes_to_gates, lane_to_qubits)
+
+    return lanes_to_gates
+
+
+def batch_tensor(
```
sserita, line 149: How is this different than doing a `Circuit.tensor_circuit` and then padding with noisy idles (I guess there is no built in to do this, but you could iterate through labels with `layer_with_idles` to build up new layers. If that works, I'd actually prefer we create a member function in Circuit that is `insert_implicit_idles_inplace` or something like that which does this.

pygsti/circuits/split_circuits_into_lanes.py
```diff
@@ -0,0 +1,265 @@
...
+
+    if isinstance(lbl, tuple):
+        return tuple(
+            _map_label_state_space_labels(member, sslbl_map)
+            for member in lbl
+        )
+
+    return lbl.map_state_space_labels(sslbl_map)
+
+def compute_qubit_to_lane_and_lane_to_qubits_mappings_for_circuit(circuit: Circuit) -> tuple[dict[int, int],
```
sserita: I'd prefer something shorter, like `compute_qubit_lane_mappings` (or with a `_for_circuit` suffix if you want).

pygsti/circuits/split_circuits_into_lanes.py
sserita: Also another thing: There is currently a concept of a "subcircuit" in `Circuit` which is a group of repeated layers. Maybe make sure this is something to do with `tensored_subcircuits` or product or something that makes it a clear distinction.

test/unit/protocols/test_protocols.py
```diff
@@ -161,7 +161,10 @@ def test_serialization(self, root_path):
             edesign.write(root)
             loaded_edesign = type(edesign).from_dir(root)
             # TODO: We don't have good edesign equality
-            self.assertEqual(set(edesign.all_circuits_needing_data), set(loaded_edesign.all_circuits_needing_data))
+            self.assertEqual(
+                set([c.serialize() for c in edesign.all_circuits_needing_data]),
+                set([c.serialize() for c in loaded_edesign.all_circuits_needing_data])
+            )
```
sserita, line 167: Why do we have to serialize now? An unexpected change in where gates are in layers is non-trivial and worth understanding.

sserita (changes requested):
I think this could be made a lot cleaner if it leveraged existing `Circuit` functionality a lot more. I've indicated some examples where that could be the case, but take a look over everything and see if you can find other places where we can avoid duplication.

test/unit/protocols/test_protocols.py
sserita: (I guess I should mention that `Circuit.serialize` makes it so that each layer only has one component, i.e. all gates are serial in time. This is unrelated to serialization having to do with file I/O.)

nkoskelo: This was done because of the lbl.concat code returning a LabelTupTup in every instance. Now that lbl.concat will return a LabelTup if there is only one component this is unnecessary.

sserita (approved):
Nice work! Feel free to merge when the CI passes.

PR #851, opened 2026-07-20
WIP: synthetic SPAM randomized benchmarking

rileyjmurray:


pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+- A single gate name, `'Gu'`, carries the ZXZ Euler angles `(alpha, beta, gamma)` of
+  one SU(2) group element as its three `args`, on a single qudit line.
+- Each sampled `(m+1, 3)` angle sequence (rows `0..m-1` Haar-random, row `m` the
+  inverting gate) is turned into `2j+1` circuits -- one per prep -- that share the
+  same `Gu` layers but differ in their explicit `rho{ell}` prep label (`ell = 0..2j`,
+  in `SpinJ.spins` order, i.e. `rho0` prepares `|j>` and `rho{2j}` prepares `|-j>`)
+  and a shared `Mdefault` POVM label.
+- `prep_index`, `seq_index`, and `euler_angles` (plus, for the character design,
+  `characters`/`charcores`) are carried as `paired_with_circuit_attrs` JSON aux data
+  rather than being recovered by re-parsing the embedded `rho{ell}`/`Mdefault` labels.
```
rileyjmurray, line 25: Hard to read.

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+
+This module ports the SU(2) synthetic-SPAM RB protocols (SSRB, SSchiRB, SSR1RB) from
+the `su2-rb-conservative` branch into pyGSTi's `Design -> DataSimulator -> Protocol ->
+Results` architecture, generalized from the branch's hardcoded spin-7/2 tooling to
+arbitrary spin `j` via `pygsti.tools.su2tools.SpinJ`.
+
+See `active-project/su2-synthetic-spam-rb-plan.md` for the normative conventions
+(Section 3) and phase-by-phase deliverables (Section 4), and
+`active-project/su2rb-label-spike-notes.md` for the Task 3.0 label/factory ergonomics
+spike that this module's circuit conventions are built on.
```
rileyjmurray, line 12: Strip out references to `active-project`.

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+  inverting gate) is turned into `2j+1` circuits -- one per prep -- that share the
+  same `Gu` layers but differ in their explicit `rho{ell}` prep label (`ell = 0..2j`,
+  in `SpinJ.spins` order, i.e. `rho0` prepares `|j>` and `rho{2j}` prepares `|-j>`)
+  and a shared `Mdefault` POVM label.
+- `prep_index`, `seq_index`, and `euler_angles` (plus, for the character design,
+  `characters`/`charcores`) are carried as `paired_with_circuit_attrs` JSON aux data
+  rather than being recovered by re-parsing the embedded `rho{ell}`/`Mdefault` labels.
+"""
+# ***************************************************************************************************
+# Copyright 2015, 2019, 2025 National Technology & Engineering Solutions of Sandia, LLC (NTESS).
```
rileyjmurray, line 28: Change year to 2026

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+    layers = [_Label(GATE_NAME, qudit_label, args=tuple(float(x) for x in row)) for row in angles]
+
+    if prep_index is not None:
+        _, two_j = _validate_spin(j)
+        dim = two_j + 1
+        if not (0 <= prep_index < dim):
+            raise ValueError("prep_index=%r is out of range for j=%s (dim=%d)" % (prep_index, j, dim))
+        layers = [_Label(_prep_name(prep_index))] + layers + [_Label(POVM_NAME)]
+
+    return _Circuit(layers, line_labels=(qudit_label,))
```
rileyjmurray, line 127: This is awfully long for something that boils down to
```python
angles = _np.asarray(angles, dtype=float)
if angles.ndim != 2 or angles.shape[1] != 3:
    raise ValueError("`angles` must have shape (m+1, 3); got %s" % (angles.shape,))
layers = [_Label(GATE_NAME, qudit_label, args=tuple(float(x) for x in row)) for row in angles]
return _Circuit(layers, line_labels=(qudit_label,))
```
in the default case.

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+
+
+def euler_angles_from_circuit(circuit):
+    """
+    Recover the `(m+1, 3)` array of ZXZ Euler angles from a `Circuit`'s `Gu` layers.
+
+    Any non-`Gu` layers (e.g. explicit `rho{ell}` prep or `Mdefault` POVM layers) are
+    ignored, so this is the inverse of `circuit_from_euler_angles` regardless of
+    whether `prep_index` was given.
```
rileyjmurray, line 136: As written, it's wrong to say "regardless of whether `prep_index` was given". The correct description is simply "if `prep_index` was not given".

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+    """
+    rows = []
+    for layer in circuit:
+        if layer.name == GATE_NAME:
+            args = layer.args
+            if len(args) != 3:
+                raise ValueError("Gate layer %s does not carry exactly 3 args" % (layer,))
+            rows.append([float(a) for a in args])
+    if not rows:
+        raise ValueError("Circuit %s contains no '%s' gate layers" % (circuit, GATE_NAME))
```
rileyjmurray, line 157: Unclear why this needs to error. You can return an ndarray with shape (0, 3).

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+    numpy array
+        Shape `(dim**2, dim**2)`.
+    """
+    dim = spinj.dim
+    composed = _np.eye(dim * dim, dtype=complex)
+    for channel in channels:
+        channel = _np.asarray(channel)
+        if channel.shape == (dim, dim):
+            channel = _unitary_to_std_process_mx(channel)
+        elif channel.shape != (dim * dim, dim * dim):
```
rileyjmurray, line 489: Define and use a `dim2 = dim ** 2` variable for better readability.

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+        rng = _np.random.default_rng(self.seed) if self.shots is not None else None
+        ideal_ops = self._ideal_computational_ops()
+
+        ds = _DataSet(collision_action="aggregate")
+        # DataSet's default repetition-count dtype is float32 (dataset.py's
+        # `Repcount_type`), which is far too coarse for the exact-probability
+        # "counts" recorded when `shots is None`: float32 has ~1e-7 relative
+        # precision, which shows up as an artifact-level discrepancy against
+        # independently-computed probabilities (e.g. from a full pyGSTi Model).
+        # Use float64 so the shots=None counts really are exact.
```
rileyjmurray, line 778: Excessive detail.

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+        raise TypeError("ell must be an int, float, or Fraction, not %s" % type(ell).__name__)
+
+    if abs(two_ell) > two_j or (two_j - two_ell) % 2 != 0:
+        valid = [j_float - i for i in range(two_j + 1)]
+        raise ValueError("ell=%r is not a valid Jz eigenstate index for j=%s (must be one of %s)"
+                          % (ell, j_float, valid))
+    return two_ell / 2.0
+
+
+_ZERO_NOISE_VARIANCE_VARIANTS = ('ssrb', 'chirb', 'r1rb', 'sschirb', 'ssr1rb')
```
rileyjmurray, line 872: Awkward variable name ("variance variants").

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+            # exactly-1 k=0 series -- that this module's analytic cross-validation
+            # tests (test_su2rb.py) need to resolve well below that level.
+            popt, pcov = _curve_fit(_decay_model, x, y, p0=p0,
+                                     bounds=([-_np.inf, 0.0], [_np.inf, 1.0]), maxfev=10000,
+                                     xtol=1e-14, ftol=1e-14, gtol=1e-14)
+        a, f = float(popt[0]), float(popt[1])
+        sigma_a, sigma_f = (float(s) for s in _np.sqrt(_np.diag(pcov)))
+        success = True
+    except (RuntimeError, _np.linalg.LinAlgError, ValueError, _OptimizeWarning):
+        a, f, sigma_a, sigma_f, success = _np.nan, _np.nan, _np.nan, _np.nan, False
```
rileyjmurray, line 1007: I'm not sure if I follow the intention here. A warning from scipy of the form "covariance could not be estimated" should be visible to the user. If such a warning is raised by our test suite then it should be suppressed selectively at the test site. The point of pyGSTi's pytest.ini erroring on scipy warnings is that our tests should be explicit about whether they're okay with an optimizer warning, and those warnings should not appear in the output of pytest when we're specifically okay with them.

pygsti/protocols/su2rb.py
```diff
@@ -0,0 +1,1365 @@
...
+            # A failed fit leaves `nan`s in `f`/`sigma_f` for just the failed irrep(s),
+            # but `p = solve(F, f)` and `cov_p` mix *every* irrep together (F is not
+            # block-diagonal), so a single failure silently poisons every entry of
+            # `p`/`cov_p` with `nan`, not just the failed irrep's. Surface that loudly
+            # rather than letting it look like an ordinary (if strange) all-nan result.
+            _warnings.warn(
+                "SyntheticSPAMRB: the per-irrep decay fit failed for irrep(s) %s; "
+                "the recovered rates and covariance (which mix all irreps via F) are "
+                "therefore entirely nan, not just the failed irrep(s)." % failed_irreps,
+                UserWarning)
```
rileyjmurray, line 1116: pyGSTi-raised warnings need to use a pyGSTi-defined class, not plain UserWarning.

pygsti/tools/matrixtools.py
```diff
@@ -964,9 +964,55 @@ def eigenvalues(m: _np.ndarray, *, assume_hermitian: Optional[bool] = None, assu
...
+        The eigenvalues of `m`.
+
+    inv_evecs : numpy array
+        The inverse of `evecs` (equal to `evecs.T.conj()` whenever `evecs`
+        is unitary, i.e. whenever the `assume_hermitian` or `assume_normal`
+        path was taken).
+    """
     if assume_hermitian is None:
-        assume_hermitian = is_hermitian(m)
+        assume_hermitian = (not assume_normal) and is_hermitian(m)
```
rileyjmurray, line 1015: Is this change necessary? If a matrix is verified as Hermitian then we should use eigh, regardless of whether assume_normal is True. (Hermitian operators are normal.) The cost of verifying Hermicity is nothing compared to an eigendecomposition.

rileyjmurray (commented):
High-level design notes:
 1. Unclear if we should include SSRB or SS_{\Xi}RB. The results of the paper were unambiguous that SSR1RB outperforms both of these.
 2. Unclear if we should restrict to gate-independent errors for the simulator. We could maybe allow a callable factory that returns an error channel for a given triplet of Euler angles. The default case of gate-independent error means the input Euler angles would be ignored.

Recurring notes:
1. Copyright statements on new files should use 2026 for the year.
1. Use type annotations.
1. Use f-strings rather than `%s` string substitutions, where possible.
2. Make sparing use of parentheticals in docstrings.
3. The expressions appearing in a `return` statement should only need trivial evaluation, if any. When debugging it's useful to inspect a function's return value with all its local variables still in the stack frame.
4. Explanatory comments for a given line of code should generally appear after that line, with a caret used to point upward at the line being discussed. For example, 
    ```python
        # Set before calling super().__init__ so that BenchmarkingDesign inserts
        # 'idealout_lists' at the front and registers 'json' auxfile types for all of
        # these (mirrors CliffordRBDesign._init_foundation / BinaryRBDesign._init_foundation).
        self.paired_with_circuit_attrs = ["euler_angles", "seq_index", "prep_index"]
    ```
    should be
    ```python
        self.paired_with_circuit_attrs = ["euler_angles", "seq_index", "prep_index"]
        # ^ Set before calling super().__init__ so that BenchmarkingDesign inserts
        #   'idealout_lists' at the front and registers 'json' auxfile types for all of
        #   these (mirrors CliffordRBDesign._init_foundation / BinaryRBDesign._init_foundation).
    ```
 5. Comments should not mention things like "active-project" or "the branch".
 6. Multi-line error messages should be built outside the constructor invocation for an error object, especially when this can be used to eliminate a mid-expression line break in source. For example,
     ```python
            raise ValueError("edesign.dim=%d does not match this simulator's spinj.dim=%d"
                              % (edesign.dim, self.dim))
    ```
    should be
     ```python
            msg = f"edesign.dim={edesign.dim} does not match this simulator's spinj.dim={self.dim}"
            raise ValueError(msg)
    ```
    As another example,
    ```python
                raise ValueError("probabilities_from_compositions produced non-negligible imaginary "
                                  "part (norm %g); statepreps/povm may not be Hermitian" % imag_norm)
    ```
    should be
    ```python
                msg = "probabilities_from_compositions produced non-negligible imaginary "
                msg += f"part (norm {imag_norm}); statepreps/povm may not be Hermitian"
                raise ValueError(msg)
    ```

PR #852, opened 2026-07-20
Add line search guard to (simpler) Levenberg-Marquardt optimizer

rileyjmurray:
This PR changes pygsti/optimizers/simplerlm.py.

**Main change**

Adds enumerative line search for the LM optimizer after computing its search direction. We need this because there are times when the norm of the search direction is huge, and existing step-size decision making leads us to jump to a wrong basin. We have to use enumerative search (rather than something fancier) because the 1-dimensional line search problem is likely to have spurious local minima.

The line search is only triggered if the norm of the search direction is sufficiently large relative to the current iterate. The meaning of "sufficient" can be set by a line search parameter. All told there are four parameters controlling line search, packed in a dict.

**Secondary change**

Introduced helper functions to make the inner loop of `simplerlm.py::simplish_leastsq` more readable. The helpers use the `nonlocal` keyword to capture outer-scope variables so that inner-scope assignment writes back to the outer-scope.

**Motivation**

After these changes the Instruments example will actually make sense. (It currently has the optimizer crap out on the CPTPLND parameterization.) It should fix most instances where people had to set small step sizes to avoid wrong-basin problems.

### DIFF
--- a/pygsti/optimize/simplerlm.py
+++ b/pygsti/optimize/simplerlm.py
@@ -171,6 +171,12 @@ class SimplerLMOptimizer(Optimizer):
         by the objective function's `.terms()` and `.lsvec()` methods (`'normal'` mode) or the
         "per-circuit quantities" computed by the objective function's `.percircuit()` and
         `.lsvec_percircuit()` methods (`'percircuit'` mode).
+
+    linesearch : dict, optional
+        Controls the optional backtracking-to-argmin line search performed on a trial
+        LM step when it looks untrusted.  See `simplish_leastsq`'s `linesearch` parameter
+        for the recognized keys (`'mode'`, `'beta'`, `'max_evals'`, `'kappa'`) and their
+        defaults; missing keys (or a `None` dict) take those defaults.
     """
 
     @classmethod
@@ -186,7 +192,8 @@ class SimplerLMOptimizer(Optimizer):
         return cls()
 
     def __init__(self, maxiter=100, maxfev=100, tol=1e-6, fditer=0, first_fditer=0, init_munu="auto", oob_check_interval=0,
-                 oob_action="reject", oob_check_mode=0, serial_solve_proc_threshold=100, lsvec_mode="normal"):
+                 oob_action="reject", oob_check_mode=0, serial_solve_proc_threshold=100, lsvec_mode="normal",
+                 linesearch=None):
 
         super().__init__()
         if isinstance(tol, float):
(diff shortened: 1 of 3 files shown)


pygsti/optimize/simplerlm.py
```diff
@@ -533,6 +595,52 @@ def simplish_leastsq(
     best_x_state = (mu, nu, norm_f, f.copy())
     # ^ here and elsewhere, need f.copy() b/c f is objfn mem
 
+    def revert_to_best_x(verb):
+        nonlocal oob_check_interval, mu, nu, norm_f
```
sserita, line 599: Ooh I forgot about nonlocal. Same as if we had passed those in as params, right?

test/unit/optimize/test_simplerlm.py
```diff
@@ -41,3 +41,98 @@ def test_simplish_leastsq_x_limits(self):
...
+            xf_none, converged_none, msg_none, *_ = lm.simplish_leastsq(
+                obj, jac, x0.copy(), max_iter=20, max_dx_scale=None, arrays_interface=ari,
+                linesearch={'mode': 'none'})
+        self.assertFalse(converged_none)
+        self.assertEqual(msg_none, "Infinite norm of objective function!")
+
+        ari = _ari.UndistributedArraysInterface(1, 1)
+        xf, converged, msg, _, _, norm_f, _ = lm.simplish_leastsq(
+            obj, jac, x0.copy(), max_iter=20, max_dx_scale=None, arrays_interface=ari,
+            linesearch={'mode': 'guarded'})
```
sserita, line 102: Is the assumption that `"mode": "always"` will pass if guarded passes? Worth checking explicitly like previous test or not?

sserita (approved):
One question on whether it makes sense to expand one test, but everything else looked good. Will leave that to your discretion and approve so you can merge whenever.

pygsti/optimize/simplerlm.py
rileyjmurray: I'm actually not sure. If we passed them as params then assignments like `norm_f = ...` wouldn't be visible to the caller. In-place mutations would be visible though (e.g., `mu *= ...`).

test/unit/optimize/test_simplerlm.py
rileyjmurray: Yeah that's the assumption. The only reason it's not hard-coded to "always" is I don't want to slow down other people's fits. Not worth checking in this case, IMO.

pygsti/optimize/simplerlm.py
sserita: Fair point, yes I forgot about that distinction.

PR #853, opened 2026-07-20
SU(2) qudit RB

rileyjmurray:
This PR implements *rank-1 synthetic-SPAM randomized benchmarking* (SSR1RB, or just R1RB) from "Randomized Benchmarking with Synthetic Quantum Circuits" (Fan, Murray, Ladd, Young, and Blume-Kohout; QST 2026).

R1RB benchmarks the global SU(2) rotations on a single spin-*j* qudit and extracts a separate, SPAM-robust decay for each irreducible component of the twirled noise — the per-irrep error rates it recovers are the quantities directly relevant to spin codes, which correct low-weight errors.

The paper defines a family of three synthetic-SPAM protocols (plain, character-weighted, and rank-1/Legendre-weighted), but its results show the rank-1 variant dominates the other two uniformly, so pyGSTi ships only R1RB. An implementation of the full three-protocol family, with different API conventions, is preserved on the `feature-synthetic-spam-rb` branch in case the other variants are ever wanted; that branch also holds the general-purpose SU(2) character machinery, of which this PR keeps just the Legendre "character core" function that R1RB uses.

### The API

The protocol lives in `pygsti/protocols/su2rb.py` as a self-contained `Design -> DataSimulator -> Protocol -> Results` family:

- `SU2QuditRBDesign` — samples circuits in batches of `2j+1` that share a gate sequence and differ only in their state prep (one per Jz eigenstate). Sequences use a hidden-first-gate convention: the appended inverting gate undoes gates `2..m`, so the net ideal operation is the Haar-random first gate. Circuits, ZXZ Euler angles, and the per-sequence Legendre weights `P_k(cos β)` of the hidden gate (`charcores`) are built in a single sampling pass and carried as JSON aux data, so the analysis never parses circuit labels.
- `SU2QuditRBSimulator` — simulates the design's circuits directly against a `SpinJ` representation (no pyGSTi `Model` involved), inserting an optional noise channel after every gate except the hidden first one. `noise_channel` accepts either a fixed superoperator or a callable factory `(alpha, beta, gamma) -> channel`, invoked once per gate with that gate's own Euler angles, so per-gate error rates can depend on which gate is being applied (a factory that ignores its arguments reproduces the fixed-channel behavior exactly; both paths are tested). The simulator has an O(1) noiseless shortcut, a per-sequence composition cache, and a SPAM-swap API (`compute_nonspam_compositions` / `probabilities_from_compositions`) that re-derives outcome probabilities under arbitrary preps and effects without re-simulating — the machinery behind the paper's SPAM-robustness figures. `jz_dephasing` and `jz_rotation` are provided as simple example noise channels; channels compose by plain matrix multiplication.
- `SU2QuditRB` / `SU2QuditRBResults` — the protocol reconstructs each sequence's prep-by-effect probability matrix, forms the Legendre-weighted per-irrep estimators, fits each irrep's depth series to a single exponential, and recovers per-irrep rates with propagated covariance by inverting the recoupling matrix. The results object exposes the decays and rates, a `rates_dataframe()` export, and a `variance_diagnostic()` sanity check.
- `predicted_zero_noise_variance(j, k)` — the paper's exact zero-noise sampling variance of the irrep-`k` estimator, useful for planning how many sequences a target precision requires.

If a per-irrep decay fit genuinely fails, the recovered rates are poisoned wholesale (the recoupling matrix mixes every irrep), so the protocol warns loudly with a new `RBFitFailureWarning` in `pygsti/tools/exceptions.py`. An unestimable covariance is *not* treated as a failure: `curve_fit`'s `OptimizeWarning` surfaces as an honest `inf` standard error on an otherwise successful fit, and `success=False` is reserved for genuine fit exceptions.

Supporting the protocol are two new tools modules:

- `pygsti/tools/su2tools.py` — the `SpinJ` class (spin-*j* representation with precomputed eigendecompositions of the generators, the Clebsch-Gordan change of basis, irrep projectors, the SPAM-synthesis matrix, and the decay-recoupling matrix), plus Haar sampling of Euler angles, batch unitary construction, composition inverses, and the Legendre character cores.
- `pygsti/tools/wignersymbols.py` — exact Clebsch-Gordan and Wigner 6-j symbols over `fractions.Fraction`, stdlib-only. The unit tests cross-check them against sympy's (via `importorskip`; `sympy` is added to the `testing_no_cython_mpi` extras, and the runtime dependency footprint is unchanged).

### Tutorial notebook

`docs/markdown/rb/SU2QuditRB.md` is a new jupytext/MyST tutorial, linked from the RB overview page and the docs TOC. It walks through the full workflow. First it builds a spin-3/2 design, simulates with composed dephasing + over-rotation noise, runs the protocol, and plots the per-irrep decay curves. Then it cross-checks the fitted decays against the exact twirled-channel eigenvalues, demonstrates SPAM robustness via the simulator's SPAM-swap API, exercises the zero-noise variance diagnostic, and round-trips everything through pyGSTi's standard I/O.

### A note on the SPAM-robustness test

The headline feature of the rank-1 weighting is that fixed SPAM error cannot bias the decays, unlike "plain" synthetic-SPAM RB's `diag(M P Mᵀ)` sandwich. The natural way to test that is to compute the unweighted estimator as a foil against the same simulated data — but that's impossible on `SU2QuditRBDesign`'s own circuits: on hidden-first-gate data the unweighted sandwich's expectation is *exactly* zero for every irrep `k > 0`, under both ideal and arbitrarily perturbed SPAM, because a Haar-random hidden gate's average conjugation is the full depolarizer regardless of what SPAM feeds into it.

So `test_su2rb.py`'s `TestSU2QuditRBSpamRobustness` instead builds a small batch of *net-identity* sequences (an inverting last gate composing to the identity, with no hidden gate) inline from public `su2tools` primitives — a few lines of numpy, touching no production code. On that convention the unweighted sandwich exhibits the textbook "plain synthetic-SPAM RB is not SPAM-robust" bias, while `SU2QuditRB` run on its own hidden-gate pipeline under the identical SPAM perturbation stays consistent with the true fidelity.

### Incidental changes

- `pygsti/circuits/circuit.py`: a docstring clarification for `insert_layer` and `insert_layer_inplace` (passing `j=-1` inserts at the second-to-last position rather than appending), which tripped me up while building the design's circuits.
- `pygsti/tools/matrixtools.py`: a stray blank line removed.
- A CHANGELOG entry under "Added".

### Testing

New unit test modules `test/unit/protocols/test_su2rb.py`, `test/unit/tools/test_su2tools.py`, and `test/unit/tools/test_wignersymbols.py` cover the design, simulator, protocol, and tools — including hand-composed probability cross-validation, serialization round-trips, the noise-factory paths, and the SPAM-robustness demonstration above.

Full `test/unit` suite, py313, no failures:

```
conda run -n py313 python -m pytest test/unit -q
# 3041 passed, 99 skipped, 2 xfailed, 59 warnings, 95 subtests passed in 284.62s (0:04:44)
```


CHANGELOG
sserita, line 1: Nice, would probably be good to do this piecemeal as we go so the release is not such a huge lift to remember what we did when.

sserita (approved):
Ahh, Clebsch-Gordon coefficients, brings back fun memories of angular momentum.

All looks like straightforward math implementation for the most part. I like the idea of sympy as analytic testing when possible and updating the CHANGELOG as we go.

Feel free to merge when no conflicts.

PR #854, opened 2026-07-21
Fix bugs in processor spec serialization: handling non-standard instrument labels

rileyjmurray:
*In what follows, "we" refers to "Fable and I."*

-------------------------

JSON serialization of a QubitProcessorSpec or QuditProcessorSpec object can lead to a corrupted nonstd_instruments dict.

Currently, keys of the nonstd_instruments dict are flattened with ':'.join(map(str, key)), which explodes plain-string names character-by-character ('Iparity' → 'I:p:a:r:i:t:y') and coerces integer sslbls to strings, while compound instrument names came back from JSON as (unhashable) lists. A round-tripped spec with instrument names then crashed in instrument_specifier with TypeError: unhashable type: 'list' (encountered via StandardGST.run). In fact no non-standard instrument survived a round trip correctly — outcomes ranged from crashes to silently substituting a standard instrument for the user's custom spec.

This PR:
  - serializes nonstd_instruments as a list of [name, spec] pairs, preserving name structure and sslbl types;
  - keeps reading the old dict format, with a best-effort repair that matches each legacy flattened key against the faithfully round-tripped instrument_names;
  - reconstructs compound instrument names consistently (as tuples) in both _from_nice_serialization implementations.

The existing regression test now asserts round-tripped values end to end, and new tests cover string-keyed custom specs, the qudit path, and the legacy on-disk format.

While working on this we found a separate latent bug in nearby code (_unserialize_instrument_member misparses instrument members that aren't exactly two rank-1 pairs). It's orthogonal to this fix; we'll open an issue for it shortly.

### DIFF
--- a/pygsti/processors/processorspec.py
+++ b/pygsti/processors/processorspec.py
@@ -29,6 +29,12 @@ from pygsti.modelmembers.operations import LinearOperator as _LinearOp
 from pygsti.modelmembers.operations import FullArbitraryOp as _FullOp
 
 
+def _tuplize(x):
+    if isinstance(x, (list, tuple)):
+        return tuple((_tuplize(el) for el in x))
+    return x
+
+
 class ProcessorSpec(_NicelySerializable):
     """
     The API presented by a quantum processor, and possible classical control processors.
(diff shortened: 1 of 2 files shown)


rileyjmurray (commented):
Leaving this comment to document that I've reviewed the tests, and they look good.

PR #733, opened 2026-04-03
Type annotations for the baseobjs

nkoskelo:
Add type annotations to baseobjs to improve the maintainability of the code.

PR #858, opened 2026-07-27
Protectedarray hardening

nkoskelo:
The 'ProtectedArray' class promises that one should not be allowed to edit the underlying array if there is a protected index which would be modified. This PR adds additional tests for safeguards.

Note, that this PR also adds that any inplace modification operation like "*=" will raise a `ValueError` even if there are no protected items. This was done as a precaution.

This was originally part of PR #733 but we are splitting this out into its own PR for ease.


sserita (approved):
Looks straightforward.

rileyjmurray (approved):

Issue #826, opened 2026-07-06
Parameterization-preserving gauge optimization bugs


rileyjmurray:
A collaborator sent me the following.

>  I remember looking at my reports of the same experiment before and after param-preserving gauge opt, and it seemed like only the error bars were affected. The error-metric point estimates looked the same in both cases for everything including instruments. But with what I am working on now, it seems like the reports are very similar except for the instruments, where the values for the raw estimates are different in the pre- and post- param-preserving reports. The entanglement infidelity, for example, is different for the instruments but not the gates.

I should investigate why this is happening.

rileyjmurray:
I believe I've sorted things out. The collaborator's code was dropping an initial step in the gauge optimization suite passed to parameterization-preserving gauge optimization. Whether that dropped step _needed to be dropped_ in some earlier context is unclear. In any case, it seems currently safe to keep that step, provided the code is run with what we have on develop (or 0.10.1; either is fine). 

One thing the investigation surfaced is that **Instruments seem to be considered SPAM** when thinking about spam_weight vs gates_weight in building a gauge optimization objective. This matters because the step of gauge optimization that was skipped put most (all?) weight on SPAM. We might want to change that down the line.

I'm leaving this issue open until I get confirmation from the collaborator that they're happy with things as they are.

rileyjmurray:
Update: the collaborator and I were able to agree on that solution. _However_, a robot found an issue with the CPTPLND Instrument and parameterization-preserving gauge optimization. Details follow.

-------------------------------

Found while wiring parameterization-preserving gauge optimization into an
instrument warm-start helper (feature-instrument-seed-tools branch, July 2026).

## Summary

`pygsti.models.explicitmodel.transform_composed_model` is the workhorse of
`_add_param_preserving_gauge_opt`. For a model containing an effect-then-gate
parameterized instrument (the structure built by `Instrument.from_effects` /
`from_cptr_superops`), the transformed model has MORE parameters than the
input: the instrument's shared ComposedPOVM error map stops being shared and
its parameters are duplicated once per member.

Minimal reproduction (1 qubit, pp basis):

```python
inst = Instrument.from_effects({'p0': E0_superket, 'p1': E1_superket}, basis)
mdl = ExplicitOpModel(QubitSpace(1), basis); mdl[('Iz', 0)] = inst
mdl.to_vector()
print(mdl.num_params)                     # 36  (12 povm errmap shared + 2 x 12 gate errmaps)
[...]

PR #862, opened 2026-07-30
WIP: more tools for Instrument fitting

rileyjmurray:
This work-in-progress PR has resources for @pcwysoc.

Lindblad-parameterized Instruments are tricky to fit. This is because when we promote from an ideal model (e.g., one with projective effects) that _usually_ introduces undesired constraints that make it possible for us to express all Lindblad-representable CPTR maps . This issue can partly be avoided by switching to a Kraus-polar Lindblad parameterization of instruments, but pathologies can also arise in that case.

This goal of this PR is to provide tools for reliable Lindblad fits of any valid instrument someone might throw at us. At the end of the day these tools just provide answers to the following question.

 > "If I don't seed at the ideal model, then how should I seed?"

**Answer 1.** Run a TP fit, project the noisy instrument members onto the set of CPTR maps (ideally using a semidefinite programming solver), then run a CPTPLND fit where all non-Instrument members are initialized with their ideal values. This is answer is encapsulated in the `refit_instruments_cptplnd` function in pygsti/protocols/gst.py. This function is temporary until we come up with a better workflow.

**Answer 2.** Provide methods that can explain _why_ a given seed Instrument will create problems in a Lindblad fit, along with functions to heuristically "patch" the instrument to address *those specific problems*. I like this approach because if we get it to work reliably we should be able to add it to the StandardGST protocol without making big changes to how that protocol works. The diagnostic functions themselves are in instruments/diagnostics.py. The functions for applying the targeted heuristic fits are in instruments/seeding.py. The one-stop shop to diagnose an instrument seed _and_ apply our heuristic repairs is `patch_instrument_seed` in instruments/seeding.py.


agent-docs/01-representation.md
```diff
@@ -152,6 +152,8 @@ The third layer has a non-obvious consequence: **representation degeneracies in
...
 
+> **TENTATIVE (branch `feature-instrument-seed-tools`, not yet merged).** The decomposed `(E_k, G_k)` become a **frozen static base** for Lindblad gate parameterizations, and the base's rank / effect-span / spectral-interval structure is an invariant of the whole fit — an ideal projective seed hard-caps the reachable member rank (construction now warns). Two new sibling modules address this: [`modelmembers/instruments/diagnostics.py`](../pygsti/modelmembers/instruments/diagnostics.py) (data-free O(1) checks plus a sampled Jacobian certificate; `diagnose_instrument`, `diagnose_model_instruments`) and [`modelmembers/instruments/seeding.py`](../pygsti/modelmembers/instruments/seeding.py) (repairs: `full_rank_base_blend`, `interior_effect_offset`, `displace_base_effects`, one-call `patch_instrument_seed`, and `project_instrument_to_cptp` with an SDP/Dykstra backend). Protocol-level, `pygsti.protocols.refit_instruments_cptplnd` warm-starts a CPTPLND fit from the CPTP projection of an existing non-Lindblad estimate. See [docs/markdown/objects/Instruments.md](../docs/markdown/objects/Instruments.md) for the workflow.
```
sserita, line 155: Remove TENTATIVE? Also these seems pretty agent-y, would be nice if they were more natural.
As a general rule, I tend to dislike agents describing things as "new" and giving documentation based on a diff of base branch code with to-be-merged code. I typically ask it for "self-contained" doc updates, or somesuch - perhaps something to configure for our AGENT docs, after discussion?

pygsti/models/explicitmodel.py
```diff
@@ -1866,24 +1866,29 @@ def transform_composed_model(mdl: ExplicitOpModel, s : _GaugeGroupElement) -> Ex
     # reverse order. For example, ComposedOp([X,Y,Z]) is applied to
     # a vector v as Z @ Y @ X @ v.
 
-    for key, rho in oldmdl.preps.items():
+    # NOTE: we rebuild from the members of `mdl` (the copy) rather than those of `oldmdl`.
+    # A rebuilt member holds a *reference* to the member it wraps, so wrapping `oldmdl`'s
+    # members would leave the returned model sharing parameterized objects with its input:
+    # calling `from_vector` on either model would then silently mutate the other.
```
sserita, line 1872: Is this an unrelated bugfix/change or something that was required by the new instrument seed infra (possibly the shared memo needs)?

If unrelated, is there an issue this fixes we can link? If not, this is probably worth calling out separately.

docs/markdown/objects/Instruments.md
sserita, line 1: Awesome :)

pygsti/modelmembers/instruments/seeding.py
sserita, line 1: Minor quibble: Is `seeding` a good name for this? Most of the content seems to be about "repairing" seeds, which is relevant to seeding Instruments... but I would have expected something more... active for the module name. Using the continuous tense (i.e. -ing suffix) seems unlike most other module conventions. Also "seeds" in most other cases are about RNG, rather than initial guesses.

IDK, maybe something like seedrepair or seedpatchers (or just repairs/patchers) seems more conventional? Not something I feel strongly about, just bringing it up for discussion.

test/unit/objects/test_explicitmodel.py
```diff
@@ -288,3 +314,67 @@ def test_tp_instrument_member_matrices_correct(self):
...
+        v = result.to_vector()
+        result.from_vector(v + 0.02 * np.random.default_rng(0).standard_normal(len(v)))
+        total = sum(result.instruments['Iz'][ek].to_dense() for ek in self.inst_keys)
+        trace_functional = np.array([np.sqrt(2), 0., 0., 0.])  # <<I| in the pp basis
+        npt.assert_allclose(trace_functional @ total, trace_functional, atol=1e-9)
+
+    def test_result_does_not_alias_input_model(self):
+        # A rebuilt member holds a reference to the member it wraps, so the rebuild
+        # must consume the *copy's* members: otherwise from_vector on the result
+        # silently mutates the input model.
```
sserita, line 421: Re: my explicitmodel change comment, it seems like this is a bug being caught and worth mentioning in the PR/release notes eventually?

sserita (changes requested):
Firstly, this is super cool and seems like a huge step forward in pyGSTi's ability to analyze what is going wrong with Instruments and proposing fixes for them.

Minor changes/discussion for docs/high-level naming, the bulk of the module looks good.

agent-docs/01-representation.md
sserita: Oh lol I'm just seeing now that this is in agent-docs only. In which case it can stay however terse it wants. Probably should still remove TENTATIVE and finalize changes.

pygsti/models/explicitmodel.py
rileyjmurray: This is related to #826. I've left a comment there about the connection. I guess that's a blocker before our next release.

test/unit/objects/test_explicitmodel.py
rileyjmurray: Agreed that this is a bugfix and needs to be in release notes.

pygsti/models/explicitmodel.py
rileyjmurray: Rebased after merging the PR that fixes 826. `explicitmodel.py` is no longer changed in this PR. Marking thread as resolved.

test/unit/objects/test_explicitmodel.py
rileyjmurray: Rebased after merging the PR that fixes 826. Neither `explicitmodel.py` nor `test_explicitmodel.py` is changed in this PR. Marking thread as resolved.

PR #863, opened 2026-08-01
Fix trace-preservation check for bases starting with identity

liuyichao82:
Hi PyGSTi team,

For bases whose first element is the identity, trace preservation requires the first row of the process matrix to be [1, 0, ..., 0].

The previous implementation checked a[1:] instead of a[0, 1:], causing trace-preserving process matrices to be misclassified and potentially selecting incorrect downstream computation paths.

I hope this fix is helpful.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## [0.10.3] - Unreleased
+### Fixed
+* Fixed the trace-preservation check for bases whose first element is the identity.
+
 ## [0.10.2] - 2026-07-30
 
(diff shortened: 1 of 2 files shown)


rileyjmurray (changes requested):
Zoinks! Thanks so much for catching this. I'm requesting changes for you to target our develop branch instead of the master branch. Please also update the `CHANGELOG` file to include a note about the fix for an as-yet-unreleased version 0.10.3.

liuyichao82:
Thanks! I've switched the PR to the develop branch and added a note about the fix to the CHANGELOG for the upcoming 0.10.3 release.

By the way, the new pyGSTi documentation looks great — I really like it!

rileyjmurray (approved):

Issue #713, opened 2026-02-02
Efficiency of embedded ops in the densitymx evotype(s)


rileyjmurray:
Tensor product tricks can be used to apply an embedded op to the relevant subsystem of a many-qubit superket. It's important that we use these tricks in forward simulation. It's not 100% clear to me that we are using them.

The relevant code for applying an embedded op lives in two places, one in the densitymx evotype and one in the densitymx_slow evotype; see below. It's not clear to me how these implementations work. It would be nice to add comments with explanations, and to run some experiments that demonstrate the scaling as the number of idling subsystems increases.

https://github.com/sandialabs/pyGSTi/blob/fb8959677f7e3ede914ff66838c437603e0c9fa2/pygsti/evotypes/densitymx/opcreps.cpp#L93-L158

https://github.com/sandialabs/pyGSTi/blob/fb8959677f7e3ede914ff66838c437603e0c9fa2/pygsti/evotypes/densitymx_slow/opreps.py#L381-L403

ping @enielse. I imagine you implemented these several years ago. Can you comment on their approach and efficiency?

rileyjmurray:
Update. Erik said that what he implemented was a bit of indexing olympics to somehow avoid being wasteful, but he didn't consciously use the tensor-product identity that I have in mind.

rileyjmurray:
Got some help from a robot friend. Here's the plan it wrote and the script it generated, both with light edits by me. Results for an Apple M3 chip are surprising. The densitymx_slow evotype ends up being much faster than densitymx as n_noop (the number of no-op qubits, aka idle qubits) increases.

### Algorithm descriptions

**Disclaimer: the descriptions that follow are LLM generated with light editing by me.**

Both `densitymx` (C++) and `densitymx_slow` (Python) use the same sparse iteration approach to apply an n_total qubit gate with n_target nontrivial target qubits and n_noop perfect idle qubits.

The approach:
- Loop over all 4^n_noop combinations of non-target (idle) basis elements
- For each combination, extract the (4^n_target)-dimensional sub-vector indexed by `_baseinds`
- Apply the embedded gate to that sub-vector
- Write the result back into the output state

The computational complexity for this approach scales as O(4^n_noop) for fixed n_target.

[...]

PR #864, opened 2026-08-04
Test coverage and small fixes, split out of the crosstalk-free GST branch

nkoskelo:
This PR carries the maintenance work that accumulated on the branch where I've been building crosstalk-free GST (XFGST) experiment designs. None of it is XFGST-specific, and the feature PR that follows needs none of it — I checked, by putting the feature module and its tests onto `develop` by themselves and running them. Splitting it out this way takes the feature PR from +3200 to +2049 lines and leaves nothing in it but the feature.

Most of what's here is test coverage, plus a relocation of the tests that assert on wall-clock time. There are three behavior changes, and they're described first because they're the part that actually needs reading.

### Behavior changes

**`Circuit.serialize` preserves `CircuitLabel` layers.** With `expand_subcircuits=False`, a `CircuitLabel` layer is now appended as-is rather than expanded into its components. The docstring already promised this ("If `False`, the circuit may contain `CircuitLabel` layers") and the code didn't do it, but it's still an observable change to a core public method.

**`EffectRepComputational.probability` no longer densifies.** It calls the new `matrixtools.zvals_int64_probability` instead of building a dense length-`4**nfactors` effect vector and taking a full dot product, so the work is `O(2**n)` rather than `O(4**n)`. This mirrors what the compiled evotypes already do in C++. It's a hot path for every forward simulation on `densitymx_slow`, which is worth more than it sounds: per #713, `densitymx_slow` can actually beat the compiled path as the number of idle qubits grows, so it isn't only a fallback.

**`compute_2Q_connectivity` and `compute_clifford_2Q_connectivity` no longer crash on `{idle}`.** `gate_num_qubits('{idle}')` equals the device size, so on a two-qubit spec the global idle clears the `== 2` arity guard — but its availability is `[None]`, so both methods reach `qubit_labels.index(sslbls[0])` and raise `TypeError: 'NoneType' object is not subscriptable`. The Clifford method's `except (KeyError, TypeError, ValueError)` doesn't save it: `clifford_symplectic_rep_of(Label('{idle}', None))` succeeds, so the crash lands after the guard rather than inside it. The idle is now special-cased and treated as available on every qubit. Today this means any two-qubit processor spec with `{idle}` in its gate set raises from both methods.

### `matrixtools`

Two substantive changes plus a large annotation pass.

`zvals_int64_probability` is the sparse inner product used by the effectrep change above. Supporting it are `_zvals_int64_indices_and_signs`, the bit-manipulation helpers `_spread_bits` and `_reverse_bits_scalar`, and a module-level `_zvals_int64_to_dense_cache` keyed on qubit count; `zvals_int64_to_dense` is rewritten around the same machinery. The cache holds index and sign arrays only — output buffers are never shared between calls, which `ZvalsInt64Tester` pins directly.

The annotation pass re-types roughly sixty-five signatures across the module. It's mechanical and behavior-free, but it does touch most of the file, so `git blame` on `matrixtools.py` gets noisier. I'm happy to pull it into a separate PR if reviewers would rather.

### Test coverage

**`test/unit/objects/test_processorspec.py` moves to `test/unit/processors/`**, matching the `pygsti/processors/` package it tests, and grows from 15 to 33 test methods. The new ones cover `resolved_availability` (both the mode matrix and the contradiction path), `subset`, `map_qudit_labels`, `rename_gate_inplace`, `gate_num_qubits`, `available_gatenames` / `available_gatelabels`, `availability_specifiers`, `compute_ops_on_qudits`, `compute_clifford_symplectic_reps`, `compute_2Q_connectivity`, and the new `GlobalIdleConnectivityTester` for the crash above. **Note for anyone with a hardcoded path to the old location.**

One of those deserves its own note. `test_compute_clifford_2Q_connectivity_only_counts_validated_sites` pins that the Clifford connectivity method validates *each availability site* rather than trusting the gate name. `nonstd_gate_symplecticreps` can register a symplectic rep against a specific `Label` — one embedding — instead of against a gate name, so name-level trust produces phantom edges that disagree with `compute_clifford_ops_on_qubits` and then crash `create_random_germ` (`pygsti/algorithms/randomcircuit.py:1736`) with `ValueError: low >= high` when one gets sampled. To be clear about what's landing: the guard itself is unchanged relative to `develop`. It was dropped and reinstated within my branch, so what this PR contributes is the regression test and an explanatory comment, not a fix. It's a randomized-benchmarking concern with no connection to XFGST.

**`test_rep_to_dense_default.py`** gains `ComputationalEffectProbabilityTester`, which is the real check on the effectrep speedup: it compares `probability` against the dense dot product it replaced, against the compiled `densitymx` evotype computing the same thing in C++, and — as a check that doesn't route through `to_dense` at all — that the outcomes of a complete computational-basis POVM sum to the trace of the state.

**`test_matrixtools.py`** gains `ZvalsInt64Tester` (both routines against an independently constructed `kron` reference, the `N == 0` scalar case, `abs_elval` handling, buffer aliasing, and cache reuse across calls with different z-values).

**`test_circuit.py`** gains nine tests: `serialize` / `parallelize` round-tripping in both the inverse and non-inverse cases, `set_line_labels` on editable and non-editable circuits, `set_occurrence`, the string properties, and the QUIL string helpers.

**`test_localnoisemodel.py`** is reformatted, and its seven bare `assert(...)` calls are now `self.assert*` calls. This adds no coverage — it's readability only, and it's about a third of the test diff. It's in its own commit if you'd like to skip it.

### Timeout-sensitive tests move to `test/performance/`

Three tests asserted on wall-clock behavior, which makes them fail under CPU load rather than on any defect:

- `GraphColoringScalingTester` — runs every algorithm under a hard per-algorithm timeout and asserts the sparse-safe ones finished.
- `test_deterministic_algorithm_colors_dense_graph` — asserts termination inside `PER_ALGO_TIMEOUT`.

On a busy machine a perfectly healthy algorithm gets killed on its budget and the failure says nothing about the code. I hit this while preparing this PR: with several suites running at once these produced up to eleven failures, and on unmodified `develop` with the box loaded, six of them fail. They now live under `test/performance/`, which the ordinary unit run does not collect.

The graph fixtures and `assess_coloring` are used by both the correctness suite and the scaling suite, so rather than duplicate them they move to `test/helpers/graphcoloring_graphs.py`. `assess_coloring` is what decides whether a coloring is *correct*, and a fix to it must not be able to land in only one of the two suites. `test/performance/` gains an `__init__.py` so it can share that module by ordinary relative import, the way the rest of the tree does.

CI keeps exactly the coverage it had. The graphcoloring change detector now also watches the relocated suite and the shared fixtures, and the conditional slow-test step runs the new path. `test/performance/` is still not part of any unscoped CI run.

### Incidental changes

- Sandia license headers added to `pygsti/circuits/split_circuits_into_lanes.py`, `test/unit/circuits/test_circuit_splitting.py`, and `test/unit/tools/test_graphcoloring.py`, none of which got one when #848 and #849 landed.
- `pygsti/models/explicitmodel.py`: a missing docstring blank line.
- `objectivefns` is now imported eagerly (`pygsti/__init__.py`, `pygsti/objectivefns/__init__.py`), so `pygsti.objectivefns.objectivefns.X` resolves without a separate import — the intent stated in commit `c4d3fb89e`. Small import-time cost, no known circularity.
- `.github/workflows/reuseable-main.yml`: drops the stale `pygsti/tools/graphcoloring.py` path from the graph-coloring change detector, since #848 turned that module into a subpackage. (The detector's other edits, and the retargeted slow-test step, are described under the relocation above.) Those performance tests will only run when we trigger the extra-tests in CI.

### Reviewer attention

1. **The `{idle}` branch sets `avail = [qubit_labels]` and then indexes `sslbls[0]` / `sslbls[1]`,** which is correct only for a two-qubit spec. It's safe today, because the idle's arity equals the device size and the enclosing `gate_num_qubits(gn) == 2` guard therefore confines the branch to two-qubit devices. The residual exposure is a user-registered two-qubit gate literally named `{idle}` on a larger spec, which would connect only the first two qubits. An assertion would make that reasoning local rather than implicit, and I'd rather add one than leave it to be rediscovered.
2. **The relocation changes what `pytest test/unit` collects.** Eleven tests (the five scaling tests, the six parametrized dense-graph cases) no longer run in the unit suite. They are not deleted and CI still runs the graphcoloring ones when that toolkit changes — but if you rely on a bare `pytest test/unit` as your local gate, that gate is now slightly narrower. Running `pytest test/performance` on an idle machine is the replacement.


.github/workflows/reuseable-main.yml
```diff
@@ -189,7 +192,7 @@ jobs:
         GRAPHCOLORING_CHANGED: ${{ steps.graphcoloring.outputs.changed }}
       run: |
         python -Ic "import pygsti; print(pygsti.__version__); print(pygsti.__path__)"
-        python -m pytest -m slow --durations=25 test/unit/tools/test_graphcoloring.py
+        python -m pytest -m slow --durations=25 test/performance/test_graphcoloring_scaling.py
```
rileyjmurray, line 195: Please change so that the performance tests only run when other performance tests are run. (This doesn't happen in our usual workflows.)

pygsti/tools/matrixtools.py
```diff
@@ -1349,18 +1367,44 @@ def minweight_match(a, b, metricfn=None, return_pairs=True,
...
+        except Exception:
+            filled = False
+
+    if not filled:
+        if pass_indices_to_metricfn:
+            for i, x in enumerate(a):
+                weightMx[i, :] = _np.ravel(_np.array([metricfn(i, j) for j, y in enumerate(b)]))
+        else:
+            for i, x in enumerate(a):
+                weightMx[i, :] = _np.ravel(_np.array([metricfn(x, y) for j, y in enumerate(b)]))
```
rileyjmurray: The added implementation complexity of this function seems not worth it. We have three invocations of `minweight_match` throughout the codebase. They're all applied to eigenvalues or eigenvectors--- objects of size O(n) obtained after an O(n^3) computation. I think the call to scipy's linear_sum_assignment function also has cubic time complexity. We shouldn't introduce this much code complexity to vectorize O(n^2) work that's sandwiched between calls that take cubic work before and after.

test/performance/test_graphcoloring_scaling.py
```diff
@@ -0,0 +1,324 @@
+#***************************************************************************************************
+# Copyright 2015, 2019, 2026 National Technology & Engineering Solutions of Sandia, LLC (NTESS).
+# Under the terms of Contract DE-NA0003525 with NTESS, the U.S. Government retains certain rights
+# in this software.
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License.  You may obtain a copy of the License at
+# http://www.apache.org/licenses/LICENSE-2.0 or in the LICENSE file in the root pyGSTi directory.
+#***************************************************************************************************
+
+"""Wall-clock scaling and algorithm-selection suite for edge coloring.
```
rileyjmurray, line 10: Unit tests that check wallclock scaling are fragile. What's the rationale for having this sort of test?

test/unit/tools/test_matrixtools.py
```diff
@@ -250,3 +250,180 @@ def sort_key(z):
...
+
+        def by_index(i, j):
+            return abs(self.A[i] - self.B[j])
+
+        by_value_weights = mt.minweight_match(
+            self.A, self.B, metricfn=by_value, return_pairs=False)
+        by_index_weights = mt.minweight_match(
+            self.A, self.B, metricfn=by_index, return_pairs=False,
+            pass_indices_to_metricfn=True)
+        self.assertArraysAlmostEqual(by_value_weights, by_index_weights)
```
rileyjmurray: This can be removed if the changes to minweight_match are reverted.

test/unit/tools/test_graphcoloring.py
```diff
@@ -783,151 +519,4 @@ def test_auto_non_bipartite_determinism(self):
...
-        printed table is the deliverable) and assert only that the timeout guard
-        protected the run and that at least one algorithm succeeded.
-        """
-        for n in [6, 10, 16]:
-            graph = make_complete_graph(n)
-            results = self._run_all(graph)
-            _print_table(f"complete K_{n}", graph, results)
-            self._assert_timeout_guard_held(results, f"K_{n}")
-            self._assert_at_least_one_valid(graph, results, f"K_{n}")
+        self.assertEqual(_canonical_coloring(r1), _canonical_coloring(r2))
```
rileyjmurray: missing blank line at EOF

rileyjmurray (changes requested):
Mostly small comments. The main issue is I don't want any wallclock time tests in `test/unit`.

test/performance/test_graphcoloring_scaling.py
nkoskelo: The rationale for this sort of test is to ensure that the implementations match the theoretical asymptotic runtime and to be able to see the difference in runtime for the algorithmic choices the user can make.

pygsti/tools/matrixtools.py
nkoskelo: Understood. I was looking at it in isolation. I am reverting those changes.

test/unit/tools/test_matrixtools.py
nkoskelo: Removed.

rileyjmurray (approved):

PR #865, opened 2026-08-05
Fix ExpErrorgenOp.to_dense to return a unitary when a Hilbert space is requested

nonohuff:
### Summary

`ExpErrorgenOp.to_dense(on_space='Hilbert')` — and `to_dense()` on any evotype whose `minimal_space` is `'Hilbert'` (e.g. `'statevec'`) — is contractually supposed to return a unitary. On `develop` it does not:

```python
from pygsti.modelmembers.operations import LindbladErrorgen, ExpErrorgenOp

eg = LindbladErrorgen.from_elementary_errorgens({('H','X'): 0.1}, evotype='statevec', state_space=1)
op = ExpErrorgenOp(eg)

op.evotype.minimal_space   # 'Hilbert'
op.to_dense().shape        # (4, 4)  <- a superoperator, not a 2x2 unitary
op.to_dense('Hilbert')     # bare AssertionError from inside LindbladErrorgen.to_dense
```

The `to_dense()` case is the worse of the two: it silently returns an object of the wrong space and shape rather than failing.

### Root cause

The cast down to a unitary is attempted in the wrong class, and currently in neither. An error generator is never itself unitary — it is the argument of the exponential, not the result — so `LindbladErrorgen.to_dense` correctly refuses `on_space='Hilbert'`:

```python
assert(on_space in ('minimal', 'HilbertSchmidt'))
```

But `ExpErrorgenOp.to_dense` forwarded its `on_space` straight through to the error generator and returned `expm(...)` unchanged, so the only place that could legitimately produce a unitary — *after* exponentiating — never tried to.

### The fix

- Ask the error generator for a `'HilbertSchmidt'` superoperator explicitly, instead of forwarding `on_space` and tripping the assert above.
- After exponentiating, cast down via `optools.superop_to_unitary` when a Hilbert space was requested, re-raising its `ValueError` with an actionable message when the operation isn't unitary (i.e. the error generator has non-Hamiltonian terms).
- Document `on_space` and add a `Raises` clause; the docstring previously described neither.

### Provenance — this is not new work

This is commit 0bdbb2830 by @sserita, *"Convert ExpErrorgenOp, not LindbladErrorgen, to unitary"* (Jul 2024), which moved this capability out of `LindbladErrorgen` into `ExpErrorgenOp`. That commit is an ancestor of the `feature-ml-*` branches only — **not** of `develop`, `beta`, or `master`:

```
$ git branch -r --contains 0bdbb2830
  origin/feature-ml-errorgenprop-refactor
  origin/feature-ml-rewrite
  origin/feature-ml-rewrite-noah
  origin/noah-feature-ml
```

So `develop` received neither half of that move. I found it riding along inside the QPANN PR (#835) and am extracting it here, with tests and a docstring added, so it can be reviewed on its own merits rather than as unrelated scope inside a large feature PR. Deleting it from #835 in the same pass.

### Correctness check

pyGSTi's Hamiltonian generator is `H_P[ρ] = -i[P, ρ]`, so `exp(0.1 · H_X)` must equal `exp(-0.1i·X) = cos(0.1)·I − i·sin(0.1)·X`. The returned 2×2 matrix matches to numerical precision.

### Tests

Adds `ExpErrorgenOpToDenseHilbertTester` to `test/unit/modelmembers/test_operation.py`:

| test | asserts |
|---|---|
| `test_hilbert_evotype_returns_unitary` | `'minimal'` and `'Hilbert'` both give a 2×2 unitary, matching the closed form above |
| `test_explicit_hilbert_schmidt_still_returns_superop` | `'HilbertSchmidt'` still returns the 4×4 superoperator |
| `test_hilbert_schmidt_evotype_unaffected` | `'densitymx'` (`minimal_space == 'HilbertSchmidt'`) is untouched |
| `test_non_unitary_errorgen_raises_informative_error` | a stochastic generator raises the informative `ValueError` |

The first and last were confirmed to fail against the unfixed code.

### Verification

- `test/unit/modelmembers` — 734 passed
- `test/unit/objects` + `test/unit/tools` — 1452 passed
- full `test/unit` with CI's own `-m "not slow"` filter — the only failure is `test_gst.py::GateSetTomographyTester::test_run_with_no_gaugeoptsuite`, which I confirmed **fails identically on a pristine `develop` worktree** and is unrelated to this change.
- `flake8 --config=.flake8-critical` — clean.

### DIFF
--- a/pygsti/modelmembers/operations/experrorgenop.py
+++ b/pygsti/modelmembers/operations/experrorgenop.py
@@ -33,6 +33,7 @@ from pygsti.modelmembers.errorgencontainer import ErrorGeneratorContainer as _Er
 from pygsti.modelmembers.torchable import Torchable as _Torchable
 from pygsti.baseobjs.polynomial import Polynomial as _Polynomial
 from pygsti.baseobjs import _compatibility as _compat
+from pygsti.tools import optools as _ot
 from pygsti import SpaceT
 
 IMAG_TOL = 1e-7  # tolerance for imaginary part being considered zero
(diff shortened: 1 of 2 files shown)


rileyjmurray:
> * full `test/unit` with CI's own `-m "not slow"` filter — the only failure is `test_gst.py::GateSetTomographyTester::test_run_with_no_gaugeoptsuite`, which I confirmed **fails identically on a pristine `develop` worktree** and is unrelated to this change.

I'm not sure what this is referencing. I just ran locally with fresh develop and saw no problems. You'll note that the four CI builds also passed just fine.

I'll review anyway and approve assuming things look good.

rileyjmurray (approved):
Looks great!

PR #868, opened 2026-08-06
Unit tests for MatrixGroup

nkoskelo:
Added unit tests for the MatrixGroup and removed a custom is integer function in favor of using `isintance(x, _numbers.Integral)`.

### DIFF
--- a/pygsti/tools/group.py
+++ b/pygsti/tools/group.py
@@ -11,11 +11,14 @@ Encapsulates a group in terms of matrices and relations
 #***************************************************************************************************
 
 from functools import reduce as _reduce
+import numbers as _numbers
 
 import numpy as _np
 
+from pygsti.tools.legacytools import deprecate as _deprecated_fn
 
-def is_integer(x):
+
+def _isint(x):
     """
     Check if `x` is an integer type.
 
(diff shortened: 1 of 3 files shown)


pygsti/tools/group.py
```diff
@@ -11,14 +11,37 @@
...
+    Parameters
+    ----------
+    x : object
+        Object to test.
+
+    Returns
+    -------
+    bool
+    """
+    return isinstance(x, _numbers.Integral)
```
rileyjmurray: Inline this helper function.

test/unit/tools/test_group.py
```diff
@@ -0,0 +1,125 @@
...
+            res_np_int = group.is_integer(np.int32(5))
+            res_float = group.is_integer(5.5)
+
+            self.assertTrue(res_int)
+            self.assertTrue(res_np_int)
+            self.assertFalse(res_float)
+
+            # Assert at least one warning was emitted and it is pyGSTiDeprecationWarning
+            dep_warnings = [warning for warning in w if issubclass(warning.category, pyGSTiDeprecationWarning)]
+            self.assertGreaterEqual(len(dep_warnings), 1)
```
rileyjmurray: Remove.

rileyjmurray (commented):
Support for exotic integers is not worth our time. I'm okay with the change to `isinstance(x, _numbers.Integral)` because it lets us deprecate the very silly `is_integer` function.

I'm doubtful that construction of a MatrixGroup object takes a measurable amount of time in people's workflows. Do you have evidence that it does? If you don't, then please revert.

Most of the added tests appear to be useful, so if nothing else the PR brings value there.

rileyjmurray (approved):

PR #871, opened 2026-08-07
WIP: Have ExplicitOpModel respect target qubit order for whole-state-space gates

rileyjmurray:
*The following was written by a robot. I've done some preliminary validation. I'm keeping this disclaimer up until I can do a more thorough inspection. Thanks to @omaupin for reporting this bug.*

*Ping @sserita and @coreyostrove. My 4x10 schedule will have me not working tomorrow. I'd welcome independent review by either of you. I think this will warrant a quick turn-around bugfix release.*

On an *n*-qubit `ExplicitOpModel`, an *n*-qubit gate placed on a permuted target ordering was built without applying the permutation. The two-qubit case is the one that bites:

```python
pspec = QubitProcessorSpec(2, ['Gh', 'Gp', 'Gcnot'], availability={'Gcnot': 'all-permutations'})
model = create_explicit_model(pspec, ideal_gate_type='CPTPLND', ideal_spam_type='CPTPLND')
np.linalg.norm(model['Gcnot:0:1'].to_dense() - model['Gcnot:1:0'].to_dense())  # 0.0
```

Both labels return the *same* superoperator — the standard CNOT — so `Gcnot:1:0` silently has its control and target exchanged. Every simulated probability, every fit, and every reported error rate involving such a gate is wrong. Symmetric gates (`Gswap`, `Gcphase`) are unaffected in practice, which is probably a good part of why this survived so long.

The symptom is not specific to `CPTPLND`: `ideal_gate_type='static'` is equally wrong, as are `'full'`, `'full TP'`, `'static unitary'`, and the rest. Nor does it depend on the gate being a CNOT — any order-sensitive gate on a model it exactly fills is affected, including non-standard ones supplied through `nonstd_gate_unitaries`.

`git blame` puts the introducing commit at `5277b1a3f` (August 2021), so this has been present in **every release from v0.9.10 (2021-10-03) through v0.10.2**. A companion write-up of the release archaeology is stored locally on Riley's machine.

## Two independent causes

Both are "no need to embed" fast paths, and both have to go.

**1. `models/modelconstruction.py` — the dimension-only shortcut.** In `_create_explicit_model`:

```python
if gate_unitary.shape[0] == state_space.udim:  # no need to embed!
    embedded_unitary = gate_unitary
else:
    embedded_unitary = _embed_unitary(state_space, inds, gate_unitary)
```

This tests total dimension, but the thing that decides whether embedding is needed is whether `inds` matches the model's own qudit ordering. When the gate fills the model, the dimensions agree for *every* permutation of `inds`, so `_embed_unitary` — which is the code that actually applies the permutation — never runs.

The correct, order-sensitive test was already sitting one scope up: this whole block is the `else` of `if inds is None or inds == tuple(qudit_labels)`. So by the time control reaches the shortcut, `inds` is *known* to be a non-trivial ordering and the shortcut can only ever fire on a case that needs permuting. It is now deleted and `_embed_unitary` is called unconditionally.

**2. `modelmembers/operations/__init__.py` — `create_from_unitary_mx` discarding its unitary.** The `'static standard'` branch:

```python
if typ == 'static standard' and stdname is not None:
    op = StaticStandardOp(stdname, basis, evotype, state_space)
```

`StaticStandardOp` reconstructs the gate from the name via `standard_gatename_unitaries()` and never looks at `unitary_mx`. So a caller that passes a correctly permuted CNOT along with `stdname='Gcnot'` gets the *unpermuted* CNOT back. The branch is now taken only when the name and the unitary agree, and otherwise falls through to the unitary-respecting preferences (`'static clifford'` / `'static unitary'`).

Validating in the callee rather than at the one call site seemed clearly right here: `create_from_unitary_mx` is shared, and silently ignoring an argument is a trap for every future caller. The check skips continuously-parameterized standard gates (`Gzr`, `Gczr`, `Gu3`), whose table entries are functions rather than matrices; incidentally that also removes an uncaught `TypeError` that `StaticStandardOp` would otherwise raise on those names.

It's a trap we'd already walked into, in fact: `test/unit/objects/test_model_updates_after_creation.py` called `create_from_unitary_mx(np.eye(2), "static standard", stdname="Gypi2")` and got a `Gypi2` back, the identity argument being a placeholder for a value it knew would be ignored. That call now passes the actual `Gypi2` unitary. It's the only such caller in the tree.

**The two fixes have to land together.** Fixing only `#1` doesn't produce a subtly wrong model — it produces a hard failure. With the permutation restored, `create_from_unitary_mx` hands the permuted unitary to `StaticStandardOp`, gets the unpermuted standard gate back as the Lindblad postfactor, and then trips the consistency assertion a few lines later:

```
ValueError: Could not create an operator of type(s) CPTPLND from the given unitary op:
{'CPTPLND': "Failure to create Lindblad operation (maybe due the complex log's branch cut?)"}
```

Worth knowing if anyone bisects this.

## What I deliberately did *not* change

`ExplicitOpModel._embed_operation` looks like a third instance of the same mistake, and I originally intended to fix it:

```python
if op_val.state_space == self.state_space and not force:
    return op_val  # if gate operates on full dimension, no need to embed.
```

The user-visible consequence is real: `model.operations[('Gcnot', 1, 0)] = <4x4 superop>` stores the operator verbatim on a 2-qubit model, but embeds (correctly, with the permutation) on a 3-qubit one — identical code, different answer depending on how many spectator qubits exist.

But the shortcut turns out to be load-bearing, and not only for the constructor. `ExplicitOpModel` stores every operation on its full state space, under keys that may name a subset of qudits or a permutation of them. `OrderedMemberDict.__setitem__` — and therefore `update`, `copy`, and `_from_nice_serialization` — re-inserts those already-embedded operations under their original keys. Removing the shortcut makes `create_explicit_model` fail outright (a 1-qubit gate's key has fewer target labels than the model has qudits, so `EmbeddedOp`'s constructor asserts), and would re-apply the permutation on every copy and every save/load round trip for the full-span case.

The ambiguity is genuine and can't be resolved at this layer: given a full-state-space operation and a target-label list, nothing distinguishes "already embedded" from "local, needs embedding." Resolving it properly needs an insertion API that says which one the caller means, threaded through the constructor, `update`, `copy`, and deserialization — a bigger change than this bug fix, and one I'd rather not smuggle in. For now the contract is spelled out in a comment at the call site, and pinned by tests (`AutoEmbedTester`) so a future reader doesn't "fix" it and break every round trip.

## Tests

- `test/unit/construction/test_modelconstruction.py::TargetQubitOrderingTester` — the main regression coverage. Every constructed gate is checked against a hand-written reference embedding (`_embed_unitary_by_hand`, built by summing over matrix units; it shares no code with pyGSTi), swept over `{'auto', 'static', 'static unitary', 'full', 'full TP', 'CPTPLND', 'H+S'}`. `'auto'` is both the default and the only entry that reaches the `StaticStandardOp` branch at all. Covers: 2-qubit models with `Gcnot` and `Gecr`; the non-standard-gate path; the 3-qubit control case that always worked; `Gswap` as the permutation-invariant foil; a forward-simulated probability rather than just `to_dense()`; and copy/serialization round trips.
- `test/unit/modelmembers/test_operation.py::CreateFromUnitaryMxStdnameTester` — unit coverage for the `stdname` check: matching name, mismatched name, mismatched name through the Lindblad postfactor recursion, wrong dimension, unknown name, and a parameterized (callable) name.
- `test/unit/objects/test_explicitmodel.py::AutoEmbedTester` — pins the auto-embed contract discussed above.

Against unfixed `develop` these produce 33 failures; all pass with the fix. Total added runtime is about 4 seconds. (3-qubit `CPTPLND` is excluded from the sweep purely on cost — that one model takes ~40s to build — with `H+S` covering the same Lindblad path.)

## Testing

Full `test/unit` suite, py313:

```
python -m pytest test/unit -q
# 3272 passed, 96 skipped, 2 xfailed, 82 warnings, 212 subtests passed in 296.19s (0:04:56)
```

### DIFF
--- a/pygsti/modelmembers/operations/__init__.py
+++ b/pygsti/modelmembers/operations/__init__.py
@@ -10,6 +10,7 @@ Sub-package holding model operation objects.
 # http://www.apache.org/licenses/LICENSE-2.0 or in the LICENSE file in the root pyGSTi directory.
 #***************************************************************************************************
 
+import functools as _functools
 import numpy as _np
 import warnings as _warnings
 
@@ -42,8 +43,36 @@ from .affineshiftop import AffineShiftOp
 from .cptrop import RootConjOperator, SummedOperator
 from pygsti.baseobjs import statespace as _statespace
 from pygsti.tools import basistools as _bt
+from pygsti.tools import internalgates as _itgs
 from pygsti.tools import optools as _ot
 from pygsti import SpaceT
+
+
+@_functools.lru_cache(maxsize=1)
+def _standard_gatename_unitaries_readonly():
+    """
+    Cached copy of :func:`standard_gatename_unitaries`, which rebuilds its dict (~3ms) on each call.
+
+    Private, and never handed out, so the usual caveat about mutating a cached mutable doesn't apply.
+    """
+    return _itgs.standard_gatename_unitaries()
+
+
+def _standard_name_matches_unitary(stdname, unitary_mx):
+    """
+    Whether `stdname` names a standard gate whose unitary is (numerically) `unitary_mx`.
+
+    `StaticStandardOp` is built from the gate *name* alone and ignores any unitary it's handed,
+    so it may only be used when the two agree.  Returns `False` for unrecognized names and for
+    continuously-parameterized standard gates (e.g. `'Gzr'`), whose entries in
+    :func:`standard_gatename_unitaries` are functions rather than matrices.
+    """
+    if stdname is None:
+        return False
+    std_unitary = _standard_gatename_unitaries_readonly().get(stdname, None)
+    if std_unitary is None or callable(std_unitary):
+        return False
+    return std_unitary.shape == unitary_mx.shape and _np.allclose(std_unitary, unitary_mx)
 
 
 def create_from_unitary_mx(unitary_mx, op_type, basis='pp', stdname=None, evotype='default', state_space=None):
@@ -54,6 +83,14 @@ def create_from_unitary_mx(unitary_mx, op_type, basis='pp', stdname=None, evotyp
     U = unitary_mx
     if state_space is None:
         state_space = _statespace.default_space_for_udim(U.shape[0])
+
+    # A `stdname` that disagrees with `unitary_mx` cannot be honored: the 'static standard' branch below
+    # builds the operation from the name alone and would silently discard `unitary_mx`.  This happens, for
+    # instance, when a standard gate is embedded into a permuted ordering of the model's qudits -- the
+    # name is still "Gcnot" but the unitary is not the standard CNOT.  Drop the name in that case and let
+    # the remaining (unitary-respecting) preferences handle it.
+    if stdname is not None and not _standard_name_matches_unitary(stdname, U):
+        stdname = None
 
     for typ in op_type_preferences:
         try:
--- a/pygsti/models/explicitmodel.py
+++ b/pygsti/models/explicitmodel.py
@@ -232,6 +232,13 @@ class ExplicitOpModel(_mdl.OpModel):
         if self.state_space is None:
             raise ValueError("Must set model state space before adding auto-embedded gates.")
 
+        # Note: a full-state-space `op_val` is taken to be *already embedded*, regardless of what
+        # `op_target_labels` says.  This is a contract, not an oversight: `ExplicitOpModel` stores
+        # every operation on its full state space, so `__setitem__`, `update` and `copy` all re-insert
+        # already-embedded operations under their original (partial or permuted) keys.  Embedding them
+        # a second time would corrupt the model on every copy and serialization round trip.  Callers
+        # that hold a *local* operation acting on `op_target_labels` must therefore either supply it on
+        # the corresponding subspace or pass `force=True`.
         if op_val.state_space == self.state_space and not force:
             return op_val  # if gate operates on full dimension, no need to embed.
 
--- a/test/unit/construction/test_modelconstruction.py
+++ b/test/unit/construction/test_modelconstruction.py
@@ -11,6 +11,8 @@ import pygsti.models.modelconstruction as mc
 import pygsti.modelmembers.operations as op
 import pygsti.modelmembers.states as st
 import pygsti.tools.basistools as bt
+import pygsti.tools.optools as ot
+from pygsti.tools.internalgates import standard_gatename_unitaries as std_unitaries
 from pygsti.processors.processorspec import QubitProcessorSpec as _ProcessorSpec
 from pygsti.baseobjs.errorgenlabel import GlobalElementaryErrorgenLabel as GEEL
 from ..util import BaseCase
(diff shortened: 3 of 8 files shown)


test/unit/construction/test_modelconstruction.py
```diff
@@ -609,6 +611,161 @@ def Isupervec(index):
...
+            for col in range(2 ** k):
+                row_bits = [0] * num_qubits
+                col_bits = [0] * num_qubits
+                for q, b in zip(spectators, spectator_bits):
+                    row_bits[q] = col_bits[q] = b
+                for i, q in enumerate(target_qubits):
+                    row_bits[q] = (row >> (k - 1 - i)) & 1
+                    col_bits[q] = (col >> (k - 1 - i)) & 1
+                embedded[flat_index(row_bits), flat_index(col_bits)] += unitary[row, col]
+    return embedded
```
rileyjmurray, line 642: I confess that correctness of this function isn't obvious to me. I think we'll want to replace with something less efficient yet clearer on the math.

test/unit/objects/test_model_updates_after_creation.py
```diff
@@ -111,8 +112,10 @@ def setup_model_for_test(self, where: Literal["op", "sp", "povm"]):
                                                                         state_space=QubitSpace(1),
                                                                         evotype='densitymx')
 
-            comp = ComposedOp([create_from_unitary_mx(np.eye(2), "static standard",
-                                                      stdname=gate_name),
+            # Note: the unitary and `stdname` have to agree -- create_from_unitary_mx's
+            # 'static standard' branch is only taken when they do.
+            comp = ComposedOp([create_from_unitary_mx(standard_gatename_unitaries()[gate_name],
+                                                      "static standard", stdname=gate_name),
```
rileyjmurray, line 118: This is the place that the PR description mentions as an existing in-repo bug. (Or, arguably a bug since `gate_name='Gypi2'` doesn't make sense with the identity matrix.)

pygsti/modelmembers/operations/__init__.py
```diff
@@ -55,6 +84,14 @@ def create_from_unitary_mx(unitary_mx, op_type, basis='pp', stdname=None, evotyp
...
         state_space = _statespace.default_space_for_udim(U.shape[0])
 
+    # A `stdname` that disagrees with `unitary_mx` cannot be honored: the 'static standard' branch below
+    # builds the operation from the name alone and would silently discard `unitary_mx`.  This happens, for
+    # instance, when a standard gate is embedded into a permuted ordering of the model's qudits -- the
+    # name is still "Gcnot" but the unitary is not the standard CNOT.  Drop the name in that case and let
+    # the remaining (unitary-respecting) preferences handle it.
+    if stdname is not None and not _standard_name_matches_unitary(stdname, U):
+        stdname = None
+
```
rileyjmurray, line 94: We'll probably want to be loud about this.

pygsti/models/explicitmodel.py
```diff
@@ -232,6 +232,13 @@ def _embed_operation(self, op_target_labels, op_val, force=False):
         if self.state_space is None:
             raise ValueError("Must set model state space before adding auto-embedded gates.")
 
+        # Note: a full-state-space `op_val` is taken to be *already embedded*, regardless of what
+        # `op_target_labels` says.  This is a contract, not an oversight: `ExplicitOpModel` stores
+        # every operation on its full state space, so `__setitem__`, `update` and `copy` all re-insert
+        # already-embedded operations under their original (partial or permuted) keys.  Embedding them
+        # a second time would corrupt the model on every copy and serialization round trip.  Callers
+        # that hold a *local* operation acting on `op_target_labels` must therefore either supply it on
+        # the corresponding subspace or pass `force=True`.
```
rileyjmurray, line 241: This can probably be struck before merge.

coreyostrove: Seconded

coreyostrove (commented):
I have reviewed the agents conclusions regarding the bugs and have also investigated this independently and agree that these two bugs are indeed bugs. As for how no one notices until now? 🤷. I suspect a combination of the relative rarity of manually constructing an explicit model (instead of using a modelpack), compounded with the additional rarity of using CNOTs when doing so. 

I'd like to request the following, though: Can we split the proposed changes here into two PRs? I am onboard with the fixes related to the usage of `_embed_unitary` for fixing the more general of the index permutation bugs. (I do still need to read the proposed tests more carefully). I would like to advocate for a very different approach to fixing the bug on the 'static standard' branch than what the agent has proposed, and don't want a fix to the first bug to be held hostage by that debate (should there be one).

PR #873, opened 2026-08-10
Fix nongauge parameter override in initial greedy germ-set test

liuyichao82:
Hi PyGSTi team,

The initial germ-set completeness check ignored `num_nongauge_params`, which could terminate germ selection early. This fix forwards the corresponding gauge-parameter count to the check.

I hope this fix is helpful.


rileyjmurray (approved):
Looks good to me!

PR #874, opened 2026-08-10
Fix crash when skipping the initial germ test

liuyichao82:
Skipping the initial germ test set `initN` to `None` and caused a `TypeError`. This change sets `initN` to `1` so germ selection can continue.


rileyjmurray (approved):
Looks good. We can merge when CI passes.

PR #876, opened 2026-08-11
Fix two UnboundLocalError bugs in errgenproptools.py's A-type-generator code

nonohuff:
### Summary

Two independent, pre-existing typo bugs in `pygsti/tools/errgenproptools.py` raise an `UnboundLocalError` for code paths involving 'A' (active) type error generators:

```python
from pygsti.tools import errgenproptools as _eprop
from pygsti.errorgenpropagation.localstimerrorgen import LocalStimErrorgenLabel as _LSE
import stim

tableau = stim.PauliString('XI').to_tableau()
errgen = _LSE('A', [stim.PauliString('X'), stim.PauliString('Y')])
_eprop.slow_bulk_alpha([errgen], tableau, ['00'])
# UnboundLocalError: cannot access local variable 'sensitivity'
```

### Root cause

1. `slow_bulk_alpha`'s "A1" branch (weight-1 'A'-type generators whose two Paulis anticommute -- i.e. *every* weight-1 A-type generator, since any two distinct single-qubit Paulis anticommute) assigned its result to `sensitivities_by_bitstring` -- the function's entire return array -- instead of the per-iteration local `sensitivity`, which is read on the very next line.
2. `error_generator_commutator`'s 'S'-'A' branch referenced an undefined local `ptup` (should be `ptup1`) in two places.

Both are plain variable-name typos, unrelated to each other, both specific to 'A'-type error-generator code paths:

- Bug 1 is masked in most environments because `bulk_alpha` is aliased to the Cython `fast_bulk_alpha` whenever the `fasterrgencalc` extension is built (the common case) -- `slow_bulk_alpha` is only reached as the pure-Python fallback.
- Bug 2 is reachable only for a degenerate `A(Q,Q)` label (two equal Paulis), which nothing in `LocalStimErrorgenLabel`'s constructor currently forbids, even though a well-formed `A_{P,Q}` generator requires P != Q (see "A Taxonomy of Small Errors", Sec. V.D).

### The fix

Two one-token variable-name corrections -- `ptup` -> `ptup1` (x2) and `sensitivities_by_bitstring` -> `sensitivity` -- with no behavioral change beyond these code paths no longer crashing.

### Tests

New tests in `test/unit/tools/test_errgenproptools.py`:

| test | asserts |
|---|---|
| `test_slow_bulk_alpha_weight_1_active_generators` | calls `slow_bulk_alpha` directly (bypassing the Cython alias, so it exercises the actual pure-Python fallback regardless of whether Cython extensions are built) for all 4 bitstrings of a weight-1 A-type generator, matching the single-label `alpha`; also sweeps 20 randomly-sampled 3-qubit A-type generators against a random Clifford tableau |
| `test_errorgen_commutator_S_A_degenerate_case` | constructs the exact degenerate `A(P,P)` label scenario (no longer raises `UnboundLocalError`), plus a well-formed-pair regression check against `error_generator_commutator_numerical` |

Both were confirmed to fail with the exact `UnboundLocalError`s described above against the pre-fix code.

### Verification

```
pytest -q test/unit/tools/test_errgenproptools.py
# 27 passed in 177.05s

pytest -q test/unit/objects/test_errorgenpropagation.py
# 25 passed in 72.14s

pytest -q test/unit/tools/test_lindbladtools.py
# 13 passed in 0.30s

flake8 pygsti/tools/errgenproptools.py test/unit/tools/test_errgenproptools.py --config=.flake8-critical
# clean
```

### Context

Found while adding 'C'/'A' error-generator support to `pygsti.extras.ml` (#835) -- not on that subpackage's own call paths (it only ever calls the singular `errgenproptools.alpha`, never `bulk_alpha`/`slow_bulk_alpha`/`error_generator_commutator`), but hit while cross-validating `alpha()` against `bulk_alpha()` and exercising the wider A-type-generator surface during that work. Per @rileyjmurray's request on #835 ("please split them off into a dedicated PR that we can land before this one"), this is extracted the same way `0bdbb2830` was extracted to #865: a small, independently-reviewable core fix rather than incidental scope inside a large feature PR. #835 has a follow-up commit removing this change from its own diff.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,7 @@
 * Fixed the `is_trace_preserving` utility function for bases whose first element is the identity. The utility function was only used in report generation and the bug did not affect the correctness of its caller.
 * Fixed a crash when the initial germ test is skipped.
 * Fixed germ selection ignoring the user-specified number of non-gauge parameters.
-* Fixed two `UnboundLocalError` bugs in `pygsti.tools.errgenproptools` affecting 'A'-type (active) error generators: `slow_bulk_alpha` assigned its weight-1 anticommuting-Pauli result to the whole return array instead of the per-iteration local (masked whenever the Cython `fasterrgencalc` extension is built, since `bulk_alpha` then aliases the Cython implementation), and `error_generator_commutator`'s 'S'-'A' branch referenced an undefined `ptup` instead of `ptup1`. (#PLACEHOLDER) <!-- TODO: fill in actual PR number before merge -->
+* Fixed two `UnboundLocalError` bugs in `pygsti.tools.errgenproptools` affecting 'A'-type (active) error generators: `slow_bulk_alpha` assigned its weight-1 anticommuting-Pauli result to the whole return array instead of the per-iteration local (masked whenever the Cython `fasterrgencalc` extension is built, since `bulk_alpha` then aliases the Cython implementation), and `error_generator_commutator`'s 'S'-'A' branch referenced an undefined `ptup` instead of `ptup1`. (#876)
 
 ## [0.10.2] - 2026-07-30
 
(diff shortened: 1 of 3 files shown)


rileyjmurray (approved):
We can merge once CI tests pass.

PR #880, opened 2026-08-12
Jamiolkowski tests

nkoskelo:
Improve the unit test coverage for `pygsti/tools/jamiolkowski.py`.

### DIFF
--- a/test/unit/tools/test_jamiolkowski.py
+++ b/test/unit/tools/test_jamiolkowski.py
@@ -5,6 +5,8 @@ from pygsti.models.modelconstruction import create_operation
 from pygsti.modelpacks.legacy import std1Q_XYI as std1Q
 from pygsti.baseobjs import statespace
 from pygsti.baseobjs import Basis
+from pygsti.baseobjs.basis import BuiltinBasis, TensorProdBasis
+from pygsti.tools.optools import unitary_to_superop
 from pygsti.tools import jamiolkowski as j
 from ..util import BaseCase, needs_cvxpy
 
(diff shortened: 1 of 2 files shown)


test/unit/tools/test_jamiolkowski.py
```diff
@@ -260,3 +292,79 @@ def test_is_cvxpy_expression_on_plain_arrays(self):
...
+        R_qutrit[2, 1] = np.sin(theta)
+        U = np.kron(X_qubit, R_qutrit)
+
+        # Convert the unitary to a superoperator in our tensor-product basis
+        self.superop = unitary_to_superop(U, self.tp)
+
+    def test_simple_equivalent_unavailable(self):
+        # Precondition check: the leakage tensor-product basis does not have a same-name
+        # builtin equivalent, which raises AssertionError during create_simple_equivalent().
+        self.assertRaises(AssertionError, self.tp.create_simple_equivalent)
```
rileyjmurray: I'm confused. This isn't invoking a function. Shouldn't this be an attribute error?

test/unit/tools/test_jamiolkowski.py
```diff
@@ -260,3 +292,79 @@ def test_is_cvxpy_expression_on_plain_arrays(self):
...
+class JamiolkowskiLeakageBasisTester(BaseCase):
+    """
+    Tests `jamiolkowski_iso` with a leakage tensor-product basis (e.g., pp ⊗ l2p1).
+    This exercises the fallback branch in `jamiolkowski_iso` when `create_simple_equivalent`
+    fails for bases with no same-name builtin equivalent.
+    """
+
+    def setUp(self):
+        # 36-dimensional qubit x qutrit tensor product basis
+        self.tp = TensorProdBasis((BuiltinBasis('pp', 4), BuiltinBasis('l2p1', 9)))
```
rileyjmurray: Please rename `self.tp` to `self.basis`.

rileyjmurray (changes requested):
I only have one hard request, plus one comment that probably warrants a change.

test/unit/tools/test_jamiolkowski.py
nkoskelo: I updated the code to be more readable.

`self.assertRaises(AssertionError, self.tp.create_simple_equivalent)` is the same as
```
with self.assertRaises(AssertionError):
     self.tp.create_simple_equivalent()
```
during the execution we check if there is a basis in the dictionary of builtin basis functions which has the name of something similar and raise an `AssertionError` if we cannot find one.

rileyjmurray: Ah! I didn't know about that syntax. thanks for clarrifying.

rileyjmurray (approved):

PR #881, opened 2026-08-13
Unify graphtools

### LINKED ISSUE #866: Unify pygsti.tools.graphcoloring and extras/ml/graphtools.py into a pygsti.tools.graphs subpackage
### Proposal

Rename the `pygsti.tools.graphcoloring` subpackage to `pygsti.tools.graphs`, and move the graph utilities currently living in `pygsti/extras/ml/graphtools.py` into it.

This issue exists to record @rileyjmurray's suggestion from review of #835:

> The develop branch now has a `pygsti.tools.graphcoloring` subpackage. In view of your graphtools.py module I suggest we rename the subpackage to `pygsti.tools.graphs` and have it include both graphcoloring and the contents of your new graphtools.py. I don't think we need to do that reorganization as part of this PR. Instead I just ask that you open a github issue pointing out the benefits of a `pygsti.tools.graphs` subpackage, linking to this comment, and saying we should restructure before the release of pygsti 0.11.x.

— https://github.com/sandialabs/pyGSTi/pull/835#discussion_r3714904755

### Current state

pyGSTi has ended up with graph code in two unrelated places, neither of which is a general home for it:

**`pygsti/tools/graphcoloring/`** (on `develop`) — six modules, ~52 KB, exposing a narrow public API:
```python
from ._dispatch import switchboard_find_edge_coloring
from ._common import check_valid_edge_coloring
```
with `_sinnamon.py`, `_vizing.py`, and `_topology.py` behind it. The package name describes one algorithm family, not a domain.

**`pygsti/extras/ml/graphtools.py`** (added in #835) — ~650 lines of graph-library-agnostic qubit-connectivity handling, none of it QPANN-specific:

| function | purpose |
|---|---|
| `qubit_graph_to_networkx` | coerce a `networkx`/`igraph`/`graph-tool` graph, `QubitGraph`, `QubitProcessorSpec`, adjacency/Laplacian matrix, or edge list to a canonical `networkx.Graph` |
| `qubit_graph_from_edges` | build a graph from an explicit edge list |
| `qubit_graph_adjacency_matrix` / `qubit_graph_laplacian` | matrix representations |
| `within_hops_matrix` / `qubits_within_hops` | graph-distance neighbourhood queries |

### Why unify

1. **`graphtools.py` is in the wrong place today.** It has no dependency on `pygsti.extras.ml` and nothing to do with neural networks — it is generic qubit-connectivity plumbing. But because `pygsti/extras/ml/__init__.py` imports `qpanns` (which imports TensorFlow at module scope), *importing these pure-numpy/networkx helpers currently requires TensorFlow to be installed*. That is a bad deal for any other part of pyGSTi that wants them.

2. **They solve adjacent halves of the same problem.** `graphcoloring` consumes a graph and colors its edges; `graphtools` is precisely the layer that turns the many things a user might call "a qubit graph" into a canonical graph object. Edge-coloring a device's connectivity — e.g. for simultaneous-gate scheduling — wants both.

3. **`graphcoloring` already has its own coercion layer.** `_topology.py` necessarily has some notion of vertices/edges/degree. Unifying lets one canonical representation serve both, instead of two independent conventions drifting apart.

4. **The name generalizes.** `pygsti.tools.graphs` accommodates future graph utilities (matching, connected components, layout) without another top-level module or another awkwardly-named package.

5. **It is cheap now and expensive later.** `graphcoloring` is new on `develop` and `graphtools` is not yet released at all, so **neither is in a tagged release**. Renaming now costs nothing in deprecation shims. Once 0.11.x ships, moving either becomes a public API break requiring a deprecation cycle.

### Suggested shape

```
pygsti/tools/graphs/
├── __init__.py        # switchboard_find_edge_coloring, check_valid_edge_coloring,
│                      # qubit_graph_to_networkx, qubit_graph_laplacian, qubits_within_hops, ...
├── _coloring/         # former graphcoloring internals (_sinnamon, _vizing, _topology, ...)
└── _connectivity.py   # former extras/ml/graphtools.py
```

with `pygsti.extras.ml.graphtools` re-exporting from the new location for the lifetime of the current branch, and `pygsti/extras/ml/{errgentools,snippers}.py` switched over to the new import path.

### Scope

Deliberately **not** part of #835 — that PR is already large and this touches code it does not own. Per the review comment above, this should land **before the release of pyGSTi 0.11.x**, while both modules are still unreleased and no deprecation shim is needed.

nkoskelo:
Fixes #866 

----------------

The following was written by a robot under Riley's direction.

----------------

This PR creates a `pygsti.tools.graphs` subpackage and moves pyGSTi's graph code into it.

That code was in two places. The edge-coloring package `pygsti.tools.graphcoloring` was named for one algorithm family rather than a domain. The qubit-connectivity helpers that #835 added as `pygsti.extras.ml.graphtools` had nothing to do with neural networks, but couldn't be imported without TensorFlow because `pygsti.extras.ml.__init__` pulls in Keras at module scope. The coloring package is now `pygsti.tools.graphs.coloring`, the connectivity helpers (`qubit_graph_to_networkx` and friends) are exported from `pygsti.tools.graphs` itself, and the vocabulary they share (`Vertex`, `Edge`, `canonical_edges`, ...) lives in one place. Neither module has appeared in a public release, so there are no compatibility shims and nothing is deprecated.

**New functionality.** `connected_supports(qubit_graph, max_size, hops)` enumerates every set of at most `max_size` qubits that is connected under the within-`hops` relation. It grows sets outward from their lowest-numbered vertex rather than testing all `C(n, w)` candidates, so its cost scales with the number of supports it returns. On a 10x10 grid with `k=4` that is 2,137 supports instead of 4,087,975 candidates. The `up_to_weight_k_*_from_qubit_graph` functions in `errgentools` now use it in place of a brute-force scan. The enumeration order matches the old scan, so error-generator indices, and hence trained QPANN parameters, are unchanged. Nothing else in `errgentools.py` changes. `check_valid_edge_coloring` now says which color, vertex and edges make a coloring invalid.

**Two things found along the way.** Review asked which of the repo's other networkx users belong in the new package. `random_connected_subgraph` in `pygsti.circuits.subcircuit_selection` is the one generic graph routine, so it moves into `pygsti.tools.graphs` (accepting any graph representation the package understands) with a re-export left behind. The tree-partitioning helpers in `pygsti.layouts.prefixtable` could follow later: they're generic rooted-tree algorithms, but about 600 lines, tied to `PrefixTable`'s node attributes, and only tested through `find_splitting_new`. That belongs in its own issue. The lanes computation in `split_circuits_into_lanes` and the crosstalk plots have nothing reusable.

Writing characterization tests for `QubitGraph` (new `test/unit/objects/test_qubitgraph.py`) turned up four bugs, all fixed here:

- The `_dirty` flag was never cleared, so every shortest-path query re-ran Floyd-Warshall.
- Floyd-Warshall ran weighted, so on a directed graph with direction names the stored direction indices became edge lengths and `shortest_path_distance` was wrong.
- `is_connected(q, q)` was False.
- The undirected branch of `is_connected_subgraph` flood-filled from node 0 rather than from the subset, which also made `connected_combos` undercount.

`subgraph` now passes the parent's direction names through as well.

**Tests.** `test/unit/tools/graphs/` covers the coloring algorithms, the shared helpers, the coercion layer, `connected_supports` against brute force, and the moved `random_connected_subgraph`. `test/performance/test_coloring_scaling.py` is the renamed scaling suite. `test_ml.py` is updated for the new import path and passes with the `ml` extra installed.


rileyjmurray (commented):
Looks okay at a quick glance. Please write a proper PR description and ping me when ready.

rileyjmurray:
The following is from Claude.

----------------------------------

Following up on the networkx survey (details in the PR description). Outside `pygsti.tools.graphs` itself, four modules use networkx:

- `circuits/subcircuit_selection.py`: `random_connected_subgraph` is a generic graph routine with no circuit dependencies. Moved into `pygsti.tools.graphs` in this PR, with a re-export left behind.
- `layouts/prefixtable.py`: the Kundu-Misra tree partitioner and its helpers (roughly 600 lines) are generic rooted-tree algorithms and would fit a `graphs.trees` module, but they are tied to `PrefixTable`'s node attributes and only tested through `find_splitting_new`. Better as its own issue than as more scope here.
- `circuits/split_circuits_into_lanes.py`: a throwaway graph for connected components. Nothing to extract.
- `extras/crosstalk/objects.py`: matplotlib drawing only.

rileyjmurray (commented):
We should resolve things marked as `KNOWN BUG` in test/unit/objects/test_qubitgraph.py.

We should consider less invasive changes to pygsti/extras/ml/errgentools.py.

rileyjmurray (approved):

PR #882, opened 2026-08-16
Bugfix Explicit Model Target Qubit Ordering

coreyostrove:
This PR is a follow-on and partial replacement to #871. On #871 I had made a request that we split off the two embedding logic related bugs identified in order to address them separately, and this PR splits off the changes for the embedding logic in `create_explicit_model` when a unitary spans the target state space, but has a different ordering of its label requiring embedding. 

### Changes:

- Bugfix for embedding logic in create_explicit_model when a unitary spans the target state space, but has a different ordering of its label requiring embedding.
- New unit tests that explicit validation and regression tests for correct behavior when using both symmetric and asymmetric two-qubit unitaries.
- Relative to #871, the `embed_unitary_by_hand` helper function in the tests has had its logic simplified for readability (as flagged by @rileyjmurray). Also cleaned up a few comments that were either long-winded, or that were left over from debugging and would not make sense outside of the context of this particular PR.

### Comments:

- A few of the embedding tests that interact with the current bug with StaticStandardOp have been marked as xfail. These will be updated once we have a permanent fix for that one.
- The changes and new tests ported here from 871 were originally authored by agentic tools. I have manually reviewed the changes and new test for correctness and sensibility. 

I am working on a proposed set of changes as an alternative to the other half of #871, and will open up a PR as soon as that is ready.


test/unit/construction/test_modelconstruction.py
rileyjmurray, line 1: There are two xfail statements of the following form in this file. I don't know what it means to "require standard-op resolution".
```python
                    if gate_type in ('auto', 'CPTPLND', 'H+S'):
                        pytest.xfail(f"gate_type={gate_type!r} for standard gates requires standard-op resolution")
```

rileyjmurray (commented):
Overall, looks great. I have one question about `test_modelconstruction.py`.

test/unit/construction/test_modelconstruction.py
coreyostrove: With just the changes on this branch these two tests xfail because of the bug related to gate names overriding the specified unitaries. I.e. because of the second bug identified on #871. #885 fixes that issue, and also removes the xfail for these cases. Resolution here just means resolution of the bug.

rileyjmurray (approved):

Issue #761, opened 2026-06-10
`Circuit.str` setter validation hole: a truncated string rep is silently accepted


rileyjmurray:
This issue brought to you by a robot, written from the robot's perspective.

------------------------------------------------

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5). Calibration up front: severity is moderate — `done_editing`
regenerates the cached string (since #692), so the inconsistency is observable only while the
circuit is editable. It is a validation hole, not persistent corruption.

### Repro

```python
from pygsti.circuits import Circuit
c = Circuit("GxGy@(0)", editable=True)
c.str = "GxGz@(0)"   # same-length mismatch: correctly rejected (ValueError)

c2 = Circuit("GxGy@(0)", editable=True)
c2.str = "Gx@(0)"    # TRUNCATED string: silently accepted
print(c2.str)        # 'Gx@(0)'  — claims 1 layer
print(len(c2))       # 2         — actually 2 layers
```

### Root cause

The setter's consistency check (circuit.py:854) iterates with `zip`:

```python
for my_layer, chk_lbl in zip(self._labels, chk_labels):
```

`zip` stops at the shorter sequence, so a string rep with *fewer* layers than the circuit
passes validation vacuously for the missing tail. While the circuit remains editable, `.str`
returns the assigned (wrong) string; any code that reads `.str` before `done_editing` (e.g.
to display or serialize an in-progress circuit) sees a circuit other than the one represented.

### Suggested fix

Length check before the zip:

```python
if len(chk_labels) != len(self._labels):
    raise ValueError(...)
```

One more line and the hole closes. A characterization test for the setter (accept/reject
matrix including truncation, extension, and metadata mismatches) is planned in an upcoming
test-infrastructure PR.

rileyjmurray:
Resolved by #887.

Issue #759, opened 2026-06-10
Editable-circuit slices and copies share mutable sublist state with the parent (latent aliasing hazard)


rileyjmurray:
This issue brought to you by a robot, written from the robot's perspective.

------------------------------------------------

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5). Calibration up front: I verified the *aliasing* by execution, but I
did **not** find a public mutator sequence that currently corrupts a sibling circuit — today's
in-place editors mostly *replace* layer lists rather than mutate them. This is a landmine
report, not an active-corruption report.

### Verified aliasing

```python
from pygsti.circuits import Circuit
p = Circuit("[Gx:0Gy:1][Gz:0]", line_labels=(0,1), editable=True)
s = p[0:2]
print(s._labels[0] is p._labels[0])   # True — same list object
```

Layer-range extraction of an editable circuit returns the parent's per-layer lists by
reference: `extract_labels` takes `self._labels[layers]` (a shallow list slice) at
circuit.py:1323 and hands it to `_fastinit`, which stores it as-is.

Similarly, `copy(editable=True)` of an *editable* circuit copies only one level
(`[sublist.copy() for sublist in self._labels]`, circuit.py:1129): the inner nested lists
that exist for compound labels (per `_label_to_nested_lists_of_simple_labels`,
circuit.py:118–127) remain shared between the copy and the original.

### Why it matters

- `_append_layer_component` (circuit.py:1232–1239) mutates a layer list **in place**
  (`self._labels[layer_index].append(...)`); any current or future code path that reaches it
  on an aliased circuit silently edits the sibling too.
- External code already mutates `_labels` raw: `tools/locking.py:79` does
  `c._labels.append(M)` — the pattern exists in the wild.
- The invariant "editing one circuit never changes another" is exactly the kind of property a
  reader assumes without checking; the aliasing makes it false in a way that depends on which
  mutator is used.

### Suggested fix

Deep-copy (two levels) the label lists on editable slice extraction and editable→editable
copies. Cost is bounded by circuit size and only on editable paths, which are not the hot
tier. A characterization test asserting slice/copy independence under each public in-place
mutator is planned regardless of when the fix lands.

rileyjmurray:
Resolved by #887.

Issue #758, opened 2026-06-10
`Circuit.__add__` drops compilable-layer indices but leaks `~` markers into the cached string — `Circuit(c.str) != c`


rileyjmurray:
This issue brought to you by a robot, written from the robot's perspective.

------------------------------------------------

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5).

### Repro

```python
from pygsti.circuits import Circuit
a = Circuit("Gx~Gy@(0)")     # layer 0 marked compilable
b = Circuit("Gz@(0)")
s = a + b
print(s.str)                          # 'Gx~GyGz@(0)'  — '~' marker present
print(s.compilable_layer_indices)     # ()             — metadata gone
print(Circuit(s.str) == s)            # False          — round-trip broken
```

### Root cause

`__add__` (circuit.py:960–986) composes the result's string rep by concatenating the operands'
*cached* string reps — including any `~`/`|` compilable-layer markers — but hard-drops
`compilable_layer_indices` (and `occurrence`, deliberately, at :983) from the resulting
object. The object and its own string rep now disagree: re-parsing the string recreates the
compilable indices that the object claims not to have, so `Circuit(c.str) != c`.

This breaks the parse(str) round-trip invariant that the text serialization formats rely on
(`CircuitList` nice-serialization, text circuit lists, dataset.txt all store `circuit.str` and
re-parse on load) — the same invariant #738 recently hardened for line labels (#495). A summed
circuit written to any of those formats comes back as a *different* circuit.

Adjacent known issue: #472 (line-label reordering during `+` breaks mirror-circuit
conventions) — worth fixing in the same area sweep.

### Suggested fix

Pick one of the two consistent behaviors:

1. **Propagate**: carry compilable indices through concatenation (second operand's indices
   shifted by `len(first)`), keeping the markers in the composed string; or
2. **Drop cleanly**: strip `~`/`|` markers from the composed string rep when dropping the
   metadata.

Option 1 seems more useful (compilability of a layer is a per-layer property that
concatenation doesn't invalidate). Either way, `parse(c.str) == c` must hold afterwards; a
generic property test for that invariant is planned as part of an upcoming
characterization-test PR.

rileyjmurray:
Resolved by #887.

Issue #746, opened 2026-06-01
Clearing the labels of a Circuit can error when it should not


nkoskelo:
On develop

```
from pygsti.circuits import Circuit

a = Circuit('Gy@(0,1)', editable=True)
print(a._line_labels)
print(a)

# Expects to error below.
a._clear_labels((0,), [0])
```

The last line should make this equivalent to the circuit.

```
b = Circuit([('Gy', 1)], line_labels=[0,1], editable=True)
print(b)
```

However, we throw an error that `Gy` straddles the line and so we cannot clear the location.
"ValueError: Cannot operate on a block that is straddled by Gy!"

rileyjmurray:
Resolved by #887. The resolution ended up just being to raise a more informative error message. (The function has a kwarg you can set which handles the error-producing scenario in a canonical way.)

PR #887, opened 2026-08-18
Circuit bugs: canonicalization, metadata on `+`, editable aliasing, `.str` validation

### LINKED ISSUE #757: Circuit equality and hash depend on construction path:  `_fastinit` skips the inner-layer sort that `__init__`/`done_editing` apply
This issue brought to you by a robot, written from the robot's perspective.

Found during a systematic audit of `pygsti/circuits/circuit.py` (executed repros at the head
of `develop`, 47b3dcae5).

Since #692 (fixing #684), logically equivalent circuits are canonicalized by sorting the
simple labels inside each layer: `__init__` applies `_sort_layer_labels` at lines 552–556, and
`done_editing` sorts at line 5050. But `Circuit._fastinit` (lines 575–580) performs **no
sorting and no validation** — it trusts callers to pass already-canonical label tuples. That
contract is undocumented, and at least one production caller violates it.

### Repro 1: direct

```python

c_fast = Circuit._fastinit((Label((('Gy',1),('Gx',0))),), (0,1), False)
c_init = Circuit([[('Gy',1),('Gx',0)]], line_labels=(0,1))
print(c_fast.tup)  # (Label((Label(('Gy', 1)), Label(('Gx', 0)))), '@', 0, 1)
print(c_init.tup)  # (Label((Label(('Gx', 0)), Label(('Gy', 1)))), '@', 0, 1)
print(c_fast == c_init, hash(c_fast) == hash(c_init))   # False False
```

### Repro 2: the production path — circuits parsed from text differ from identical circuits built in code

`io/stdinput.py` constructs **every** circuit parsed from strings/files via `_fastinit`
(stdinput.py:129 and :329), passing the parser's output in source order:

```python
from pygsti.io import stdinput

sip = stdinput.StdInputParser()
c_parsed = sip.parse_circuit("[Gy:1Gx:0]@(0,1)", create_subcircuits=False)
c_built  = Circuit("[Gy:1Gx:0]@(0,1)")
print(c_parsed == c_built)                    # False
print(hash(c_parsed) == hash(c_built))        # False
```

The same text yields two unequal, differently-hashing Circuits depending on whether it came
through `StdInputParser` (dataset/text-file loading) or the `Circuit` constructor. Since
Circuits key `DataSet.cirIndex` and the layout caches, a dataset loaded from disk can silently
fail to match circuits constructed programmatically — a wrong-lookup failure mode, not an
error.

(Note `Circuit.__init__` with a *string* argument does sort, because the constructor applies
`_sort_layer_labels` after parsing; only the `_fastinit` parser path skips it.)

### Suggested fix

Either (a) canonicalize in `_fastinit` when `not editable` (needs benchmarking — `_fastinit`
is the 0.6µs hot construction tier, ~29x faster than `__init__`, and #692's review set the
precedent of demanding before/after timing evidence here), or (b) have both parser
implementations emit layer labels pre-sorted (`fastcircuitparser.pyx` and
`slowcircuitparser.py` must stay in lockstep), or (c) document and enforce the already-canonical
precondition with a debug-mode assertion. Option (b) keeps the hot path free.

Related: #684, #692, #472.

rileyjmurray:
What follows is written by a robot.

-------------------------
This PR fixes #757, #758, #759 and #761, and proposes closing #746 as not-a-bug with a better error message.

Prior to these changes, all four fixed bugs were pinned by passing tests in `test/unit/circuits/test_circuit_known_bugs.py`. Every pin is flipped here, and each one moves to the file that describes the behavior it now documents rather than staying in a file about bugs.

One commit per issue, in increasing order of risk. Most of them change something a user could notice; *Reviewer attention* below collects those in one place, worst first.

### #757 — circuits parsed from text didn't match circuits built in code

The one I'd read most carefully. Since #692 (fixing #684) equivalent circuits are canonicalized by sorting the simple labels inside each layer. `__init__` and `done_editing` do it; `Circuit._fastinit` doesn't — it goes straight to `_bare_init` with no sort and no validation, on an undocumented precondition that callers pass canonical tuples. `io/stdinput.py` violated that at lines 129 and 329, which are the only `_fastinit` callers outside `circuit.py`, so every circuit parsed from a string or file was built in source order:

```python
sip.parse_circuit("[Gy:1Gx:0]@(0,1)", create_subcircuits=False) == Circuit("[Gy:1Gx:0]@(0,1)")  # False
```

Circuits key `DataSet.cirIndex` and the layout caches, so a dataset loaded from disk could silently fail to match circuits built in code. Wrong lookup, not an error.

I fixed this in the parsers rather than in `_fastinit`, per the issue's option (b). `_fastinit` is the ~0.48 µs hot construction tier, roughly 300× cheaper than `__init__` at ~140 µs, and putting a per-call sort there would tax every internal construction to fix two callers. Both `fastcircuitparser.pyx` and `slowcircuitparser.py` now call `Label.with_sorted_inner_labels()` when they build a bracketed layer, so the canonical order is defined in exactly one place and the two parsers can't drift — splitting behavior by whether the Cython extension built would be worse than the bug.

Sorting unconditionally turned out to cost real time. For an already-canonical layer `with_sorted_inner_labels` still rebuilds the label (~2.1 µs per two-component layer), and `__init__` then re-sorts anyway because `expand_subcircuits` drops the `is_sorted` flag. That doubled `parse_via_stdinput` from 28 µs to 56 µs. Both parsers now skip the rebuild when a layer's components are already strictly increasing in sslbls — an O(n) check, true of every circuit string pyGSTi writes itself. Strictly-increasing rather than non-decreasing is deliberate: duplicate sslbls must still reach the sorter so it can raise.

### #758 — `+` dropped compilable indices but kept the `~` in the string

`__add__` composed the result's string rep by concatenating the operands' cached strings, markers and all, while hard-dropping `compilable_layer_indices`. The object and its own string rep disagreed, so `Circuit(c.str) != c` — the round-trip invariant that `CircuitList` nice-serialization, text circuit lists and `dataset.txt` all depend on, and the same one #738 hardened for line labels (#495).

Compilability is a per-layer property that concatenation doesn't invalidate, so I went with the issue's suggested option 1 and propagate: the right operand's indices shift past the left operand's layers. `__radd__` and the Label-tuple form of `+` do the same, so the operator is consistent in both directions.

The string rep needed more care than just keeping the concatenation. `_op_seq_to_str` marks whichever set is smaller — `~` on the compilable layers or `|` on the uncompilable ones — and each operand made that choice for its own layer count, so splicing two rendered strings can produce one carrying *both* markers, which the parser rejects outright with "contains both barrier and compilable layer joining". When the combined index set is non-empty the result's string rep is therefore left unset and regenerated from the object on first access, which is canonical by construction. When it's empty — nearly always — the cached strings are still concatenated verbatim, so user-authored spellings survive and the hot path is untouched.

### The writer-side sibling of #758: no marker at all when every layer is compilable

Found while fixing #758, and folded in rather than filed separately: it's the same invariant in the same function, and leaving it out meant #758's round-trip fix landed with an asterisk.

When every layer is compilable the uncompilable set is *empty*, so the smaller-set rule above picks `|` and marks nothing — and a reader can't tell that from a circuit with no compilable layers at all:

```python
c = Circuit([('Gx', 0)], line_labels=(0,), compilable_layer_indices=(0,))
c.str                 # 'Gx:0@(0)'  — no marker
Circuit(c.str) == c   # False
```

Single-layer circuits are the sharpest case, since their only non-empty index set is the full one: before this, such a circuit could never round-trip. The fix prefers `~` whenever the uncompilable set is empty. That condition is reached only when the uncompilable set is empty and the circuit is non-empty, so every other input takes the branch it took before and every string that already carried a marker is byte-identical — by construction rather than by testing.

Worth knowing when reading this, because it's what makes the bug easy to miss: the loss is writer-only. `Circuit('Gz~') == Circuit(Circuit('Gz~').str)` is `True` and always was, because `.str` hands back the cached input string and the renderer never runs. It's the paths that *generate* the string — `done_editing`, `+`, plain construction — that dropped the markers.

Blast radius is small: `io/stdinput.py` is the only thing in the package that sets `compilable_layer_indices`, and only ever reflecting the input text, so nothing inside pyGSTi produces an all-compilable circuit. The change is reachable from user code that sets the indices explicitly, and from `+` on parsed circuits.

### #759 — editable slices and copies shared mutable state with the parent

The issue calls this a landmine and says no public mutator sequence corrupts a sibling. One does, and it's an ordinary thing to write:

```python
p = Circuit("[Gx:0Gy:1][Gz:0][Gy:1]", line_labels=(0,1), editable=True)
s = p[0:3]
s.compress_depth_inplace()
# p: [Gx:0Gy:1]Gz:0Gy:1@(0,1)  ->  [Gx:0Gy:1][Gz:0Gy:1][]@(0,1)
```

`compress_depth_inplace` reaches `_shift_gates_forward_inplace`, which calls `_append_layer_component` and `_remove_layer_component` — the primitives that edit a layer list in place rather than replacing it. Slice a sub-circuit, optimize it, and the circuit it came from quietly becomes a different circuit.

`extract_labels` took a shallow list slice and handed it to `_fastinit`; `copy(editable=True)` copied one level, leaving the nested lists for compound labels shared. Both now rebuild every list level, sharing only the immutable `Label` leaves. The change is confined to the editable tier — the static path still shares its tuples, so `static_copy` and `layer_slice` are unaffected.

`test_circuit_aliasing.py` asserts independence across 24 public in-place mutators against both a slice and a copy. Only `compress_depth_inplace` corrupted the parent before this change, which corroborates the issue's observation that the other editors replace layer lists rather than mutate them.

### #761 — `.str` setter accepted truncated and extended string reps

The consistency check used `zip`, which stops at the shorter sequence, so both directions passed vacuously and the circuit misreported itself until `done_editing` regenerated the string. A length check ahead of the zip closes it. The accept/reject matrix is now complete: exact match accepts; same-length mismatch, truncation, extension and a static circuit all raise.

### #746 — proposing to close

The report expects clearing line 0 of a two-line `Gy` to restrict the gate to line 1. I don't think that's right: restricting an n-qubit gate to a subset of its qubits isn't well defined in general, and it's only meaningful for a parallel compound of single-qubit gates, which `_layer_components` would have decomposed before the straddle check ran.

What the report does expose is a bad error. `"Cannot operate on a block that is straddled by Gy!"` doesn't mention that `clear_straddlers` exists, and someone who finds it would reasonably expect "restrict" when it actually deletes the gate outright — I checked, and `clear_straddlers=True` on that circuit leaves `_labels == [[]]`. So neither branch does what the reporter wanted and the message gave no way to discover that. The error now names the option and says what it does, and `clear_labels`'s docstring matches.

Both straddle guards that *have* an escape hatch now name it. The third, in `insert_idling_layers_inplace`, has no such option to name, and whether shifting should get one is a separate question I left alone.

### Reviewer attention

Five changes are observable from outside the library, each with a CHANGELOG entry. Worst first, by how hard the failure would be to attribute rather than by how likely it is. Details are in the sections above.

1. **#757: a circuit parsed from non-canonical text now hashes differently.** An on-disk dataset keyed by one silently changes identity. Anything already canonical is byte-identical — see *Testing*.
2. **All-compilable circuits render differently**, `'Gx:0Gy:0@(0)'` → `'Gx:0~Gy:0~@(0)'`, so a stored string for one no longer matches. No other string moves.
3. **#757, narrower: `[Gx:0Gy:0]` and `[Gx:0Gy:qa]` are now rejected at parse time.** A hand-written input file containing either used to load and now raises.
4. **#758: `a + b` reports both operands' compilable layers where it previously reported none.** Downstream consumers of compilability see a different answer for summed circuits.
5. **#761: assigning a string rep with the wrong number of layers now raises.** Any code this breaks was already producing a self-inconsistent circuit, which is why it's last.

Deliberately out of scope: #472 (line-label reordering during `+`), which the #758 report suggests sweeping in here.

### One follow-up to file

Also found while doing this work, and pinned as current behavior so it can't regress silently:

- **`_fastinit` and `__init__` disagree about nested compound layers**, which `__init__` flattens. Reproduces from a hand-built, already-sorted label with no parser involved, so it's a separate divergence rather than a remnant of #757 — but it's the same wrong-lookup failure mode, reachable through `StdInputParser` for hand-written `[[...]]` layer syntax. Unlike the marker bug above I did *not* fold this one in: there's no scoped fix. It needs a decision on whether nested compound layers should be representable at all — `_op_seq_to_str` can't render them distinctly, so arguably the parser should flatten at parse time and the nesting should never reach a `Circuit` — and then a shared canonicalization step that `__init__` and both parsers call. This is the second divergence found behind `_fastinit`'s undocumented precondition, which is an argument for eventually documenting it.

### Testing

`test/unit` in full, plus the two recorded baselines that exist to catch exactly this class of drift:

- **Circuit corpus replay** (`test/performance/circuit_corpus.py`): all 22634 fingerprints match the committed baseline, empty allowlist. Read that as bounding collateral damage rather than as confirming either fix — the corpus is generated through constructors, so every circuit in it was already canonical (#757), and it holds no circuit with `compilable_layer_indices` at all, a documented coverage gap, so it cannot witness the marker change.

- **Golden fixtures** (`test_circuit_golden_fixtures.py`): pass unchanged, including `golden_dataset.pkl.gz` and `compressed_golden.pkl`, written by an older code version — no on-disk identities moved, so nothing needed regenerating. These *do* cover compilable indices: `compilable_tilde` (1 of 3 layers) and `compilable_pipe` (2 of 3), one on each marker branch and neither all-compilable, so both pre-existing branches still render as before.

Microbenchmarks, best of 7, `test/performance/circuit_microbench.py`, same interpreter and machine:

| benchmark | before | after |
|---|---|---|
| `fastinit` | 0.486 µs | 0.440 µs |
| `init_check_true` | 158.4 µs | 146.6 µs |
| `init_check_false` | 147.8 µs | 134.7 µs |
| `parse_via_init` | 77.8 µs | 74.1 µs |
| `parse_via_stdinput` | 28.0 µs | 27.7 µs |
| `static_copy` | 0.217 µs | 0.201 µs |
| `cached_hash` | 0.061 µs | 0.055 µs |
| `add_nostr` | 17.2 µs | 15.8 µs |
| `add_withstr` | 17.6 µs | 16.8 µs |
| `layer_slice` | 0.629 µs | 0.562 µs |

Everything is at or below baseline, but the uniform few-percent improvement is run-to-run variation, not a speedup — nothing here plausibly makes `cached_hash` faster. What the table shows is the absence of a regression. The one figure that moved for a real reason is `parse_via_stdinput`, which the unconditional-sort version of the #757 patch took to 56.4 µs before the already-canonical fast path brought it back.

New and relocated tests:

- `test_circuit_aliasing.py` (new): slice/copy independence across 24 public in-place mutators, plus the `compress_depth_inplace` witness. Fails in 5 places without the #759 fix.
- `test_circuit_identity_contract.py`: gains the `stdinput` path in `test_construction_paths_agree`, the parsed-matches-constructed and both-parsers-agree assertions, the duplicate-sslbls rejection, `_fastinit`'s canonical-input precondition, and the nested-layer pin.
- `test_circuit_metadata_policy.py`: the `add` row flips, plus index propagation, mixed-marker regeneration, and the unchanged no-compilable path.
- `test_circuit_str_roundtrip.py`: summed-circuit round-trip and index-concatenation properties, and the deliberate `max_size=n_layers - 1` bound that kept the all-compilable case out of the strategy is gone — so the property tests now generate it, and 1-layer circuits carrying indices are covered for the first time. Four example tests name the specific behaviors, including the parsed-vs-generated distinction.
- `test_circuit.py`: two `#746` characterization tests for the message and the removal semantics.
- `test_circuit_known_bugs.py`: now pins no open bugs. I kept the file and its convention docstring, with the str-setter accept/reject matrix as its remaining content since that grew out of the #761 pins and has no better home.


rileyjmurray (commented):
Leaving this comment as a note that I've reviewed the changes.

pygsti/circuits/circuitparser/slowcircuitparser.py
```diff
@@ -16,6 +16,33 @@ def _to_int_or_strip(x):
...
+        if prev is not None:
+            try:
+                if not prev < sslbls:
+                    return False
+            except TypeError:
+                return False  # incomparable sslbls; let the sorter raise
+        prev = sslbls
+    return True
+
+
```
sserita, line 47: If this is the same between the two and we explicitly need them to stay in lockstep, can we drop one of them and import/use from the other (probably fast imports slow?)

test/unit/circuits/test_circuit_known_bugs.py
sserita, line 1: Could we not shift these now contract-enforcing rather than bug-catching tests into test_circuit or test_circuit_str_roundtrip (or elsewhere) and nix this file?

sserita (changes requested):
One code dedup question, one test organization question, but otherwise looks good to me.

test/unit/circuits/test_circuit_known_bugs.py
rileyjmurray: I support shifting the flipped tests to other files.

The only reason for keeping this file after moving tests would be to establish that it's the home for bugs that get identified during characterization tests. We might want another serious round of characterization tests before the inevitable overhaul of the Circuit class.

pygsti/circuits/circuitparser/slowcircuitparser.py
rileyjmurray: Interesting suggestion about fast importing slow! I'll have it try that and then benchmark before/after.

sserita: I guess I didn't consider perf implications of that, I was just thinking from a code structure POV. Could end up being a bad suggestion if there is a hit to perf.

test/unit/circuits/test_circuit_known_bugs.py
sserita: Fine with keeping an empty file as a landing place for it, just want ONLY broken/xfail tests in there.

pygsti/circuits/circuit.py
```diff
@@ -902,7 +926,10 @@ def __contains__(self, x: Union[LabelsLike, int]):
     def __radd__(self, x: Union[Circuit, Sequence[_Label]]):
         if not isinstance(x, Circuit):
             assert(all([isinstance(l, _Label) for l in x])), "Only Circuits and Label-tuples can be added to Circuits!"
-            return Circuit._fastinit(x + self.layertup, self._line_labels, editable=False)
+            # our layers all shift right by len(x), so our compilable indices shift with them
+            shifted = tuple(i + len(x) for i in self._compilable_layer_indices_tup)
```
coreyostrove: compute len(x) once.

pygsti/circuits/circuit.py
coreyostrove, line 1301: Is there another latent compilable index bug here? Should the compilable indices follow the extracted labels? This is a design question. I am leaning toward no, but in that case we should document this behavior since some paths return an actual circuit.

coreyostrove (changes requested):
A couple questions and a minor request inline below. Quick confirmation, did you run the microbenchmarks in the table above or did the agent? If the agent, have you validated that they were in fact run for real? I am reminded of @kmrudin's recent experience of requesting some benchmarks from and agent and within seconds having the agent output a table with runtimes of tens of minutes.

pygsti/circuits/circuitparser/slowcircuitparser.py
rileyjmurray: Turns out there was no performance benefit to the fast parser implementation. Cython only buys performance when we have type annotations. Those functions didn't have any. I've changed to import from the slow parser.

pygsti/circuits/circuit.py
rileyjmurray: I also lean no. The documentation did actually mention this, just at the end in a totally buried way. I've moved this point up to just before the parameter descriptions.

sserita (approved):
Approving as my asks were resolved

coreyostrove (approved):
Reviewed latest changes. LGTM, thanks, @rileyjmurray!

Issue #766, opened 2026-06-10
`Label.IS_SIMPLE` conflates two independent properties


rileyjmurray:
This issue brought to you by a robot, written from the robot's perspective.

------------------------------------------------

Found while auditing `pygsti/baseobjs/label.py` and `pygsti/circuits/circuit.py` during the `bugfix-circuit-batch` work (line numbers from that branch head, `4fcccd5ed`; the structural facts are identical on `develop`). This is a *design* issue for the planned `Circuit`/`Label` refactor, not a request to change runtime behavior on `develop`. It changes only the label **classification API**; it does not touch the frozen identity contract (`tup`/`hash`/`__eq__`/string grammar).

It is the root cause behind one bug already fixed on the batch branch (commit `2aff09799`, "Preserve CircuitLabels in `map_names_inplace` and `map_state_space_labels_inplace`") and two latent traps described below.

### TL;DR

`Label.IS_SIMPLE` (read via the `is_simple` property, `label.py:228`) is asked to answer **two different questions** that happen to give the same answer for every label class *except* `CircuitLabel`:

- **A — "Is this a layer-atom?"** i.e. does it occupy a single layer slot as an opaque unit, so the within-layer machinery must **not** expand its `.components` as parallel co-layer siblings? `circuit.py` consumers want this one.
- **B — "Is this a primitive op?"** i.e. is it a single operation with **no nested sub-label structure**, reconstructible from its `name`/`sslbls`(/`args`) and resolvable as one op in a model? The model- and budget-layer consumers want this one.

`CircuitLabel` is the one class where the answers diverge: a sub-circuit box **is** a layer-atom (A = true) but is **not** a primitive op (B = false — it wraps a whole subcircuit). Because the single flag can only hold one value, `CircuitLabel.IS_SIMPLE = True` (`label.py:1384`) is *correct for A* and *wrong for B*, and every B-consumer either needs an `isinstance(..., CircuitLabel)` bypass or carries a latent bug.

The refactor should replace the one flag with a pair of clearly-named booleans, one per axis.

### The two questions, with consumer evidence

**Axis A — layer-atom (horizontal / parallel decomposition).** The within-layer routines read `lbl.components` as *the gates simultaneously sharing this layer* and use the flag to decide whether to descend:

- `_label_to_nested_lists_of_simple_labels` (`circuit.py:118-127`): `[lbl]` if simple, else recurse into `lbl.components`.
[...]

PR #889, opened 2026-08-18
Label semantics: `time` is a duration, and `CircuitLabel.depth` is not a layer index

### LINKED ISSUE #754: LabelX.time can have multiple meanings
In label.py, `CircuitLabel` suggests that the time property of a label can be in relative or absolute terms and that it is the users specified. However, we do not have a way to differentiate the two meanings at the level of a label.

I propose that we add a `duration` property. That way we can use `duration` as a relative time property to the previous layer, and `time` as the absolute time compared to the start of the execution of the circuit.

rileyjmurray:
This PR is mostly documentation and two small behavior fixes. It closes #754 and resolves minor bugs discovered along the way.

#### `Label.time` is a duration (#754)

#754 says the meaning of `time` is undetermined — "relative or absolute" — and proposes adding a `duration` property so that `duration` becomes the relative quantity and `time` the absolute one. The implementation isn't actually ambiguous. Every aggregation site treats `time` as a duration: `LabelTupTupWithTime.init` takes the `max` over components (a layer's gates run in parallel), `CircuitLabel.__new__` takes the `sum` (a box's components are sequential layers), and `Circuit.duration` sums the layers. Those three rules only cohere under a duration reading.

The time-dependent map forward simulator settles it. In `mapforwardsim_calc_densitymx.pyx` it starts a clock at `t0 = datarow_time[kinit]` and then does `t += gl.time` per layer, calling `op.set_time(t)` as it goes. Label `time` is the increment; the absolute clock comes from the `DataSet`. The "absolute time from the start of execution" that #754 wants already exists in pyGSTi — it lives on data rows, not on labels.

So the ambiguity is confined to one docstring, which is the one #754 quotes. The six identical `time : float` parameter blocks across the `Label` classes now say the duration, not the moment, and the `Label` class docstring gains a `Notes` section stating the aggregation rules, distinguishing label durations from `DataSet` timestamps, and recording that `==`/`hash` ignore `time` while `str` and `to_native` preserve it. `Circuit.duration` gains a real docstring — it previously had only `# similar to depth()`. And the forward-simulator line above carried the comment `# time in gate label == gate duration?`; the question mark is now answered.

I've closed #754 as answered-by-the-code rather than implemented. An absolute per-label start time is a legitimate thing to want; the argument here is only against spelling it `time`, since `Circuit.duration` already exists and is built out of `time`-as-duration.

#### `CircuitLabel.time` ignored `reps`

Found while writing the above. When `time` is not given explicitly, `CircuitLabel.__new__` inferred it as the sum of its components' times and dropped `reps` entirely, so a box of layers totalling 7.0 with `nreps=3` reported `time == 7.0`. That is inconsistent with `CircuitLabel.depth`, which is `sum(component depths) * reps`, and with the sequential reading that justifies the sum in the first place. It's now `reps * sum(...)`. An explicitly passed `time` is still taken as given.

This has no consumers in-tree — `Circuit.duration` is the only reader and nothing in the library calls it — and no test covered the inferred path with `reps > 1`, which is presumably how it survived.

#### Five loops mixed up `Circuit.depth` and `Circuit.num_layers`

`Circuit.depth` counts executed layers, descending into `CircuitLabel` boxes; `num_layers` counts top-level slots. The split is documented, and both have consumers that need what they give — `CircuitLabel.depth` is hand-overridden precisely so a box reports its executed length. But the per-layer accessors (`layer_label`, `layer_label_with_idles`, `layer`, `_layer_components`) index top-level layers, so they are bounded by `num_layers`, and several callers loop `range(circuit.depth)` over them. On any circuit retaining a box, that raises on the second iteration.

The affected sites are `circuit_budget` and the `circuit_budgets` derivative loop in `wildcardbudget.py`, `ErrorRatesModel._circuit_cache` in `oplessmodel.py`, the remaining-width 2Q gate count in `subcircuit_selection.py`, and the layer encoding in `extras/ml/encoding.py`. Each is now guarded with

```python
if circuit.depth != circuit.num_layers:
    circuit = circuit.expand_subcircuits()
```

Expansion rather than iterating `range(num_layers)`, because these consumers genuinely want one entry per executed layer: `encoding.py`'s padding arithmetic is measured against `depth`, and an error rate or wildcard budget should count an `nreps=3` box three times. The guard leaves the common boxless path untouched.

`Circuit.layer_label`'s assertion message said *"Circuit is only of depth 1"* while checking `num_layers`, which points anyone who hits this at the wrong quantity; it and the docstring now name `num_layers` and suggest `expand_subcircuits()`.

Worth knowing if you go looking for this: `Circuit.default_expand_subcircuits` is `True`, so constructing a static circuit directly flattens the box and the divergence disappears. You only retain one via `editable=True` followed by `done_editing()`.

#### Small changes:
 * `agent-docs/known-debt.md` gains an entry for issue #766's `IS_SIMPLE` conflation. That issue is deferred to the `Circuit`/`Label` rewrite rather than addressed here; I've commented there with the reasoning, and with a correction that this PR occasions — #766 says its latent asserts never fire because boxes are "usually expanded out before reaching these paths," but in fact the `depth`/`num_layers` mismatch above kills the loop first, so fixing that makes them reachable for the first time.
 * `Circuit.layer_labels`' non-static branch was `layer_lbl if isinstance(layer_lbl, _Label) else _Label(layer_lbl)`; it's now just `_Label(layer_lbl)`. `Label.__new__` returns a `Label` input unchanged, so the guard was redundant. Three stray double-blank-lines inside the class body removed while I was there.
 * New `CircuitDepthVsNumLayersTester` in `test/unit/circuits/test_circuit.py` pins the `depth`/`num_layers` contract, the improved assertion message, and the two fixed consumers that can be exercised without optional dependencies.
 * Two tests in `test/unit/objects/test_label.py` cover the inferred `CircuitLabel.time` across `reps` and `Circuit.duration` over a box.
 * `test/unit/objects/test_label.py test/unit/circuits/` goes from 372 to 380 passed, all eight new; the golden fixtures are unchanged.


coreyostrove (approved):
No changes requested, but I was not aware of the behavior involving CircuitLabels in the Circuit construction wherein they are flattened by default. I find that unexpected and think we should determine whether this is an intentional behavior/design choice (as it is being pinned to by some of the new tests), or unintentional and should be changed. Please open an issue to track this question.

PR #894, opened 2026-08-20
Move AGENTs.md to the top-level directory. Update agent-docs.

rileyjmurray:
For a few months now pyGSTi has had an AGENTS.md file at `agent-docs/AGENTS.md`. This PR moves that file to a top-level `AGENTS.md`.

The move of AGENTS.md necessitated some changes to other files in `agent-docs`. Once I got started in that I decided to do a comprehensive update. The update burned more tokens than I would have liked, so the new file `agent-docs/META.md` gives instructions for different kinds of updates to `agent-docs`.


agent-docs/known-debt.md
rileyjmurray, line 1: The renumbering here was a bad idea. Removing resolved _content_ is wise, but we need the headings in place since we refer to the numbered issues elsewhere in `agent-docs`.

Issue #205, opened 2021-08-25
Rethink inline javascript for reports for JupyterLab compatibility


sserita:
**Is your feature request related to a problem? Please describe.**
The current report workspaces heavily uses inline javascript for its interactive rendering. This works great in regular Jupyter notebooks, but JupyterLab silently suppresses the JS execution. If more users are moving towards JupyterLab, this is something we should provide an alternative for.

**Describe the solution you'd like**
Maintain the same level of display functionality as classic Jupyter notebooks in JupyterLab.

**Describe alternatives you've considered**
- Try to find an extension/setting that reenables the current JS execution in JupyterLab, and require it for interactive report generation.
- Rework the JS execution into something that is natively enabled in JupyterLab.

Not sure which is the easier solution yet. The former seems like less work in the PyGSTi codebase but may not be easy to find/require. More likely, we should make sure to keep this in mind the next time we spend significant effort on upgrading the reports.

**Additional context**
I have not searched exhaustively yet, but this thread is relevant: https://github.com/jupyterlab/jupyterlab/issues/3118

rileyjmurray:
Update on this. I haven't been able to get notebook reporting to work even if I try to avoid jupyterlab. Here's what I see when I install nbclassic:

![image](https://github.com/user-attachments/assets/78af3497-f6b5-4281-87cf-15dde5c70962)

There might still be a way to avoid this. I'm just not super familiar with the python notebook ecosystem. In fact, even when _only_ installing nbclassic from pip, I see the following when launching the server. Note that it mentions JupyterLab being loaded and linked.
<img width="1058" alt="image" src="https://github.com/user-attachments/assets/58536417-0c92-4a29-bd4f-0e4688c9d8aa" />

sserita:
Yes, I have recently been getting reports of this breaking even if avoiding JupyterLab. However, there are some known workarounds by reaching directly into the plotly objects to pull out the plots - I wonder if we can turn that into less of a hack (or at least a canonicalized hack) and finally close this issue. I'll put this on our agenda to discuss the next time we do issue re-triage (which might be next week).

sserita:
This might need to be split into two: short-term fix for report notebooks, and long-term fix to handle more complicated things like switchboards.

PR #895, opened 2026-08-20
Restructure the web docs around readers, and ship them executed

rileyjmurray:
This PR began as a plan to reorganize the tutorial notebooks, whose directory names (`objects/`, `gst/`, `reporting/`, ...) mirrored pyGSTi's internal code taxonomy rather than anything a reader needs. Once the pages were organized for readers, the fact that the hosted site shows every code cell with no output became the biggest remaining defect, and fixing that pulled in CI, ReadTheDocs plumbing, and a handful of library bugs that the newly-executed pages exposed.

The result is 198 files changed: the tutorial corpus goes from 106 pages to 84, every page now executes in CI and ships its output to the hosted site, and the example HTML reports the tutorials generate are published with the book at `/reports/`. I don't expect anyone to read the whole diff, and the review guide below says where I think attention is best spent. The rendered site is the primary review artifact: **https://pygsti.readthedocs.io/en/docs-preview/**, staged from this branch by the staging workflow this PR adds.

## TL;DR. (Really. This is the short version.)

The site is now organized into four parts, identifiable by the sidebar top-level sectioning and by the URL prefixes. 
 * `start/` (9 pages, a linear on-ramp);
 * `guides/` (protocol chapters — GST, RB, volumetric benchmarks, drift — plus shared chapters for the measurement workflow, models, and analysis);
 * `advanced/` (internals, conventions, simulation engines, specialist protocols);
 * the API reference.

Depth is capped at 2 everywhere. The function-based driver API pages are retired in favor of one migration table; nine pages are new, aimed at holes the old corpus never filled (getting your own data in, choosing a protocol, judging fit quality); and the 16-page GST chapter got a page-by-page revision against its own executed output.

<details>
<summary>Page accounting: what was retired, merged, and added</summary>

- **Retired.** Stubs, self-declared-broken pages, and duplicates, each verified before removal (the deletion commits record the evidence: one page's stated purpose calls a function that no longer exists; another ships the text "This example doesn't work"). The four function-API pages are retired deliberately: the rest of the docs teach the Protocol API, and `FromFunctionAPI` maps every driver function to its replacement.
- **Merged.** 22 near-duplicate secondaries folded into their primaries: the two volumetric pages, the four QPANN pages, the two MPI pages, the two error-bar pages, and so on. The planned RB merge was cancelled on inspection — those protocols are genuinely distinct, and merging would have re-buried three of them.
- **New.** `start/YourOwnData` (how to get your data into pyGSTi — previously answered nowhere), `ChooseAProtocol`, `FirstBenchmark`, `ReadingResults`, `KeyConcepts`, `guides/gst/JudgingTheFit`, `guides/gst/BadFits`, `guides/rb/InterleavedRB` (previously the unannounced second half of a page about a different protocol), and `advanced/migration/FromFunctionAPI`. GST is now split across four chapters.
- **Revised.** All 16 `guides/gst/` pages were executed, their outputs digested, and the pages corrected against what actually printed: wrong counts fixed, claims that executed output contradicted rewritten, dead weight cut, cross-links added.
- **Corpus-wide.** Every page title is now sentence case. The internal `agent-docs/` links into `docs/markdown/` were repointed to the new locations (47 links across six files).
</details>

## Executed output, in CI and on the site

CI was already executing all the notebooks under `pytest --nbval-lax` and throwing the results away — nbval never compared an output against anything, since no notebook uses its output-checking tags. A new driver, `docs/execute-notebooks.py`, replaces that step: same green criterion, roughly the same 40 minutes, but the outputs are kept, normalized for byte-stability, and shipped to the generated branch that ReadTheDocs builds. Executed `.ipynb` never touch a source branch. Any branch can now be staged to ReadTheDocs before merging, which is how the preview link above exists.

<details>
<summary>Pipeline details: the driver, the artifact flow, reports, and build cost</summary>

- **The driver** runs each notebook with its own directory as cwd, runs the eight notebooks sharing `tutorial_files/` as a serial producer-before-consumer chain and the rest in parallel, takes a per-cell timeout, always persists what did run, and normalizes run-to-run noise out of the output. It exists because `nbconvert` isn't in `.[docs]` and `jupytext --execute` hardcodes `timeout=None`; its docstring has the full rationale.
- **The artifact flow.** The job that executes uploads `docs/markdown/**/*.ipynb`; the committing job downloads it, deletes each `.md` that has an executed pair (Sphinx wants one source per document), and pushes the generated branch. It counts the notebooks after download and fails loudly below 70 rather than pushing a branch with none.
- **Staging.** `docs/stage-rtd-branch.py` plus `stage-docs-branch.yml` / `docs-staging.yml` let any branch be assembled and pushed to a preview branch that ReadTheDocs builds. The production path (develop) and the preview path share the same implementation, so a preview is evidence about what merging would produce.
- **Reports.** Executing the tutorials leaves ~25 pyGSTi HTML reports in scratch directories. `docs/collect-reports.py` gathers them, and they ship to the site as a 5 MB tarball rather than the 594 MB they stage to, mostly by writing them with `connected=True` and deduplicating the JS libraries. The "see the report" links on the site resolve for the first time.
- **Build cost.** ReadTheDocs was at 53 minutes against a 60-minute ceiling. Capping the rendered sidebar depth (the API reference was putting 339 KB of `<nav>` into each of 1848 pages) takes a cold build from 20 to 12 minutes and the site from 810 MB to 208 MB. `docs/build-prose.py` gives contributors a build that skips the API reference and finishes in seconds.
</details>

## Library changes

A docs PR has no business changing the library silently, so here is the complete list, stratified by user exposure.

Higher exposure:

- **Workspace figures render again in JupyterLab 4 and Notebook 7** (#205). Every in-notebook figure depended on RequireJS, which modern front-ends don't ship, so figures simply never appeared. The new `pygsti/report/notebook_setup.py` loads the JS dependencies with plain script tags. The trade, stated plainly: classic Notebook (6.x, EOL) support is dropped, and that's where these figures used to work.
- **`sum_of_negative_choi_eigenvalues*` renamed to `abs_sum_of_negative_choi_eigenvalues*`** (three functions in `pygsti.tools.jamiolkowski`). The old names suggested a non-positive return value; the functions return absolute values. The old names remain as wrappers that emit `pyGSTiDeprecationWarning`.
- **`chp_sim` is now the default CHP backend in tooling**, with a new `[chp]` extra. Aaronson's C implementation stays supported via `pygsti.evotypes.chp.chpexe`.
- **`ExplicitOpModel.print_info` removed** (deprecated since 2022).

Lower exposure, found because executed pages made the symptoms visible:

- Report construction no longer emits spurious `RuntimeWarning`s (two divide-by-zero sites, plus a numpy warning raised from inside cvxpy, silenced in `sdptools`).
- The report's data-comparison section now honors `verbosity=0`.
- Protocol checkpoint objects have informative, deterministic reprs. Previously `print(checkpoint)` gave `<...object at 0x...>` immediately after tutorial prose calling the object "useful to inspect".
- The `@deprecate` decorator now applies `functools.wraps`, so deprecated functions keep their names and docstrings in the API reference.

Unit tests accompany the repr and rename changes; the CHANGELOG records the rename, the removal, and the figure fix.

## Where to spend review attention

1. **Line-level review: everything outside `docs/markdown/`.** About 35 files — the library changes above, the five workflow files, and the new `docs/*.py` tooling. Small in lines, and it's where the consequences live. I'd examine `pygsti/report/workspace.py` + `notebook_setup.py` and `.github/workflows/reuseable-main.yml` in detail.
2. **Decision-level review, no diff-reading required.** Three decisions here are hard to reverse after release, and I want them vetoed now or owned together:
   - **Old doc URLs will 404.** I built a redirect mechanism covering every removed URL, verified it end-to-end, and then removed it (`872284e03`): its lasting value was driving the one-time link rewrite, and carrying a redirect stub generator forever seemed worse than a clean break. Since the website just went live this summer I figure there's no harm in breaking URLs. Call me out on that if you think I'm wrong!
   - **Classic Notebook support for in-notebook figures is dropped** (see above).
   - **The hosted docs are assembled only by CI**: executed notebooks are committed to a generated branch and ReadTheDocs builds from there, so the site is only as fresh as the last green execution run.
3. **Domain spot-checks.** The revised GST chapter makes quantitative claims that I verified against executed output but would value a (real) physicist's eye on: the robust-GST (TVD) conclusion numbers; FewerCircuits' statement that neither per-germ nor global fiducial-pair reduction reliably yields the smaller design; the loosened optimizer tolerance in TwoQubitGST (a 7-10x speedup, justified in the page); and KeyConcepts' account of the H/S/C/A error-generator taxonomy.
4. **Everything else: skim the built site rather than the diff.** The preview renders exactly what this branch ships.

## What we (the robot and I) checked

All 76 code-bearing pages execute cleanly under the driver, most recently in the staging run that built the preview above. The GST chapter was also re-executed after every revision wave, and once more after merging the latest `develop` (16/16 clean, 261 s total). Unit CI is green on the branch.

<details>
<summary>Verification details</summary>

- Byte-stability of executed output, measured over ten full runs: 48 of 76 notebooks come out byte-identical across independent cold runs. The residue is unseeded randomness inside pyGSTi plus one MPI page with nondeterministic rank interleaving; it was measured and accepted at ~2.4 MB of packed git history per fully re-executed release.
- Intra-corpus links: 0 broken and 0 self-links at the last full sweep (283 links); the handful added since were each checked to resolve. Toc: 0 dangling entries, 0 orphans.
- Redirect-era checks (before the mechanism was removed): 125/125 stubs emitted with 0 unresolvable targets, including on incremental builds over a dirty output directory.
- The new pages were adversarially verified claim-by-claim against the library; among the defects caught before shipping: a per-circuit loglikelihood threshold that ignored multiple comparisons, a backwards claim about what Clifford RB benchmarks directly, and a "nothing in your existing scripts will break" claim that one broken driver function falsified.
- Local runs used a dedicated conda env with an editable install of this branch; the GST-chapter digests (per-cell outputs and timings) are retained outside the repo.
</details>

## Deliberately out of scope

- The remaining page splits from the original plan (Circuit's label reference, ModelNoise's parameterization reference, Results' developer half, FileIO's serialization taxonomy, DataSimulation's freeform framework) and the long tail of per-page revision outside the GST chapter.
- Cross-page `#anchor` links don't resolve in this Jupyter Book configuration. New links in this PR were written page-level to avoid the bug; a handful of pre-existing anchor links remain broken and need their own investigation.
- Library follow-ups, to be filed as issues: `DeprecationWarning` from `@_deprecated_fn` is invisible outside `__main__` under default filters, for every user; iterative GST prints an alarming, unexplained `WARNING:` on stderr when it converges by hitting its iteration cap; `run_linear_gst` raises `TypeError` for all inputs; and IdleTomography's long-admitted factor-of-2 discrepancy needs a physics-level answer before that page rejoins the promoted reading path.


pygsti/tools/sdptools.py
```diff
@@ -30,9 +31,26 @@
...
+    whether it fires at all depends on what was last in that memory. Seen with cvxpy 1.9.2.
+    """
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', _EMPTY_REDUCE_WARNING_PATTERN, RuntimeWarning)
+            return fn(*args, **kwargs)
+    return wrapper
+
+
```
sserita, line 53: This is potentially untested. Low-impact though so 🤷

pygsti/tools/jamiolkowski.py
sserita, line 1: Fine with the renames

.github/workflows/docs-staging.yml
sserita, line 1: Is this and stage-docs-branch.yml getting blown away as we merge? Won't be relevant on master.

docs/markdown/guides/gst/TwoQubitGST.md
```diff
@@ -0,0 +1,92 @@
...
+                                               mdl_datagen, num_samples=1000, seed=2020)
+
+# ---- NOTE: you can stop and restart the python session at this point; everything you need is saved to disk ---- 
+
+#Load in the "data object" which packages together the dataset and experiment design
+data = pygsti.io.read_data_from_dir("../../../example_files/My2QExample")
+```
+
+## Step 4: run GST
+Just like for 1-qubit GST, we use the `StandardGST` protocol to compute the GST estimates.  We loosen the optimizer's convergence tolerance (`optimizer={'tol': 1e-3}`) to keep this demonstration quick — the fit below finishes in well under a minute, versus several minutes at the default tolerance — but you should drop the `optimizer` argument for a real analysis.  Runtime materially increases with sequence length; if your fits are taking too long, pyGSTi supports MPI-acceleration — see the [Parallelism tutorial](Parallelism).
```
sserita, line 60: I'm fine with doing this in the tutorial. However, I would also recommend including a one-line comment in the cell below. I often skip prose and only look at cells, so would be nice to catch other users who may do the same.

docs/markdown/guides/gst/RobustGST-TVD.md
```diff
@@ -246,11 +252,18 @@ report = construct_standard_report(
...
 report_dir += '/robust-gst-report'
 print('HTML report will be written to ... ')
 import os
 print(os.getcwd() + '/' + report_dir + '/main.html\n\n')
-report.write_html(report_dir, verbosity=0)
-
+report.write_html(report_dir, connected=True, verbosity=0)
 ```
+
+Trained and tested on clean data (first table), the two objectives agree closely: `-logl` scores 465.6, `tvd` scores 475.7 — a roughly 2% cost for the robust objective when there is nothing to be robust to.
```
sserita, line 263: I abstain on this, @coreyostrove knows this stuff much better than me.

docs/markdown/start/KeyConcepts.md
```diff
@@ -0,0 +1,66 @@
...
+Writing a noisy gate as $G = e^{L} G_{\text{target}}$ makes $L$ the *error generator*: the thing that vanishes when the gate is perfect. Error generators are more useful than the maps themselves for most purposes, because they compose additively to first order and their pieces have physical names.
+
+pyGSTi sorts those pieces into four types, labeled by their first tuple element:
+
+- **H** — Hamiltonian (coherent) errors. A systematic over- or under-rotation is an H term.
+- **S** — stochastic errors. The diagonal of the non-Hamiltonian block; a depolarizing channel is all-S.
+- **C** — correlation errors: the symmetric (real) half of that block's off-diagonal part.
+- **A** — active errors: the antisymmetric (imaginary) half of the same off-diagonal part.
+
+The names are pyGSTi's, following the taxonomy of [arXiv:2103.01928](https://arxiv.org/abs/2103.01928). Note that C and A are two halves of one thing, not a diagonal term and an off-diagonal term: both index a *pair* of basis elements, and they split the off-diagonal block between them.
```
sserita, line 34: One minor suggestion: I would make sure to call this HSCA "elementary error generators" somewhere here. It is not the only basis that works for error generators, but it is the one in the taxonomy AND the one explicitly called out in various functions/classes in the codebase.

docs/stage-rtd-branch.py
sserita, line 1: Similar question to the staging Actions: Is this a general thing we want in master, or is it enough to keep in the commit history in case we need to do something like this again?

docs/markdown/guides/gst/FewerCircuits.md
sserita, line 1: This is actually trickier and probably worth explicit discussion.

I would definitely lean towards at least rejecting the agent's (apparently not well-based) claim that PGGFPR is more effective at multi-qubit (line 42, second to last sentence). It might be true, but Corey's original line in 875c5b0 said nothing about the multi-qubit context. I believe the agent decided to say this because a) we comment that PGGFPR is recommended and approaches informatic-theoretic lower bounds and b) awkwardly, this doesn't pan out in our example (to be discussed in a second).
So it decided "it must be better in multiqubit" with no good justification.

## PGGFPR awkwardness

That brings us to the second awkwardness: PGGFPR is not the best (and in fact, not great) when compared to the other methods on this problem.

### Regression check
These counts are stable since the original 2023 check-in of these tutorials, i.e. it is not an introduced behavior in the meantime.

### Sweeping over tolerances
Twiddling tolerance knobs don't fix this discrepancy.
1. PGGFPR's `inv_trace_tol` is not a helpful knob here. The search is stopping on the rank condition, not the inv trace metric.
1. A GFPR `tol` sweep to tighter tolerances results in larger FPR sets, as expected.
1. A PFPR `base_loweig_tol` sweep to tighter tolerances actually results in **smaller FPR sets**. What?

```python
import numpy as np
import pygsti
import pygsti.circuits as pc
from pygsti.modelpacks import smq1Q_XY

target_model = smq1Q_XY.target_model()
prep_fiducials = smq1Q_XY.prep_fiducials()
meas_fiducials = smq1Q_XY.meas_fiducials()
germs = smq1Q_XY.germs()
maxLengths = [1, 2, 4, 8, 16, 32]
opLabels = list(target_model.operations.keys())

def l32(fid_pairs):
    s = pc.create_lsgst_circuit_lists(opLabels, prep_fiducials, meas_fiducials, germs, maxLengths, fid_pairs=fid_pairs)
    return len(s[-1])

# (a) PGGFPR's inv_trace_tol is a no-op: the search is rank-bound, not trace-bound.
germ_vecs, _ = pygsti.alg.germ_set_spanning_vectors(target_model, germs, float_type=np.double, verbosity=0)
for itt in [10, 50, 200]:
    d = pygsti.alg.find_sufficient_fiducial_pairs_per_germ_global(
        target_model, prep_fiducials, meas_fiducials, germ_vector_spanning_set=germ_vecs,
        inv_trace_tol=itt, verbosity=0)
    print(f"inv_trace_tol={itt:>3d}: per-germ pairs={[len(p) for p in d.values()]}, L=32 circuits={l32(d)}")

print()

# (b) GFPR is seed-sensitive; the demo's seed=1234 is its best result of 5 tested.
for seed in [1234, 42, 7, 2021, 99]:
    fid_pairs = pygsti.alg.find_sufficient_fiducial_pairs(
        target_model, prep_fiducials, meas_fiducials, germs,
        search_mode="random", n_random=10, seed=seed, verbosity=0,
        mem_limit=int(2 * (1024) ** 3), minimum_pairs=2)
    print(f"GFPR seed={seed:>5d}: L=32 circuits={l32(fid_pairs)}")

print()

# (c) PFPR: tightening base_loweig_tol below the demo's own 0.25 shrinks the design further.
for tol in [0.25, 0.1, 0.05]:
    d = pygsti.alg.find_sufficient_fiducial_pairs_per_germ(
        target_model, prep_fiducials, meas_fiducials, germs,
        search_mode="random", constrain_to_tp=True, n_random=100, min_iterations=50,
        base_loweig_tol=tol, num_soln_returned=1, type_soln_returned='best', retry_for_smaller=True,
        seed=1234, verbosity=0, mem_limit=int(2 * (1024) ** 3))
    print(f"PFPR base_loweig_tol={tol}: L=32 circuits={l32(d)}")
```

Output:

```
inv_trace_tol= 10: per-germ pairs=[6, 6, 4, 2], L=32 circuits=142
inv_trace_tol= 50: per-germ pairs=[6, 6, 4, 2], L=32 circuits=142
inv_trace_tol=200: per-germ pairs=[6, 6, 4, 2], L=32 circuits=142

GFPR seed= 1234: L=32 circuits=125
GFPR seed=   42: L=32 circuits=136
GFPR seed=    7: L=32 circuits=128
GFPR seed= 2021: L=32 circuits=141
GFPR seed=   99: L=32 circuits=134

PFPR base_loweig_tol=0.25: L=32 circuits=150
PFPR base_loweig_tol=0.1: L=32 circuits=122
PFPR base_loweig_tol=0.05: L=32 circuits=111
```

### Fisher information check
Everything is amplificationally complete at first glance, i.e. GFPR is not cheating by losing sensitivity.

```python
import numpy as np
import pygsti
import pygsti.circuits as pc
from pygsti.tools.edesigntools import calculate_fisher_information_matrices_by_L
from pygsti.modelpacks import smq1Q_XY

target_model = smq1Q_XY.target_model()
prep_fiducials = smq1Q_XY.prep_fiducials()
meas_fiducials = smq1Q_XY.meas_fiducials()
germs = smq1Q_XY.germs()
maxLengths = [1, 2, 4, 8, 16, 32]
opLabels = list(target_model.operations.keys())
full_structs = pc.create_lsgst_circuit_lists(opLabels, prep_fiducials, meas_fiducials, germs, maxLengths, fid_pairs=None)

# (a) approx=False (the default) is numerically unstable here; approx=True is properly PSD.
for approx in [False, True]:
    fims = calculate_fisher_information_matrices_by_L(target_model, full_structs, maxLengths, num_shots=100, verbosity=0, approx=approx)
    eig32 = np.sort(np.linalg.eigvalsh(fims[32]))
    print(f"approx={approx!s:5s}: min eigenvalue={eig32[0]: .3e}, max eigenvalue={eig32[-1]:.3e}")

print()

# (b) with approx=True fixed, every design (incl. knob-tuned ones) is amplificationally complete.
Ngauge = target_model.num_gauge_params
full_fims = calculate_fisher_information_matrices_by_L(target_model, full_structs, maxLengths, num_shots=100, verbosity=0, approx=True)
full_eig32 = np.sort(np.linalg.eigvalsh(full_fims[32]))
gap_threshold = np.sqrt(max(full_eig32[Ngauge - 1], 1e-12) * full_eig32[Ngauge])
print(f"Full design's own gap: gauge tops out at {full_eig32[Ngauge - 1]:.3e}, real starts at {full_eig32[Ngauge]:.3e}")

def is_complete(fid_pairs, name):
    structs = pc.create_lsgst_circuit_lists(opLabels, prep_fiducials, meas_fiducials, germs, maxLengths, fid_pairs=fid_pairs)
    fims = calculate_fisher_information_matrices_by_L(target_model, structs, maxLengths, num_shots=100, verbosity=0, approx=True)
    eig32 = np.sort(np.linalg.eigvalsh(fims[32]))
    n_below = int(np.sum(eig32 < gap_threshold))
    print(f"{name:22s}: L=32 circuits={len(structs[-1]):4d}, eigenvalues below gap={n_below} (expect {Ngauge})")

is_complete(None, "Full (no reduction)")
d_pfpr_tight = pygsti.alg.find_sufficient_fiducial_pairs_per_germ(
    target_model, prep_fiducials, meas_fiducials, germs,
    search_mode="random", constrain_to_tp=True, n_random=100, min_iterations=50,
    base_loweig_tol=0.05, num_soln_returned=1, type_soln_returned='best', retry_for_smaller=True,
    seed=1234, verbosity=0, mem_limit=int(2 * (1024) ** 3))
is_complete(d_pfpr_tight, "PFPR (tol=0.05)")

germ_vecs, _ = pygsti.alg.germ_set_spanning_vectors(target_model, germs, float_type=np.double, verbosity=0)
d_pggfpr = pygsti.alg.find_sufficient_fiducial_pairs_per_germ_global(
    target_model, prep_fiducials, meas_fiducials, germ_vector_spanning_set=germ_vecs, verbosity=0)
is_complete(d_pggfpr, "PGGFPR")
```

Output:

```
approx=False: min eigenvalue=-1.931e+06, max eigenvalue=1.284e+14
approx=True : min eigenvalue=-1.132e-02, max eigenvalue=1.284e+14

Full design's own gap: gauge tops out at 1.247e+00, real starts at 1.364e+04
Full (no reduction)   : L=32 circuits= 568, eigenvalues below gap=16 (expect 16)
PFPR (tol=0.05)       : L=32 circuits= 111, eigenvalues below gap=16 (expect 16)
PGGFPR                : L=32 circuits= 142, eigenvalues below gap=16 (expect 16)
```

###  Initial digging in

Agent started digging into this, but it got into the weeds very quickly. @coreyostrove, would love your insight on this.

TL;DR: There are 18 amplifiable directions. PGGFPR seems to prefer short germs for parameter amplification, while greedy PFPR prefers using the longer germs. This ends up meaning that we are adding a lot more circuits at low depth for PGGFPR; specifically, we are adding more circuits that happen to not overlap with the base LGST circuits for PGGFPR, whereas the few low-germ-power fid pairs for PFPR almost overlap with LGST circuits completely.

TL;DR: This is a low-germ-power effect. PGGFPR will actually be more efficient than PFPR eventually at sufficiently (although in this case unreasonably) high L. So this is just an unfortunate showcase. Unless there is something that could be down to get the low-L savings also in teh PGGFPR case.

```python
import pygsti
import pygsti.circuits as pc
from pygsti.modelpacks import smq1Q_XY

target_model = smq1Q_XY.target_model()
prep_fiducials = smq1Q_XY.prep_fiducials()
meas_fiducials = smq1Q_XY.meas_fiducials()
germs = smq1Q_XY.germs()
maxLengths = [1, 2, 4, 8, 16, 32]
opLabels = list(target_model.operations.keys())

germ_vecs, _ = pygsti.alg.germ_set_spanning_vectors(target_model, germs, float_type=np.double, verbosity=0)
d_pggfpr = pygsti.alg.find_sufficient_fiducial_pairs_per_germ_global(
    target_model, prep_fiducials, meas_fiducials, germ_vector_spanning_set=germ_vecs, verbosity=0)
d_pfpr_greedy = pygsti.alg.find_sufficient_fiducial_pairs_per_germ_greedy(
    target_model, prep_fiducials, meas_fiducials, germs, seed=1234, verbosity=0)

# PGGFPR loads the most pairs onto the shortest (most-repeated) germs; PFPR-greedy does the opposite.
def per_germ_cost(fid_dict, name):
    print(f"--- {name} ---")
    total = 0
    for germ in germs:
        pairs = fid_dict[germ]
        structs = pc.create_lsgst_circuit_lists(opLabels, prep_fiducials, meas_fiducials, [germ], maxLengths, fid_pairs={germ: pairs})
        marginal = len(structs[-1]) - 56  # subtract the always-shared 56-circuit LGST base
        total += marginal
        print(f"  germ len={len(germ)}: {len(pairs)} pairs -> {marginal} circuits beyond the shared base")
    print(f"  TOTAL circuits beyond shared base: {total} (+ 56 shared base = {total + 56} L=32 circuits)")

per_germ_cost(d_pggfpr, "PGGFPR")
per_germ_cost(d_pfpr_greedy, "PFPR (greedy)")
```

Output:

```
--- PGGFPR ---
  germ len=1: 6 pairs -> 29 circuits beyond the shared base
  germ len=1: 6 pairs -> 30 circuits beyond the shared base
  germ len=2: 4 pairs -> 19 circuits beyond the shared base
  germ len=3: 2 pairs -> 8 circuits beyond the shared base
  TOTAL circuits beyond shared base: 86 (+ 56 shared base = 142 L=32 circuits)
--- PFPR (greedy) ---
  germ len=1: 4 pairs -> 11 circuits beyond the shared base
  germ len=1: 4 pairs -> 16 circuits beyond the shared base
  germ len=2: 5 pairs -> 21 circuits beyond the shared base
  germ len=3: 7 pairs -> 23 circuits beyond the shared base
  TOTAL circuits beyond shared base: 71 (+ 56 shared base = 127 L=32 circuits)
```

```python
import numpy as np
import pygsti
import pygsti.circuits as pc
from pygsti.modelpacks import smq1Q_XY

target_model = smq1Q_XY.target_model()
prep_fiducials = smq1Q_XY.prep_fiducials()
meas_fiducials = smq1Q_XY.meas_fiducials()
germs = smq1Q_XY.germs()
maxLengths = [1, 2, 4, 8, 16, 32]
opLabels = list(target_model.operations.keys())
pspec = target_model.create_processor_spec()

germ_vecs, _ = pygsti.alg.germ_set_spanning_vectors(target_model, germs, float_type=np.double, verbosity=0)
d_pggfpr = pygsti.alg.find_sufficient_fiducial_pairs_per_germ_global(
    target_model, prep_fiducials, meas_fiducials, germ_vector_spanning_set=germ_vecs, verbosity=0)
d_pfpr_greedy = pygsti.alg.find_sufficient_fiducial_pairs_per_germ_greedy(
    target_model, prep_fiducials, meas_fiducials, germs, seed=1234, verbosity=0)

def stage_totals(d):
    totals, news, prev = [], [], None
    for i, L in enumerate(maxLengths, start=1):
        structs = pc.create_lsgst_circuit_lists(opLabels, prep_fiducials, meas_fiducials, germs, maxLengths[:i], fid_pairs=d)
        design = pygsti.protocols.GateSetTomographyDesign(pspec, structs)
        total = len(design.all_circuits_needing_data)
        news.append(total - prev if prev is not None else 0)
        totals.append(total)
        prev = total
    return totals, news

LGST_COUNT = 56  # the L=1 stage: always these same 56 circuits, kept regardless of FPR method
pfpr_totals, pfpr_news = stage_totals(d_pfpr_greedy)
pggfpr_totals, pggfpr_news = stage_totals(d_pggfpr)

print("--- PFPR (greedy) ---")
for L, total, new in zip(maxLengths, pfpr_totals, pfpr_news):
    print(f"  L={L:>3d}: total={total:4d}  (LGST={LGST_COUNT}, beyond-LGST={total - LGST_COUNT:3d})  new-this-stage={new}")

# Per-stage diff vs. PFPR (PGGFPR minus PFPR); confirm the diffs sum to the real total gap between the two edesigns.
print("--- PGGFPR ---")
diffs = []
for L, total, new, pf_new in zip(maxLengths, pggfpr_totals, pggfpr_news, pfpr_news):
    diff = new - pf_new
    diffs.append(diff)
    print(f"  L={L:>3d}: total={total:4d}  (LGST={LGST_COUNT}, beyond-LGST={total - LGST_COUNT:3d})  new-this-stage={new}  (vs PFPR: {'+' if diff >= 0 else ''}{diff})")

print()
print("Sum of per-stage diffs:", sum(diffs))
print("Actual gap (PGGFPR total - PFPR total):", pggfpr_totals[-1] - pfpr_totals[-1])
```

Output:

```
--- PFPR (greedy) ---
  L=  1: total=  56  (LGST=56, beyond-LGST=  0)  new-this-stage=0
  L=  2: total=  60  (LGST=56, beyond-LGST=  4)  new-this-stage=4
  L=  4: total=  70  (LGST=56, beyond-LGST= 14)  new-this-stage=10
  L=  8: total=  89  (LGST=56, beyond-LGST= 33)  new-this-stage=19
  L= 16: total= 108  (LGST=56, beyond-LGST= 52)  new-this-stage=19
  L= 32: total= 127  (LGST=56, beyond-LGST= 71)  new-this-stage=19
--- PGGFPR ---
  L=  1: total=  56  (LGST=56, beyond-LGST=  0)  new-this-stage=0  (vs PFPR: +0)
  L=  2: total=  70  (LGST=56, beyond-LGST= 14)  new-this-stage=14  (vs PFPR: +10)
  L=  4: total=  88  (LGST=56, beyond-LGST= 32)  new-this-stage=18  (vs PFPR: +8)
  L=  8: total= 106  (LGST=56, beyond-LGST= 50)  new-this-stage=18  (vs PFPR: -1)
  L= 16: total= 124  (LGST=56, beyond-LGST= 68)  new-this-stage=18  (vs PFPR: -1)
  L= 32: total= 142  (LGST=56, beyond-LGST= 86)  new-this-stage=18  (vs PFPR: -1)

Sum of per-stage diffs: 15
Actual gap (PGGFPR total - PFPR total): 15
```

sserita (changes requested):
A lot of my comments are on a high-level. I'll try to add line-specific things in the review. LMK if you need this in another format (including a MD that logs all investigations/decisions that are being reported out here):

## Decision-level review
- Old doc URLS will 404: 👍 
- Drop Notebook 6.x support: 👍 
- Hosted docs only reassembled on green CI run: 👍

## Proposed TOC

There were a couple of what I felt where characterization mismatches. Happy to discuss further:

- Model packs in Start here seemed excessive. Everything an average user needs to know about modelpacks is in the relevant tutorials (i.e. getting models and edesigns from them).
- Similarly, Explicit Models (which is also sort of a misnomer as a category header since the implicit models notebook is under it) doesn't need to be in characterization guides. At no point do the tutorials do crazy ExplicitModel creation in the guides here.
- IBMQ running is not actually a new/different characterization - it's just a way to run one through an external platform. It potentially belongs with the other interop notebook for Cirq. (I expect possible pushback here, not a hill I'll die on but worth considering)
- Simulating RB data is probably better served under the RB guide rather than under forward simulation.
- Obviously the FAQ, but also I think the Cite PyGSTi page, are misplaced. I propose renaming the last section to Reference and putting these as siblings to the API documentation (API docs last so we don't lose the FAQ, etc when API expands).
- Finally, I think it is worth splitting Advanced Capabilities from Internal. You can probably do most of the specialist protocols without knowing TOO much more about the deep internals. And grouping all internal together makes sense to me as a category.

All in all, I propose the following split:

```
Start here (unchanged except one removal)
- Index, Install, ChooseAProtocol, FirstGST, FirstBenchmark, YourOwnData, ReadingResults, KeyConcepts
- TargetModels → moved to Internals
Characterization guides (three removals, no additions)
- workflow/ — unchanged
- models/ — Models.md ("Explicit models") removed → moved to Internals. MultiQubitModels and ModelNoise stay here, but this chapter now needs a new landing page since Models.md was it (flagging this as a real gap for whoever implements it, not something I've resolved — could be a short new stub, or promoting/retitling one of the two survivors).
- gst/ — unchanged, four chapters as before
- rb/ — gains SimulatingRBData (moved from advanced/simulation/)
- benchmarks/, drift/ — unchanged
- analysis/ — loses Troubleshooting (FAQ) → moved to Reference. Remaining children: Reports, Metrics, GaugeFreedom, ErrorBars, ModelTesting.
- hardware/IBMQ → moved to Internals (Capabilities)
Advanced capabilities (new part, half of the old "Advanced topics and internals")
- specialist/Index (RPE, Idle Tomography, Parity Benchmarking)
- ml/QPANN
- interop/Cirq
- interop/IBMQ — moved in from guides/hardware/, as a sibling top-level entry next to Cirq, not merged under a shared "interop" landing page
Internals (new part, half of the old "Advanced topics and internals")
- models/Operators + CustomOperators, CustomPOVMs, TyingParameters
- conventions/Bases + StateSpaces
- simulation/ForwardSimulators + StabilizerCHP, ErrorGeneratorPropagation, ErrorGeneratorPolynomials
- extending/LowLevelGST + GaugeOptReference
- figures/WorkspaceFigures + Switchboards
- TargetModels (Model packs) — moved in from start/
- Models (Explicit models) — moved in from guides/models/
Reference (renamed from "API reference")
- api.rst (unchanged)
- Troubleshooting (FAQ) — moved in from guides/analysis/
- Citing — moved in from the tail of advanced/
- FromFunctionAPI (migration table) — moved in from advanced/migration/ (this could stay in Advanced capabilities also)
```

docs/markdown/guides/gst/FewerCircuits.md
sserita: Apologies in advance for you trying to get through this comment. 😅

.github/workflows/docs-staging.yml
rileyjmurray: I think it's worth keeping docs-staging around even after the merge. It'll be useful for future revisions of the website. Open to removing though.

docs/markdown/guides/gst/TwoQubitGST.md
rileyjmurray: The "this" you refer to is explaining the tolerance?

docs/stage-rtd-branch.py
rileyjmurray: I'm reasonably confident that we'll want some locally executable tooling to help with RTD. Reason being: the runtimes seen in the notebooks are really misleading when run on crappy CI machines. I'm going to want the `master` web docs to include notebook output generated on Peregrine. I'll probably modify the notebooks to be beefier along the way. Maybe they'll use some machine introspection to decide quantities like GST max-max-lengths and then print those as outputs.

pygsti/tools/sdptools.py
rileyjmurray: I don't think we need dedicated tests. It earned its keep by silencing warnings we legitimately encountered. If someone wants to try and remove years down the line then they're welcome to do so.

sserita: Agreed, just thought I'd mention it

docs/markdown/guides/gst/TwoQubitGST.md
sserita: Yes, both loosening and then explaining in prose. I just think the extra comment line makes it more likely that those who skip prose also recognize the non-standard loosening

docs/markdown/guides/gst/FewerCircuits.md
sserita: My extra TL;DR: This is probably not blocking for this PR. The question is whether it rises to the level of a new issue, or if this is a known thing.

PR #896, opened 2026-08-21
Keep protocol checkpoints out of the source tree during tests

rileyjmurray:
### Problem

A full run of `test/unit`, `test/integration` and `test/test_packages` leaves checkpoint JSON scattered around the checkout. Adding the notebook suite brings the total to 26 directories and ~155 MB. Everything is gitignored, so none of it was ever at risk of being committed, but it's a lot of disk and it makes the working tree tedious to look at.

`GateSetTomography`, `StandardGST`, `ModelTest` and `IBMQExperiment` each default their checkpoint location to a directory relative to the *current working directory* — `./gst_checkpoints/`, `./standard_gst_checkpoints/`, `./model_test_checkpoints/`, `./ibmqexperiment_checkpoint/`. That's a fine default for someone running a fit interactively. Under pytest it means the artifacts land wherever pytest was invoked from, normally the repo root.

### Solution

This PR uses pytest's `conftest.py` mechanism to effectively override the default checkpoint path (`None`) to a temporary directory that's cleaned up after execution.

Destination is pytest's `--basetemp` when one was given, so checkpoints follow the same lifecycle as everything else pytest writes; otherwise a private `mkdtemp` cleaned up at session end. Each xdist worker runs `pytest_configure` separately and gets its own directory, so there's no contention. `--checkpoint-redirect=off` disables the whole thing, and `PYGSTI_TEST_KEEP_CHECKPOINTS=1` keeps the directory around for inspection. The session header reports where things went:

```
protocol checkpoints redirected to: /tmp/pygsti-test-checkpoints-msw2sb6e
```

This PR also changes an existing file `docs/conftest.py` written by a previous agent. That file existed to handle the fact that some notebooks required shared artifacts, but we don't want to have to run the notebooks in a specific order during tests. This PR's change is to pass `disable_checkpointing=True`. That's the right move since none of the example notebooks use the checkpoints as intentional artifacts.

This PR _does not_ cover the rest of the notebook suite's checkpointing artifacts. There's plenty of those, but they all get written to `docs/`, so they aren't cluttering the main repo tree.

### How the robot found the full set

**The following is written by the robot in first person.**

I wrote a pytest plugin that intercepts `Path.mkdir` / `os.mkdir` / `os.makedirs` / write-mode `open`, attributes each write to a test nodeid, and separately diffs the repo tree across a session to catch writes from subprocesses the in-process hooks can't see. Run with its allowlist disabled — counting sanctioned scratch dirs too — it reported for `test/unit` + `test/integration`:

```
   187 writes   31 tests  gst_checkpoints            <- unsanctioned
   138 writes   14 tests  standard_gst_checkpoints   <- unsanctioned
    55 writes    6 tests  model_test_checkpoints     <- unsanctioned
   239 writes    7 tests  test/     (all under test_packages/temp_test_files)
     2 writes    1 tests  .hypothesis
     2 writes    1 tests  .pytest_cache
```

and for `test/test_packages`, 1590 writes under `test/` (1523 of them in `temp_test_files`, 67 in the three checkpoint dirs) plus two to `.pytest_cache`. So checkpoints really are the only artifact escaping into the tree, which is why this PR is scoped to them.

### Notes

Two incidental things that turned up and are *not* addressed here, in case anyone wants them later: seven `test/unit` tests write into `test/test_packages/temp_test_files/`, which is a unit test reaching into the legacy suite's scratch directory; and nothing ever cleans `temp_test_files/` up, so it just accumulates and gets overwritten.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,137 @@
+"""Repo-wide pytest configuration.
+
+Keeps protocol checkpoints out of the source tree during test runs.
+
+``GateSetTomography``, ``StandardGST``, ``ModelTest`` and ``IBMQExperiment``
+each default their checkpoint location to a directory *relative to the current
+working directory* (``./gst_checkpoints/``, ``./standard_gst_checkpoints/``,
+``./model_test_checkpoints/``, ``./ibmqexperiment_checkpoint/``). That's a
+reasonable default for someone running a fit interactively, but under pytest it
+means ~50 tests deposit checkpoint JSON wherever pytest happened to be invoked
+from — normally the repo root — and a full notebook run scatters ~150 MB across
+the ``docs/markdown`` source directories.
+
+This plugin redirects only the *default*: when a caller passes
+``checkpoint_path=None`` (i.e. expresses no opinion) and hasn't disabled
+checkpointing, the path is rewritten into a throwaway directory outside the
+source tree. Callers that pass an explicit ``checkpoint_path`` are untouched, so
+the tests that deliberately exercise checkpoint write/resume — notably
+``test/test_packages/drivers/test_drivers.py`` — keep testing exactly what they
+tested before. Checkpointing itself still runs; only the destination changes.
+
+Disable with ``--checkpoint-redirect=off``. Set
+``PYGSTI_TEST_KEEP_CHECKPOINTS=1`` to leave the directory behind for inspection
+instead of deleting it at session end.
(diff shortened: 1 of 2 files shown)


coreyostrove (commented):
This sort of meta-programming goes over my head. I will assume you have a handle on this, @rileyjmurray.

PR #897, opened 2026-08-21
Bugfix Approx Logarithm Failures

coreyostrove:
# Fix spuriously large error bars on pi-rotation gate angles

## Problem

This problem was found while investigating an issue reported by @pcwysoc in which anamolously large error bars for the rotation angle of a Hadamard gate were reported by the general_decomposition function in `pygsti.report.reportables`, and found in the gate decompositions table of GST reports. This problem was isolated to the failure of the computation of the matrix logarithm, and the underlying pathology will generally hold for gates whose target has a `-1` eigenvalue (any pi rotation: Hadamard, X(pi), CNOT, CZ). In the case reported by @pcwysoc error bars up to three orders-of-magnitude too large were reported.

For such a target to have a real-valued logarithm, as is required for extracting the rotation angle, it must have negative eigenvalues that are paired (See Walter J. Culver Proceedings of the American Mathematical Society, Vol. 17, No. 5. (Oct., 1966), pp. 1146-1151). This is true at the target, but for the noisy estimated gate this degeneracy is broken,  so generically it has no real logarithm at all. The original fallback for this case was to use the function  `pygsti.tools.matrixtools.approximate_matrix_log`, which used an L-BFGS-B optimization to find approximate real-valued logarithms. This function was failing, and this was being reported as a large value of the 'log inexactness' reported by `general_decomposition`, which corresponds to the error picked up from a round trip through the logarithm and re-exponentiation. Additionally, the output of this optimization was not numerically stable in general, and large jumps in the solution would result in finite differences computed by `ConfidenceRegionFactoryView` as part of error bar construction to blow up.

## Fix
New `matrixtools.approximate_matrix_log_BCH`: take the logarithm of the near-identity error map `inv(target)@ m`, which is unique and smooth, then recombine with the analytically known `log(target)` via BCH. Truncated BCH is a polynomial in the matrix entries, so it is smooth by construction. Used in the `-1`-eigenvalue branch of `general_decomposition`; other gates are untouched.

On a user-reported Hadamard estimate (95% CI):

| | before | after |
|---|---|---|
| angle |0.9999583578541491 ± 0.12358208448741723 | 0.9999652938476625 ± 7.52e-05 |
| log inexactness |0.002597049819903578 ± 4.75471751032029 | 0.0021073765899969656 ± 2.25e-04 |
| runtime | 12.1 s | 0.8 s |

(While running these tests I also just notices the error bar pathologies unsurprisingly extended to the rotation axis uncertainty. From before the fix 'Gh:0 axis': ReportableQty([ 7.06958993e-01 -1.13832251e-05  7.07254538e-01] +/- [1.4709484  0.10589819 1.47032276]).)

Another symptom of this issue was instability in the value of finite difference derivatives with step size eps.  `d(angle)/d(eps)` is now stable to 6 significant figures across four decades of step size.

## Notes

- **Order is hardcoded to 3.** Order empirically found not to help. Order 5 found empirically to be unsafe, since BCH converges only while the spectral radius of `ad_log(target)` is under `2*pi` — exactly attained by a single-qubit pi rotation and exceeded by some multi-qubit targets. In testing nonmonotonicity was observed.

## Tests

`test/unit/report/test_reportables.py` (new) pins the reported gate in a one-gate model, reproducing the full pipeline; reverting the fix fails 3 of its 4 regression tests. `test/unit/tools/test_matrixtools.py` adds round-trip, order-convergence, domain-guard, and order-validation coverage.

### DIFF
--- a/test/unit/tools/test_matrixtools.py
+++ b/test/unit/tools/test_matrixtools.py
@@ -8,53 +8,7 @@ import pygsti.tools.matrixtools as mt
 from ..util import BaseCase
 
 
-# Hamiltonian generators in the normalized Pauli-product basis, as real superoperator generators.
-_HX = np.zeros((4, 4)); _HX[2, 3] = -1; _HX[3, 2] = 1
-_HZ = np.zeros((4, 4)); _HZ[1, 2] = -1; _HZ[2, 1] = 1
-
-
-def _pi_rotation_superop():
-    """X(pi) as a superoperator in the normalized Pauli-product basis: eigenvalues (1, 1, -1, -1)."""
-    return np.diag([1.0, 1.0, -1.0, -1.0])
-
-
-def _pi_over_2_rotation_superop():
-    """Z(pi/2) as a superoperator; eigenvalues (1, 1, +i, -i), so its real logarithm is unique."""
-    return spl.expm((np.pi / 2) * _HZ)
-
-
-def _anisotropically_damped(target, rates=(0.01, 0.02, 0.03)):
-    """Damp the three Pauli axes at *different* rates.
-
-    Applied to a pi-rotation this splits the doubly-degenerate -1 eigenvalue into two distinct
-    negative real eigenvalues, so the result has no real matrix logarithm at all.
-    """
-    return np.diag([1.0] + [1.0 - r for r in rates]) @ target
-
-
-def _coherently_perturbed(target, scale=1.0):
-    """Apply a real, trace-preserving error generator with a coherent part that does not commute
-    with log(target), so that the BCH truncation error is actually visible."""
-    L = scale * (0.3 * _HZ + 0.15 * _HX + np.diag([0.0, -0.01, -0.02, -0.03]))
-    return spl.expm(L) @ target
-
-
-# A Hadamard gate estimate and its target, taken from a failing case reported by an end user. The
-# estimate has two distinct real negative eigenvalues and so has no real matrix logarithm.
-_HADAMARD_ESTIMATE = np.array([
-    [1.0, 0.0, 0.0, 0.0],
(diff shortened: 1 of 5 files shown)


test/unit/tools/test_matrixtools.py
```diff
@@ -6,6 +8,52 @@
...
+    [-0.00047897732274857033, -0.0026434875205955393, -0.9881533456738786, -0.0028483281874282394],
+    [-0.000516711221855366, 0.9890358028125664, 0.0028055709936753892, 0.00026052111853187057]])
+
+_HADAMARD_TARGET = np.array([
+    [1.0, 0.0, 0.0, 0.0],
+    [0.0, 0.0, 0.0, 1.0],
+    [0.0, 0.0, -1.0, 0.0],
+    [0.0, 1.0, 0.0, 0.0]])
+
+
```
rileyjmurray: I'm not a fan of introducing so many top-level symbols in this file, especially at the start of the file.

I suggest splitting the BCH-related tests off into their own class that comes after `MatrixToolsTester`. We can rename `MatrixToolsTester` to `GeneralMatrixToolsTester` if you'd like.

Once split off, we can stuff these free functions into static methods of the dedicated test class.

rileyjmurray (changes requested):
This is extremely cool! I have one requested organizational change.

test/unit/tools/test_matrixtools.py
coreyostrove: Sounds good, I'll report back once I have made these changes.

rileyjmurray: @coreyostrove I went ahead and applied the fixes. Will merge once tests pass.

Issue #870, opened 2026-08-07
LocalNoiseModel: enforce Label prefixes in OrderedMemberDicts


rileyjmurray:
This issue concerns how we fetch model members from LocalNoiseModels. It's cumbersome to do that by comparison to ExplicitOpModels. A partial fix will be included in PR #735. I'm filing this issue for follow-up.

#### OrderedMemberDict

Both ExplicitOpModel and LocalNoiseModel store their model members in a pygsti-defined data structure called OrderedMemberDict.

These data structures have special rules for their keys. 

The first rule concerns indexing that triggers `__getitem__`, `__setitem__`, or `__del__`. In any of these functions you can pass a string, or tuple, or Label. However, the first thing that the dispatched function does is convert the would-be key into a Label. So, the contents of `OrderedMemberDict.keys()` are always Labels.

The second rule concerns the `prefix` argument passed to its `__init__`.
```python
    def __init__(self, parent, default_param, prefix, flags, items=None):
```
The docstring for `__init__` says that `prefix` must be a string. However, the actual implementation lets `prefix` be a string _or_ None. When non-None, all Label objects in `OrderedMemberDict.keys()` are constrained so that `Label.has_prefix(prefix)` is True. No constraints are applied when `prefix is None`.

#### ExplicitOpModel

The ExplicitOpModel constructor sets its OrderedMemberDicts as follows. I only want to draw attention to the `prefix` arguments, which are set by variables`prep_prefix`, `povm_prefix`, `gate_prefix`, and `instrument_prefix`. These default to "rho", "M", "G", and "I".
```python
        self.preps = _OrderedMemberDict(self, default_prep_type, prep_prefix, flagfn("state"))
        self.povms = _OrderedMemberDict(self, default_povm_type, povm_prefix, flagfn("povm"))
        self.operations = _OrderedMemberDict(self, default_gate_type, gate_prefix, flagfn("operation"))
        self.instruments = _OrderedMemberDict(self, default_instrument_type, instrument_prefix, flagfn("instrument"))
        self.factories = _OrderedMemberDict(self, default_gate_type, gate_prefix,
                                            {'auto_embed': True, 'match_parent_statespace': True,
                                             'match_parent_evotype': True, 'cast_to_type': "factory"})
```
We can fetch a model member from an ExplicitOpModel by indexing into it. Here's the (slightly compacted) current code for that.
```python
    def __getitem__(self, label):
        if self._strict:  # a static class variable, set in source to False.
[...]

PR #903, opened 2026-09-01
Issue 870 prefixes

niketh-keshavan:
**summary**

closes https://github.com/sandialabs/pyGSTi/issues/870

- allow `OrderedMemberDict` prefixes to be string, tuple of strings, or `None`
- enforce namespaces (`rho`, `M`, `G`, `I`) across all `LocalNoiseModel` member dictionaries
- preserve implicit operation labels
- apply the same prefix policies across construction, copying, pickling, and nice deserialization
- add regression coverage for the above ^

**testing**

- issue-specific tests: all 24 passes
- model-construction tests: 50 passed, 4 pre-existing skips
- remaining failures reproduce on the current `upstream/develop`

These contributions are under the terms of the Apache 2.0 license.


test/unit/objects/test_implicitmodel.py
```diff
@@ -41,7 +41,9 @@ def _test_getitem(base: BaseCase, m: ImplicitOpModel):
         base.assertIs( m['Mdefault'],    m.povm_blks['layers']['Mdefault']         )
         base.assertIs( m['Gx'],          m.operation_blks['gates']['Gx']           )
         base.assertIs( m[('Gx', 'qb0')], m.operation_blks['layers'][('Gx', 'qb0')] )
-        base.assertIs( m['Gx:qb0'],      m.operation_blks['layers'][('Gx', 'qb0')] )
+        base.assertIs( m[Label(('Gx', 'qb0'))], m.operation_blks['layers'][('Gx', 'qb0')] )
```
rileyjmurray, line 45: Does this PR make the ` m['Gx:qb0']` syntax unsupported?

pygsti/models/memberdict.py
```diff
@@ -42,15 +42,21 @@ def __init__(self, prefix, items=None):
...
-                           "beginning with the prefix '%s'" % self._prefix)
+        if self._prefix is not None:
+            if isinstance(self._prefix, tuple):
+                has_prefix = any(key.has_prefix(p) for p in self._prefix)
+                prefix_desc = "one of the prefixes " + ", ".join(f"'{p}'" for p in self._prefix)
+            else:
+                has_prefix = key.has_prefix(self._prefix)
+                prefix_desc = "the prefix '%s'" % self._prefix
+            if not has_prefix:
+                raise KeyError("All keys must begin with %s" % prefix_desc)
```
rileyjmurray: Please check that the docstring for this function is correct. If it is, then please update its type annotation for `key`.

pygsti/models/localnoisemodel.py
```diff
@@ -170,21 +193,7 @@ def __init__(self, processor_spec, gatedict, prep_layers=None, povm_layers=None,
...
-        self.povm_blks['layers'] = _OrderedMemberDict(self, None, None, flags)
-        self.operation_blks['gates'] = _OrderedMemberDict(self, None, None, flags)
-        self.operation_blks['layers'] = _OrderedMemberDict(self, None, None, flags)
-        self.instrument_blks['layers'] = _OrderedMemberDict(self, None, None, flags)
-        # ^ Unclear why instrument_blks should only be keyed by `layers`.
-        #
-        #     I'll grant that it seems weird to key by `gates`, but the things
-        #     stored in operation_blks['gates'] have direct instrument analogs
-        #     that aren't suitable for instrument_blks['layers'].
-        #
```
rileyjmurray, line 185: This comment technically still applies. Please move it to just below the corresponding line in the definition of the `_LOCALNOISE_MEMBER_PREFIXES` dict literal.

pygsti/models/localnoisemodel.py
```diff
@@ -33,6 +33,29 @@
...
+def _init_localnoisemodel_member_dicts(mdl, modelmembers=None):
+    """Initialize the seven standard member dictionaries of a `LocalNoiseModel`."""
+    flags = {'auto_embed': False, 'match_parent_statespace': False,
+             'match_parent_evotype': True, 'cast_to_type': None}
+    for blk_name, key, prefix in _LOCALNOISE_MEMBER_PREFIXES:
+        blk = getattr(mdl, blk_name)
+        items = modelmembers.get(f'{blk_name}|{key}', []) if modelmembers is not None else []
+        blk[key] = _OrderedMemberDict(mdl, None, prefix, flags, items)
+
+
```
rileyjmurray: I think I'd prefer for these to be static members of the LocalNoiseModel class. Is there a reason to define at the module level?

pygsti/models/localnoisemodel.py
```diff
@@ -33,6 +33,29 @@
...
+def _init_localnoisemodel_member_dicts(mdl, modelmembers=None):
+    """Initialize the seven standard member dictionaries of a `LocalNoiseModel`."""
+    flags = {'auto_embed': False, 'match_parent_statespace': False,
+             'match_parent_evotype': True, 'cast_to_type': None}
+    for blk_name, key, prefix in _LOCALNOISE_MEMBER_PREFIXES:
+        blk = getattr(mdl, blk_name)
+        items = modelmembers.get(f'{blk_name}|{key}', []) if modelmembers is not None else []
+        blk[key] = _OrderedMemberDict(mdl, None, prefix, flags, items)
```
rileyjmurray: This is honestly pretty hard to read. Please reduce the number of logical operations that are happening on a given line.

rileyjmurray: The variable name here is a little misleading. Most of the dicts are indeed suffixed with `_blk`, but that's not universally true (see `factories`). Please rename `blk` and `blk_*` variables to `mmdict` and `mmdict_*`.

rileyjmurray (changes requested):
This is an excellent start! I have a few comments, including questions and requested changes.

test/unit/objects/test_implicitmodel.py
niketh-keshavan: `ImplicitOpModel.__getitem__` should take string keys -> label keys so `'Gx:qb0'` -> `Label(('Gx', 'qb0'))` but I should revert the test case

pygsti/models/localnoisemodel.py
niketh-keshavan: No specific reason, moved into `LocalNoiseModel`

pygsti/models/localnoisemodel.py
```diff
@@ -136,6 +136,37 @@ class LocalNoiseModel(_ImplicitOpModel):
...
+        flags = {'auto_embed': False, 'match_parent_statespace': False,
+                 'match_parent_evotype': True, 'cast_to_type': None}
+        if modelmembers is None:
+            modelmembers = {}
+
+        for mmdict_name, key, prefix in self._member_prefixes:
+            mmdict = getattr(self, mmdict_name)
+            serialization_key = f'{mmdict_name}|{key}'
+            items = modelmembers.get(serialization_key, [])
+            mmdict[key] = _OrderedMemberDict(self, None, prefix, flags, items)
```
rileyjmurray: Shoot. My requested change wasn't right on the semantics. What I asked to rename to `mmdict` isn't actually a dict of modelmembers. It's really a dict-of-dicts. The semantics of the triplets in `_member_prefixes` are something like (`attr_name`, `inner_name`, `prefix_for_inner_keys`). Variable names like that are probably clearer. Please don't feel obligated to use those exact names though.

Some in-line type annotations would be helpful here. **Setting aside the variable name question**, the line
```python
mmdict = getattr(self, mmdict_name)
```
could be annotated as
```python
mmdict : dict[str, _OrderedMemberDict] = getattr(self, mmdict_name)
```

Sorry for the hassle on this point. It's just that when things like `getattr` start getting thrown around I want to try extra hard to keep code easy to read.

niketh-keshavan: yeah makes sense =D

two options I can think of:

1. keep the metadata loop how it is (centralized `_member_prefixes` tuple) but just rename the loop variables more explicitly and add type annotation on `getattr`
2. maybe more extreme, but I could see completely eliminating `getattr` and just iterating over the bound dictionary elements themselves

~~lmk which one you prefer and I'll get on it~~
going with 1

rileyjmurray: Whoops! I saw your message yesterday, but I failed to respond. Either course of action sounds reasonable to me. I see you've gone ahead with 1. We'll just do that.

rileyjmurray (changes requested):
Okay, the code looks good. Can you please update the changelog? Cite this PR number with a note under the "changed" heading for the next release.

rileyjmurray (approved):
Approved! I'll wait for either @coreyostrove or @sserita to also approve before merge.

PR #904, opened 2026-09-01
Switch IBMQExperiment.submit() default from Session to Batch

### LINKED ISSUE #899: Switch from IBMQuantumSession to IBMQuantumBatch
We should switch ibmqexperiment.py to use an IBMQuantumBatch not an IBMQuantumSession by default. If there's a lot of IBM-side processing (e.g., out of the box dynamical decoupling) the session hangs and causes us to be charged for far more QPU usage than we actually use.

sserita:
Closes [#899](#899); folds in several related fixes raised in that issue's discussion thread.

A few questions for @pcwysoc:

1. What did you have in mind with the `IGate` comment on the issue?
1. DD motivated this issue, but there were no asks about that. Did you want that addressed here as well, and if so, how?
1. Mid-circuit-measurement handling (`measure_2` stripped from basis gates, `ConvertToMidCircuitMeasure` pass) looks fully in place since [#679](#679), with dedicated tests passing. Does that cover it, or is something still broken here?

## Session -> Batch default, with backward compatibility

`submit()` now defaults to creating a `Batch` runtime-mode object instead of a `Session`, since `Batch` mode only bills for actual QPU time, not IBM-side preprocessing. A new `ibmq_runtime_mode` parameter accepts either a `Session` or `Batch`; the older `ibmq_session` parameter still works but is deprecated and just forwards to it:

```python
# pygsti/extras/ibmq/ibmqexperiment.py:532-536
caller_supplied_mode = ibmq_runtime_mode is not None
if ibmq_runtime_mode is None:
    ibmq_runtime_mode = _Batch(backend=ibmq_backend)

sampler = _Sampler(mode=ibmq_runtime_mode)
```

A self-created runtime mode is also closed once submission finishes; a caller-supplied one is left for the caller to manage:

```python
# pygsti/extras/ibmq/ibmqexperiment.py:643-645
if not caller_supplied_mode:
    ibmq_runtime_mode.close()
```

## Reduced checkpointing overhead in `retrieve_results()`

Checkpointing here has gone through two rounds: it originally wrote after every retrieved result (slow due to sheer frequency), which [#679](#679) fixed by moving the write to once at the end of the method -- but that end-of-method write used the full `self.write()` (re-serializing the entire on-disk state), regressing per-call cost.

This combines both fixes: still one write at the end, but only the newly-retrieved data, via a new `checkpointing_mode` parameter (`"data"` by default):

```python
# pygsti/extras/ibmq/ibmqexperiment.py:418-422
if checkpointing_mode == "data":
    self.data.write(self.checkpoint_path, edesign_already_written=True)
elif checkpointing_mode == "full":
    self.write()
elif checkpointing_mode == "none":
    ...
```

`"full"` restores the post-[#679](#679) always-full-write behavior; `"none"` skips the checkpoint write entirely, since `submit()` already persists job IDs and calibration data incrementally.

## Batch sizing: raised default, hard-limit validation, opt-in `"auto"` mode

`circuits_per_batch`'s old default of `75` came from a per-job circuit-count cap no longer in IBM's current docs.

**The default is now `3000`**, and `transpile()` validates every batch against IBM's actual published hard limits before submission:

```python
# pygsti/extras/ibmq/ibmqexperiment.py:836-843
execution_count = len(batch) * self.num_shots
if execution_count > MAX_EXECUTIONS_PER_JOB:
    raise ValueError(
        f"Batch {batch_idx} would submit {execution_count:,} executions "
        f"({len(batch)} circuits x {self.num_shots} shots), exceeding the IBM "
        f"hard limit of {MAX_EXECUTIONS_PER_JOB:,} per job. "
        f"Set ignore_batch_limit_checks=True to bypass this check."
    )
```

`circuits_per_batch` also accepts `"auto"`, which estimates a cost-aware batch size from the backend's own gate-duration data instead of a fixed circuit count, targeting a specific job duration:

```python
# pygsti/extras/ibmq/ibmqexperiment.py:789-792
per_circuit_time = isa_circuit.estimate_duration(ibmq_backend.target) + rep_delay
raw_n = int(TARGET_JOB_DURATION_SECONDS // (self.num_shots * per_circuit_time + CIRCUIT_OVERHEAD_SECONDS))
capped_n = MAX_EXECUTIONS_PER_JOB // self.num_shots
self.circuits_per_batch = max(1, min(raw_n, capped_n))
```

## Testing

Full `test/unit/extras/ibmq/test_ibmqexperiment.py` suite passes (31/31), including new tests for each change above. The tutorial (`docs/markdown/interfaces/IBMQExperiment.md`) is updated to match and re-verified via `pytest --nbval-lax`.


pygsti/extras/ibmq/ibmqexperiment.py
rileyjmurray, line 1: It looks like the `transpile` function is over 300 lines long. Is there a way to define helper functions to encapsulate some of its logic and make the overall function more readable?

pygsti/extras/ibmq/ibmqexperiment.py
```diff
@@ -307,13 +328,35 @@ def monitor(self):
...
+              (ProtocolData's data/ directory), avoiding re-serialization of
+              edesign, processor_spec, and circuit batches. This is fast for
+              large experiments.
+            - "full": Re-serialize the entire on-disk experiment state (the
+              original/legacy behavior prior to this parameter).
+            - "none": Skip checkpoint write entirely. This is safe because
+              `submit()` already persists job_ids.json and calibration data
+              incrementally, so results can always be re-retrieved on demand
+              later once the caller knows the jobs are complete.
+            Defaults to "data".
```
rileyjmurray: Please use type annotation `Literal['data', 'full', 'none']`.

rileyjmurray: I'd prefer that the documentation of `checkpointing_mode` explain the benefit of checkpointing, rather than providing a non-default `'none'` option that's simply described as "safe."

pygsti/extras/ibmq/ibmqexperiment.py
```diff
@@ -307,13 +328,35 @@ def monitor(self):
...
     
-    def retrieve_results(self):
+    def retrieve_results(self, checkpointing_mode="data"):
         """
         Gets the results of the completed jobs from IBM Q, and processes
         them into a pyGSTi DataProtocol object (stored as the key 'data'),
         which can then be used in pyGSTi data analysis routines (e.g., if this
         was a GST experiment, it can input into a GST protocol object that will
         analyze the data).
+
```
rileyjmurray: Maybe mention that checkpointing_mode has no effect if `self.disable_checkpointing` is False.

pygsti/extras/ibmq/ibmqexperiment.py
```diff
@@ -372,10 +415,21 @@ def qiskit_bitstring_to_outcome_label(bs, ordered_data_indices, mcm_dict):
...
+            else:
+                raise ValueError(
+                    f"Invalid checkpointing_mode '{checkpointing_mode}'. "
+                    f"Allowed values are: 'data', 'full', 'none'."
+                )
 
 
-    def submit(self, ibmq_backend, start=None, stop=None, ignore_job_limit=True, wait_time=5, max_attempts=10, ibmq_session=None):
+    def submit(self, ibmq_backend, start=None, stop=None, ignore_job_limit=True, wait_time=5, max_attempts=10,
+               ibmq_session=None, ibmq_runtime_mode=None):
```
rileyjmurray: Type annotations for new function arguments plz

rileyjmurray (changes requested):

pygsti/extras/ibmq/ibmqexperiment.py
sserita: Split out some functions. LMK if you envisioned a different split/it is still not readable enough.

pygsti/extras/ibmq/ibmqexperiment.py
sserita: Docstring updated to mention why I called "none" as safe a bit more explicit

pygsti/extras/ibmq/ibmqexperiment.py
sserita: Done

pygsti/extras/ibmq/ibmqexperiment.py
sserita: The whole file was missing type annotations. Now we have them all :)

rileyjmurray (approved):
Approved!

PR #905, opened 2026-09-03
Tensor-product gauge groups: gauge optimization that acts on each subsystem independently

rileyjmurray:
The following is written by a robot, and edited by me. It's been revised from the original PR description following changes made after a round of review.

----------------

This PR adds `TensorProductGaugeGroup`, a gauge group whose elements are Kronecker products of per-subsystem gauge transformations. The motivating case is local unitary gauge freedom, `U(d)` on each of `k` qudits. When we gauge-optimize a multi-qubit estimate over the full unitary or TP group, the optimizer is free to absorb crosstalk or entangling error into a non-local gauge choice. Restricting to local transformations removes that freedom.

### The abstraction

`TensorProductGaugeGroup(factors, state_space, basis)` takes a(ny) sequence of ordinary `GaugeGroup` objects, one per contiguous run of state-space labels. A two-qubit factor next to a one-qubit factor on a three-qubit space is fine too, as is a mix of qubits and qubits.

An element's transform matrix is `(C_1 S_1 C_1^{-1}) ⊗ ... ⊗ (C_k S_k C_k^{-1})`, where the `S_j` are the factor elements' matrices in their own bases and each `C_j` is a fixed change of basis from factor `j`'s basis to the model basis's component on that factor's labels. This is the same operator as `C (S_1 ⊗ ... ⊗ S_k) C^{-1}` with `C = C_1 ⊗ ... ⊗ C_k`, but only factor-sized operators are ever built. For `pp` on qubits, and for the `TensorProdBasis` objects our model constructors already produce, every `C_j` is the identity and is skipped. The model basis has to be a tensor product over the factors for this to make sense: a `TensorProdBasis` with one component per label, or the builtin `pp`, which splits as `pp(d1 d2) = pp(d1) ⊗ pp(d2)`. The builtin `gm` or `std` of the full dimension on several qudits is rejected (`gm` on two qutrits is not `gm ⊗ gm`), since the change of basis it would need couples the factors and is a dense operator on the whole space.

Feature entry points:
 * Two convenience constructors, `TensorProductGaugeGroup.local_unitary(state_space, basis)` and `.local_tp(...)`, build one factor per state-space label.
 * `ExplicitOpModel.default_gauge_group` now accepts the strings `'local unitary'` and `'local tp'` alongside `'unitary'` and `'tp'`.

### Small supporting changes

- `OpGaugeGroup` exposes a public `state_space` property and `OpGaugeGroupWithBasis` a public `basis` property, so every factor can be interrogated uniformly.
- `UnitaryGaugeGroup` passes its state space into the Lindblad error generator instead of stashing it as an instance attribute, which would now collide with the property.
- The `'tp'` branch of the `default_gauge_group` setter referenced `self.evotypes`, which doesn't exist. It would have raised on first use. Fixed while I was in there.
- `TensorProdBasis` gets a Notes section recording that its transform matrices don't exploit the tensor-product structure. That's why the group does the change of basis factor by factor rather than asking the basis for it.
- `InverseGaugeGroupElement.from_vector` called the wrapped element's `from_vector` with no argument, so it would have raised on first use. Nothing in the gauge-optimization path calls it, but the tensor-product tests exercise inverse elements, so I fixed it and added a test that the vector reaches the wrapped element.

### Testing

Unit tests in `test/unit/objects/test_gaugegroup.py` check the element against `np.kron` of the factor matrices and against `unitary_to_superop` of the product unitary, check the derivative against central finite differences on both the identity and non-identity change-of-basis paths (the latter with factors in `l2p1` on a model in `gm ⊗ gm`), and cover `wrt_filter` ordering, mixed factor types, multi-label factors, the constructor's rejection cases, and nice-serialization round trips.

End-to-end tests in `test/unit/algorithms/test_gaugeopt_correctness.py` use two-qubit full-TP models built from a processor spec. A local-unitary gauge perturbation is recovered by both L-BFGS-B and least squares, including when the only entangler is `Gcnot` on `(1, 0)`; that case checks that the Kronecker order follows the state-space labels and not the gate's target order. The local element is also checked gate by gate against the full-space unitary element for `U_a ⊗ U_b`. Finally, an entangling `ZZ` gauge perturbation is left in place by the local group and removed by the full unitary group, which is the behavioral reason this class exists.

Regression over `test/unit/{objects,algorithms,leakage,protocols}` on py313: 1715 passed, 14 skipped, 1 xfailed.

### What this PR doesn't do

- Factors must act on contiguous runs of labels, in state-space order. A factor on qubits 0 and 2 with another on qubit 1 would need a qudit permutation, analogous to `level_partition` in `DirectSumUnitaryGroup`. The constructor rejects that layout with a message saying so.
- The model basis must be a tensor product over the factors. Factors in one basis and a model in another are fine, since that's what the per-factor `C_j` are for, but a model in the builtin `gm` or `std` of the full dimension, on more than one factor, raises with a message pointing at `Basis.cast('gm', state_space)`. Supporting it would mean a dense change of basis on the full space, which defeats the point of keeping everything factor-sized.
- Only single-block state spaces are supported. Leakage-style spaces (direct sums of tensor products) raise. Composing with `DirectSumUnitaryGroup` is the natural next step, but I didn't want to widen the scope here.
- `set_default_gauge_group_for_member_type` is untouched. It maps member parameterizations to groups, and locality is a property of the state space, not of the member type.

### Where to look

**The change-of-basis direction**, in `TensorProductGaugeGroup._factor_change_of_basis`. For each factor the group sets `C_j = factor_basis.create_transform_matrix(model_basis)`, where `model_basis` is the model basis's component on that factor's labels, and the element then computes `C_j @ S_j @ inv(C_j)` before taking the Kronecker product. Getting `C_j` and `inv(C_j)` swapped is the easy mistake here, and it's invisible on qubits in `pp`, because there `C_j` is the identity and gets skipped entirely. The test that would catch a swap is `test_deriv_wrt_params_with_change_of_basis`: two qutrits with `UnitaryGaugeGroup` factors in `l2p1`, assembled into `gm ⊗ gm`. Each `C_j` is a 9-by-9 orthogonal matrix that isn't the identity, so the swapped version would be `C_j.T @ S_j @ C_j` and would disagree with `unitary_to_superop(kron(U_a, U_b), gm ⊗ gm)`. The same test checks the element against conjugating the whole Kronecker product by the full-space `C_1 ⊗ C_2`, and checks the derivative by finite differences along that path, since the conjugation is applied to every derivative column too. If you'd rather not trust one test, `Basis.create_transform_matrix`'s docstring ("transforms a vector from this basis to `to_basis`") is the contract I'm relying on.

**The label-matching rule**, in `_match_factors_to_label_runs`. This is what decides which state-space labels each factor acts on, and therefore whether the Kronecker order matches the model's basis order. The same rule, minus the warning, lines the model basis's components up with the labels in `_model_bases_for_factor_runs`, so it also decides which piece of the model basis each factor's `C_j` is computed against. The rule is: walk the block's labels left to right, and give each factor the shortest run of unclaimed labels whose Hilbert-space dimensions multiply to that factor's `udim`. Two things to check:

- Every factor consumes at least one label, even a factor of dimension 1. Without that clause a `TrivialGaugeGroup` on a dimension-1 label claimed nothing, the next factor started one label too early, and the whole layout shifted. `test_local_constructors_with_dimension_one_label` covers this on a `(2, 1, 3)` space.
- The rule is greedy and has no notion of which labels a factor "meant" to act on. A dimension-2 factor placed where a dimension-1 label sits will claim that label and the next qubit together, since `1 * 2 = 2`. That's dimensionally consistent and I believe it's the right behavior, but it's the one place where the constructor accepts a layout the caller might not have intended. Mismatches that can't be reconciled this way, and leftover labels after the last factor, both raise with the offending labels named.

### DIFF
--- a/pygsti/models/gaugegroup.py
+++ b/pygsti/models/gaugegroup.py
@@ -27,7 +27,7 @@ from pygsti.baseobjs import (
 )
 from pygsti.baseobjs import _compatibility as _compat
 from pygsti.modelmembers import operations as _op
-from pygsti.baseobjs.basis import Basis as _Basis, BuiltinBasis as _BuiltinBasis
+from pygsti.baseobjs.basis import Basis as _Basis, BuiltinBasis as _BuiltinBasis, TensorProdBasis as _TensorProdBasis
 from pygsti.baseobjs.nicelyserializable import NicelySerializable as _NicelySerializable
 from pygsti.evotypes.evotype import Evotype as _Evotype
 from pygsti.tools.optools import superop_to_unitary, unitary_to_superop
@@ -359,6 +359,17 @@ class OpGaugeGroup(GaugeGroup):
         GaugeGroup.__init__(self, name)
 
     @property
+    def state_space(self) -> _StateSpace:
+        """
+        The state space that elements of this gauge group act on.
+
+        Returns
+        -------
+        StateSpace
+        """
+        return self._operation.state_space
+
+    @property
     def num_params(self) -> int:
         """
         Return the number of parameters (degrees of freedom) of this gauge group.
@@ -415,6 +426,17 @@ class OpGaugeGroupWithBasis(OpGaugeGroup):
     def __init__(self, operation: _LinearOperator, elementcls, name: str, basis):
         self._basis = basis
         super().__init__(operation, elementcls, name)
+
+    @property
+    def basis(self) -> Union[_Basis, str]:
+        """
+        The basis in which elements of this gauge group express their transform matrices.
+
+        Returns
+        -------
+        Basis or str
+        """
+        return self._basis
 
     def _to_nice_serialization(self):
         state = super()._to_nice_serialization()
@@ -802,11 +824,10 @@ class UnitaryGaugeGroup(OpGaugeGroupWithBasis):
             raise ValueError()
         evotype = _Evotype.cast(str(evotype), default_prefer_dense_reps=True)  # since we use deriv_wrt_params
         errgen = _op.LindbladErrorgen.from_operation_matrix(
-            _np.eye(state_space.dim), "H", basis, mx_basis=basis, evotype=evotype
+            _np.eye(state_space.dim), "H", basis, mx_basis=basis, evotype=evotype, state_space=state_space
         )
         operation = _op.ExpErrorgenOp(errgen)
         OpGaugeGroupWithBasis.__init__(self, operation, UnitaryGaugeGroupElement, "Unitary", basis)
-        self.state_space = state_space
 
     def compute_element(self, param_vec) -> UnitaryGaugeGroupElement:
         return super().compute_element(param_vec)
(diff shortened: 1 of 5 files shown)


pygsti/models/gaugegroup.py
```diff
@@ -1434,3 +1455,387 @@ def _from_nice_serialization(cls, state):
         basis = state['basis'] if isinstance(state['basis'], str) else _Basis.from_nice_serialization(state['basis'])
         level_partition = state.get('level_partition', None)
         return cls(subelements, basis, level_partition)
+
+
+def _match_factors_to_label_runs(factor_udims, labels, label_udims):
```
coreyostrove, line 1460: Suggestion. As far as I can tell there is nothing special about udims for the purposes of this matching, and so it should be possible to do this matching with respect to hilbert-schmidt dimension too. The latter would reduce the need to change this to handle cases where udim is not well defined, but dim is, when direct sum state space support eventually gets added. E.g. kite models.

pygsti/models/gaugegroup.py
```diff
@@ -1434,3 +1455,387 @@ def _from_nice_serialization(cls, state):
...
+    Walks the labels of a single tensor-product block in order and gives factor ``j`` the shortest
+    run of not-yet-assigned labels whose Hilbert-space dimensions multiply to ``factor_udims[j]``.
+    Returns a tuple of index tuples, one per factor. Raises ValueError if a factor's dimension
+    cannot be matched this way or if labels are left over after the last factor.
+    """
+    runs = []
+    i = 0
+    for j, fudim in enumerate(factor_udims):
+        start, prod = i, 1
+        while (prod < fudim or i == start) and i < len(labels):  # always consume at least one label
```
coreyostrove, line 1473: In the PR message you indicated an allowed but possibly unexpected edge case where a udim>1 factor could consume a dimension 1 factor. It is fine if you want to allow this, but if a user hits a case where this happens odds are it is because of an accidental error. Please add logic to raise a warning when this is detected.

pygsti/models/gaugegroup.py
```diff
@@ -1434,3 +1455,387 @@ def _from_nice_serialization(cls, state):
...
+        self.factors = factors
+        self.state_space = state_space
+        self.basis = basis
+        self.factor_bases = resolved
+        self._param_dims = _np.array([f.num_params for f in factors], dtype=int)
+        self._num_params = int(_np.sum(self._param_dims))
+        self._change_of_basis = self._compute_change_of_basis(resolved, basis)
+        super().__init__(name)
+
+    def _resolve_factor_basis(self, j, factor, explicit, basis, labels):
```
coreyostrove, line 1560: This isn't public, so doesn't need a docstring, but could you please add a comment explaining what this does? To save a future dev from needed to reverse engineer it.

pygsti/models/gaugegroup.py
```diff
@@ -1434,3 +1455,387 @@ def _from_nice_serialization(cls, state):
...
+                f"Pass `factor_bases` explicitly.") from e
+
+    @staticmethod
+    def _compute_change_of_basis(factor_bases, basis):
+        """
+        The matrix C with S_model = C @ kron(S_1, ..., S_k) @ inv(C), or None when C is the identity.
+        """
+        tp_basis = _TensorProdBasis(factor_bases) if len(factor_bases) > 1 else factor_bases[0]
+        C = tp_basis.create_transform_matrix(basis)
+        if hasattr(C, 'toarray'):
```
coreyostrove, line 1590: I'm not a fan of the hasattr pattern our robot friends love. I happen to know that this is effectively checking whether the transform is a scipy sparse array, but someone unfamiliar with the scipy.sparse API wouldn't know the intention here. This should be replaceable with an instance check for the scipy.sparse.sparray base class (if you want to be sparse array type agnostic).

pygsti/models/gaugegroup.py
```diff
@@ -1434,3 +1455,387 @@ def _from_nice_serialization(cls, state):
...
+        state_space = _StateSpace.cast(state_space)
+        if state_space.num_tensor_product_blocks != 1:
+            raise ValueError("Local gauge groups require a state space with a single tensor-product block.")
+        labels = state_space.tensor_product_blocks_labels[0]
+        udims = state_space.tensor_product_blocks_udimensions[0]
+        if isinstance(basis, _TensorProdBasis) and len(basis.component_bases) == len(labels):
+            bases = list(basis.component_bases)
+        else:
+            name = basis if isinstance(basis, str) else basis.name
+            bases = [_Basis.cast(name, udim ** 2) for udim in udims]
```
coreyostrove, line 1613: Similar comment to the suggestion I made above. Requesting the dimensions directly will probably make this eventually easier to integrate with cases where udim is not well-defined.

pygsti/models/gaugegroup.py
```diff
@@ -1434,3 +1455,387 @@ def _from_nice_serialization(cls, state):
...
+
+    change_of_basis : numpy.ndarray, optional
+        Matrix ``C`` such that the transform matrix in the model's basis is
+        ``C @ kron(S_1, ..., S_k) @ inv(C)``.  ``None`` means the identity.
+    """
+
+    def __init__(self, factor_elements, change_of_basis=None):
+        self.factor_elements = tuple(factor_elements)
+        self._C = change_of_basis
+        self._Cinv = None if change_of_basis is None else _np.linalg.inv(change_of_basis)
```
coreyostrove, line 1726: A few comments regarding this inverse.

1. This should be computed using the relevant `Basis` object's `reverse_transform_matrix` method analogously to the use of `create_transform_matrix` in `_compute_change_of_basis` above. The basis object natively caches this which will reduce recomputation.
2. Given we create a new instance of this object with every `compute_element` call, we should store the inverse `change_of_basis` as an attribute of the `TensorProductGaugeGroup` and add the option to pass it in to this `__init__`. 
3. In typical cases I would expect that C would also have a tensor product structure compatible with the factors.  This is not being taken advantage of, and the use of dense linear algebra will at some point (10ish qubits?) bottleneck the `transform_matrix` method.
4. Corollary to 3. The `TensorProductBasis` should itself be taking advantage of tensor product structure of the component bases in the construction of the transform matrices. As far as I can tell it is still using the base `Basis` classes implementations, which doesn't. This too will create an avoidable bottleneck and should be remedied as part of this process.

test/unit/algorithms/test_gaugeopt_correctness.py
```diff
@@ -342,6 +342,90 @@ def test_lbfgs_frodist(self):
...
+        gg = TensorProductGaugeGroup.local_unitary(target.state_space, target.basis)
+        rng = np.random.default_rng(seed)
+        el = gg.compute_element(strength * rng.normal(size=gg.num_params))
+        model = target.copy()
+        model.transform_inplace(el)
+        model.default_gauge_group = gg
+        check_gate_metrics_are_nontrivial(gate_metrics_dict(model, target), tol=1e-4)
+        return model, el
+
+    def _check_recovery(self, target, cnot_qubits_desc):
```
coreyostrove, line 375: `cnot_qubits_desc` is not used by this method.

test/unit/algorithms/test_gaugeopt_correctness.py
```diff
@@ -342,6 +342,90 @@ def test_lbfgs_frodist(self):
...
+        for seed in range(2):
+            model, _ = self._prep(target, seed)
+            for method, alg_tol, test_tol in (('L-BFGS-B', 1e-7, 1e-5), ('ls', 1e-15, 1e-6)):
+                out = gop.gaugeopt_to_target(model, target, method=method, tol=alg_tol,
+                                             spam_metric='frobenius', gates_metric='frobenius', verbosity=0)
+                metrics = gate_metrics_dict(out, target)
+                check_gate_metrics_near_zero(metrics, test_tol)
+
+    def test_recovers_local_unitary_gauge(self):
+        self._check_recovery(self._two_qubit_target((0, 1)), 'Gcnot:0:1')
```
coreyostrove, line 385: Per comment above, it doesn't look like the cnot string is used by _check_recovery. What is the intended purpose of passing this in?

test/unit/algorithms/test_gaugeopt_correctness.py
```diff
@@ -342,6 +342,90 @@ def test_lbfgs_frodist(self):
...
+
+    def test_recovers_local_unitary_gauge_with_out_of_order_cnot(self):
+        # The entangler targets (1, 0). Its superoperator lives in the full-space basis whose
+        # Kronecker order follows the state-space labels, not the gate's target order.
+        target = self._two_qubit_target((1, 0))
+        self.assertIn(pgl.Label('Gcnot', (1, 0)), target.operations)
+        self._check_recovery(target, 'Gcnot:1:0')
```
coreyostrove, line 392: See above comments.

coreyostrove (changes requested):
A few requests for changes and consideration left below. Arigato gozaimashita.

coreyostrove (changes requested):
A few requests for changes and consideration left below. Arigato gozaimashita.

coreyostrove (changes requested):
A few requests for changes and consideration left below. Arigato gozaimashita.

coreyostrove (changes requested):
A few requests for changes and consideration left below. Arigato gozaimashita.

pygsti/models/gaugegroup.py
rileyjmurray: Done

pygsti/models/gaugegroup.py
rileyjmurray: Raising a DubiousTargetWarning

pygsti/models/gaugegroup.py
rileyjmurray: Done. scipy.sparse.issparse though, since that catches spmatrix too.

pygsti/models/gaugegroup.py
rileyjmurray: done

test/unit/algorithms/test_gaugeopt_correctness.py
rileyjmurray: fixed

test/unit/algorithms/test_gaugeopt_correctness.py
rileyjmurray: fixed

pygsti/models/gaugegroup.py
rileyjmurray: Revised so that bases are computed without reference to udims. However, this function also needs to return a StateSpace object for each basis. The revised implementation still relies on udims to build ExplicitStateSpace objects. There's a comment indicating as much, so it still serves the desired purpose of making future extension  of this class easier.

pygsti/models/gaugegroup.py
rileyjmurray: 1. Done. The `TensorProductGaugeGroup` class now forwards the appropriately-constructed inverse operator to `TensorProductGaugeGroupElement.__init__`; the latter function will compute the inverse operator from`change_of_basis` if the inverse is omitted.
2. Agreed, and fixed.
3. It'll be rare that C has tensor product structure and be anything other than the identity matrix. The revisions to a better job of detecting when $C=I$.
4. I agree that TensorProductBasis should ideally be using the tensor product structure when building the transform matrices. I think that's out of scope for this PR, though.

coreyostrove: Regarding point 3, I don't see how this wouldn't be the case. The gauge group elements have a tensor product structure with each factor element in principle being specified in its own unique basis. So, I can imagine cases where one would want to be doing a factor-wise change of basis, but not one where you'd want to do an arbitrary change of basis which couples the factors. So, given the limitation on dense linear algebra, why allow that as an option?

coreyostrove: Regarding point 4, that is fine, this is an existing issue. But could you please document this?

pygsti/models/gaugegroup.py
coreyostrove, line 1972: Should also serialize _Cinv.

coreyostrove (changes requested):
Left a comment on one of the threads from the previous round of reviews, and a new (minor) request below.

pygsti/models/gaugegroup.py
rileyjmurray: For 3 this just comes down to providing support for something that's not a great idea but is mathematically valid and easy to accomodate. I'll remove support for this if you prefer.

For 4 --- sure thing.

coreyostrove: > I'll remove support for this if you prefer.

What is "this" referring to? The support for non-tensor product changes of basis, or changes of basis writ large?

rileyjmurray: My instinct is only the support for non-tensor product changes of basis. But if implementing that in a meaningful way is nontrivial (e.g., it's equivalent to handling point 4) then I'd be inclined to remove the ability to change basis altogether. If you think the balance tips in the direction of "do 4 in this PR" I'm also okay with that. I'm sure it'll pay dividends in the near(ish) future.

coreyostrove: I have no opposition to following your instincts as you've relayed them as the plan of record.

PR #906, opened 2026-09-03
Make TorchForwardSimulator go vroom vroom

rileyjmurray:
Edit (Sep 17, 2026): revised after @sserita's review. #917, now merged into this branch, changed how the circuit-store cache is keyed, and the review asked for two tests, a docstring pointer, and a slicing fix. The cache-semantics, tests, and open-items sections below describe the branch as it stands today. The measurements are unchanged.

I want `TorchForwardSimulator` to be useful on large GST problems. Before this PR it was correct, but its CPU-mode performance wasn't competitive with MPI-accelerated `MapForwardSimulator`. What follows is written by the robot, with editing by me; "we" refers to the robot and me.

The attached Zip folder contains presentations the robot made at my direction. Interested readers should check out `draft-1/slides.html`. [Link: presentations.zip](https://github.com/user-attachments/files/31810208/presentations.zip)

## Context

The test instance throughout is leakage GST on a six-level (qubit ⊗ qutrit) model with the CPTPLND parameterization. Such models are easy to build: take a 2-qubit modelpack and apply `promote_bb_to_bt` from `pygsti.leakage.models`. I used the `smq2Q_XYICNOT` model and edesign; the promoted CPTPLND model has 10,080 parameters. At L = 16 one Jacobian took 24.8 seconds on a GH200. `MapForwardSimulator` on 32 MPI ranks of the same host takes 45 seconds for that Jacobian, so the GPU path was already the fastest option and still far too slow to be useful. (I didn't test CPU-only `TorchForwardSimulator` on the GH200 box, but I could have.)

The starting point was a conversation with an NVIDIA engineer who looked at `TorchForwardSimulator`. He observed that we had far too many kernel launches: one GEMV per gate application.

## Findings

The launch count is worse than "one GEMV per gate application" makes it sound. The L = 16 circuit list has 98,577 gate applications, so a probability call issued about 10⁵ launches of a 36 × 36 matvec. The Jacobian came from pushing `torch.func.jacfwd` (or `jacrev`) through that same Python loop, which kept every launch and multiplied the working set by the parameter count; the profiler counted 430,411 launches for one Jacobian call. The GPU was idle almost all of the time, and no tuning inside the matvec could have helped.

Fixing that exposed three more problems, none visible while launch overhead dominated.

1. `LindbladErrorgen.torch_base` contracted the complex term-superoperator stack with an `einsum` that, under the `vmap` used by `torch.func`, vectorized into a batched GEMV running at about 3.7 TFLOP/s. That one kernel was 44% of all GPU time in a CPTPLND Jacobian. We didn't set out to touch `LindbladErrorgen`; the profiler sent us there.
2. Copying the Jacobian to the host went through `J.cpu().numpy()`, which allocates a buffer the size of the Jacobian and then copies it a second time into the caller's array. For a 2.63 GiB Jacobian that round trip cost several times the transfer itself.
3. `StatelessModelCircuitStore` was rebuilt on every call, walking every circuit in Python, though nothing about it depends on parameter values. The `TODO` in `_bulk_fill_dprobs` already said as much.

After this PR the same warm Jacobian call takes 120 ms (208× faster), with results unchanged to about 1e-9 relative against finite differences.

### Things users will notice

**The simulator is much faster, and nothing else about it should look different.** Same constructor, same `bulk_fill_probs` / `bulk_fill_dprobs` contract, same dtypes and devices.

### Things with low exposure that reviewers should still look at

**The circuit store is cached across calls, keyed on object identity.** `StatelessModelCircuitStore` and its evaluation plan depend only on the structure of the (layout, model) pair, so an optimizer now pays for them once per layout rather than once per iteration. The semantics:

1. The cache holds one entry, keyed on the identity of the layout, the identity of the model, and the identity of each parameterized member together with its label. All of these must match for reuse.
2. Everything in the key is held by weak reference, so the cache can't keep a layout, a model, or a member alive.
3. Member identity is the right key because the store bakes each member's type and `stateless_data`, and because `set_all_parameterizations` and `convert_members_inplace` always install new member objects under the same labels. An earlier version of this branch keyed on `(label, type, num_params)` per member instead. @sserita found that converting CPTPLND to GLND at the target point preserves both type and parameter count while changing which Lindblad coefficient blocks are active, so the stale store was reused silently; his script shows a 0.07 absolute error in the probabilities. #917 made the change.
4. The key does not detect in-place mutation of a static member's dense matrix, whose value is baked into `stateless_data`, and it does not detect `ComposedOp.append`/`insert`/`remove`, which change a container's factors without replacing the container. We consider the first out of contract; the docstring says to reassign `model.sim` if you do it. The second is a residual gap that we haven't confirmed and haven't closed; `_model_signature`'s docstring records it.

**`LindbladErrorgen.stateless_data` returns a different tuple.** It was `(G, blocks, param_dims)` with `G` a complex `(N, d², d²)` stack. It is now `(G_re, G_im, superop_shape, blocks, param_dims, block_is_complex)` with `G` pre-split into real `(N, d²·d²)` matrices. The only consumer is `LindbladErrorgen.torch_base`, and every wrapper in the Torchable protocol passes the tuple through opaquely, so we believe nothing else is affected. Grep for `stateless_data` if you doubt that; we did.

### Things that are internal

**Batched prefix and suffix DAG evaluation** (`pygsti/forwardsims/torchfwdsim_plan.py`, new). Circuits are inserted into a prefix DAG and a suffix DAG over op-label sequences. Each DAG is evaluated breadth-first, one dense matrix-matrix product per (depth, gate label) pair. Nodes are numbered so that every (depth, gate) group is contiguous, so each step writes into a slice of the state tensor and only the source side needs a gather. On the L = 16 design this is 245 batched matmuls instead of 98,577 matvecs, and prefix sharing cuts the arithmetic by 9.4×. `MapForwardSimulator` gets the same prefix sharing from `PrefixTable`, but walks its table one state propagation at a time; the suffix DAG has no counterpart there, since `MapForwardSimulator` differentiates by finite differences. The module docstring says this too.

**An analytic Jacobian with respect to the dense superoperators.** Write `R_k` for the prefix state after `k` gates and `L_k` for the suffix state before gate `k+1`. Then `∂p_c/∂M_g` is a sum of rank-one terms `L_k ⊗ R_{k−1}` over the occurrences of gate `g` in circuit `c`; repeated germs mean a gate usually occurs many times in one circuit. The plan assembles that tensor with a chunked scatter-add of outer products per gate (temporaries capped at 256 MiB) and contracts it against `∂M_g/∂θ_g`. That second factor still comes from `torch.func.jacrev` on each member's `Torchable.torch_base`, so PyTorch still differentiates every parameterization; it just no longer differentiates the circuit loop. Prep and POVM blocks fall out of the same two DAGs. Instruments whose expanded members share a parameter block accumulate into that block before it's written.

**The error-generator contraction is a real GEMM.** Since only the real part of the contraction is kept, `Re(Σᵢ cᵢ Gᵢ) = Re(c)·Re(G) − Im(c)·Im(G)`, so `torch_base` is now two 2-D real matmuls: half the flops, and a 2-D matmul vectorizes into a GEMM. `Im(G)` is kept only for coefficient blocks whose `block_data` is complex (the Hamiltonian block's is real). On the headline Jacobian this took the kernel from 26.6 ms to negligible and the whole call from 144 ms to 115 ms at the time.

**Device-to-host copies write straight into the caller's array.** `_copy_to_numpy` wraps the destination with `torch.from_numpy` so the copy engine writes into it directly, falling back to the old path for arrays that can't be wrapped. Probabilities requested alongside the Jacobian are now produced on-device by the same plan and copied out the same way.

## Measurements

All numbers are from one GH200 in float64. "Warm" is the median of calls after the first against a fixed layout, which is what GST pays per optimizer iteration. Each (instance, simulator) pair ran in a fresh process with CUDA synchronized around the timed region. "Before" is pyGSTi at `ed579556c`. `bt6` is the qubit ⊗ qutrit model and `2q` the plain two-qubit model, both on the `smq2Q_XYICNOT` design.

| instance | HS dim | params | circuits | elements | before | after | speedup |
|---|---|---|---|---|---|---|---|
| `2q-hs-L8` | 16 | 240 | 6,131 | 24,524 | 8.42 s | 51 ms | 165× |
| `bt6-hs-L8` | 36 | 560 | 6,131 | 24,524 | 8.73 s | 58 ms | 151× |
| `bt6-hs-L16` | 36 | 560 | 8,740 | 34,960 | 16.0 s | 69 ms | 231× |
| `bt6-cptp-L8` | 36 | 10,080 | 6,131 | 24,524 | 14.8 s | 88 ms | 169× |
| `bt6-cptp-L16` | 36 | 10,080 | 8,740 | 34,960 | 24.8 s | 120 ms | 208× |

Kernel launches on `bt6-cptp-L16`: `_bulk_fill_dprobs` went from 430,411 to 1,555, and `_bulk_fill_probs` from 107,632 to 518. `_bulk_fill_probs` on its own went from 1.98 s to 26 ms.

Of the 98.5 ms in the profiled headline call, about 32 ms is the per-member `torch_base` derivatives, 39 ms is the DAG replay plus the rank-one assembly and contraction, and 10 ms is copying the 2.63 GiB Jacobian to the host. GPU kernels account for 34 ms of the 72 ms device phase. In our assessment the device side is now latency-bound; two further kernel-level optimizations barely moved the wall clock, and we stopped there.

For scale only: `MapForwardSimulator` on this Jacobian takes 153 s on one MPI rank and bottoms out at 45.2 s on 32 ranks. Those are single warm calls with `PYGSTI_USE_SHARED_MEMORY=0` (the container's `/dev/shm` is 64 MB), so they aren't the default GST-under-MPI configuration.

## Correctness

Torch on CPU and GPU agree with `MatrixForwardSimulator` and with a finite-difference Jacobian to relative error between 4e-10 and 3e-9 on every instance above (all 240 columns on `2q-hs-L8`, subsets of columns on the others). That is the FD error `MatrixForwardSimulator` itself shows, so the residual is the finite differences. Probabilities agree to about 2e-15 absolute. On a one-qubit LSGST model against the analytic `MatrixForwardSimulator` Jacobian, float64 agrees to 3e-15 and float32 to 1.6e-7, on both devices. Note that different simulators build different `Layout` subclasses with different element orders, so every cross-simulator comparison canonicalizes to (circuit, outcome) order first.

## Tests

Seven test functions are new in `test/unit/objects/test_forwardsim.py`, each skipped when PyTorch isn't installed.

* `test_torch_multiple_state_preps` uses a model with two state preps, which needs two roots in the prefix DAG, and compares probabilities and Jacobian to `MatrixForwardSimulator`.
* `test_torch_plan_batches_gate_applications` checks that the plan issues far fewer matmuls than there are gate applications, and that its probabilities equal a plain per-circuit, per-layer evaluation.
* `test_torch_circuit_store_is_reused_across_calls` checks that the store persists across calls on one layout, isn't shared across layouts, and still tracks parameter changes.
* `test_torch_circuit_store_is_invalidated_by_reparameterization` converts a model from full TP to CPTPLND in place and compares the Jacobian to a fresh simulator's.
* `test_torch_circuit_store_invalidated_by_type_preserving_reparameterization`, from #917, does the same for the three conversions that preserve type and parameter count: CPTPLND/GLND, H/S, and H+S/H+s.
* `test_torch_plan_instrument_jacobian` runs circuits with `Instrument` and `TPInstrument` layers, one of them with two instrument layers, and compares the Jacobian element-wise to `MatrixForwardSimulator`. It also checks that the plan routes both expanded members into one column block. The instrument consistency Testers already ran the torch simulator on these circuits, but they only check colinearity across simulators.
* `test_torch_plan_scattered_output_rows_jacobian` builds a dataset layout in which one circuit's outcomes are reversed relative to `povm.keys()` and another's are a strict subset. Every group in the plan then takes the scattered-write path rather than the contiguous slice. Probabilities and Jacobian rows are compared per (circuit, outcome) to `MatrixForwardSimulator`. Flipping the gather indices in that path makes the test fail.

We ran `test/unit/objects`, `test/unit/modelmembers`, and `test/unit/algorithms` together at the current head: 2159 passed, 20 skipped, 1 xfailed. We have not run `test/test_packages`.

## How we'd suggest reviewing this

Read the module docstring of `torchfwdsim_plan.py` first; it derives the rank-one Jacobian formula and is the fastest way to decide whether you believe the approach. Then `_jacobian_group` and `_accumulate_a` in the same file, where that formula meets `index_add_` and where an indexing mistake would hide if there is one. `_circuit_store` and `_model_signature` in `torchfwdsim.py` are short and deserve a skeptical read against the cache semantics above. The `lindbladerrorgen.py` change is mechanical once you accept the real-part identity.

## What we haven't done

We haven't run an end-to-end GST fit on this branch. Every timing and check above isolates a single `bulk_fill_dprobs` call against a fixed layout; a real fit is the only thing that exercises the store cache across L-stages and across the reparameterizations `set_all_parameterizations` performs. We'd want that before a release.

The timings were taken at `7c3d1ef32`. The head now also runs a per-call cache check, a handful of `is` comparisons, one per parameterized member. We expect that to be unmeasurable against 120 ms, but we haven't re-timed it.

`StatelessModelCircuitStore.circuit_probs_from_free_params` and `default_to_reverse_ad` remain but the simulator no longer routes through them; we left them rather than decide their fate here. Sparse-superop `LindbladErrorgen` representations remain unsupported on the torch path, as before, and we did nothing for multiple GPUs or nodes.

@sserita points out that the suffix-DAG Jacobian applies to `MapForwardSimulator` as well, since `deriv_wrt_params` and most `adjoint_acton` implementations already exist, and that `OpCRep_ExpErrorgen::adjoint_acton` is the blocker for the CPTPLND case. That is #918, with a proof of concept, and not part of this PR.

### DIFF
--- a/pygsti/forwardsims/torchfwdsim.py
+++ b/pygsti/forwardsims/torchfwdsim.py
@@ -394,6 +394,24 @@ class TorchForwardSimulator(ForwardSimulator):
         self._store_cache = None
         super(ForwardSimulator, self).__init__(model)
 
+    @staticmethod
+    def _model_signature(model):
+        """
+        A cheap structural fingerprint of `model`'s parameterized members.
+
+        The store bakes each member's *type* and its `stateless_data`, so reusing a store is only
+        sound while the members themselves are structurally unchanged. Object identity of the
+        model is not sufficient: `set_all_parameterizations` and `convert_members_inplace` replace
+        member objects under the same labels on the same model object, which would otherwise leave
+        the cached types and stateless data stale.
+
+        This does not, and cannot cheaply, detect in-place mutation of a *static* member's dense
+        matrix, whose value is baked into `stateless_data`. Mutating a member pyGSTi considers
+        static is out of contract; call `model.sim = TorchForwardSimulator(...)` again if you do it.
+        """
+        return tuple((lbl, type(obj), obj.num_params)
+                     for lbl, obj in model._iter_parameterized_objs())
+
     def _circuit_store(self, layout) -> StatelessModelCircuitStore:
         """
         The StatelessModelCircuitStore for `layout`, reused across calls.
@@ -402,18 +420,22 @@ class TorchForwardSimulator(ForwardSimulator):
         data; building its evaluation plan walks every gate application. Neither depends on the
         model's parameter *values*, so a GST optimization -- which calls this simulator once or
         twice per iteration against a fixed layout -- should pay for them once, not once per
-        iteration. The cache holds weak references so it can't keep a layout or model alive, and
-        is keyed by object identity: a structurally different layout or model is a different
-        object, and `get_free_params` re-checks the modelmember labels on every call.
+        iteration.
+
+        The cache holds weak references so it can't keep a layout or model alive. It is keyed by
+        the identity of the layout and model *and* by :meth:`_model_signature`, so that replacing
+        modelmembers in place (a different parameterization under the same label) invalidates it.
         """
+        signature = self._model_signature(self.model)
         cached = self._store_cache
         if cached is not None:
-            layout_ref, model_ref, store = cached
-            if layout_ref() is layout and model_ref() is self.model:
+            layout_ref, model_ref, cached_signature, store = cached
+            if (layout_ref() is layout and model_ref() is self.model
+                    and cached_signature == signature):
                 return store
         store = StatelessModelCircuitStore(self.model, layout, self.dtype, self.device)
         try:
-            self._store_cache = (_weakref.ref(layout), _weakref.ref(self.model), store)
+            self._store_cache = (_weakref.ref(layout), _weakref.ref(self.model), signature, store)
         except TypeError:  # pragma: no cover - a layout or model that can't be weak-referenced
             self._store_cache = None
         return store
--- a/pygsti/forwardsims/torchfwdsim.py
+++ b/pygsti/forwardsims/torchfwdsim.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 from typing import Tuple, Optional, Dict, Any, Type, TYPE_CHECKING
 from pygsti.modelmembers.torchable import Torchable
 from pygsti.forwardsims.forwardsim import ForwardSimulator
+from pygsti.forwardsims.torchfwdsim_plan import TorchEvalPlan
 
 if TYPE_CHECKING:
     from pygsti.baseobjs.label import Label
@@ -29,6 +30,7 @@ if TYPE_CHECKING:
     import torch
 
 import warnings as warnings
+import weakref as _weakref
 
 
 try:
@@ -88,6 +90,8 @@ class StatelessModelCircuitStore:
         from pygsti.modelmembers.instruments.instrument import Instrument as _Instrument
         from pygsti.modelmembers.instruments.tpinstrument import TPInstrument as _TPInstrument
 
+        self.dtype = dtype
+        self.device = device
         circuits = []
         self.outcome_probs_dim = 0
         povm_effect_orders = {}
@@ -119,6 +123,9 @@ class StatelessModelCircuitStore:
                 circuits.append(c)
                 self.outcome_probs_dim += c.outcome_probs_dim
         self.circuits = circuits
+        self.povm_effect_counts = {lbl: len(order) for lbl, order in povm_effect_orders.items()}
+        self.model_dim = model.dim
+        self._eval_plan = None
 
         # We need to verify assumptions on what layout.iter_unique_circuits() returns.
         # Looking at the implementation of that function, the assumptions can be
@@ -227,29 +234,49 @@ class StatelessModelCircuitStore:
 
         return torch_bases
 
+    @property
+    def eval_plan(self) -> TorchEvalPlan:
+        """
+        The batched :class:`TorchEvalPlan` for this store's circuits, built on first use.
+
+        The plan depends only on the *structure* of the (model, layout) pair -- circuit contents,
+        which modelmembers are parameterized, and how many parameters each has -- so one plan
+        serves every parameter value a GST optimization visits.
+        """
+        if self._eval_plan is None:
+            self._eval_plan = TorchEvalPlan(
+                self.circuits, self.param_metadata, self.instrument_expansions,
+                self.povm_effect_counts, self.outcome_probs_dim, self.dtype, self.device
+            )
+        return self._eval_plan
+
     def circuit_probs_from_torch_bases(self, torch_bases: Dict[Label, torch.Tensor]) -> torch.Tensor:
         """
         Compute the circuit outcome probabilities that result when all of this StatelessModelCircuitStore's
         StatelessCircuits are run with data in torch_bases.
-    
+
         Return the results as a single (vectorized) torch Tensor.
         """
-        probs = []
-        for c in self.circuits:
-            superket = torch_bases[c.prep_label]
-            superops = [torch_bases[ol] for ol in c.op_labels]
-            povm_mat = torch_bases[c.povm_label]
-            if c.effect_row_indices is not None:
-                povm_mat = povm_mat[c.effect_row_indices]
-                # ^ select (and order) the POVM rows to match c.effect_labels; advanced
-                #   indexing keeps autograd flowing to the POVM's parameters.
-            for superop in superops:
-                superket = superop @ superket
-            circuit_probs = povm_mat @ superket
-            probs.append(circuit_probs)
-        probs = torch.concat(probs)
-        return probs
-    
+        return self.eval_plan.probs(torch_bases, self.model_dim)
+
+    def circuit_jacobian_from_free_params(self, free_params: Tuple[torch.Tensor],
+                                          probs_out=None) -> torch.Tensor:
+        """
+        The Jacobian of :meth:`circuit_probs_from_free_params` at `free_params`, as a dense
+        ``(outcome_probs_dim, params_dim)`` Tensor.
+
+        Unlike ``torch.func.jacfwd(circuit_probs_from_free_params)``, this does not differentiate
+        through the circuit-evaluation loop. It differentiates each modelmember's
+        ``torch_base`` map on its own (a small, per-member autodiff problem) and applies the chain
+        rule with an explicitly-derived expression for the derivative of a circuit's outcome
+        probabilities with respect to the dense superoperators. See
+        :mod:`pygsti.forwardsims.torchfwdsim_plan`.
+        """
+        num_params = sum(int(fp.numel()) for fp in free_params)
+        return self.eval_plan.jacobian(free_params, self.model_dim, num_params,
+                                       probs_out=probs_out)
+
+
     def circuit_probs_from_free_params(self, *free_params: Tuple[torch.Tensor], enable_backward=False) -> torch.Tensor:
         """
         This is the basic function we expose to pytorch for automatic differentiation. It returns the circuit
@@ -346,12 +373,37 @@ class TorchForwardSimulator(ForwardSimulator):
 
         self.dtype  = dtype
         self.device = device
+        self._store_cache = None
         super(ForwardSimulator, self).__init__(model)
+
+    def _circuit_store(self, layout) -> StatelessModelCircuitStore:
+        """
+        The StatelessModelCircuitStore for `layout`, reused across calls.
+
+        Building the store walks every circuit in Python and bakes each modelmember's stateless
+        data; building its evaluation plan walks every gate application. Neither depends on the
+        model's parameter *values*, so a GST optimization -- which calls this simulator once or
+        twice per iteration against a fixed layout -- should pay for them once, not once per
+        iteration. The cache holds weak references so it can't keep a layout or model alive, and
+        is keyed by object identity: a structurally different layout or model is a different
+        object, and `get_free_params` re-checks the modelmember labels on every call.
+        """
+        cached = self._store_cache
+        if cached is not None:
+            layout_ref, model_ref, store = cached
+            if layout_ref() is layout and model_ref() is self.model:
+                return store
+        store = StatelessModelCircuitStore(self.model, layout, self.dtype, self.device)
+        try:
+            self._store_cache = (_weakref.ref(layout), _weakref.ref(self.model), store)
+        except TypeError:  # pragma: no cover - a layout or model that can't be weak-referenced
+            self._store_cache = None
+        return store
 
     def _bulk_fill_probs(self, array_to_fill, layout, split_model = None) -> None:
         assert self.model is not None
         if split_model is None:
-            smcs = StatelessModelCircuitStore(self.model, layout, self.dtype, self.device)
+            smcs = self._circuit_store(layout)
             free_params = smcs.get_free_params(self.model, self.dtype, self.device)
             torch_bases = smcs.get_torch_bases(free_params)
         else:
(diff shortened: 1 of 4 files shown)


pygsti/forwardsims/torchfwdsim_plan.py
```diff
@@ -0,0 +1,575 @@
...
+    J[(c,e), theta_g] = A_g[(c,e), :, :] . (d M_g / d theta_g)
+
+The second factor is a small per-modelmember object of shape ``(d^2, n_params_g)`` that we get
+from ``torch.func.jacrev`` on ``Torchable.torch_base`` -- so PyTorch still does all the
+differentiation of the parameterization, which is the whole point of the Torch simulator. The
+first factor is built by one scatter-add of outer products, and the contraction is one large
+matrix-matrix product. Both map perfectly onto a GPU.
+
+The result is a Jacobian computed in a few hundred kernel launches instead of ~10^5 * n_params.
+"""
```
sserita, line 54: Worth a sentence noting how this relates to MapForwardSimulator's PrefixTable (layouts/prefixtable.py)?

Two things are still genuinely new/different: GPU-batched (depth, gate) grouping instead of PrefixTable's serial per-node walk, and an analytic Jacobian via the suffix DAG — MapForwardSimulator computes its own Jacobian via O(n_params) finite differences, with no backward pass at all.

**Also worth noting:** this same suffix-DAG trick looks structurally applicable to MapForwardSimulator too (its deriv_wrt_params() and most adjoint_acton implementations already exist) — but OpCRep_ExpErrorgen::adjoint_acton is currently an assert(false) stub, which is the actual blocker for the common CPTPLND case.
Filed as a follow-up issue (with working PoC): #918.

pygsti/forwardsims/torchfwdsim_plan.py
```diff
@@ -0,0 +1,575 @@
...
+
+        def write(cols, block):
+            """Write a (n_dense, n_cols) block into the right rows/columns of `out`."""
+            if g.contiguous_out:
+                out[g.out_slice, cols] = block
+            else:
+                out[g.out_rows, cols] = block.index_select(0, g.dense_rows)
+
+        # --- gate blocks: A_g . (dM_g / dtheta) ------------------------------------------------
+        # Instrument members share a column block, so accumulate per member label first.
```
sserita, line 512: This looks right, but tests don't actually include any Instruments - worth having an explicit check for that?

pygsti/forwardsims/torchfwdsim_plan.py
```diff
@@ -0,0 +1,575 @@
...
+            out = torch.empty(self.outcome_probs_dim, dtype=self.dtype, device=self.device)
+        for g in self.groups:
+            fwd = g.forward_states(torch_bases, dim, self.dtype, self.device)
+            povm_mat = torch_bases[g.povm_label]
+            dense = (fwd.index_select(0, g.circ_end_fwd) @ povm_mat.transpose(-2, -1)).reshape(-1)
+            if g.contiguous_out:
+                out[g.out_slice] = dense
+            else:
+                out[g.out_rows] = dense[g.dense_rows]
+        return out
```
sserita, line 409: Agent reports the fallback branch (for a scattered write of non-contiguous outcome elements, computed on line 295) here is untested - worth a test?

pygsti/forwardsims/torchfwdsim_plan.py
```diff
@@ -0,0 +1,575 @@
...
+
+        # --- POVM block: R_L contracted with (d E / d theta) ------------------------------------
+        _, jac, cols = derivs[g.povm_label]
+        n_v = cols.stop - cols.start
+        if n_v > 0:
+            w = jac.reshape(n_eff, dim, n_v)
+            block = torch.einsum('ca,ean->cen', r_end, w).reshape(n_dense, n_v)
+            write(cols, block)
+
+    def _accumulate_a(self, g, fwd, bwd, ci, lnode, rnode, dim, eff_arange):
```
sserita, line 555: Looks right and an independent verification also came out correct.

pygsti/forwardsims/torchfwdsim.py
```diff
@@ -346,41 +391,79 @@ def __init__(self, model : Optional[ExplicitOpModel] = None, use_gpu: Optional[b
...
-            J_func = torch.func.jacfwd(smcs.circuit_probs_from_free_params, argnums=argnums) # type: ignore
+        probs_out = None
+        if pr_array_to_fill is not None:
+            probs_out = torch.empty(smcs.outcome_probs_dim, dtype=self.dtype, device=self.device)
 
-        J_val = J_func(*free_params)
-        J_val = torch.column_stack(J_val)
-        array_to_fill[:] = J_val.cpu().detach().numpy()
+        J_val = smcs.circuit_jacobian_from_free_params(free_params, probs_out=probs_out)
+        _copy_to_numpy(J_val, array_to_fill)
```
sserita, line 466: Minor inconsistency: This doesn't slice like line 453. Currently correct while callers pass same-sized arrays, but could be an obscure bug source if that was not the case.

pygsti/modelmembers/operations/lindbladerrorgen.py
```diff
@@ -751,43 +751,83 @@ def to_dense(self, on_space: SpaceT='minimal'):
...
     def stateless_data(self, real_dtype: _torch.dtype, device: _torch.Device):
         """
-        Returns `(G, blocks)` where `G` is the constant `(N, d2, d2)` term-superoperator stack
-        (`combined_lindblad_term_superops`) as a torch tensor, and `blocks` is a list, in
-        `coefficient_blocks` order, of `blk.stateless_data(...)`.
-        
-        The error generator is linear in the (per-block) block_data, so `torch_base` rebuilds it as
-        `einsum('i,ijk->jk', concat(block torch_bases), G).real` -- the differentiable analog of
-        `to_dense('HilbertSchmidt')`.  Dense path only.
+        Returns `(G_re, G_im, superop_shape, blocks, param_dims, block_is_complex)`, the constant
```
sserita, line 754: The optimization to keep only real part checks out, everything else falls out mechanically like you said

test/unit/objects/test_forwardsim.py
```diff
@@ -631,3 +631,172 @@ def test_map_fwdsim(self):
...
+    assert model.sim._circuit_store(other_layout) is not first
+
+    # The Jacobian must still track parameter changes through the reused store.
+    J2 = np.empty_like(J)
+    model.from_vector(model.to_vector() + 0.01)
+    model.sim.bulk_fill_dprobs(J2, layout, None)
+    assert not np.allclose(J, J2)
+
+
+def test_torch_circuit_store_is_invalidated_by_reparameterization():
```
sserita, line 772: This missed the CPTPLND -> GLND-like reparameterization cases. See main review comment and #917

sserita (changes requested):
The core math checks out, excited to see these speedups hit develop.
More critically: I think this can also be applied to MapForwardSim too because it has a similar prefix table structure. Had the agent cook up a PoC in #918.

However, there is at least one bug that needs to be fixed where a stale cache happens for a set_parameterizations from CPTPLND -> GLND (among others). Further explained and solution proposed in #917.

My other comments are places where we can add tests, or explicit approval of things you mentioned as critical pieces of the PR.

test/unit/objects/test_forwardsim.py
rileyjmurray: Resolved by merge of 917 into this branch.

pygsti/forwardsims/torchfwdsim_plan.py
rileyjmurray: Note made

pygsti/forwardsims/torchfwdsim_plan.py
rileyjmurray: It was indeed untested. 29e954e02 adds test_torch_plan_scattered_output_rows_jacobian to check this case.

pygsti/forwardsims/torchfwdsim_plan.py
rileyjmurray: The instrument consistency Testers did run the torch simulator on instrument circuits, but NOT in the tests for evaluation planning.

29e954e02 adds test_torch_plan_instrument_jacobian, which compares element-wise against MatrixForwardSimulator for Instrument and TPInstrument.

pygsti/forwardsims/torchfwdsim.py
rileyjmurray: Fixed in 499410471; it now slices to outcome_probs_dim like the probabilities path.

sserita (approved):
Looks great, thanks!

PR #909, opened 2026-09-08
Change default offline mode

coreyostrove:
This PR makes 2 changes:

First, it changes the default behavior for pyGSTi HTML/notebook report generation to assuming the presence of an active internet connection. The previous default behavior assumed the lack of such a connection, but as a result produces additional artifacts in the form of embedded javascript resources, and an `offline` directory with assets required for report rendering when offline. This adds a nontrivial on-disk storage overhead in service of (an important) edge case. I have also renamed the related kwarg from `connected` to `enable_offline_mode` to make this more evocative for users who do need offline viewing functionality. (Note that the name change also flips the boolean value corresponding to offline mode in the old naming scheme).

I am currently migrating to a new computer, and this change was inspired in part by the fact that I got annoyed by the amount of time my file transfer utility was having to spend transferring `offline` directories for the several hundred reports sitting on my HD.

Second, this PR expires a number of deprecated functions within `pygsti.report.factory`. Namely: `create_general_report`, `create_standard_report`, `create_nqnoise_report`, and `create_report_notebook`. These functions have been formally deprecated since 0.9.9, and long since replaced with either new factory functions (`pygsti.report.factory.construct_*`) or methods of the `Report` class (`write_html` and `write_notebook`). So this is just some spring cleaning since I noticed it while editing the files for change 1.


rileyjmurray (commented):
I'm not a fan of changing the kwarg name (`connected` --> `enabled_offline_mode`) _and_ semantic default behavior when the semantic default behavior could be expressed by flipping the old kwarg's default value (`connected=False` --> `connected=True`). What's the chance that this breaks people's code? The benefit of the more evocative name should be weighed against the chance that people's workflows will break.

rileyjmurray (approved):

rileyjmurray (changes requested):
Whoops. Approved too soon. We need a changelog entry for this.

PR #911, opened 2026-09-09
Split edesigntools into a pygsti.tools.edesign subpackage

rileyjmurray:
*I've revised this slightly from text the robot gave me. Past this point, "I" is first-person from the robot's perspective.*

This PR is a move plus one deletion. `pygsti/tools/edesigntools.py` was 883 lines holding three unrelated concerns, of which the Fisher-information block was about 730. The module is now a package of the same name, `pygsti/tools/edesigntools/`, following the `pygsti/tools/graphcoloring/` precedent. Two of the three concerns each get their own private module behind the package `__init__.py`, with every function body copied verbatim:

 * `_runtime.py` (121 lines) — `calculate_edesign_estimated_runtime`
 * `_fisher.py` (743 lines) — the three `calculate_fisher_information_*` functions and their three private helpers

The third concern, `pad_edesign_with_idle_lines`, is deleted rather than moved; see below.

I'm doing this now as groundwork. In the next day or so I'll open a PR that adds significant experiment-design reduction helpers, and they'll go in this new subpackage.

### Removing `pad_edesign_with_idle_lines`

Review turned up that this function had no consumer anywhere: not in pygsti, not in pygsti's notebooks, and not in loqs. Rather than carry dead code into the new layout, the PR removes the function and its associated tests. This is the only behavior change in the PR.

### The facade

Because the package keeps the module's name, there is no compatibility shim and nothing to deprecate. `edesigntools/__init__.py` re-exports the moved functions, not copies of them, so `import pygsti.tools.edesigntools` and `from pygsti.tools.edesigntools import ...` work exactly as before. `pygsti/tools/__init__.py` is untouched: it still star-imports from `edesigntools`, which now resolves to the package `__init__.py`. Since that facade declares `__all__`, the star-import picks up only the public names, which guards against the kind of import leak-rot described next.

### One intentional namespace change

`pygsti.tools.ceil` no longer exists. It was never meant to be public. The `from math import ceil` that the Fisher chunking needs leaked out through the star-import. The package `__init__.py` now declares `__all__`, which removes it.

Every bare `ceil(` call site in `pygsti/` and `test/` imports the name directly (`algorithms/fiducialpairreduction.py:21`, `layouts/prefixtable.py:15`, `tools/edesigntools/_fisher.py:15`); the only other occurrence in the tree is prose in a `tools/graphcoloring/_sinnamon.py` docstring; and `ceil` was the sole name the star-import leaked. So there is nothing to fix and nothing to announce, and I have not added a CHANGELOG entry or a release note for it. `test/unit/tools/test_edesign_layout.py` asserts the name is gone, as `__all__` hygiene rather than as a compatibility promise.

### Testing and docs

`test/unit/tools/test_edesigntools_layout.py` is new. It pins the arrangement above: that the public names are the *same objects* whether reached through `pygsti.tools` or `pygsti.tools.edesigntools`; that each private module owns its own names and the split duplicated nothing; and that the three import spellings used in `docs/markdown/` still work. The tests that covered `pad_edesign_with_idle_lines` go away with the function.

Since the rest is a move, the check that matters is that it really is one. I parsed develop's `edesigntools.py` and the two new modules with `ast` and diffed each function's source span: all seven surviving definitions are byte-identical, none were added, and the only one dropped is `pad_edesign_with_idle_lines`.


pygsti/tools/edesign/_padding.py
coreyostrove, line 1: `pad_edesign_with_idle_lines` reads to me as something that'd be better rehomed as a method of the various `ExperimentDesign` classes.

coreyostrove (commented):
Left a comment about possible rehoming of one of these utilities. I don't have very strong feelings vis-a-vis the subpackage restructuring. Though I will say I don't love the leading underscore on the implementation file names.

pygsti/tools/edesign/_padding.py
rileyjmurray: Opted to remove the function after discussion with Stefan.

PR #913, opened 2026-09-11
Enforce gate-name namespaces in `CloudNoiseModel` and `ProcessorSpec`

rileyjmurray:
*The following was drafted by a robot and verified by me.*

---

The develop CI run [34407975127](https://github.com/sandialabs/pyGSTi/actions/runs/34407975127) failed the notebook regression on six cells of `docs/markdown/objects/ModelMemberGraph.ipynb`, all cascading from the first. The tutorial builds a `QubitProcessorSpec` with a gate named `mygate`, and #903 gave `LocalNoiseModel`'s member dictionaries their standard prefixes, so `operation_blks['gates']` refuses that key now. Renaming the gate to `Gmygate` fixes the tutorial; every other tutorial already follows the convention (`Gcustom`, `Gwait`).

Fixing the tutorial is one line. The rest of this PR is here because the hole #903 closed for `LocalNoiseModel` was still open in two other places, and both of them accepted `mygate`.

### Changes

**`CloudNoiseModel` enforces the same namespaces.** It built all nine of its member dictionaries with `prefix=None`, so `create_cloud_crosstalk_model` took a gate named `mygate` and filed it under `operation_blks['gates']` without complaint. The class now declares `rho`, `M`, `G` or `{`, and `I` the way `LocalNoiseModel` does, including its two `cloudnoise` blocks, whose keys are Labels built from processor-spec gate names and so belong to the operation namespace.

**`ProcessorSpec` validates gate names.** `QuditProcessorSpec`'s docstring says "All gate names must start with 'G'" and nothing checked it, so a bad name survived construction and surfaced much later as a `KeyError` from inside model building. That's the failure mode that took down the tutorial. `gate_names`, `nonstd_gate_unitaries` and the arity dict are now checked at construction, and `rename_gate_inplace` applies the same rule to its new name. A name passes if it begins with `G` or is a brace-wrapped implicit name such as `{idle}`. Non-string names are checked with `Label.has_prefix`, the same predicate the member dictionaries apply, which is what lets the empty layer label through: `pygsti.leakage` renames the implicit idle to `Label(())`.

**`_init_member_dicts` moves to `ImplicitOpModel`.** #903 put it on `LocalNoiseModel`, and `CloudNoiseModel` wants the same thing, so both subclasses inherit it and declare a `_member_prefixes` tuple instead. `CloudNoiseModel` loses the nine-line dictionary block it carried in `__init__` and again in `_from_nice_serialization`.

### What this breaks

Code that names a gate outside these two namespaces now fails at `ProcessorSpec` construction. For crosstalk-free models that code was already broken as of #903, and this only moves the error earlier. For cloud-noise models it worked, so this is a real behavior change and the CHANGELOG entry says so.

`test_rename_gate_inplace` renamed `Gx` to `MyGx`, which the new rule rejects. That test is about renames propagating across the four dictionaries, so it now uses a compliant name, and the rejection gets its own test.

### Testing

`test/unit` passes (3505 passed, 60 skipped, 2 xfailed), as does `test/integration`. So do the cloud-noise-heavy suites under `test_packages` that run in my environment: `test_calcmethods1Q`, `test_fogi_gst`, `test_labels`, `test_idt`. Serialization round-trips, `copy`, and pickling all preserve the new prefixes for both model classes, with matching probabilities before and after.

I ran the full notebook regression. Every notebook on develop passes except the four under `docs/markdown/ml/`, which want a TensorFlow I don't have installed.

New coverage: prefix enforcement on all nine `CloudNoiseModel` dictionaries, the `{...}` namespace, the `Label` forms of the name check, and the three `ProcessorSpec` arguments. The new `CloudNoiseModelTester` starts on the migration that `test_implicitmodel.py`'s TODO asks for without finishing it. Its `test_getitem` deliberately doesn't reuse `ImplicitOpModelMixin._test_getitem`, which asserts that `m[layer_label]` *is* the stored `operation_blks['layers']` entry: true for a `LocalNoiseModel`, and false for a `CloudNoiseModel`, which composes the layer op from the target gate and its cloud noise on demand.


sserita (approved):
Looks fine to me. I did remember seeing that in 870 (going from prefix=None to an actual prefix), but didn't realize this would be the downstream consequence. Not a wrong consequence necessarily. The clear error message means this should hopefully not be too annoying for users with non-leading-G gate names.

PR #915, opened 2026-09-11
A greedy block D-optimal experiment design kernel in edesigntools

rileyjmurray:
This adds `pygsti/tools/edesigntools/blockdopt.py` and exports its three public functions from `pygsti.tools`. 

`block_linear_dopt(A, block_size, max_blocks)` takes a transposed Jacobian, treats each candidate's `block_size` outcome columns as one indivisible block, and greedily selects the blocks that most increase `0.5*logdet(I + J_S^T J_S)`. The method's derivation appears in its docstring rather than this PR description.

The PR ships two helper functions: `greedy_candidate_scores` and `greedy_path_log_volumes`. The former answers "what would this candidate add?". The latter answers "what is this selection order worth?" from Cholesky of the information matrix, sharing no arithmetic with the kernel.

On its own this PR adds no user-facing entry point. I have a follow-up PR with the entry points planned. We can merge this PR (possibly with modifications) or close it and only introduce the functionality along with the API entry points.

**Tests**

Two types of tests establish correctness. `ColumnPivotedQRTester` uses the fact that at `block_size == 1` this algorithm is Businger-Golub column-pivoted QR of the augmented matrix `vstack([A, eye(n)])`, so `scipy.linalg.qr(..., pivoting=True)` is an exact reference: the kernel has to match LAPACK's pivots on the nose and its score curve to rounding, at every budget, on eight problems and both QR paths. `BruteForceEnumerationTester` covers general `block_size` by enumerating small problems outright, evaluating the objective with Cholesky over every ordering of the candidates and reading the greedy order off the resulting table. 

**Implementation notes**

The implementation pays attention to memory layout and buffer management. This is intentional. My original implementation of this method was in C++, but I figure portability of (nearly) pure-python is preferable to a C++ implementation for now. 

`_batched_qr_r` works out whether `scipy.linalg.qr` accepts a stack of matrices by trying it once at import rather than by comparing version strings, because `pyproject.toml` doesn't pin scipy and the batched path needs 1.18 or newer. `test_batched_qr_probe_rejects_a_nonbatched_qr` pins that the probe returns False instead of raising, and the fallback is held to the same references as the batched path.


pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+    an all-zero block gains exactly zero, so blocks that carry no information are
+    simply ranked after every block that carries some.  There is no singular
+    case to handle, and non-finite input is rejected up front.
+
+    The `bled` reference implementation
+    (`bled.reference_impls.block_linear_dopt`) instead QRs an `(m + b_sz) x m`
+    workspace per candidate.  The two are algebraically the same greedy
+    selection, and agree on selections and log-volumes to rounding; this
+    formulation is about 5x faster and uses a third of the memory at 1926
+    candidates with 42 parameters and 8 outcomes.
```
rileyjmurray, line 193: This needs to be removed.

pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+    for i in range(W.shape[0]):
+        out[i], = _spl.qr(W[i], mode='r', check_finite=False)
+    return out
+
+
+def block_linear_dopt(A, block_size, max_blocks, *, return_scores=False):
+    """Deterministic greedy block D-optimal selection.
+
+    Selects column blocks of `A` one at a time, each time taking the block that
+    most increases `0.5 * logdet(I + sum_{i in S} A_i A_i^T)`.
```
rileyjmurray: Add a remark that 
```
logdet(I_m + sum_{i in S} A_i A_i^T)
```
equals
```
logdet(I_b + sum_{i in S} A_i^T A_i)
```

pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+
+    The identity block in `P_i` is a fixed unit ridge on the information matrix.
+    Callers who want a different ridge `lambda` should scale `A` by
+    `lambda**-0.5` before calling, which turns the objective into
+    `0.5 * logdet(lambda * I + J_S^T J_S)` up to an additive constant.  Because
+    `P_i^T P_i = I_b + G_i G_i^T` is at least `I_b`, every candidate has a finite
+    nonnegative gain: a rank-deficient block gains less than a full-rank one and
+    an all-zero block gains exactly zero, so blocks that carry no information are
+    simply ranked after every block that carries some.  There is no singular
+    case to handle, and non-finite input is rejected up front.
```
rileyjmurray: Unless there's a compelling reason, comments should only describe what we do -- not what we don't do or decided not to do. Much of this material can be removed based on this principle.

pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+    n_rem = n_candidates
+    log_volume = 0.0
+
+    for it in range(num_blocks):
+        # ---- Step 1: QR of every remaining panel, b_sz columns each. -------
+        # mode='r' wraps LAPACK geqrf and returns the full (m + b_sz) x b_sz
+        # factor; the leading b_sz x b_sz upper triangle is rho_i.
+        R = _batched_qr_r(buf[:n_rem].transpose(0, 2, 1))[:, :b_sz, :]
+
+        # ---- Step 2: gain = sum_j log|rho_jj|; lowest-index maximiser. -----
```
rileyjmurray: Use American English spelling, here and elsewhere.

pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+    blocks = _candidate_blocks(A, block_size)                       # (n_cand, m, b)
+    return blocks @ blocks.transpose(0, 2, 1)
+
+
+def _half_logdet(M):
+    """Batched `0.5 * logdet(M)` over a stack of ridged information matrices.
+
+    Every `M` handed here is `I_m` plus a sum of Gram matrices, so its
+    eigenvalues are at least 1 and `logdet` is nonnegative; there is no
+    sign to test.
```
rileyjmurray: This being the case, we can just compute Cholesky decompositions, then log-sum of the diagonal of each Cholesky factor. The scipy implementation of slogdet uses LU decomposition.

pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+    block_pivots : numpy.ndarray
+        int64, length `min(n_candidates, max_blocks)`.  0-based candidate
+        indices in selection order.
+
+    scores : numpy.ndarray
+        Only returned if `return_scores` is true.  float64, same length.
+        `scores[k]` is the objective value of the block chosen at step `k`,
+        i.e. `sum_j log|R_jj|` of the workspace of that block, which equals
+        `0.5 * logdet(I_m + sum_{i in S_k} A_i A_i^T)` with `S_k` the first
+        `k+1` selected blocks.  The curve is nondecreasing.
```
rileyjmurray: Please change the API so scores are always returned.

pygsti/tools/edesigntools/blockdopt.py
```diff
@@ -0,0 +1,372 @@
...
+    eigenvalues are at least 1 and `logdet` is nonnegative; there is no
+    sign to test.
+    """
+    return 0.5 * _np.linalg.slogdet(M)[1]
+
+
+def greedy_candidate_scores(A, block_size, selected=()):
+    """Score of every candidate block given an already-selected prefix.
+
+    This is the objective :func:`block_linear_dopt` maximises at each step,
```
rileyjmurray: American English

test/unit/tools/test_blockdopt.py
```diff
@@ -0,0 +1,527 @@
...
+
+def _design(seed, ncand, b, p, dtype=np.float64, order="C"):
+    """Random `A = J^T` of shape `(p, ncand * b)` in the requested layout."""
+    rng = np.random.default_rng(seed)
+    J = rng.standard_normal((ncand * b, p))
+    return np.array(J.T, dtype=dtype, order=order)
+
+
+class BlockDoptKernelTester(BaseCase):
+    def assert_tol_greedy_path(self, A, b, piv, rel_tol=1e-8, label="selection"):
```
rileyjmurray, line 42: Add blank line before function definition.

CHANGELOG
rileyjmurray, line 1: Revise to indicate that this is a low-level utility function that users are unlikely to benefit from. We use it internally, and just happen to not keep it private.

rileyjmurray (commented):
Leaving comments to pull down locally

Issue #919, opened 2026-09-15
Orientation: simultaneous GST designs, patches, stitching, and design reduction


rileyjmurray:
This issue is a reader's guide to the simultaneous-GST subsystem for reviewers of the pull requests that finish it for 0.11.0. The subsystem arrived in #735 and has grown in #912, #914 and #915; none of it is in a release yet except the pieces called out below.

Minor note: replace "qubit" with "qudit" where it appears below.

**What a `SimultaneousGSTDesign` is.** It runs a one-qubit and a two-qubit GST design at the same time on disjoint parts of a device. Its inputs are a `QubitProcessorSpec`, the two component `GateSetTomographyDesign`s (which must share their germ powers), and an edge coloring of the connectivity graph. It is a `GateSetTomographyDesign` itself, so it holds one circuit list per maximum length, nested.

**Patches and colorings.** An edge coloring partitions the two-qubit edges into *color patches*. Each patch is a set of edges that can be driven simultaneously because no two share a qubit. Qubits not covered by a patch's edges are free for one-qubit circuits. The design stores the coloring as `color_patches` and derives `vertices`, `edges`, `neighbors` and `deg` from the processor spec. `make_simultaneous_gst_design` computes a coloring for you with `pygsti.tools.graphcoloring`.

**Stitching.** Say we're doing 5-qubit GST simultaneous, and we've decided on a 1-qubit design, a 2-qubit design, and which pairs of qubits get the 2-qubit design. We still haven't decided what circuits we'll try to run on the 5-qubit device. The task of tensoring together suitable circuits from the sub-designs to run on the 5-qubit device is called *stitching.* 

A *stitcher* tensors circuits from the two component designs onto one patch at a time and produces one circuit list per germ power. On `develop` today the stitcher is a module-level function, `assign_the_designs_with_mapping`, chosen by a random per-germ-power schedule, and its output is checked by `assert_circuit_lists_match_color_patches`, which reads each circuit's patch off the circuit content (which edges its multi-qubit gates act on). #912 made that check content-based rather than positional, which is why a design can now be cut down to a subset of its circuits by the ordinary truncation methods. PR #920 turns the stitcher into an object (`CircuitStitcher`) so that a design can record and serialize the rule that built it.

[...]

PR #920, opened 2026-09-15
Simultaneous GST: make circuit stitcher an object

rileyjmurray:
See #919 for a orientation on the simultaneous-GST subsystem  (patches, colorings, stitchers). This subsystem hasn't appeared in a public release so we can change the API at will. This PR is one such API change.

**The trouble**

`circuit_stitcher` was a bare callable, which failed in three ways. A callable can only be recorded by qualified name, so a lambda or a notebook-defined stitcher came back from `from_dir` as `None` with a warning, and the design had quietly lost the record of how it was built. Options went through `**stitcher_kwargs`, a catch-all that accepts a misspelling in silence: `share_same_shape_schedule=False` (singular) did nothing. And the random state was a live `Generator`, which `from_dir` cannot restore, which is why the old docstring said a loaded design "reproduces its circuits but cannot re-stitch them."

**The changes**

`CircuitStitcher` is a `NicelySerializable` base class; subclasses implement `_stitch` and nothing else. `RandomizedPatchStitcher` wraps `assign_the_designs_with_mapping` and is the default. `CallableStitcher` adapts a plain function for one-off use. Options are constructor arguments, so a wrong name raises.

Randomness is injected at `stitch` time rather than held by the stitcher. That is what lets the design record `stitch_seed` and makes `restitch()` work on a loaded design. The stitcher does not own its seed because `make_simultaneous_gst_design` splits one user seed into independent coloring and stitching streams, and that coordination belongs to the factory.

`assert_circuit_lists_match_color_patches` moves inside `CircuitStitcher.stitch`, so the check covers direct stitcher calls and not only construction. #912's content-based validator is what made that possible.

The stitching machinery moves to `pygsti/protocols/_stitchers.py`. `simultaneous_gst.py` keeps the design class and re-exports the two names that were importable from it before.

**Tests**

`CircuitStitcherTester` covers casting, the injected rng, output validation and serialization, including a subclass defined outside pyGSTi and the documented non-round-trip of a lambda-backed `CallableStitcher`. `SeedRecordTester` covers the seed record. `SerializationTester` gains a write/load/`restitch()` round trip. `test/unit/protocols` passes: 260 passed, 6 skipped.


CHANGELOG
```diff
@@ -6,6 +6,7 @@
...
+* **Pluggable circuit stitching for `SimultaneousGSTDesign`.** How the one- and two-qubit designs are combined into circuit lists is now an object: subclass `CircuitStitcher` and implement `_stitch`, or pass a plain callable and have it wrapped for you. The default is `RandomizedPatchStitcher`, which assigns patches at random subject to a shared schedule across same-shaped patches. Stitchers are `NicelySerializable`, so a design records the one that built it, survives `write`/`from_dir` with its configuration intact, and can reproduce its circuits from a reloaded design via the new `SimultaneousGSTDesign.restitch()` — provided its seed was an int or a `SeedSequence` rather than a live `Generator`, which cannot be recorded.
```
rileyjmurray, line 9: These can and should be consolidated.

pygsti/protocols/_stitchers.py
```diff
@@ -0,0 +1,987 @@
...
+        state['kwargs'] = self.kwargs
+        return state
+
+    @classmethod
+    def _from_nice_serialization(cls, state):
+        name = state.get('func')
+        if name is None:
+            raise ValueError("Cannot restore a CallableStitcher: its function had no "
+                             "module and qualified name to record. Subclass "
+                             "CircuitStitcher for a stitcher that reloads.")
```
sserita: This is a bad behavior change IMO. I know we can change the API, but an unresolvable lambda/closure stitcher now crashes `from_dir()` entirely, instead of degrading gracefully like the pre-PR code did.
Fixed in #922.

pygsti/protocols/_stitchers.py
```diff
@@ -0,0 +1,987 @@
...
+        CircuitStitcher
+        """
+        if isinstance(obj, CircuitStitcher):
+            if kwargs:
+                raise TypeError(
+                    f"Options passed alongside an already-built {type(obj).__name__}: "
+                    f"{sorted(kwargs)}. Set them on the stitcher instead.")
+            return obj
+        if callable(obj):
+            return CallableStitcher(obj, **kwargs)
```
sserita, line 151: Passing a stitcher class instead of an instance (forgetting `()`) isn't caught here -- silently misinterpreted as a callable, fails later with a confusing, unrelated error. Fixed in PR #922.

pygsti/protocols/_stitchers.py
```diff
@@ -0,0 +1,987 @@
...
+    def cast(cls, obj, **kwargs):
+        """`obj` as a :class:`CircuitStitcher`: itself, or a callable wrapped in one.
+
+        Parameters
+        ----------
+        obj : CircuitStitcher or callable
+            A callable is adapted by :class:`CallableStitcher`, which also carries the
+            serialization caveat that comes with one.
+
+        **kwargs
```
sserita, line 132: Docstring says extra kwargs are "ignored when `obj` is already a stitcher", but the code actually raises `TypeError` in that case (and there's a test for it) -- just a stale docstring, worth a one-line fix.

pygsti/protocols/_stitchers.py
```diff
@@ -0,0 +1,987 @@
...
+    verbosity : int, optional
+        If greater than 0, display a progress bar over germ powers while
+        stitching. Defaults to 0 (silent), so that library calls and test
+        suites produce no output.
+
+    **kwargs
+        Ignored. Accepted so this stitcher matches the generic
+        ``circuit_stitcher(oneq, twoq, vertices, color_patches, **kwargs)``
+        calling convention used by ``SimultaneousGSTDesign``, allowing
+        it to be swapped with other stitchers that take extra options.
```
sserita, line 878: A misspelled `**kwargs` option here is still silently swallowed. Probably fine since that's the back-compat/`CallableStitcher` path rather than the new API, but could be worth either noting or throwing a warning?

pygsti/protocols/simultaneous_gst.py
```diff
@@ -133,34 +98,31 @@ def _normalize_coloring(coloring: Mapping[int, Sequence[Edge]]) -> Dict[int, Lis
...
-                       "unaffected." % (name, e))
-        return None
-    return cast(CircuitStitcher, resolved)
+def _recordable_seed(seed: Optional[SeedLike]) -> Optional[Union[int, Dict[str, Any]]]:
+    """`seed` reduced to something JSON can hold, so a design can re-stitch after a reload.
+
+    An int is itself; a ``SeedSequence`` becomes its entropy and spawn key, which is
+    enough to rebuild it. A live ``Generator`` has consumed state that cannot be
+    recorded, so it becomes None and the design loses the ability to re-stitch -- pass an
+    int if that matters.
```
sserita, line 107: `restitch()` is silently non-reproducible whenever `stitch_seed` is `None` (which you've noted here). But this is also the default unseeded path, so worth a clear warning IMO. Fixed in  PR #922.

pygsti/protocols/simultaneous_gst.py
```diff
@@ -8,81 +8,46 @@
...
+# stitcher-agnostic output validator (which anyone writing their own stitcher is
+# expected to run). The remaining helpers are deliberately left out.
 __all__ = [
     'SimultaneousGSTDesign',
     'make_simultaneous_gst_design',
+    'CircuitStitcher',
+    'RandomizedPatchStitcher',
     'assign_the_designs_with_mapping',
     'assert_circuit_lists_match_color_patches',
 ]
```
sserita, line 48: I generally agree with don't re-export internal things; however, `CircuitStitcher.cast()` forwards `**kwargs` to a wrapped callable, but `SimultaneousGSTDesign.__init__` never passes any kwargs into `cast()` -- so there's no public way to use that feature. The only path is constructing `CallableStitcher(func, **kwargs)` directly, which needs the private `_stitchers` import.

Worth re-exporting `CallableStitcher`, or adding a `circuit_stitcher_kwargs` passthrough on `SimultaneousGSTDesign` if we don't want to expose it?

sserita (changes requested):
Really like the move from bare function to a class, and it fixes a bunch of problems.

I have a few small changes noted below, but also some more critical bugfixes in #922. You can cherry pick those if you only want to accept some of them, but I think they should all at least be addressed.

pygsti/protocols/_stitchers.py
```diff
@@ -872,10 +872,9 @@ def assign_the_designs_with_mapping(
         suites produce no output.
 
     **kwargs
-        Ignored. Accepted so this stitcher matches the generic
-        ``circuit_stitcher(oneq, twoq, vertices, color_patches, **kwargs)``
-        calling convention used by ``SimultaneousGSTDesign``, allowing
-        it to be swapped with other stitchers that take extra options.
+        Accepted for compatibility with legacy callable stitchers, but ignored
+        with a warning naming the unused options. Use the named parameters above
+        to configure this helper.
```
rileyjmurray: We don't have compatibility concerns. SimultaneousGSTDesign hasn't appeared in any public release. Probably best to remove this.

pygsti/protocols/_stitchers.py
```diff
@@ -894,6 +893,11 @@ def assign_the_designs_with_mapping(
         supported; truncate the longer design (or rebuild both with the same
         ``max_lengths``) before calling.
     """
+    if kwargs:
+        _warnings.warn(
+            f"assign_the_designs_with_mapping ignored unused keyword arguments: {sorted(kwargs)}. "
+            "Check the option names or configure a RandomizedPatchStitcher instead.",
+            stacklevel=2)
```
rileyjmurray: See above, re: removing kwargs.

pygsti/protocols/simultaneous_gst.py
```diff
@@ -34,7 +34,7 @@
 
 # This module is star-imported into ``pygsti.protocols``, so ``__all__`` is kept
 # to the documented public surface: the design class, its convenience
-# constructor, the default circuit stitcher and the stitcher base class
+# constructor, the default circuit stitcher, callable adapter and stitcher base class
```
rileyjmurray: Since the free-function-handle approach never appeared in the public API of a release, it's worth considering removing it from the public API here.

rileyjmurray (commented):
We should reevaluate all changes to see where we're prematurely committing to API stability for a feature that was never released.

sserita (approved):
Sorry that "do we expose CallableStitcher" thing turned into a much more involved edit than I thought it would be 😅 
But I like the end result! Approved :)

rileyjmurray:
> Sorry that "do we expose CallableStitcher" thing turned into a much more involved edit than I thought it would be 😅 But I like the end result! Approved :)

No apology needed! The changes are worth it, IMO :)

